PR #2, opened 2026-07-07
feat(bcs): generalize leader election provider wiring

questchaos:
Introduce a public leader_election configuration surface that describes lease timing and provider-specific options without depending on any private cache or database implementation. The bootstrap composition root now passes provider options into inventory-registered leader election factories and reports provider lookup errors with leader terminology.

Keep the open-source binary on the standalone leader implementation by default while allowing linked provider crates to supply their own LeaderElectionPort and lifecycle service. The standalone leader crate now exposes only the public leader contract helpers it needs.

Clean up private infrastructure terminology from cache, database, and repository code. Cache registrations no longer carry zone labels, Redis cache setup connects only from Redis config, persistent bot/group/relation stores use provider-neutral cache/db/logical_db naming, and private integration notes are removed from the public tree.

Preserve DB-backed bot status cache prefix behavior by threading cache.redis.key_prefix into PersistentBotRepo status keys, with bcs: as the compatibility default.

Tests: cargo test --package bcs-config-api --package bcs-cache-local --package bcs-cache-redis --package bcs-db-api --package bcs-db-local --package bcs-bot-store --package bcs-bot --package bcs-group-store --package bcs-relation-store --package bcs-leader-election --package bcs-test-support --package bcs


zh-bupt (approved):
LGTM.

PR #3, opened 2026-07-07
fix(bcs): use refs/heads/dev raw URL and bash -c form for install command

zh-bupt:
- Point install.sh/install.md URLs to refs/heads/dev in the frontend access templates so curl can fetch the correct branch.
- Switch the self-service command from `curl | bash -s -- --token` to `bash -c "$(curl ...)" -s --token`, which leaves stdin free for the terminal so the script's interactive prompts (reuse credentials, bot-name) work instead of reading the script body via the pipe.
- Drop the internal npm registry override from install.sh/install.md and switch the plugin package to @avernet-plugin/openclaw-channel-bcn.

### DIFF
--- a/src/bcs/docs/install-instructions/install.md
+++ b/src/bcs/docs/install-instructions/install.md
@@ -92,7 +92,7 @@ Examples:
 ## Step 3: Install the Plugin
 
 ```bash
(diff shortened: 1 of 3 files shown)


questchaos (approved):
LGTM

questchaos (approved):
LGTM

PR #4, opened 2026-07-07
feat(bcs): add pluggable security gateway registry

questchaos:
Introduce a security gateway provider registry in the public BCS bootstrap so non-noop providers can be supplied by linked plugins instead of being hardcoded in the public binary.

Replace the provider-specific top-level security_gateway fields with a generic providers.<name> configuration map. This keeps public configuration provider-neutral, lets private plugins parse their own options, and rejects legacy top-level provider options instead of silently ignoring them.

Keep the open-source default on the noop security gateway and make missing non-noop providers fail startup with a clear InvalidConfig error rather than falling back to noop. This prevents deployments from believing a real security gateway is active when the provider plugin is not linked.

Tests: cargo test --manifest-path ocb-public/src/bcs/Cargo.toml --package bcs security_gateway --lib

### DIFF
--- a/src/bcs/configs/bcs-config-example.toml
+++ b/src/bcs/configs/bcs-config-example.toml
@@ -191,12 +191,10 @@ default_page_limit = 100000000
 max_page_limit = 100000000
 
 [security_gateway]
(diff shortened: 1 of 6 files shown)


liveandevil (commented):
lgtm

liveandevil (approved):
lgtm

PR #6, opened 2026-07-07
docs: update Avernet introduction to reflect evolving open-source scope

dxwinux:
- problem
  - code is not ready, while news is already there. 
-  solution
   - need some notes to avoid unnecessary misunderstanding.


carolynli (approved):
LGTM

PR #7, opened 2026-07-07
feat(bcs): e2e line-coverage orchestration with parallel pre-push gate

zh-bupt:
Add an end-to-end coverage pipeline for the BCS server, exercised by the e2e-test suite and gated on the pre-push hook. e2e_coverage.sh starts the instrumented bcs plus 5 demo bots via singlebox --standalone, runs e2e-test/e2e.sh, SIGTERMs bcs to flush profraw, stops bcs_bots, then aggregates Cobertura plus a text table via cargo llvm-cov report --package bcs. e2e is a 100% gate; --bcs-min additionally gates line coverage.

prepare_bcs_coverage_bin builds instrumented bcs/bcs-cli into an isolated target/cov-e2e dir with -Cinstrument-coverage/-Clink-dead-code and exports BCS_BIN/BCS_CLI_BIN plus the runtime LLVM_PROFILE_FILE. --force-rebuild bypasses the cache so the pre-push gate exercises the pushed source. Build-time profraw from build scripts and proc-macros is redirected to src/bcs/target/tmp, keeping the source tree clean; all profraw is cleared after aggregation, and *.profraw is gitignored. bcs_bin_path/bcs_cli_path honor the coverage overrides so check_bcs_*_binary, bots/bcs onboarding, and start_bcs_bots.sh work after rm -r target. A preflight step reclaims busy bcs/bot ports by SIGTERM/SIGKILL so re-runs don't fail, without touching the frontend (e2e needs none).

The bot stack starts in --standalone mode so profiles live under <checkout>/.standalone-openclaw per-checkout instead of the shared ~/.openclaw-* that --local stamps with the first checkout's workspace; without this the e2e-coverage gate fails in a second checkout/worktree ("profile exists but does not match this singlebox local stack").

The BCS pre-push gate (src/bcs/scripts/pre_push.sh, mirroring ci_test.sh's --base/--head contract) runs the unit (nextest, fast-fail) and e2e-coverage gates concurrently with fail-fast: the first gate to exit non-zero terminates the other's whole process tree (_kill_tree via pgrep, portable without setsid) and blocks the push. Output is line-tagged via FIFO ([unit]/[e2e]) so the interleaved streams stay atomic; a per-gate PASSED/FAIL/KILLED summary block is printed at the end, with full logs under scripts/.dependencies/logs/prepush/. The generic scripts/ci/pre_push.sh just computes the push range, gates on OCB_PRE_PUSH_ENABLE_BCS, and delegates. OCB_PRE_PUSH_ENABLE_BCS=0 / OCB_PRE_PUSH_ENABLE_BCS_E2E=0 opt out (the latter selects serial unit-only).

Tests: rm -r src/bcs/target then bash src/bcs/scripts/e2e_coverage.sh --bcs-min 30 --force-rebuild (e2e 38/38, ~37.8% > 30%, 0 stray profraw); scripts/ci/pre_push.sh --base HEAD~2 --head HEAD over a src/bcs/ range delegates to src/bcs/scripts/pre_push.sh -> unit 2066/2066 + e2e 38/38 @37.76% > 30%, exit 0; reproduced the ~/.openclaw-* collision in .worktrees/bugfix and confirmed it passes with --standalone (40/40). Synthetic both-pass and fail-fast stubs verify tagging and process-tree teardown.

### DIFF
--- a/scripts/ci/pre_push.sh
+++ b/scripts/ci/pre_push.sh
@@ -39,12 +39,6 @@ matches_any() {
   printf '%s\n' "$changed_files" | grep -Eq "$pattern"
 }
 
-run_required() {
-  echo ""
-  echo "== required: $* =="
-  "$@"
-}
-
 echo "base: $base"
 echo "head: $head"
 
@@ -55,7 +49,9 @@ fi
 
 if matches_any '^src/bcs/'; then
   if [[ "${OCB_PRE_PUSH_ENABLE_BCS:-1}" == "1" ]]; then
-    run_required "$repo_root/src/bcs/scripts/ci_test.sh" --base "$base" --head "$head" --fast-fail
+    echo ""
+    echo "== required: src/bcs/scripts/pre_push.sh --base $base --head $head =="
+    "$repo_root/src/bcs/scripts/pre_push.sh" --base "$base" --head "$head"
   else
     echo "bcs changes detected; BCS/BCN CI gate skipped (OCB_PRE_PUSH_ENABLE_BCS=0)"
   fi
--- a/scripts/modules/bcs.sh
+++ b/scripts/modules/bcs.sh
@@ -713,7 +713,7 @@ onboard_default_openclaw_if_connected() {
         return 0
     fi
 
-    local bcs_cli="${BCS_DIR}/target/debug/bcs-cli"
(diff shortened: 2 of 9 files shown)


questchaos (approved):
LGTM

carolynli (approved):
LGTM

PR #8, opened 2026-07-07
feat(bcs): add pluggable user directory support

questchaos:
Introduce a public user_directory configuration contract with provider-specific option maps so deployments can select linked directory plugins without hard-coding provider settings in bootstrap.

Add a UserDirectoryFactory registry alongside the existing cache, database, leader election, LLM, and security gateway registries. Wire server startup to load the configured provider, pass its option map through to the factory, and fail fast when an enabled provider is missing from the binary.

Document the disabled-by-default public configuration shape and update tests that construct BcsConfig directly so they remain explicit about the new field.

### DIFF
--- a/src/bcs/configs/bcs-config-example.toml
+++ b/src/bcs/configs/bcs-config-example.toml
@@ -138,6 +138,12 @@ max_keep_days = 7
 hyper = "warn"
 tower_http = "warn"
 
+[user_directory]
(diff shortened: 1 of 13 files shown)


RaymondHX (approved):
LGTM

vzvince (approved):

PR #17, opened 2026-07-08
ci(unit-tests): add backend and engine test jobs

totalfrank:
## What
Adds `Backend unit tests` and `Engine unit tests` jobs to the Unit Tests workflow, so the newly imported `src/backend` and `src/engine` subtrees get CI coverage.

## How
Mirrors the existing BCS job pattern:
- **Change detection** — diff the subtree against the PR base; skip the job when nothing under `src/backend` / `src/engine` changed, otherwise run the full suite.
- **Runner** — invokes the vendored `scripts/ci_test.sh`, which runs pytest with coverage and enforces the gates via `scripts/ci/report_check.py`:
  - backend: 100% case pass · ≥75% line · ≥80% changed-line
  - engine: 100% case pass · ≥70% line · ≥90% changed-line
- **Python** pinned to 3.12 (backend targets `>=3.12,<3.13`; engine `>=3.12`).
- **uv** with lockfile-scoped caching.

## Scale note
A change touching one file under a subtree runs that subtree's entire suite (~6,700 backend / ~1,700 engine test functions). This is the first time these suites run in CI, so the initial run may surface env-dependent failures or coverage-gate misses to tune.

## Follow-up
Once the check names appear, add `Backend unit tests` and `Engine unit tests` to the `protect-dev` ruleset's required status checks so they gate merges (currently only the BCS checks are required).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,3 +12,5 @@
 /docs/arch/ @carolynli @xianmuyq
 /src/bcs/ @carolynli @vzvince @questchaos @RaymondHX
 /src/plugin/ @RaymondHX @vzvince @carolynli @questchaos
+/src/backend/ @totalfrank @xianmuyq
+/src/engine/ @totalfrank @xianmuyq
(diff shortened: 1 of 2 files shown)


carolynli (approved):
LGTM

PR #34, opened 2026-07-09
feat(models): add extra dict to Model for transparent relay field for…

CapQing:
…warding

Model DTO gains an extra: dict[str, Any] field. The API router merges non-empty extra into the top-level response dict so engine-specific fields (e.g. runtime from the relay) reach the frontend without requiring a DTO field per key.


totalfrank (dismissed):
LGTM

xianmuyq (dismissed):

xianmuyq (approved):

PR #36, opened 2026-07-09
feat(cron): support service bot cron runtime instances

XIEAN99:
## Summary
- Route cron router operations through explicit runtime-operation methods instead of adding new behavior to `forward_request`.
- Support service bot runtime-stage and device-scoped cron operations, including multi-instance fan-out for run/toggle and instance-aware run history.
- Preserve legacy `forward_request` behavior for existing internal callers and extend resolver tests for the runtime paths.


totalfrank (approved):
LGTM.

PR #46, opened 2026-07-09
fix(cron): route runtime APIs outside forward_request

XIEAN99:
Preserve legacy forward_request behavior for existing cron relay callers.

Move service bot runtime operations into explicit operation methods.

Keep cron list stage-scoped while running and run history can be queried by runtime stage or device.

Add coverage for runtime fan-out, run history, and validation paths.


totalfrank (approved):
LGTM

PR #58, opened 2026-07-09
fix: fan out service bot cron by runtime devices

XIEAN99:
Resolve service bot cron runtime targets from the runtime device list instead of the public instance list, so BaaS/Teclaw multi-instance published stages can route by device_uuid without health checks.

Keep personal bots on the legacy draft path, reject published-stage params for personal bots, and keep Arca/local published stages as single-target routing.

Restore multi-instance update/run/runs responses to the documented results[] shape, with per-result runtime metadata and no instance health fields.


totalfrank (approved):
LGTM

PR #75, opened 2026-07-10
fix(cron): skip Hermes bots in runtime queries

XIEAN99:
Exclude Hermes bots before cron target construction so list and running queries do not access device, publish, resolver, or adapter layers. Keep supported-engine behavior unchanged.


totalfrank (approved):
LGTM.

PR #87, opened 2026-07-10
fix(cron): preserve verify targets after online release

XIEAN99:
Resolve verify targets from validating publish records first, then fall back to the latest successful publish record.\n\nKeep cron list and stage-specific operations aligned after a service bot reaches online.


totalfrank (approved):
LGTM

PR #93, opened 2026-07-11
refactor(runtime): separate deploy profile from data env

FreddieSun:
## Summary

- Make `DeployProfile` the only implementation/DI selector.
- Keep runtime/data Env limited to `dev / pre / prod`; singlebox now launches with `DEPLOY_PROFILE=singlebox SERVER_ENV=dev`.
- Let each profile YAML overlay own profile-specific component fields, including workspace placement, instead of adding another environment axis.
- Select YAML overlays and test/singlebox bindings explicitly by Profile.
- Reject legacy `SERVER_ENV=singlebox` at startup, remove the Device/Schema alias, and add a canonical static guard preventing `singlebox` from re-entering the Env axis.

## Review order

@totalfrank please read the Spec first, then the implementation:

1. `docs/superpowers/specs/2026-07-11-profile-env-separation-design.md`
2. `docs/superpowers/plans/2026-07-11-profile-env-separation.md`
3. Config/Profile commit, DI split commit, then Runtime/Env commit

The long-term coverage PR train is recorded in `docs/superpowers/plans/2026-07-11-singlebox-coverage-pr-train.md`.

## Review feedback status

- Profile-to-overlay ownership: fixed; YAML providers now own profile-specific component values.
- Workspace folder selection: fixed; the extra `WORKSPACE_ENV_FOLDER` environment variable was removed.
- Relative engine workspace paths / `get_bolt_base_dir`: intentionally split into #107 and assigned to @FreddieSun and @totalfrank for a dedicated follow-up PR.

## Validation

- GitHub Actions: Backend, BaaS, BCS, Engine, BCS E2E, and Singlebox Coverage all passed.
- Real pre-push gate passed after rebasing onto the latest `dev`.
- Real singlebox coverage passed after the full standalone stack started.
- Teardown verified: the stack was stopped after coverage collection.

BCS has no diff in this PR. Its existing concurrent integration-test baseline is documented separately and is not hidden as a result of this refactor.

## Follow-up

After merge and mirror sync, update the outer OCB `ocb-public` gitlink and remove its retired `TestingAccessModule` corp-test expectation.


src/backend/src/agentclaw/community/core/config/yaml_provider.py
```diff
@@ -24,52 +22,10 @@
...
-      专属块（token_exchange / arca_sandbox / dima / …）都不设值，社区 overlay 提供
-      自己需要的中性配置，因此合并后仍是纯中性配置。
-    - corp / test / singlebox：``application.yaml`` + 由 SERVER_ENV 选定的 overlay
-      （``application-{prod,dev,singlebox,…}.yaml``），与 sofapy_base 在 prod 用
-      SERVER_ENV 选 yaml 的逻辑一致。corp 专属值由各 env overlay 承载。
-    """
-    overlay_name = _select_overlay_name()
-
+def _load_yaml_configs(
+    overlay_name: str = "application-dev.yaml",
```
totalfrank, line 58: Why allowing call site to specify it? This should be the yaml provider's own logic.

Basically yaml provider looks at profile to decide which is the overlay.

It's OK if you want to pass the profile explicitly. But generally I think it's a bad idea for this to take the overlay file directly.

src/backend/src/agentclaw/community/core/config/yaml_provider.py
```diff
@@ -114,10 +66,13 @@ def _deep_merge(base: dict, override: dict) -> dict:
 
 
 class YamlConfigProvider:
-    """Loads :class:`AppConfig` from local ``application*.yaml`` files."""
+    """Load AppConfig from the neutral base plus one explicit overlay."""
+
+    def __init__(self, overlay_name: str = "application-dev.yaml") -> None:
+        self.overlay_name = overlay_name
```
totalfrank: Same thing, profile decides overlay. Not call site decides.

src/backend/src/agentclaw/community/core/workspace/path_factory.py
```diff
@@ -128,15 +128,20 @@ def _get_aidesktop_root() -> Path:
     return _get_config_value("aidesktop_root", DEFAULT_AIDESKTOP_ROOT, "AIDESKTOP_ROOT")
 
 
-def _get_aidesktop_env() -> str:
-    """读取当前环境名（dev/pre/prod 等）。"""
+def _get_aidesktop_env_folder() -> str:
+    """Return the physical workspace folder selected by the active Profile."""
+    explicit = os.getenv("WORKSPACE_ENV_FOLDER")
```
totalfrank: Works but might not be the best approach.

This explicit overwrite brings cognitive load.

For example, when I read the code here, I will be concerned for

1. In prod, I should not set up this env var. Where should I go check to make that not happen?
2. In testing/sandbox, am I setting up this properly? Where should I go and check?

I think the better approach to do it still using overlays so that we don't have this explicit overlay here. Everything for an env is in its overlay file and it's much easier for me to check.

src/backend/src/agentclaw/community/core/workspace/path_factory.py
```diff
@@ -128,15 +128,20 @@ def _get_aidesktop_root() -> Path:
...
     from agentclaw.community.utils.env_utils import get_current_env
-    return get_current_env()
+
+    return f"aidesktop_{get_current_env()}"
 
 
 def get_bolt_base_dir() -> Path:
     """bolt_data 根目录: {aidesktop_root}/aidesktop_{env}/bolt_data"""
-    return _get_aidesktop_root() / f"aidesktop_{_get_aidesktop_env()}" / "bolt_data"
+    return _get_aidesktop_root() / _get_aidesktop_env_folder() / "bolt_data"
```
totalfrank, line 151: Might not need to do it here. But we might not need this get_bolt_base_dir at all.

Because these are just for referencing the files in the engine container and we are half way there migrate the file path to be relative rather than absolute.

Just a FYI.

totalfrank (commented):
A few comments need to call out, please take a look

src/backend/src/agentclaw/community/core/config/yaml_provider.py
FreddieSun: Updated in 5c30bec. The composition root now passes the semantic DeployProfile object, and YamlConfigProvider owns the profile-to-overlay mapping. Raw application-*.yaml filenames are no longer accepted by the provider API. A structural Protocol avoids a Core-to-DI dependency.

src/backend/src/agentclaw/community/core/config/yaml_provider.py
FreddieSun: Updated in 5c30bec. YamlConfigProvider now receives the semantic profile and resolves its own overlay internally; config_bootstrap no longer maps profiles to physical YAML filenames.

src/backend/src/agentclaw/community/core/workspace/path_factory.py
FreddieSun: Updated in 5c30bec. Removed WORKSPACE_ENV_FOLDER from both Backend and BaaS launch scripts. Backend and BaaS now declare workspace.env_folder: aidesktop_singlebox in their active singlebox overlays and resolve it through their config providers. Live singlebox startup confirmed both sides use the same path.

src/backend/src/agentclaw/community/core/workspace/path_factory.py
FreddieSun: Acknowledged. I kept get_bolt_base_dir in this PR to avoid expanding the Profile/Env separation into the relative-path migration. RSYNC_TARGET_DIR was made lazy only to avoid config reads before composition-root registration; the broader relative-path cleanup remains follow-up work.

FreddieSun: Tracked as #107: https://github.com/inclusionAI/Avernet/issues/107, assigned to @FreddieSun and @totalfrank. This remains intentionally out of scope for PR #93 and will be implemented in a dedicated follow-up PR after the Profile/Env separation lands.

src/backend/src/agentclaw/community/core/config/yaml_provider.py
FreddieSun: Post-rebase commit reference: 7a1eacd. The implementation and ownership described above are unchanged; PR #93 is now rebased onto dev at f8a4bfb.

src/backend/src/agentclaw/community/core/config/yaml_provider.py
FreddieSun: Post-rebase commit reference: 7a1eacd. Profile-to-overlay resolution remains owned by YamlConfigProvider.

src/backend/src/agentclaw/community/core/workspace/path_factory.py
FreddieSun: Post-rebase commit reference: 7a1eacd. The environment-variable override remains removed and workspace.env_folder remains overlay-owned in Backend and BaaS.

totalfrank (approved):
LGTM.

cassiuscai (approved):

PR #95, opened 2026-07-11
fix(bcs): reset thinking accumulation at agent stream boundaries

vzvince:
## Summary

- Reset the per-run thinking accumulator when any non-thinking agent stream event arrives.
- Handle real `ToolCallStart` and `ToolCallEnd` boundaries instead of requiring `ChatEventState::Delta`.
- Add regression coverage for tool start/end segmentation and provider result-state mapping.

## Root Cause

Thinking segmentation was gated on `ChatEventState::Delta`. Provider SSE maps tool `start` and `result` phases to `ToolCallStart` and `ToolCallEnd`, so those events never cleared the thinking buffer and later thinking segments included earlier content.

## Impact

Frontend thinking snapshots now remain cumulative only within the current thinking segment and reset correctly across tool calls.

## Validation

- `cargo test --manifest-path src/bcs/Cargo.toml -p bcs-message-flow -p bcs-provider-http`
- Verified the consuming `bcs-internal` binary still compiles in the integration workspace.


carolynli (approved):
LGTM.

PR #112, opened 2026-07-13
fix(singlebox): land BaaS-backed device runtime on dev

FreddieSun:
## Summary

- land the effective changes from merged PR #98 onto the latest `dev`
- route the singlebox device lifecycle through the real BaaS provider
- keep test and singlebox DI profiles separated
- seed the singlebox BaaS template mapping through an explicit lifecycle participant

## Lineage

PR #98 was merged into the intermediate `codex/profile-env-separation` branch rather than `dev`. This PR reapplies its single effective commit on top of the current `dev` so the implementation actually reaches the main development line.

## Verification

- targeted DI/config tests: 30 passed
- Backend community regression: 7419 passed
- pre-push SAST, Backend suite, and BCS build passed; real singlebox setup reached BCN plugin installation, then was stopped after the local `npm install` made no progress for several minutes

### DIFF
--- a/src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
+++ b/src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
@@ -1,0 +1,168 @@
+"""Singlebox BaaS-only device runtime bindings."""
+
+from __future__ import annotations
+
+from typing import cast  # noqa: UP035 - injector binding key matches provider side
+
+from injector import Binder, Module, inject, provider, singleton
+
+from agentclaw.community.core.bot_management.repository.protocol import BotRepository
+from agentclaw.community.core.bot_management.services.teclaw_status_reconciler import (
+    TeclawStatusReconciler,
+)
+from agentclaw.community.core.devices.protocols import BotQueryProtocol
+from agentclaw.community.core.devices.repository.protocol import DeviceBindingRepository
+from agentclaw.community.core.devices.services.arca_bot_create_baas_rollout_policy import (
+    ArcaBotCreateBaasRolloutDecision,
+    ArcaBotCreateBaasRolloutPolicy,
+)
+from agentclaw.community.core.devices.services.baas_device_accessor import (
+    BaasDeviceAccessor,
+)
+from agentclaw.community.core.devices.services.baas_device_service import (
+    BaasDeviceService,
+)
+from agentclaw.community.core.devices.services.device_accessor import DeviceAccessor
+from agentclaw.community.core.devices.services.device_service import (
+    BAAS_DEVICE_PROVIDER,
+    DeviceService,
+)
+from agentclaw.community.core.devices.services.device_service_router import (
+    DeviceServiceRouter,
+)
+from agentclaw.community.core.notify.protocol import NotifyBotLister
+from agentclaw.community.core.system_config import SystemConfigService
+from agentclaw.community.di import config as cfg
+from agentclaw.community.core.service_bot.repository.bot_publish_repository import (
+    BotPublishRepositoryProtocol,
+)
+from agentclaw.community.core.service_bot.services.baas_service import BaasService
+from agentclaw.community.plugin_api.device_adapter_transport import (
+    DeviceAdapterTransport,
+)
+from agentclaw.community.plugin_api.device_connection_manager import (
+    DeviceConnectionManagerPlugin,
+)
+from agentclaw.community.plugin_api.passport import PassportPlugin
+from agentclaw.community.plugin_api.sandbox_runtime import SandboxRuntimeClient
+from agentclaw.community.di.modules.infrastructure.singlebox.template_config import (
+    SingleboxBaasTemplateConfigLifecycle,
+)
+
+
+class _SingleboxAllBaasRolloutPolicy(ArcaBotCreateBaasRolloutPolicy):
+    """Singlebox has no ARCA provider, so every allocation is BaaS-owned."""
+
+    def __init__(self) -> None:
+        pass
+
+    def decide(
+        self,
+        *,
+        user_id: str,
+        bot_type: str | None,
+        engine_type: str | None = None,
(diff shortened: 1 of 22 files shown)


src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
```diff
@@ -0,0 +1,168 @@
...
+        from agentclaw.community.plugins.local.device_connection_manager import (
+            NoopDeviceConnectionManagerPlugin,
+        )
+
+        binder.bind(
+            DeviceConnectionManagerPlugin,
+            to=NoopDeviceConnectionManagerPlugin,
+            scope=singleton,
+        )
+        binder.bind(DeviceAccessor, to=BaasDeviceAccessor, scope=singleton)
```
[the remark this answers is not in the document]
FreddieSun: Addressed in 64e684d. I added an explicit singleton binding for BaasDeviceAccessor and expose that concrete singleton through a provider for DeviceAccessor. This follows the repository concrete-to-protocol alias pattern and guarantees both injection keys resolve to the same object; binding both keys directly to the class would still create separate Injector-scoped instances. The DI test now asserts object identity. Verified with 69 targeted Device/BaaS/DI tests, plus the real pre-push Backend suite (7419 passed) and Singlebox coverage gate.

src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
```diff
@@ -0,0 +1,175 @@
...
+    """Singlebox has no ARCA provider, so every allocation is BaaS-owned."""
+
+    def __init__(self) -> None:
+        pass
+
+    def decide(
+        self,
+        *,
+        user_id: str,
+        bot_type: str | None,
```
totalfrank: No | None for all of them. I believe they are all required.

src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
```diff
@@ -0,0 +1,175 @@
...
+            passport_plugin=passport_plugin,
+            sandbox_client=sandbox_client,
+            publish_repo=bot_publish_repo,
+            bot_repo=bot_repository,
+        )
+
+    @singleton
+    @provider
+    @inject
+    def teclaw_status_reconciler(
```
totalfrank, line 151: Do we still need this? I think with the queue task service, this is no longer required.

src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
```diff
@@ -0,0 +1,175 @@
...
+    def notify_bot_lister(self, bot_repository: BotRepository) -> NotifyBotLister:
+        from agentclaw.community.core.notify.local_bot_lister import (
+            LocalNotifyBotLister,
+        )
+
+        return LocalNotifyBotLister(bot_repository=bot_repository)
+
+    @singleton
+    @provider
+    def device_adapter_transport(self) -> DeviceAdapterTransport:
```
totalfrank: Nit: you can just bind this instead of using provider. But it's fine either way, it's just binder way seems neater.

src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/template_config.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class SingleboxBaasTemplateConfigLifecycle(LifecycleBase):
+    """Install an idempotent BaaS template map after SQLite bootstrap."""
+
+    def __init__(
+        self,
+        *,
+        config_service: SystemConfigService,
+        template_uuid: str | None,
```
totalfrank: This must be provided right? If so, let's not do | None.

totalfrank (commented):
Minor comments, PTAL.

src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
FreddieSun: Teclaw has not migrated to TaskQueueService yet: TeclawProvisionService still calls the process-local TeclawStatusReconciler directly, while only the regular BaaS create/restart lifecycle uses the durable queue. Removing this singlebox binding now would activate the real in-process scheduler in singlebox rather than remove the restart gap. Keeping this no-op binding in this PR and tracking the full Teclaw migration separately in #116, including removal of the reconciler and all profile-specific bindings.

src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
FreddieSun: Fixed in e5ff877. The rollout policy contract now requires concrete `str` values for bot, engine, and template type. Optional request/repository values are normalized once at the `DeviceServiceRouter` boundary before entering the policy, and all implementations/tests now follow the required contract.

src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/devices.py
FreddieSun: Applied in e5ff877. `DeviceAdapterTransport` is now bound directly to `InMemoryDeviceAdapterTransport` as a singleton in `configure()`, and the one-line provider method has been removed.

src/backend/src/agentclaw/community/di/modules/infrastructure/singlebox/template_config.py
FreddieSun: Fixed in e5ff877. `template_uuid` is now required end-to-end (`str`, not optional); the lifecycle constructor rejects non-strings explicitly and tests cover both the empty-string missing-config path and accidental `None` input.

totalfrank (approved):
LGTM.

xxxxpenny (approved):

PR #137, opened 2026-07-13
fix(bcs-cli): render current bot groups

RaymondHX:
## What changed

- Make `bcs-cli list-groups --mine` understand the per-bot group response fields: `group_id`, `coordinator_bot`, and `group_strategy`.
- Preserve compatibility with the global/legacy fields: `id`, `driver_bot`, and `mode`.
- Update the E2E fixture to match the real `/bots/{bot_uuid}/groups` response and fail if `unknown` is printed.

## Why

The per-bot endpoint returns a different item shape from `/groups`. The CLI rendered the per-bot result using the global response fields, so every row appeared as `unknown [unknown] driver=unknown`.

## User impact

`bcs-cli list-groups --mine` now prints the actual group ID, strategy, and coordinator.

## Validation

- `cargo test --package bcs-cli --test e2e` (7 passed, 1 ignored)
- `cargo check --package bcs-cli --all-targets`
- `git diff --check`

### DIFF
--- a/src/bcs/crates/tools/bcs-cli/src/main.rs
+++ b/src/bcs/crates/tools/bcs-cli/src/main.rs
@@ -2670,14 +2670,18 @@ async fn main() -> Result<()> {
             for group in groups {
                 let id = group
                     .get("id")
+                    .or_else(|| group.get("group_id"))
                     .and_then(|v| v.as_str())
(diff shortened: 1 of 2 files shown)


vzvince (approved):
LGTM

PR #160, opened 2026-07-14
fix(channel): inject context for group-bound sessions

RaymondHX:
## Summary

- inject the standard BCS session context when an IM channel is bound to an existing BCS group
- preserve direct-bot context projection without exposing synthetic group details
- wire `SystemMessageService` into the channel runtime and cover new and reused session behavior

## Root cause

Channel-created sessions bypassed the regular group session creation route, so they never emitted `SystemMessageEvent::SessionContext`. Direct bot and group targets therefore shared the same no-initial-context behavior.

## Behavior

- Group target: inject context once before the first user message
- Bot target: do not inject group context
- Reused sessions: do not inject duplicate context
- State-machine target: unchanged
- Context injection failure: log the failure and continue dispatching the user message

## Verification

- `cargo test -p bcs-channel` (40 passed)
- `cargo check -p bcs`
- `git diff --check`


questchaos (approved):
LGTM

PR #161, opened 2026-07-14
fix: preserve default cli scope during mcp sync

panqiwp:
## Summary
- Preserve default CLI scope when synchronizing MCP configuration
- Avoid losing default CLI registration during MCP sync

## Test
- Not run (not requested)


totalfrank (approved):
LGTM.

PR #167, opened 2026-07-14
unify stream converter and clean up legacy config

xxxxpenny:
- Remove BcnStreamConverter; route all SSE streams through
      DefaultStreamConverter to simplify the converter pipeline
- Add heartbeat support to StreamChunk and SSE comment frames
      in SseEvent.to_sse
- Add ignore_case option to AsyncChatClient / pool for
      case-insensitive sessionKey matching
- Inject SystemConfigManageService into AsyncChatClientPool for
      runtime config access
- Clean up obsolete sidecar, k8s_template, and mist config blocks
      from application YAMLs
- Drop legacy SQL DDL scripts and f.yml oneapi stub
- Add opentelemetry-instrumentation-openai dependency
- Update tests to use DefaultStreamConverter; remove obsolete
      test_default_converter.py


pfmiles (approved):
ok

PR #183, opened 2026-07-15
feat(baas): remove legacy allow-all from parse_policy

phoenixliu:
BREAKING CHANGE: parse_policy now treats None/empty policy and missing allowed_bots key as deny-all (403) instead of allow-all. Historical compatibility logic has been removed as data migration is complete.

- _policy.py: None/empty → deny-all; missing allowed_bots → deny-all
- dependencies.py: update validate_policy docstring
- test_policy.py: 4 cases updated from allow-all to deny-all
- test_dependencies.py: 3 validate_policy cases updated to expect 403


xxxxpenny (approved):
+1

PR #190, opened 2026-07-15
fix(publish): describe terminal statuses in process() instead of raising

totalfrank:
## Summary

`PublishFlowService.process()` routes every non-advance-point status through `_describe_publish_status`, which only consulted `_DESCRIBE_STATUS_MESSAGES`. A publish record can reach a terminal state (`UPGRADED`/`RELEASED`) concurrently with an in-flight `process()` call — superseded by a newer publish, or taken offline — and those statuses are deliberately absent from that table, so `process()` raised `PublishStatusInvalidError("Unknown publish status: ...")` for a perfectly known state.

This makes `_describe_publish_status` fall back through `_SYNC_ONLY_STATUS_MESSAGES` (the same table `describe_publish` already uses), so a superseded/offline record reports a friendly message instead of raising.

## Changes

- `_describe_publish_status` now falls back to `_SYNC_ONLY_STATUS_MESSAGES` when a status isn't in `_DESCRIBE_STATUS_MESSAGES`. Only `UPGRADED`/`RELEASED` change behavior — `DRAFT`/`VALIDATING` are intercepted as advance points and never reach this path, and the CAS-loss re-read only ever sees post-advance statuses.
- Made `_describe_publish_status` a `@staticmethod` (it uses no instance state).
- Updated the `_SYNC_ONLY_STATUS_MESSAGES` comment to reflect that both describe paths now share the fallback.
- Added a parametrized test asserting `process()` describes `UPGRADED`/`RELEASED` without mutating or raising.

## Testing

`uv run python -m pytest tests/community/core/service_bot/services/test_publish_flow_service.py` — 121 passed.

## Notes

This came out of a review that raised five points; the other four are intentionally not code changes:
- The `is_online_release_recorded` guard in `publish_flow/tasks.py` is required (online release runs within `ONLINE_PUB` with no self-advance, so it needs the `ext.publish.online` marker as its idempotency checkpoint; verify release doesn't need one because its `BUILT → VALIDATE_PUB` transition serves that role).
- `refresh_publish_handle` and every other method in `PublishFlowService` are used in production, so there's no dead code to remove.
- The submit-time vs poll-time `BUILDING` message wording ("Build started …" vs "Build in progress …") is a deliberate, consistent pattern (mirrored for `ONLINE_PUB`); left pending a product decision.

https://claude.ai/code/session_01Bkr9bgEZhmm2ZkchDNZ6Lh

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bkr9bgEZhmm2ZkchDNZ6Lh)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/service_bot/services/publish_flow_service.py
+++ b/src/backend/src/agentclaw/community/core/service_bot/services/publish_flow_service.py
@@ -117,9 +117,10 @@ _DESCRIBE_STATUS_MESSAGES = {
     PublishStatus.SUCCESS: "Publish complete",
 }
 
-# The read-only /sync status report covers every status, including the ones
-# process() handles as advance points or rejects — keep these out of
-# _DESCRIBE_STATUS_MESSAGES so process() behavior stays unchanged.
+# Statuses ``process()`` never advances *through*: DRAFT/VALIDATING are its advance
+# points (intercepted before describe-dispatch); UPGRADED/RELEASED are terminal.
+# Both describe paths (_describe_publish_status, describe_publish) fall back through
+# this table so a superseded/offline record reports a friendly message, not an error.
 _SYNC_ONLY_STATUS_MESSAGES = {
     PublishStatus.DRAFT: "Draft, publish not started",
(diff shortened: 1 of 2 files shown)


FreddieSun (approved):
LGTM

PR #223, opened 2026-07-16
fix(bcs): fix openclaw "unknown channel id: bcs" error when oboarding local openclaw

zh-bupt:
fix(bcs): fix openclaw "unknown channel id: bcs" error when oboarding local openclaw


RaymondHX (approved):
LGTM

PR #231, opened 2026-07-16
feat(engine): add git author session create fields

maicaisongxiaocong:


src/engine/src/engine/community/core/session/models.py
```diff
@@ -68,6 +68,8 @@ class SessionCreateRequest(BaseModel):
     cwd: str | None = None
     uuid: str | None = None
     extInfo: dict[str, Any] | None = None
+    gitAuthorName: str | None = None
```
totalfrank: These are engine specific fields right? We can't keep adding fields to engine agnostic domain model.

Add a payload field, and put whatever you need inside it(preferrablely a namespace under that payload so it's more clear).

Something like this: {
  "ai_coding":  {
    ...
  }
}

totalfrank (changes requested):

totalfrank (approved):

totalfrank (commented):
LGTM

PR #105, opened 2026-07-12
refactor(service-bot): decompose publish_flow_service + durable task pipeline

totalfrank:
## Summary

Refactors the 3185-line `publish_flow_service.py` into a maintainable package and
turns the publish pipeline into a **durable, crash-safe** task flow. Driven via
Spec-Driven Development — see `src/backend/specs/2026-07-12-publish-flow-service-refactor/`
(spec/plan/tasks).

Behavior is preserved except the intentional changes noted below: the pipeline
now advances through the persisted task queue, `/process` moves the status
forward synchronously at the user-driven gates, and `/sync` becomes a read-only
status report.

## Structural refactor (goals #1–#5)

- **Decomposed** the monolith into a facade (~950 lines) + a `publish_flow/`
  package: `ext_state`, `provider_behavior`, `build_stage`, `release_stage`, and
  mixins (`progress_sync`, `restart`, `scale`, `stage_status`, `rollback_ops`,
  `baas_publish_ops`, `device_binding`, `publish_ext`, `eval_publish`,
  `upgrade_resolution`). Every module is now under the repo's 1000-line cap; the
  `test_no_oversized_modules.py` allowlist entry is removed.
- **Provider-agnostic:** a `ProviderBehavior` seam + router replaces all the
  scattered `active_engine == "teclaw"` / `provider == TECLAW` branches.
- **De-duplicated:** the near-identical verify/online release methods collapse
  into one `ReleaseStageRunner.first_release` + `.upgrade_release`, parameterized
  by a `StageSpec` (preserving the verify-first-release "no version" quirk).
- **Clean component boundaries:** the stage runners take their real dependencies
  explicitly (`PublishExtState`, services, routers) instead of reaching into
  facade privates; the release runner consumes the four release-record ops
  (`create_release_binding` / `record_release_ext` / `approve_baas_publish` /
  `refresh_publish_handle`, now public) through a narrow `ReleaseRecordOps`
  protocol. The task handlers likewise use only the facade's public contract.
  No component crosses the underscore boundary.
- **Single-writer per resource:** publish-record status/ext writes live only in
  `PublishExtState` (+ `PublishExtMixin` domain ops); device-binding writes only
  in `DeviceBindingMixin` (create / activate / release); BaaS workflow calls
  (approve + progress query) only in `BaasPublishOpsMixin`. Feature mixins
  orchestrate.
- **Renames:** `_upgrade_last_publish` → `_mark_previous_publish_superseded`,
[...]

PR #256, opened 2026-07-19
fix(service_bot): clear stale online release refs on rollback to DRAFT

### LINKED ISSUE #255: Service bot: online release skipped when re-publishing a rolled-back version (stale ext.publish.online)
## Summary

For a service bot, after **rollback to a previous version** followed by re-driving that version's online publish, the online release phase (`PublishFlowService.execute_release_phase`) is **silently skipped**. The publish record marches to `SUCCESS` and the UI reports the version is live, but the live BaaS bot instance is never upgraded — it keeps running the rollback target. Recorded state and the actually-deployed bot diverge, with no error surfaced.

## Where

`src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/tasks.py` — `PublishOnlineReleaseHandler._run`:

```python
if status == PublishStatus.ONLINE_PUB and not self._flow.is_online_release_recorded(publish_id):
    release_result = await self._flow.execute_release_phase(record, operator)   # skipped
```

## Root cause

`is_online_release_recorded(publish_id)` returns `True` whenever the record's `ext.publish.online` marker is present (`publish_flow_service.py:973-984`). That marker is written on every successful online release (`publish_flow/release_stage.py:286`).

On rollback, `rollback_publish` (`publish_rollback_mixin.py:153-166`) moves the **current** version's record back to `DRAFT` but does **not** clear its `ext` — so the `ext.publish.online` marker from that version's *earlier* go-live survives. Nothing on the re-build / re-publish path clears it either. When the record is re-driven and reaches `ONLINE_PUB`, `is_online_release_recorded` returns `True` (stale marker) → the guard is `False` → `execute_release_phase` is skipped. The poll then reads the stale/old BaaS publish id and drives the record to `SUCCESS`, so the failure is silent.

The `execute_release_phase` call is the **only** step that talks to the BaaS layer to re-point the live online bot/binding at the version being published (via `upgrade_async`). Skipping it means production is never actually updated.

## Is this a regression?

Yes. Introduced by the durable-task-pipeline refactor (PR #105, `ea8f8208`). Before it, `process()` at `VALIDATING` called `_execute_release_phase` **unconditionally** — there was no `is_online_release_recorded` gate. (The `ext.publish.online` *field* pre-dates #105 as a data record of the BaaS publish id, but it was never read to skip the release.)

## Why not just remove the guard

The guard is load-bearing elsewhere and must stay:
- **Crash-resume idempotency** (`tasks.py`): the online release runs *within* `ONLINE_PUB` with no self-advance, so a lease-expiry / pod-crash re-run of the task can only be told apart from a fresh run by the marker. Without it, the re-run creates a **second BaaS bot**.
- **Retry strategy** (`publish_flow_service.py:884`): the marker decides BaaS-restart (poll failed after create) vs. re-running the release work (create itself failed).

The marker is correct *within a single publish lifecycle*. The defect is that it **leaks across lifecycles** because rollback re-uses the same record.

## Proposed fix

At the rollback boundary — where `rollback_publish` resets the current record to `DRAFT` — clear the stale online release refs (`ext.publish.online` and `ext.binding.online`) so the record re-enters DRAFT as a genuinely un-released version. Both are cleared together for consistency (a record should not be "published on one and not the other"). The next re-publish then runs `execute_release_phase`, and `_should_upgrade_online` (previous record `SUCCESS`) takes the upgrade path, re-pointing the shared online bot/binding at the version. This is the one seam that only fires on rollback (a fresh record starts at DRAFT with empty ext, and `ONLINE_PUB` re-entered via retry must keep the marker).

totalfrank:
## What & why

Fixes #255.

After a service-bot **rollback**, re-driving the rolled-back version's online publish silently skipped the online release: the publish record advanced to `SUCCESS` while the live BaaS bot was never upgraded — recorded state and the deployed bot diverged, with no error.

**Root cause.** `rollback_publish` moves the current version's record back to `DRAFT` but left its `ext.publish.online` (BaaS publish id) and `ext.binding.online` (device binding) markers in place. `is_online_release_recorded()` reads `ext.publish.online` as the crash-resume idempotency guard, so on a later re-publish the `online_release` task saw the stale marker and skipped `execute_release_phase` — the only step that re-points the live online bot/binding at the version via `upgrade_async`.

This is a regression from the durable-task-pipeline refactor (PR #105): before it, `process()` at `VALIDATING` called the release phase unconditionally.

## The fix

When the current record re-enters `DRAFT` during rollback, clear both the online publish id and online binding refs so it becomes a genuinely un-released version. Both are dropped together for consistency (never one without the other).

The guard is deliberately **kept** — it's load-bearing for crash-resume idempotency (a lease-expiry re-run must not create a second BaaS bot) and for retry strategy selection. The bug was that the marker leaked *across* publish lifecycles because rollback reuses the record; clearing it at the rollback→DRAFT seam is the one place that only fires on rollback (a fresh record starts DRAFT-empty, and `ONLINE_PUB` re-entered via retry must keep the marker). The shared binding still lives on the rollback target's record, so the next upgrade re-resolves it from `last_pub_id`.

## Changes

- `publish_rollback_mixin.py`: clear `ext.publish.online` and `ext.binding.online` when resetting the current record to `DRAFT`.
- `test_bot_publish_service.py`: new `test_rollback_publish_clears_online_release_refs` asserting the online refs are cleared while `verify`-stage refs and other ext keys are preserved.

## Testing

`test_bot_publish_service.py`, `test_publish_tasks.py`, `test_publish_flow_service.py` — 233 passed. The new test fails on the pre-fix code and passes with the fix (confirmed it exercises the change).

https://claude.ai/code/session_01PWJ3sZiNMa3r7Xq6CJK4UR

---
_Generated by [Claude Code](https://claude.ai/code/session_01PWJ3sZiNMa3r7Xq6CJK4UR)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/service_bot/services/publish_rollback_mixin.py
+++ b/src/backend/src/agentclaw/community/core/service_bot/services/publish_rollback_mixin.py
@@ -6,6 +6,7 @@ from datetime import datetime
 from typing import TYPE_CHECKING, Callable
 
 from agentclaw.community.core.service_bot.repository.models import PublishStatus
+from agentclaw.community.core.service_bot.types import PublishStage
 from agentclaw.community.log import get_logger
 
 if TYPE_CHECKING:
@@ -158,6 +159,19 @@ class PublishRollbackMixin:
             "rollback_reason": reason,
             "target_publish_id": current_record.last_pub_id,
         }
+        # Clear this version's online release refs before it re-enters DRAFT. The refs
+        # (ext.publish.online = BaaS publish id, ext.binding.online = device binding)
+        # are from this version's earlier go-live; left in place, a later re-publish
+        # would see is_online_release_recorded() == True and skip execute_release_phase
+        # — the upgrade that re-points the shared online bot/binding at this version —
+        # so production would never actually change. Drop both together so the record
+        # is consistently un-published (never one without the other). The shared binding
+        # still lives on the rollback target's record, so the next upgrade re-resolves
+        # it from last_pub_id.
+        for _section in ("publish", "binding"):
+            _refs = current_ext.get(_section)
+            if isinstance(_refs, dict):
+                _refs.pop(PublishStage.ONLINE.value, None)
         self.update_publish_status_with_ext(
             publish_id=publish_id,
             target_status=PublishStatus.DRAFT,
--- a/src/backend/tests/community/core/service_bot/services/test_bot_publish_service.py
+++ b/src/backend/tests/community/core/service_bot/services/test_bot_publish_service.py
@@ -2046,6 +2046,62 @@ class TestRollbackPublish:
         assert second_ext["migration_path"] == "/tmp/build"
         assert second_ext["rollback_restored_from"] == 3
 
+    @pytest.mark.asyncio
+    async def test_rollback_publish_clears_online_release_refs(self):
+        """回滚后当前版本回到 DRAFT，应清除其线上发布/绑定引用。
+
+        否则再次发布时 is_online_release_recorded() 会因残留的 ext.publish.online
+        判定为已发布，从而跳过 execute_release_phase（不再执行 upgrade 把共享的线上
+        bot/binding 指向本版本）。清除后，重新发布才会真正重新执行 upgrade。
+        """
+        mock_repo = Mock()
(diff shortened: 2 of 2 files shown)


FreddieSun (approved):
lgtm

PR #204, opened 2026-07-15
feat(bcs): session collection — collect/uncollect/collected list

zh-bupt:
## Summary

Implements session collection for BCS: a bot (or a human acting through an owned bot) can collect/uncollect a session and list the sessions a given bot has collected within a group. Squashed from the `feat/session-collection2` feature branch (15 commits → 1); spec/plan design docs intentionally excluded from this commit.

## Changes

- **service-api**: add `collect` / `uncollect` / `list_collected_by_group` to the session repo port + application service, with default impls.
- **session application service**: `collect` / `uncollect` / `list_collected_by_group` orchestration.
- **stores**: memory + MySQL `collect`/`uncollect`/`list_collected_by_group`. MySQL `collect` idempotency via **SELECT-before-UPDATE** (changed-rows safe).
- **migrations**: `collected` column — MySQL migration `004_add_session_collection.sql`, plus SQLite path for both fresh and existing DBs (`ensure_sqlite_session_collected_column` repair on legacy tables, version-4 migration records progress).
- **HTTP**: collected-list filter (`?collected=true&participant=<bot>`) on `GET /groups/{id}/sessions`, plus `POST /sessions/{sid}/collect` and `DELETE /sessions/{sid}/collect`. Router wiring included.
- **tests**: session repo contract harness assertions, HTTP contract tests covering all three endpoints + idempotency + human-caller ownership/`participant` enforcement, and e2e coverage (human-caller uncollect via `?participant=` query string).

## API surface (frontend-facing)

| Method | Path | Purpose |
|---|---|---|
| `POST` | `/sessions/{sid}/collect` | Collect a session (`participant` in JSON body; required for human callers) |
| `DELETE` | `/sessions/{sid}/collect` | Uncollect (`participant` via query string; required for human callers) |
| `GET` | `/groups/{id}/sessions?collected=true&participant=<bot>` | List sessions a bot has collected in a group |

Collection is a `(session_id, bot_uuid)` pair. Both collect/uncollect are **idempotent**.

## Notes

- The `sqlite_migration_count` helper is `#[allow(dead_code)]`-annotated: it is genuinely used by `bcs-admin`, but the function lives in `migrations.rs` which other test crates pull in via `#[path=...]`, triggering a dead_code warning in those compile contexts.
- Design spec and implementation plan docs are kept as local untracked files and are **not** in this commit by design.

PR #260, opened 2026-07-20
feat(bcs): order collected session list by collect event + surface collected_at

zh-bupt:
## Summary

The collected session list previously sorted by **session creation time**, with no record of when a collection actually happened. This PR introduces a per-participant `collected_at` timestamp so the list is ordered by **collect event (newest first)** and the event time is surfaced to clients.

Follow-up to #204 (session collection feature), addressing the "query collected sessions ordered by collect time" requirement.

## Semantics

- **collect** — `collected_at = CASE WHEN collected=0 THEN NOW() ELSE collected_at END`: only the first collect (0 → 1) records the time; repeat collects are idempotent and keep the original event time.
- **uncollect** — clears `collected_at`, so a later re-collect records a fresh time.
- **list** — `ORDER BY COALESCE(sp.collected_at, s.gmt_create) DESC`: sessions collected before this migration ran fall back to creation time.

## Changes

- **domain**: `Session` gains `collected_at: Option<u64>` (serde `skip_serializing_if = "Option::is_none"`); populated only by the collected-list store context.
- **mysql store**: `select_cols_prefixed` includes `collected_at_ms`; `row_to_session` reads it (opt — absent → `None` on non-join queries); `collect`/`uncollect`/`list` updated for the timestamp + `COALESCE` ordering.
- **memory store**: `collected` set → `HashMap<(session_id, bot_uuid), u64>`; `collect` is `or_insert` (no refresh); list fills `collected_at` and sorts by it desc.
- **migrations**: MySQL `005_add_session_collection_timestamp.sql` (standard `ADD COLUMN` — no MariaDB-only `IF NOT EXISTS`); SQLite fresh DDL + version-5 migration + ensure-repair adds both `collected` and `collected_at` to legacy DBs.
- **tests**: HTTP contract mock tracks `collected_at` via a monotonic counter; new `collected_list_ordered_by_collect_event_desc` test; contract harness asserts `collected_at` is `Some`; ~23 `Session` literals updated; SQLite migration assertions 4 → 5.

## Verification

- `bcs-http` collection contract: **8 passed** (incl. new ordering test)
- `bcs-session-store`: 11 unit + 2 conformance passed
- `bcs` migrations: 8 passed (incl. new `collected_at` column assertion)
- `bcs-admin` migrate: 15 passed
- `bcs-organization-store` conformance: no warnings (prior `sqlite_migration_count` dead-code fix from #204 holds)
- Full workspace `--tests` build clean

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/tests/cli_service_integration.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/tests/cli_service_integration.rs
@@ -98,6 +98,7 @@ impl SessionManagementService for MockServiceSessions {
             updated_at: 1_700_000_000_000,
             completed_at: None,
             meta: cmd.params.meta.clone(),
+            collected_at: None,
         };
         *self.last.lock().await = Some(session.clone());
         Ok(CreateOrReactivateOutcome { session, created: true })
--- a/src/bcs/crates/adapters/http/bcs-http/tests/cli_session_integration.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/tests/cli_session_integration.rs
@@ -67,6 +67,7 @@ impl SessionManagementService for MockSessions {
             updated_at: 1_700_000_000_000,
             completed_at: None,
             meta: cmd.params.meta.clone(),
+            collected_at: None,
         };
         *self.last_created.lock().await = Some(session.clone());
         Ok(CreateOrReactivateOutcome { session, created: true })
--- a/src/bcs/crates/adapters/http/bcs-http/tests/group_messages_contract.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/tests/group_messages_contract.rs
@@ -373,6 +373,7 @@ fn test_session(session_id: &str, group_id: &str, participants: Vec<Participant>
         meta: None,
         current_msg_seq: 0,
         participant_join_seq: None,
+        collected_at: None,
     }
 }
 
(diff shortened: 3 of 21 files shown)


zh-bupt:
Review feedback addressed (amended into the commit):

**[high] MySQL DOUBLE → i64 decode failure** — Root cause confirmed: `UNIX_TIMESTAMP(sp.collected_at)` on a `datetime(3)` column returns a `DOUBLE`, which `db_get_column_opt::<i64>` cannot decode (`as_i64()` returns `None` for `F64` → `Conversion` error). Because `collected_at_ms` had been added to the shared `select_cols_prefixed`, this also broke **ordinary `list_by_group`** on MySQL/OceanBase — every row failed `row_to_session` and got dropped, hence the e2e `✗ session list did not contain created session`. (SQLite was unaffected since `strftime(...)` already yields `INTEGER`.)

Fix:
- Removed `collected_at_ms` from the shared `select_cols_prefixed`; ordinary `list_by_group` no longer selects it (restores prior behavior).
- The collected-list query builds a dedicated column list that **CASTs** the timestamp to `SIGNED` (MySQL) / keeps `INTEGER` (SQLite), so the value always decodes to `i64`.
- `row_to_session` now reads `collected_at_ms` **tolerantly** (absent or non-integer → `None`); a best-effort timestamp never fails the whole row.

**[medium] Redundant SQLite index** — `idx_collected_at (env, group_id, bot_uuid, collected, collected_at)` covers `idx_collected` as a prefix. Removed `idx_collected` from the SQLite ensure-repair; the MySQL counterpart (migration 004) is left in place rather than dropped here, since `DROP INDEX IF EXISTS` is not standard MySQL/OceanBase and connection-pooled multi-statement DDL is fragile — noted in the 005 migration comment.

Verification: SQLite-side tests green (8 http contract incl. ordering, session-store, migrations, bcs-admin). MySQL path exercised by CI e2e on re-run.

questchaos (approved):
LGTM

PR #270, opened 2026-07-20
feat(bcs): surface per-session collected state on the session list

zh-bupt:
## Summary

The session list previously returned no indication of whether the caller's participant had collected each session — clients had to issue a separate `collected=true` query to know. This PR surfaces per-session collected state directly on the list response, and includes the underlying fix that makes `collected_at` actually populate on MySQL.

Supersedes #268 (the standalone COALESCE fix is folded in here).

## Session list behavior

| Request | Per-item fields |
|---|---|
| `GET /groups/{id}/sessions?participant=<bot>` | `collected` (bool); `collected_at` (epoch ms) present only when collected=true |
| `GET /groups/{id}/sessions` (no participant) | neither field added — collected is per-(session, bot) and has no value without a participant |
| `GET /groups/{id}/sessions?collected=true&participant=<bot>` | `collected: true` + `collected_at` on every item (all are collected) |

## Implementation

- **port**: `SessionRepoPort::collected_at_map(session_ids, bot_uuid) -> Vec<(sid, collected_at_ms)>` with a default empty impl. MySQL implements a single `SELECT session_id, collected_at FROM ... WHERE bot_uuid=? AND collected=1 AND session_id IN (...)` (CAST to SIGNED/INTEGER, same DOUBLE-decode reason as the collected-list query); memory filters its HashMap.
- **application**: `SessionManagementService::collected_at_map` forwarding default + `SessionService` delegating to the repo.
- **http**: after building ordinary list items, if `participant` is set, one batched `collected_at_map` call injects `collected`/`collected_at` per item. The `collected=true` branch injects `collected: true` on each item.
- **collected field**: injected at the HTTP serialization layer; domain `Session` does not gain a `collected` field (only `collected_at`).

## Also fixes: collected_at never written on MySQL

`collect` used `SET collected = 1, collected_at = CASE WHEN collected = 0 THEN NOW() ELSE collected_at END` — dialect-dependent in a silent, data-corrupting way:

- **SQLite** evaluates all SET RHS against the pre-update row → `collected` reads 0 → CASE true → writes `NOW()`. Worked.
- **MySQL/OceanBase** evaluate a single UPDATE's SET left-to-right → `collected` is already 1 when the CASE reads it → CASE always false → `collected_at` stays `NULL` forever.

Replaced with `SET collected = 1, collected_at = COALESCE(collected_at, NOW())`, which expresses first-collect-writes / repeat-collect-keeps idempotency via `collected_at`'s own NULL-ness and is identical across MySQL and SQLite. A comment documents the trap so the CASE form isn't reintroduced.

## Verification

- HTTP contract: 9 passed (incl. new `ordinary_list_with_participant_surfaces_collected_fields`, and `collected=true` branch asserting `collected: true`)
- session-store conformance: collected_at_map matches the list view, empty for a non-collector
- Full workspace `--tests` build clean
- MySQL path exercised by CI e2e


vzvince (approved):
LGTM

PR #283, opened 2026-07-21
feat(engine): add session favorites and harden session recovery

evilcoda:
## Summary

- Add an engine-agnostic SQLite-backed session favorites API, including persistent favorite records, listing with the standard session response shape, and best-effort cleanup when a session is deleted.
- Preserve private-chat continuity by resolving saved session keys before pagination while retaining the legacy first-100 fallback for engines that do not yet implement exact filtering.
- Surface unavailable OpenClaw session gateways as retryable service-unavailable responses instead of internal errors.

## Impact

- Frontend integration is additive: existing private-chat requests and responses remain unchanged.
- OpenClaw and Hermes can accurately recover older saved sessions; engines without exact-key filtering keep their prior lookup behavior.

## Validation

- `62 passed`: session router, session favorites router, and session favorite repository tests; targeted coverage `96%`.
- `79 passed`: OpenClaw session adapter and session port tests.
- `ruff check` for all changed engine source and tests.

### DIFF
--- a/src/engine/docs/session-favorites-api.zh-CN.md
+++ b/src/engine/docs/session-favorites-api.zh-CN.md
@@ -1,0 +1,156 @@
+# 会话收藏接口
+
+## 适用范围
+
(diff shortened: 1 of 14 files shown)


FreddieSun (approved):
LGTM

PR #285, opened 2026-07-21
perf(aicoding): speed up NAS file tree scan

maicaisongxiaocong:
## Summary

- replace the AICoding file-tree `os.walk` plus per-file `stat` scan with GNU `find`
- preserve the existing full-tree request contract and filtering behavior
- keep the `size` field while leaving it unset (`null`) to avoid NAS metadata lookups
- emit and parse NUL-delimited type/path pairs so whitespace and newlines in paths remain safe
- do not follow symlinks; expose symlinks as non-directory nodes

## Behavior

- `.git` and `node_modules` directories are pruned at every level
- root `skills` and `.repos` directories are pruned before traversal
- project-local `skills` and `.repos` remain visible
- directories remain sorted before files, then alphabetically

## Tests

- `142 passed, 1 warning`
- `git diff --check`
- Python `compileall`

The warning is the existing Starlette/httpx deprecation warning.

### DIFF
--- a/src/engine/src/engine/community/api/aicoding_sessions/tests/test_router.py
+++ b/src/engine/src/engine/community/api/aicoding_sessions/tests/test_router.py
@@ -274,6 +274,26 @@ def test_file_tree_success(client, workspace_svc):
     assert body["tree"][0]["children"][0]["name"] == "README.md"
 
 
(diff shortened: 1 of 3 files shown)


FreddieSun (approved):
lgtm

PR #287, opened 2026-07-21
fix(bcn): finalize queued replies from agent events

RaymondHX:
## Summary

- bind each queued OpenClaw Agent run ID to the original BCS run ID through `onAgentRunStart`
- keep queued `chat.send` contexts after dispatcher settlement and derive delta/final/error delivery from Agent events
- prevent same-session events from being claimed through an unsafe session-only fallback
- send `NO_REPLY` only after a successful terminal lifecycle with no visible assistant text
- add terminal error/timeout cleanup and bump the BCN plugin package to `1.0.18`

## Root cause

`dispatchReplyWithBufferedBlockDispatcher` may settle while a same-session message is still queued. Treating dispatcher settlement as terminal finalized the BCS run too early, emitted `NO_REPLY`, and removed the context before the actual Agent run started.

## Validation

- `npm run lint`
- `npm run prepublishOnly`
- `npm run test-local` (50 passing)
- `npm run cov` (76.57% statements, 60.84% branches)
- `bash scripts/test_bcn_plugin_source.sh`
- `bash scripts/test_singlebox_default_mode.sh`

The repository pre-push Rust gate was also attempted: 2,356 tests passed and 27 unrelated BCS/CLI process-start tests failed. The branch was pushed with `--no-verify` as requested.

### DIFF
--- a/src/bcs/crates/plugins/openclaw-channel-bcn/README.md
+++ b/src/bcs/crates/plugins/openclaw-channel-bcn/README.md
@@ -48,7 +48,7 @@ Select with the flag or env var (flag wins):
 BCN_PLUGIN_SOURCE=npm ./scripts/singlebox.sh
 
 # pin a version in npm mode (default: latest)
-BCN_PLUGIN_SOURCE=npm BCN_PLUGIN_VERSION=1.0.16 ./scripts/singlebox.sh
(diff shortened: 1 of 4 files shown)


vzvince (dismissed):

questchaos (approved):
LGTM

PR #292, opened 2026-07-21
ci: derive Plugin API coverage from offline evidence

FreddieSun:
## Summary

- derive Plugin API coverage from coverage.py JSON plus checked-in AST symbols
- validate manifest evidence against real Plugin Protocol implementations under plugins/
- remove runtime coverage proxy injection from Core/business DI and add an architecture guard
- keep Router evidence as runtime middleware hits

This is the foundation for refactoring #124-#129. Those PRs will be stacked on this branch so their diffs contain only module-specific coverage work.

## Verification

- 59 focused tests passed
- Backend full suite: 8,342 passed
- singlebox acceptance: 32 passed
- singlebox coverage gate passed
- manifest validation, shell gate, Ruff, bash syntax and diff checks passed

## Notes

The first local gate attempt selected Python 3.14 for BCSFuse; its pinned faiss-cpu wheel supports up to CPython 3.12. Recreating the generated BCSFuse venv with Python 3.12 allowed the full unchanged pre-push gate to pass.


totalfrank (dismissed):
LGTM.

carolynli (approved):
LGTM.

xianmuyq (approved):

Issue #197, opened 2026-07-15
Publish service is not idempotent under crash/retry: BaaS publish ids minted before any persisted intent, approvals never re-driven, restart/offline/rollback/eval unguarded


totalfrank:
# Summary

The service-bot publish pipeline (verify release, online release, upgrade, restart, scale, offline, rollback, eval, approval) is **not crash-safe**. The process can die at any point, and the durable task queue is explicitly **at-least-once** (lease reclaim → re-run; `task_queue/services/registry.py:38-41`), so every handler must be idempotent — but almost none of the operations actually are.

The recurring anti-pattern: **every operation asks BaaS to mint a publish workflow id first, and only persists that id (and its follow-up obligations, like approve) in separate, later statements.** A crash in between either loses the id (orphan/duplicate BaaS bots, workflows stuck unapproved forever) or strands the record with no driver. `is_online_release_recorded` (`publish_flow/tasks.py:219`, `publish_flow_service.py:959-970`) looks like an idempotency guard but only covers one narrow window of one operation — see below.

Goal of this issue (to be executed via SDD):
1. Real idempotency/crash-safety for **all** publish-service operations.
2. Clean up the worst code-quality offenders in this pipeline (multiple 100–190-line methods).

# Scope decisions ([discussion](https://github.com/inclusionAI/Avernet/issues/197#issuecomment-4978735223))

- **Concurrent double-claim race (lease-expiry overlap / duplicate enqueue): descoped.** The lease auto-renew mechanism makes it very unlikely; if it materializes it gets fixed on the infra side. This issue targets **sequential crash-resume idempotency** only.
- **Operation ledger: dedicated table**, not ext keys.
- **Atomic advance+enqueue + sweep/reconciler: postponed** → tracked in #198, low priority.

# What exists today (and is sound)

- Status transitions are a conditional-UPDATE CAS on `status` (`plugins/bot_publish_repository.py:308-364`); `update_status_with_ext` writes status+ext in **one** statement. The user-driven `process()` advance points (DRAFT→BUILDING, VALIDATING→ONLINE_PUB) are double-submit-safe via this CAS (`publish_flow_service.py:323-353`).
- The build phase is genuinely re-run-safe: versioned rsync target / stage-scoped OSS keys, artifact pointers merged under the BUILDING→BUILT CAS (`publish_flow/build_stage.py:61-149`). This is the model the rest of the pipeline does **not** follow.
[...]

PR #297, opened 2026-07-21
refactor(publish): make RELEASED offline an idempotent no-op

totalfrank:
## What

`offline_publish` previously re-enqueued a durable offline destroy whenever it was
called on an already-`RELEASED` record (the "crash-resume" block added in #197).
This replaces that with a plain **idempotent no-op**: the `RELEASED` early-return
stays, but the re-enqueue is dropped.

```python
if current_status == PublishStatus.RELEASED:
    # idempotent no-op: return success, do not flip status, do not enqueue
    return {"success": True, "bot_destroyed": False, ...}
```

## Why

The re-enqueue was only ever exercised on a **hard crash**. Tracing the callers:

- The only path that re-drives `offline_publish` is the durable `APPROVAL_TRIGGER_TASK`,
  and only via a lease reclaim after an abrupt process death.
- A **soft** failure (an exception between the `SUCCESS→RELEASED` flip and
  `enqueue_offline_destroy`) is caught and swallowed in `_trigger_offline`, so the
  handler returns `Complete()` and the task is never retried — the re-enqueue never
  ran for that case anyway.

So the re-enqueue guarded only the narrow hard-crash window while adding a durable
task on every RELEASED call. Keeping the early-return preserves the useful properties
(a duplicate submit or a durable-task re-run lands cleanly instead of falling through
to `_resolve_offline_stage` and raising); dropping the re-enqueue removes machinery
that didn't robustly cover the leak it was meant to.

Proper recovery of the flip-then-crash window (record `RELEASED` but destroy never
enqueued) is deliberately **out of scope** here and will be handled separately as part
of the broader offline idempotency fix — e.g. not swallowing in `_trigger_offline`, or
a reconciler that sweeps `RELEASED` records whose stage binding is not yet released.

## Tests

- Renamed `test_offline_publish_released_record_reenqueues_destroy` →
  `test_offline_publish_released_record_is_noop`; it now asserts no status flip and no
  `enqueue_offline_destroy` call.
- `test_bot_publish_service.py` + `test_publish_approval_service.py`: 132 passed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BQp2ePgrPn113aaDaneYwi)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/service_bot/services/bot_publish_service.py
+++ b/src/backend/src/agentclaw/community/core/service_bot/services/bot_publish_service.py
@@ -937,25 +937,21 @@ class BotPublishService(PublishRollbackMixin):
         # Step 2: 根据状态判断 stage
         current_status = publish_record.status
 
-        # (#197) Crash-resume: a prior online offline already flipped
-        # SUCCESS→RELEASED but may have crashed before the durable destroy was
-        # enqueued (the flip and the enqueue are not atomic). Re-enqueue so the
-        # online bot is never leaked. execute_offline_destroy short-circuits when
-        # the destroy already COMPLETED, so this re-enqueue is a true no-op then.
+        # RELEASED is terminal — the offline already ran. Return an idempotent
+        # no-op so a duplicate submit or a durable-task re-run lands cleanly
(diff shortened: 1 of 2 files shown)


FreddieSun (approved):
LGTM

PR #305, opened 2026-07-22
fix(harness): map gmt_create/gmt_modified in HarnessPatch _row_to_domain

wienyang:
_row_to_domain omitted gmt_create/gmt_modified when converting the ORM row to PatchDefinition, so the dataclass fell back to its datetime.utcnow default. API consumers (e.g. /diagnose/dim-history) therefore received a freshly-minted UTC timestamp instead of the DB row's real creation time. Pass the ORM columns through.


FreddieSun (approved):
LGTM

PR #312, opened 2026-07-22
feat(bcs): add recoverable state machine judging

questchaos:
## Summary

Add a recoverable `judging` phase for state machine bot nodes so BaaS can receive the bot response immediately while judge evaluation continues asynchronously.

## Changes

- Add the `judging` node status and expose it in the admin panel.
- Run judge evaluation in a background worker after the bot returns its final output.
- Persist judge claim tokens, lease expiration, and decisions for safe recovery.
- Fence stale workers from writing judge results.
- Allow interrupted post-judge processing to resume without calling the LLM again.
- Publish frontend bot events asynchronously with a bounded timeout.
- Add MySQL and SQLite schema migrations and indexes.
- Preserve existing behavior for bots that do not use the BaaS judging flow.
- Update integration documentation and test configuration.

## Problem Fixed

Previously, a state machine node could remain `running` indefinitely after the LLM judge returned successfully. Failures or stale workers during post-judge processing could leave the node unclaimable, while frontend event publication could delay the transition into judge processing.

The new flow transitions the node to `judging`, returns control to BaaS immediately, and completes judge processing through a leased, recoverable background workflow.

## Database Changes

The migration adds:

- `judge_claim_token`
- `judge_lease_until_ms`
- `judge_decision_json`
- An index for locating claimable judging nodes

### DIFF
--- a/src/bcs/crates/services/bcs-collaboration-runtime/src/runtime.rs
+++ b/src/bcs/crates/services/bcs-collaboration-runtime/src/runtime.rs
@@ -19,25 +19,23 @@ use bcs_protocol::{
 };
 use bcs_service_api::{
     BotDeliveryCommand, BotDeliveryKind, BotDeliveryPort, BotDeliveryTarget,
-    BotRegistryCoreService,
-    CancelStateMachineRunCommand,
-    ChatEventState, CollaborationDefinitionRecord, CollaborationEventRepoPort,
-    CollaborationRuntimeError, CollaborationRuntimeService, ConfigureGroupRuntimeCommand,
-    ConfigureGroupRuntimeOutcome, CreateOrReactivateCommand, DefinitionYamlSource,
-    FrontendDeliveryCommand, FrontendDeliveryKind,
-    FrontendDeliveryPort, FrontendDeliveryTarget, GroupCoreService, GroupRuntimeBindingRepoPort,
-    GroupCollaborationDefinitionView, HandleBotTerminalEventCommand,
-    HandleBotTerminalEventOutcome, MAX_COLLABORATION_DEFINITION_YAML_BYTES,
-    MessageLogContent, MessageLogEventType, MessageLogMode, MessageLogStatus,
-    MESSAGE_LOG_SCHEMA_VERSION, MSG_LOG_TARGET, message_log_json,
-    NewSessionParams, PatchGroupCollaborationDefinitionCommand, RunFallbackDelivery,
+    BotRegistryCoreService, CancelStateMachineRunCommand, ChatEventState,
+    CollaborationDefinitionRecord, CollaborationEventRepoPort, CollaborationRuntimeError,
+    CollaborationRuntimeService, ConfigureGroupRuntimeCommand, ConfigureGroupRuntimeOutcome,
+    CreateOrReactivateCommand, DefinitionYamlSource, FrontendDeliveryCommand, FrontendDeliveryKind,
+    FrontendDeliveryPort, FrontendDeliveryTarget, GroupCollaborationDefinitionView,
+    GroupCoreService, GroupRuntimeBindingRepoPort, HandleBotTerminalEventCommand,
+    HandleBotTerminalEventOutcome, JudgeArtifact, JudgeDecision, JudgeEvaluatorPort, JudgeRequest,
+    MAX_COLLABORATION_DEFINITION_YAML_BYTES, MESSAGE_LOG_SCHEMA_VERSION, MSG_LOG_TARGET,
+    MessageLogContent, MessageLogEventType, MessageLogMode, MessageLogStatus, NewSessionParams,
+    PatchGroupCollaborationDefinitionCommand, RunFallbackDelivery, ServiceError,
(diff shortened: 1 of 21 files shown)


src/bcs/crates/services/bcs-collaboration-store/src/lib.rs
```diff
@@ -1539,67 +1922,395 @@ impl StateMachineRunRepoPort for MySqlCollaborationStore {
...
+        lease_until_ms: u64,
     ) -> ServiceResult<bool> {
         let sql = format!(
             "UPDATE bcs_state_machine_node_runs \
-             SET status = 'retry_scheduled', attempt = ?, \
-                 delivery_request_id = NULL, bot_delivery_run_id = NULL, \
-                 artifact_text = NULL, error_message = NULL, started_at_ms = NULL, \
-                 completed_at_ms = NULL, timeout_deadline_ms = NULL, \
-                 {} \
+             SET judge_lease_until_ms = ?, {} \
```
vzvince, line 1949: This renewal CAS validates the token and status but not whether the existing lease is still live. `complete_judging_node_attempt` and `fail_judging_node_attempt` have the same omission, as do the memory-store implementations. A paused worker can resume after expiry and renew/complete/fail before another worker replaces its token, so expiry alone does not fence it. Pass `now_ms` (preferably DB time) to every claim-authorized mutation and require `judge_lease_until_ms > now`; renewal must not resurrect an expired lease.

src/bcs/crates/services/bcs-collaboration-runtime/src/runtime.rs
```diff
@@ -1344,6 +1364,332 @@ impl CollaborationRuntime {
...
+                        &compiled,
+                        &run,
+                        &node.node_id,
+                        &outcome,
+                        completed_at,
+                    )
+                    .await?;
+                    self.dispatch_ready_targets(&compiled, &group, &run, &node.node_id, &outcome)
+                        .await?;
+                    self.complete_run_if_done(&compiled, &run).await?;
```
vzvince, line 1592: `complete_run_if_done` persists the run as `completed` before session completion/callback dispatch is durable, and it logs/swallow session persistence errors. Claim discovery only selects parent runs still marked `running`, then this continuation clears the judge recovery fields. A crash or session-store failure can therefore leave a running session and undispatched callback with no reclaimable work. Keep a durable continuation checkpoint/outbox claimable until all required post-judge effects complete, and propagate persistence failures.

src/bcs/crates/services/bcs-collaboration-runtime/src/runtime.rs
```diff
@@ -1624,6 +1970,27 @@ impl CollaborationRuntimeService for CollaborationRuntime {
...
+        let duplicate_judging_final = node.as_ref().is_some_and(|node| {
+            matches!(cmd.state, ChatEventState::Final)
+                && run.status == StateMachineRunStatus::Running
+                && node.status == StateMachineNodeStatus::Judging
+                && node.attempt == correlation.attempt
+                && node.delivery_request_id.as_deref()
+                    == Some(correlation.delivery_request_id.as_str())
+                && correlation.assignee_bot_id == cmd.bot_id
+        });
+        if duplicate_judging_final {
```
vzvince, line 1982: This duplicate-final path returns before `publish_state_machine_bot_event`. If the first callback persisted `judging` but crashed before publication—or the detached publication failed—the provider retry is acknowledged without re-enqueuing the bot output, so the frontend event is permanently lost. Persist a publication checkpoint/outbox, or idempotently re-enqueue publication from this path until delivery is acknowledged.

src/bcs/crates/services/bcs-collaboration-runtime/src/runtime.rs
```diff
@@ -1715,109 +2073,93 @@ impl CollaborationRuntimeService for CollaborationRuntime {
...
-                        &text,
-                    )
-                    .await?;
-                match evaluation {
-                    JudgeEvaluationResult::Outcome(outcome) => {
-                        let completed_at = bcs_protocol::now_ms();
-                        if self
-                            .runs
-                            .complete_node_attempt(
+                if compiled_node_has_judge(&compiled, &correlation.node_id) {
```
vzvince, line 2076: This queues every judged terminal event asynchronously, not only BaaS/provider callbacks. WebSocket dispatch and provider events call the same service method, and `HandleBotTerminalEventCommand` has no source/execution-policy field. That changes non-BaaS judging semantics despite the PR's compatibility claim. Add an explicit source/policy and apply this path only where intended, or document and compatibility-test the global contract change.

vzvince (commented):
The core `running -> judging` transition and persisted-decision recovery are well structured, and the focused tests plus CI are green. I found four recovery/compatibility gaps that should be addressed before merge: expired workers are not fully fenced, terminalizing the run can disable post-judge recovery before session completion is durable, duplicate provider callbacks cannot recover a lost frontend publication, and the asynchronous behavior also affects WebSocket/non-BaaS callers. Inline comments include concrete failure paths and suggested fixes.

vzvince (approved):
LGTM

questchaos:
https://github.com/inclusionAI/Avernet/pull/344

PR #317, opened 2026-07-22
feat(backend): harden caller identity switches

yhzhang35:


src/backend/src/agentclaw/community/adapters/http/token_exchange/router.py
```diff
@@ -152,6 +155,20 @@ async def get_iam_token(
             status_code=400,
             headers=headers,
         )
+    if is_test_exchange and not bot_id:
```
totalfrank, line 157: Do not keep growing domain logic in http layer. Do a small refactoring to move all this logic to service layer.

totalfrank (commented):
Move domain logic to service layer please.

totalfrank (approved):
LGTM.

PR #325, opened 2026-07-22
feat(paas): add optional mode param to update_outbound_operation_rule

xxxxpenny:
Propagate the mode query param from the router down through facade, ABC, protocols, and all service implementations to the ArcaPaasService sandbox.update_outbound_rule call. Defaults to REPLACE when not provided.

- Router: add mode query param (OutBoundOperationRuleUpdatedMode | None)
- Facade/ABC/Protocol: add mode param to signature and pass through
- All impls (arca/teclaw/k8s/local/mock/sigma/poolab/standalone): add mode param
- ArcaPaasService: use APPEND when mode is explicitly set, otherwise REPLACE
- Update test_facade and test_facade_coverage assertions for new param


pfmiles (approved):
ok

PR #331, opened 2026-07-22
fix(harness): stop softening LLM-disabled scans into a high health_score

wienyang:
When LLM service is down, every diagnostic is skipped and only LLM01 ("LLM 服务未启用", INFO) findings are produced. The penalty fallback in _compute_score counted these as content defects: 6×INFO = -12 → score 88, misreporting an un-scanned bot as healthy.

LLM01 is not a content defect — it means the diagnostic never ran. Skip it in the penalty branch; when *all* findings are LLM01 (no real diagnosis ran), return 0 instead of a dressed-up 88. Real content findings mixed with LLM01 are penalised exactly as before (LLM01 neither adds penalty nor softens the score).

No None introduced, so _compute_grade's >= comparisons are unaffected; schema/DB/frontend contract unchanged.

Tests:
- test_llm_disabled_findings_not_scored_as_content: 6 LLM01 → 0 (was 88)
- test_llm_disabled_ignored_when_real_findings_present: LLM01 + CRITICAL → 80

### DIFF
--- a/src/backend/src/agentclaw/community/core/harness/services/content_scanner.py
+++ b/src/backend/src/agentclaw/community/core/harness/services/content_scanner.py
@@ -277,13 +277,24 @@ class ContentScanner:
         if file_avg_scores:
             return int(sum(file_avg_scores) / len(file_avg_scores))
 
-        # Fallback: penalty-based scoring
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #340, opened 2026-07-22
fix: move heartbeat after converter to prevent idle timeout starvation

xxxxpenny:
The heartbeat timer was placed before the SSE converter, timing out based on upstream chunk arrival. When upstream sends high-frequency tool update/progress events that the converter filters to None, the timer keeps getting reset (chunks arrive), but no SSE bytes are written downstream. This starves the heartbeat and the BCS socket goes silent for 120s, triggering an idle timeout.

Move heartbeat logic to after the converter, so it fires based on actual SSE byte output intervals. Extract a reusable with_sse_heartbeat() wrapper in api/sse and apply it in both bcn_router and message_router. Remove the old _with_heartbeat from _runner, which returned StreamChunk(type=heartbeat) before conversion.

Changes:
- Add api/sse/_heartbeat.py with with_sse_heartbeat() + HEARTBEAT_SSE
- Remove _with_heartbeat and _StreamEnd from _runner.py
- Wrap _convert_stream() with with_sse_heartbeat in bcn_router
- Wrap _convert_stream() with with_sse_heartbeat in message_router
- Replace TestWithHeartbeat with TestWithSseHeartbeat (7 tests, 100% cov)

### DIFF
--- a/src/baas/src/secbaas/community/adapters/web/routers/bcn_downlink/bcn_router.py
+++ b/src/baas/src/secbaas/community/adapters/web/routers/bcn_downlink/bcn_router.py
@@ -54,7 +54,13 @@ from secbaas.community.api.bcn import (
     ChatSendInput,
 )
 from secbaas.community.api.bot_runtime import BotBindingNotFoundError
-from secbaas.community.api.sse import SseConverterFactory, SseEvent, StreamChunk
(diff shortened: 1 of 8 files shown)


pfmiles (approved):
ok

PR #348, opened 2026-07-22
fix(bcs): expose async judge progress and cover provider timeout

questchaos:
Persist provider artifacts before asynchronous judging and derive extensible running-node sub-statuses for the API and panel.

Exercise the callback-cancellation regression with a local HTTP Provider and blocking Judge in the instrumented singlebox E2E coverage suite.


RaymondHX (approved):
LGTM

PR #354, opened 2026-07-22
fix: avoid reusing incomplete singlebox bot sessions

FreddieSun:
## Summary

- Only reuse a persisted BCS bot identity in singlebox when the session has a non-empty token and matches the active BCS URL.
- Ignore persisted plugin sessions without a reconnect token, while preserving token-only sessions whose bot UUID is not assigned yet.
- Add regressions for incomplete and stale session files.

## Root cause

A stale or incomplete `.bcs/session.json` could supply only `bot_uuid`. The launcher wrote it as `channels.bcs.botId`, then every fresh bot attempted `bot.connect` with that same ID and no token, causing a BCS registration conflict.

## Validation

- `bash scripts/test_bcn_plugin_source.sh`
- BCN plugin lint, tests (52 passed), and build
- Real `bash scripts/ci/singlebox_coverage.sh`
  - Backend acceptance: 40 passed
  - BCS E2E: 454 passed, 0 failed
  - BCS coverage: line 42.76%, method 38.47%, router/CLI 100%
- CI artifact verification passed

> Note: the repository pre-push Rust nextest gate times out in existing WebSocket/human-regression integration tests under default parallel execution. The real singlebox coverage flow above passed; this PR does not modify Rust test code.

### DIFF
--- a/scripts/test_bcn_plugin_source.sh
+++ b/scripts/test_bcn_plugin_source.sh
@@ -226,6 +226,59 @@ test_stack_script_has_npm_branch() {
   assert_contains "$src" '[ "$BCN_PLUGIN_SOURCE" != "npm" ]'
 }
 
+test_session_bot_uuid_requires_usable_session() {
+  local tmp; tmp="$(mktemp -d)"
+  local funcs="${tmp}/stack-session-functions.sh"
+  local profile_root="${tmp}/profiles"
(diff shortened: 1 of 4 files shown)


RaymondHX (approved):
LGTM

totalfrank (approved):
LGTM.

PR #355, opened 2026-07-22
fix(bcs-cli): select compiled URL by runtime environment

vzvince:
## What changed

- Allow distributions to compile pre-release and production BCS defaults into one `bcs-cli` binary.
- Select the compiled default at runtime using the existing environment resolution chain.
- Preserve URL precedence: explicit CLI URL, environment URL, session URL, compiled distribution default, then localhost.
- Keep public builds backward compatible when no compiled defaults are provided.
- Add tests for public and distribution build modes, environment priority, URL selection, and override precedence.

## Why

After internal packaging moved away from the previous BCS source layout, the CLI lost its environment-aware default URL behavior. Some Provider-integrated bots cannot guarantee that `BCS_API_BASE_URL` is present at runtime, so the CLI could incorrectly fall back to `127.0.0.1:21000`.

The fix lets a distribution build one artifact containing both defaults while still choosing pre-release or production only when the binary runs.

## User impact

- `pre` and `prepub` environments select the pre-release compiled default.
- Other or missing runtime environments select the production compiled default.
- Explicit URL environment variables and session configuration continue to override compiled defaults.
- Unconfigured public builds continue to use the local default.

## Validation

- `cargo test --package bcs-cli` in public build mode: 146 passed, 1 existing ignored.
- `cargo test --package bcs-cli` with both compiled defaults: 146 passed, 1 existing ignored.
- `cargo check --package bcs-cli --all-targets` passed.
- `git diff --check` passed.

### DIFF
--- a/src/bcs/crates/tools/bcs-cli/CONTEXT.md
+++ b/src/bcs/crates/tools/bcs-cli/CONTEXT.md
@@ -28,6 +28,13 @@
 ## Configuration
 
 - CLI args, optional config files, and tokens are read at tool entrypoints only.
+- BCS API URL precedence is `--url`, `BCS_API_BASE_URL`, `MOLTIS_BCS_URL`,
+  `$BOT_DATA_DIR/.bcs/session.json`, a runtime-selected compiled distribution
+  default, then the local default.
+- Distributions may compile both `BCS_CLI_DEFAULT_PRE_URL` and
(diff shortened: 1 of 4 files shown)


RaymondHX (approved):
LGTM

carolynli (approved):
LGTM.

PR #360, opened 2026-07-23
fix(service_bot): route online retries through the release path; extract shared deploy atom

totalfrank:
## Problem

PR #341 fixed the online-release **gate** (a rolled-back-then-re-promoted record no longer skips its release), but the same `is_online_release_recorded` predicate was also consulted by **retry** to choose BaaS-restart vs re-running the release. Those are different questions: the retry decision turned on shared-bot *liveness* — the wrong axis for "does the release work still need to run" — so an online retry could take a recovery path that didn't match what actually failed.

Worse, the predicate could not tell "deploy landed" from "deploy failed": a release op is `COMPLETED` at bookkeeping time, and nothing ever recorded the workflow's later failure — so a failed deploy read as live, and a naive "always re-run" retry would have skipped the re-issue and looped FAILED forever.

## Fix

1. **Ledger reflects deploy outcome** — new `fail_by_workflow(publish_id, baas_publish_id, error)` outcome-correction (permits `COMPLETED → FAILED`), applied by `_handle_sync_failure` in all three wait paths. A failed deploy no longer reads as live and no longer supersedes a genuinely live earlier release.
2. **Retry re-route** — an `ONLINE_PUB`-pre-failure retry always re-enqueues `online_release`; the gate + ledger decide run-vs-skip and first-release-vs-upgrade. `_retry_uses_baas_restart` is deleted; only `VALIDATE_PUB` (verify cannot re-enter; symmetry deferred) and `SUCCESS` (release IS current; the gate would skip BaaS) restart. The `ext.retry` poll-redirect flag is written only on restart branches — previously a release-path retry left it set and the poll synced a stale restart workflow (stranding).
3. **Rename** — `is_online_release_recorded` → `is_current_online_deployment`, a gate-only predicate; the skip-if-current check never applies to restart, so a restart always re-deploys via BaaS.
4. **Shared deploy atom** — `acquire_deploy_workflow` (open op → acquire workflow with uniform `BOT_NOT_FOUND` classification → validate ids) now underpins `first_release`, `upgrade_release`, and `execute_restart`. Restart's `BOT_NOT_FOUND` leg becomes a crash-safe recreate: abandon the RESTART op, run a fresh `FIRST_RELEASE` op with a **new** bot + **new** binding (closing the documented orphan window), plus a finalize-sweep for the one crash window the re-delivery cannot revisit.
5. **Production bugs found by the new tests** — `BaasService._post_teclaw` dropped the HTTP response when normalizing errors, so the teclaw path could never classify the server's 404 `{"detail":{"error_code":"BOT_NOT_FOUND"}}`: the gone-bot fallbacks only ever worked for ARCA. Fixed by preserving `.response` on the raised error.

## Cross-publish-boundary DI test suite (new)

`tests/community/e2e/publish_boundary/` — production wiring end-to-end (real app, per-test injector, real `PublishFlowService`/task handlers/teclaw build), with a new **stateful LocalBaas** at the HTTP seam (per-bot workflow timelines, monotonic ids, controllable outcomes, bot deletion) so the real `BaasService` code runs. Durable tasks drained deterministically via `TaskWorker.run_once()`.

Scenarios: L0 full lifecycle; R1–R6 (wait-failure retry re-issues, failed deploy never supersedes, retry×restart interleaving, verify/SUCCESS retries still restart, redelivery idempotency); C1–C5 (upgrade chain, rollback-then-re-promote — the original #5984 stranding asserted dead — restart-always-hits-BaaS, restart recreate on gone bot, kind coexistence).

## Out of scope

The verify release/retry flow is untouched (`VALIDATE_PUB` retries stay on restart); verify↔online symmetry is deferred to a follow-up spec.

## Verification

- [x] Spec / plan / tasks: `src/backend/specs/2026-07-23-online-release-retry-fix/` (all 14 acceptance criteria checked)
- [x] Group reviews (A/B/C): all findings fixed, incl. a crash-window residue in the recreate leg (regression-tested)
- [x] `tests/community/core/service_bot/` + `tests/community/e2e/` + affected trees green (3162 targeted; full backend gate 8633 passed)
- [x] Pre-push gates vs `origin/REL20260723`: SAST ✅, backend ci_test ✅ (100% pass rate, 97.40% changed-line coverage)

https://claude.ai/code/session_01VxbFr1w7mC7v2KdsxMyWjE

### DIFF
--- a/src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/restart_mixin.py
+++ b/src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/restart_mixin.py
@@ -337,11 +337,20 @@ class RestartMixin:
         complete without issuing anything.
 
         The condition is exact: a non-terminal FIRST_RELEASE op for this
-        record/stage whose recorded workflow id IS ``ext.publish.<stage>`` — the
-        ext write demonstrably landed, so completing the op is pure bookkeeping
-        catch-up. ``complete`` is a CAS from ID_RECORDED; anything else no-ops.
-        The returned id lets ``execute_restart`` return that workflow instead of
-        issuing a redundant second deploy (idempotent task redelivery)."""
+        record/stage whose recorded workflow id IS **both** ``ext.publish.<stage>``
+        and ``ext.restart.<stage>``. Requiring the ``restart`` match too is what
+        keeps this specific to a restart-recreate redelivery: ``_recreate_restart_target``
+        writes the same workflow id into publish AND restart, whereas a *normal*
+        online first release (which can also crash after ``record_release_ext``
+        writes ``ext.publish.online`` but before ``complete_operation``, and whose
+        stage is restartable) never touches ``ext.restart`` — so without this
+        clause a restart issued in that window would finalize the release op and
+        return without actually restarting. That release op is finalized by the
+        online_release task's own redelivery, not here.
+
+        ``complete`` is a CAS from ID_RECORDED; anything else no-ops. The returned
+        id lets ``execute_restart`` return that workflow instead of issuing a
+        redundant second deploy (idempotent task redelivery)."""
         op = self._publish_operation_repo.get_latest_by_kind(
             publish_id, str(PublishOperationKind.FIRST_RELEASE), stage_enum.value
         )
@@ -351,6 +360,8 @@ class RestartMixin:
         if op.state in terminal:
             return None
         if op.baas_publish_id != (ext.get("publish") or {}).get(stage_enum.value):
+            return None
+        if op.baas_publish_id != (ext.get("restart") or {}).get(stage_enum.value):
             return None
         logger.info(
             "[PublishFlowService._finalize_dangling_recreate_op] completing "
--- a/src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/restart_mixin.py
+++ b/src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/restart_mixin.py
@@ -6,7 +6,8 @@ from agentclaw.community.core.service_bot.repository.models import (
     PublishStatus,
 )
 from agentclaw.community.core.service_bot.services.publish_flow.operation_runner import (
-    PublishOperationError,
+    TargetBotGoneError,
+    acquire_deploy_workflow,
 )
 from agentclaw.community.core.service_bot.types import PublishStage
 from agentclaw.community.log import get_logger
@@ -187,10 +188,12 @@ class RestartMixin:
         the in-doubt restart workflow instead of issuing a second one.
 
         Re-resolves everything from ``publish_id`` (the durable task carries only
-        ids), then: open a ``restart`` op keyed to the bot, acquire the workflow
-        (upgrade; BOT_NOT_FOUND falls back to a create with the same artifact),
-        dual-write ``ext.restart.<stage>`` for the status read, refresh the teclaw
-        read handle, and complete the op. Returns ``{success, message}``."""
+        ids), then runs the deploy atom for a ``restart`` op keyed to the bot
+        (upgrade interface; a BOT_NOT_FOUND abandons the op and hands off to
+        :meth:`_recreate_restart_target` — a fresh, crash-safe first-release op
+        with a new bot + binding), dual-writes ``ext.restart.<stage>`` for the
+        status read, refreshes the teclaw read handle, and completes the op.
+        Returns ``{success, message}``."""
         stage_enum = PublishStage(stage)
         publish_record = self._publish_service.get_publish_by_id(publish_id)
         if not publish_record:
@@ -221,16 +224,8 @@ class RestartMixin:
         # so restarting a non-latest stage never delivers another stage's channels.
         delivery = self._ext_state.compose_stored(publish_record.ext or {}, stage_enum)
 
-        op = self._operation_runner.open_operation(
-            publish_id=publish_id,
-            kind=PublishOperationKind.RESTART,
-            stage=stage_enum,
-            bot_uuid=bot_uuid,
-            operator=operator,
-        )
-
         async def _issue():
-            restart_result = await self._build_service.upgrade_async(
+            return await self._build_service.upgrade_async(
                 bot_uuid=bot_uuid,
                 bot=bot,
                 user_id=publish_record.owner_id,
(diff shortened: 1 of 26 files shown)


src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/restart_mixin.py
```diff
@@ -221,16 +225,14 @@ async def execute_restart(
...
-            kind=PublishOperationKind.RESTART,
-            stage=stage_enum,
-            bot_uuid=bot_uuid,
-            operator=operator,
-        )
+        # A prior recreate that crashed between its ext write and its
+        # complete_operation left a dangling op; this re-delivery resolves the
+        # NEW bot from ext and takes the happy path, so nothing else would ever
+        # finish that bookkeeping — finalize it here before deploying.
+        self._finalize_dangling_recreate_op(publish_id, stage_enum, ext)
```
totalfrank: This finalizes the dangling recreate op but then continues into the normal restart issuance below. In the crash window this comment describes, `ext.restart.<stage>` already points at the recreate `FIRST_RELEASE` workflow, so a task redelivery should be able to complete the bookkeeping and return that existing workflow. Continuing here opens a new `RESTART` attempt and sends an extra BaaS `upgrade_async` to the newly-created bot; the new test even enshrines this as workflow 902 after the original recreate workflow 901. That breaks the at-least-once idempotency guarantee for this crash window and can leave two concurrent deploy workflows for one restart request. Please make `_finalize_dangling_recreate_op` report that it completed a recreate and short-circuit `execute_restart` with the existing `ext.restart`/`ext.publish` id instead of issuing another restart.

totalfrank (commented):
Reviewed the publish-flow changes against the PR head. I found two actionable issues: one correctness problem in restart recreate redelivery, and one lockfile hygiene/reproducibility problem. Leaving this as a non-blocking comment review.

src/backend/uv.lock
```diff
@@ -78,25 +78,25 @@ dev = [
 [[package]]
 name = "aiofiles"
 version = "25.1.0"
-source = { registry = "https://mirrors.aliyun.com/pypi/simple" }
-sdist = { url = "https://mirrors.aliyun.com/pypi/packages/41/c3/534eac40372d8ee36ef40df62ec129bee4fdb5ad9706e58a29be53b2c970/aiofiles-25.1.0.tar.gz", hash = "sha256:a8d728f0a29de45dc521f18f07297428d56992a742f0cd2701ba86e44d23d5b2" }
+source = { registry = "https://pypi.org/simple" }
```
totalfrank: This PR does not change dependencies, but the lockfile rewrites package sources from the repo's configured default Aliyun index (`src/backend/pyproject.toml`) to PyPI and adds resolver metadata across the file. That makes the diff noisy and can change locked install behavior in environments that rely on the configured mirror. Please regenerate with the project index configuration or drop the `uv.lock` churn so this PR only carries the publish-flow changes.

src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/restart_mixin.py
totalfrank: Good catch — fixed in `3d596ed`. You're right that continuing into normal issuance broke idempotency for this at-least-once redelivery.

`_finalize_dangling_recreate_op` now returns the recreate's workflow id, and `execute_restart` short-circuits with it (refresh the read handle, return that `restart_publish_id`) instead of opening a second `RESTART` op and issuing another `upgrade_async`. So a redelivery of the crashed recreate converges on the single existing workflow (the recreate's `901`) — no second concurrent deploy.

The test that enshrined `902` is rewritten to assert the opposite: on redelivery `upgrade_async` is wired to raise if called (and asserted not-awaited), no new `RESTART` attempt is opened (the abandoned one stays), the `FIRST_RELEASE` op finalizes to `COMPLETED`, and exactly one deploy workflow exists on the recreated bot.

---

src/backend/uv.lock
totalfrank: Reverted in `3d596ed` — `src/backend/uv.lock` is now byte-for-byte identical to `origin/REL20260723`, so the PR carries no lockfile diff.

The churn was unintentional: the sandbox this ran in can't reach the repo's configured Aliyun mirror, so a local `uv sync` fell back to PyPI and rewrote the index sources. No dependencies changed, so there was nothing for the lockfile to carry.

---

src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/restart_mixin.py
```diff
@@ -311,3 +318,129 @@ def _mutate(latest_ext: dict) -> None:
...
+        issuing a redundant second deploy (idempotent task redelivery)."""
+        op = self._publish_operation_repo.get_latest_by_kind(
+            publish_id, str(PublishOperationKind.FIRST_RELEASE), stage_enum.value
+        )
+        if op is None or op.baas_publish_id is None:
+            return None
+        terminal = {s.value for s in PublishOperationState.terminal()}
+        if op.state in terminal:
+            return None
+        if op.baas_publish_id != (ext.get("publish") or {}).get(stage_enum.value):
```
totalfrank, line 362: The short-circuit now fixes the duplicate restart workflow for the recreate redelivery case, but this guard only checks `ext.publish.<stage>`. A normal online release can also be left as a non-terminal `FIRST_RELEASE` with `ext.publish.online` already written if it crashes after `record_release_ext` and before `complete_operation`; since `restart_bot` supports `ONLINE_PUB`, a restart request in that window would incorrectly finalize that release op and return without actually restarting. Can we also require `(ext.get("restart") or {}).get(stage_enum.value) == op.baas_publish_id` before returning the workflow? That keeps this path specific to restart-recreate redelivery, where `_recreate_restart_target` wrote both `publish` and `restart` to the same workflow.

totalfrank (commented):
Re-checked the follow-up fix. The lockfile cleanup is good, and the restart recreate redelivery now short-circuits instead of issuing the duplicate restart workflow. I found one remaining edge in that shortcut guard: it should verify the `ext.restart` marker too so it only applies to restart-recreate redelivery, not a normal release op that crashed after writing `ext.publish`.

src/backend/src/agentclaw/community/core/service_bot/services/publish_flow/restart_mixin.py
totalfrank: Agreed — fixed in `2a031ce`. The `restart` clause is exactly the right discriminator.

`_finalize_dangling_recreate_op` now also requires `(ext.get("restart") or {}).get(stage) == op.baas_publish_id`. `_recreate_restart_target` writes the same workflow id into both `ext.publish` and `ext.restart`; a normal online first release writes only `ext.publish` and never touches `ext.restart`, so its crash-window op no longer matches — a restart issued in that window proceeds to actually restart the existing bot, and the dangling release op is left for the `online_release` task's own redelivery to finalize.

Added a regression test (`test_restart_does_not_finalize_normal_release_crash_window`): an ID_RECORDED `FIRST_RELEASE` with `ext.publish.online` set but no `ext.restart` entry → the restart issues a fresh RESTART deploy and returns that workflow (not the release's `850`), and the release op stays `ID_RECORDED`.

---

totalfrank (commented):
Re-reviewed the latest head after the restart-recreate follow-up. I don’t have any remaining actionable findings.

Checked specifically:
- `uv.lock` is no longer part of the PR diff.
- `_finalize_dangling_recreate_op` now requires both `ext.publish.<stage>` and `ext.restart.<stage>` to match the `FIRST_RELEASE` workflow, so the short-circuit is scoped to restart-recreate redelivery.
- Added/updated crash-window tests cover both the idempotent recreate redelivery and the normal-release false-positive case.
- Focused test run passed locally against the archived PR head: `test_restart_recreate_crash_before_complete_resumes_idempotently` and `test_restart_does_not_finalize_normal_release_crash_window`.

FreddieSun (approved):
LGTM

PR #362, opened 2026-07-23
fix(bcs): scope channel bindings by environment

RaymondHX:
## Summary

- bind ChannelBinding repositories to the runtime environment
- scope every binding read and mutation by `env`
- reject cross-environment creates and add DB/memory regression coverage

## Root cause

`bcs_channel_bindings.env` was written on create but omitted from repository lookups. When pre-production and production shared the table, `/channels/bindings/by-target` could return a production binding for the same target.

## Impact

Pre-production binding queries, inbound account lookup, listing, status/config updates, and deletion can now access only pre-production bindings. The HTTP contract and response shape are unchanged, and no database migration is required.

## Validation

- `cargo test -p bcs-channel-store`
- `cargo test -p bcs-channel`
- `cargo test -p bcs --lib`
- `cargo test -p bcs-http --test channel_provider_routes_contract`
- `cargo check --workspace`
- `git diff --check`

Strict Clippy remains blocked by pre-existing warnings outside this change.


vzvince (approved):
LGTM

PR #364, opened 2026-07-23
feat(bcs): auto-join humans for session chat

RaymondHX:
## Summary

- Automatically add an authenticated Human caller to a running Session as `Observer + Present` before handling `POST /sessions/{session_id}/chat` when they are not already a participant.
- Bind Human messages to the authenticated caller's `actor_id`, so the request `from` field cannot override the sender identity; Bot callers remain fail-closed and must already be Session participants.
- Validate and route Session chat against Session participants rather than the parent Group participant list, and document the updated contract.

## Why

The endpoint previously rejected authenticated Human callers who were not already Session participants. Even after adding a Human to the Session, the message flow still validated the sender against the parent Group, preventing a Session-only Human from sending under their own identity.

## Validation

- `cargo test -p bcs-http`
- `cargo check -p bcs-http -p bcs-message-flow --all-targets`


vzvince (approved):
LGTM

PR #365, opened 2026-07-23
fix(bcs): reject invalid explicit bot credentials

vzvince:
## Summary

- Treat the presence of `X-BCS-Bot-Token` as an explicit request to authenticate as a bot.
- Return bot authentication and container-validation errors instead of silently falling back to the human OAuth identity.
- Preserve human OAuth fallback when no explicit bot-token header is supplied, and keep valid bot credentials higher priority when both identities are available.
- Add regression coverage for stale bot tokens, container mismatches, mixed bot/human credentials, and human-only OAuth requests.

## Root cause

`resolve_actor_caller` discarded every error from bot authentication. In office/OAuth mode, the CLI can send both a human OAuth `Authorization` header and `X-BCS-Bot-Token`; an invalid explicit bot credential therefore selected the human actor and returned the wrong actor's groups.

## Impact

`GET /groups/my` now fails closed when an explicit bot credential is invalid, while requests that only carry a valid human OAuth identity continue to resolve as `human_<staff_no>`.

## Contract impact

- No endpoint or response-schema change.
- No service or plugin API change.
- This corrects authentication precedence for the existing endpoint.

## Verification

- `cargo test -p bcs-http`
- `git diff --check`

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/groups.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/groups.rs
@@ -1633,11 +1633,18 @@ async fn resolve_actor_caller(
     headers: &HeaderMap,
     uri: &Uri,
 ) -> Result<String, HttpAdapterError> {
-    if let Ok(bot_id) = authenticated_bot_from_headers(state, headers).await {
-        if !bot_id.trim().is_empty() {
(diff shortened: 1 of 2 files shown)


RaymondHX (approved):
LGTM

PR #380, opened 2026-07-23
fix(bcs-cli): restore discover JSON output

vzvince:
## Summary

- Restore JSON output as the default for `bcs-cli discover`.
- Restore `bcs-cli discover --no-json` for human-readable interactive output.
- Make `DiscoverBotsExtendedResponse` serializable and add end-to-end coverage using the real CLI process.

## Background

The JSON renderer, command-local `--no-json` option, and their tests were present in `ERV10158366-20260705` but were omitted during the initial `ocb-public` import. The current discover handler therefore printed human-readable text unconditionally, despite the CLI still defaulting to structured output.

## Compatibility

This change does not modify the discover HTTP request or response schema. It only changes how the CLI renders the response it already receives, so old CLI/new server and new CLI/old server combinations remain compatible.

## Test Plan

- `cargo test --package bcs-cli`: 148 passed, 1 existing ignored.
- `cargo check --package bcs-cli --all-targets`: passed.
- Added E2E coverage for default JSON output and `discover --no-json` text output.

### DIFF
--- a/src/bcs/crates/contracts/bcs-protocol/src/http/bots.rs
+++ b/src/bcs/crates/contracts/bcs-protocol/src/http/bots.rs
@@ -163,7 +163,7 @@ pub struct DiscoverBotsResponse {
 }
 
 /// Extended response from bot discovery with visibility and friendship info.
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


RaymondHX (dismissed):
LGTM

RaymondHX (approved):
LGTM

PR #384, opened 2026-07-23
fix(bcs): expose async judge progress and cover provider timeout

zh-bupt:
This PR improves state-machine progress visibility and asynchronous Judge handling after a bot returns its result. Nodes remain in the durable `running` state while `sub_status` distinguishes between `awaiting_response` and `judging`, allowing the side panel to clearly show whether it is waiting for the bot or evaluating the returned response. For HTTP/BaaS-style integrations, final callback processing is detached from the request lifetime so that a request timeout cannot cancel the Future and leave the Judge workflow stuck. The PR also adds an E2E regression scenario using a real HTTP Provider mock and a blocking Judge, and includes it in the instrumented singlebox LLVM coverage suite. No database migration is required because `sub_status` is derived from the existing node status and artifact. The full E2E suite passed 487/487 assertions, with 46.44% line coverage, 41.16% method coverage, and 100% HTTP endpoint and bcs-cli coverage.


carolynli (approved):
LGTM.

PR #386, opened 2026-07-23
fix(bcs): unify provider callback timeout

vzvince:
## Summary

- Define a single `DEFAULT_PROVIDER_CALLBACK_TIMEOUT_MS` contract in `bcs-service-api`.
- Use the shared one-hour timeout for HTTP Provider downlink requests, ordinary group callback contexts, and system-message callback contexts.
- Remove the system-message five-minute deadline that could expire callback correlation while the Provider still had a valid one-hour execution window.
- Add contract and regression coverage for the shared timeout while preserving `chat.inject` and failed-delivery behavior.

## Root Cause

BCS advertised `timeout_ms=3600000` to HTTP Providers, but system-message `chat.send` runs stored callback context for only five minutes. A Provider callback arriving after five minutes could therefore receive `404 run_not_found` even though it was still within the advertised execution window.

## Impact

HTTP Provider `chat.send` callbacks now use one authoritative one-hour default across delivery and callback correlation. This change does not add durable run-context storage and does not change cancellation, duplicate-terminal, A2A, collaboration-runtime, or frontend reconnect behavior.

## Validation

- Not run during the publish step; the branch was pushed with `--no-verify` as requested.


RaymondHX (approved):
LGTM

PR #403, opened 2026-07-23
fix(bcs): emit human join event for session chat

RaymondHX:
## Summary
- emit HumanJoined when an authenticated Human is auto-added by POST /sessions/{sid}/chat
- keep the authenticated Human as the message sender
- add an HTTP contract assertion for the emitted system-message event

## Tests
- cargo test -p bcs-http --test group_messages_contract


vzvince (approved):
LGTM

PR #404, opened 2026-07-23
fix(baas): preserve configured engine venv symlink

FreddieSun:
## Summary

- Preserve the configured engine interpreter symlink when BaaS starts a local adapter.
- Add a regression test for virtualenv interpreter symlinks.

## Root cause

`Path.resolve()` converted `LOCAL_ENGINE_PYTHON` from the engine virtualenv entry point to its uv-managed base interpreter. The base interpreter does not include `uvicorn`, so the adapter never listened on port 20010 and BaaS publish failed.

## Impact

Singlebox BaaS now starts per-Bot adapters with the configured engine virtualenv and its installed dependencies. No public contract or configuration schema changes.

## Validation

- `DEPLOY_ENV=test ../../../src/baas/.venv/bin/python -m pytest tests/unit/plugins/sandbox/arca/test_process_manager_coverage.py tests/unit/plugins/sandbox/arca/test_process_manager_env.py -q`
- `../../../src/baas/.venv/bin/python -m ruff check src/secbaas/community/plugins/sandbox/arca/local_proc/_process_manager.py tests/unit/plugins/sandbox/arca/test_process_manager_coverage.py`

### DIFF
--- a/src/baas/src/secbaas/community/plugins/sandbox/arca/local_proc/_process_manager.py
+++ b/src/baas/src/secbaas/community/plugins/sandbox/arca/local_proc/_process_manager.py
@@ -1105,7 +1105,9 @@ class LocalProcessManager:
         """
         configured = os.environ.get("LOCAL_ENGINE_PYTHON")
         if configured:
-            candidate = Path(configured).expanduser().resolve()
+            candidate = Path(configured).expanduser()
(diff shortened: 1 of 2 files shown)


cassiuscai (approved):

totalfrank (approved):
LGTM.

PR #408, opened 2026-07-23
fix(baas): preserve configured engine venv symlink

FreddieSun:
## Summary

Backport the complete singlebox BaaS/BCN startup fix set to `REL20260723`.

## Changes

- Preserve `LOCAL_ENGINE_PYTHON` virtualenv symlinks so BaaS adapters run with the interpreter that provides `uvicorn`.
- Ignore incomplete saved BCN sessions without a reconnect token, so a stale session cannot pin multiple local bots to one prior identity.

## Root causes

- Resolving the configured virtualenv interpreter followed its symlink to the uv-managed base interpreter, which lacks `uvicorn`; BaaS publish therefore failed before the adapter listened on port 20010.
- A partial `.bcs/session.json` could be reused without a reconnect token and retain an obsolete Bot identity, causing conflicts when the local five-Bot stack started.

## Validation

- `DEPLOY_ENV=test ../../../src/baas/.venv/bin/python -m pytest tests/unit/plugins/sandbox/arca/test_process_manager_coverage.py tests/unit/plugins/sandbox/arca/test_process_manager_env.py -q` — 112 passed, 3 skipped.
- `npx egg-bin test test/bcs-ws-client.test.ts --grep 'incomplete saved session|pending session'` — passed.
- Ruff check for the BaaS changes passed.

### DIFF
--- a/scripts/test_bcn_plugin_source.sh
+++ b/scripts/test_bcn_plugin_source.sh
@@ -226,6 +226,59 @@ test_stack_script_has_npm_branch() {
   assert_contains "$src" '[ "$BCN_PLUGIN_SOURCE" != "npm" ]'
 }
 
+test_session_bot_uuid_requires_usable_session() {
+  local tmp; tmp="$(mktemp -d)"
+  local funcs="${tmp}/stack-session-functions.sh"
+  local profile_root="${tmp}/profiles"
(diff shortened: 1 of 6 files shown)


RaymondHX (approved):
LGTM

phoenixliu (approved):

totalfrank (approved):
LGTM

PR #418, opened 2026-07-23
feat: manage skills pool migration quarantine

### LINKED ISSUE #377: [Skills Pool P3] 管理 Migration Quarantine 七日生命周期
## What to build

把数据面切换前的旧内容作为 Migration Quarantine 管理，只用于取证和人工恢复，不作为正常自动回滚源。系统必须以 Pool 激活时间、后续新镜像成功重启或重新发布证据以及当前迁移健康状态共同判定清理资格，默认至少保留七天。

## Acceptance criteria

- [ ] 原子切换时旧内容进入与 Bot 和 migration generation 可关联的隔离区，并记录创建与来源证据。
- [ ] 隔离内容不参与正常读写、mapping、locator 解析或自动回滚。
- [ ] Pool 激活不足七天时不得自动清理。
- [ ] 未经历至少一次新镜像成功重启或重新发布及 reconciliation 时不得自动清理。
- [ ] Bot 处于失败、人工修复或显式回滚流程时不得自动清理。
- [ ] 同时满足全部条件后，后台清理幂等执行并留下审计记录。
- [ ] 重复任务、任务中断和隔离目录已不存在时不会误删其他 Bot 或其他 migration generation 的内容。
- [ ] 运维查询可以看到隔离区年龄、重启证据、阻塞原因和预计清理资格。
- [ ] 时间可控的自动化测试覆盖七日前后、重启证据缺失、失败状态和成功清理。

## Blocked by

- #376

FreddieSun:
## Summary
- persist generation-scoped Migration Quarantine evidence when Pool cutover commits
- require seven-day retention plus a verified post-activation runtime reconciliation before cleanup
- atomically fence cleanup on retention, latest READY evidence, layout/generation, and cleanup lease
- preserve MySQL microsecond ordering and make FAILED win equal-timestamp runtime evidence races
- perform idempotent Engine cleanup with exact-generation and symlink safety checks
- expose the single-record operational eligibility seam; batch/API aggregation remains in #378

## Verification
- Backend full suite: `8888 passed`
- Backend focused Skills Pool/quarantine suite: `90 passed`
- Backend architecture suite: `103 passed`
- Engine full suite: `1985 passed, 5 deselected`
- Ruff check and format: passed for all changed Python files
- Spec and standards reviews: no blocking findings

## Dependency
- Rebased onto the final `dev` implementation of #376/#417; this PR now targets `dev` directly.

Closes #377


totalfrank:
## Review — Skills Pool Migration Quarantine (#418)

Reviewed the full diff (32 files, +2408/-80) against `dev` at `e4106d1`. This is a carefully-built change; the concurrency/fencing design is the strong point. Comments below, then a few questions worth resolving before merge.

### What it does
- At Pool cutover-commit, persists the old `skills-local` layout as a generation-scoped **Migration Quarantine** row (`ac_skill_migration_quarantine`), keyed by `(env, entity_id, bot_id, migration_generation)`.
- Requires **7-day retention + a verified post-activation runtime reconciliation** (READY probe whose identity matches `preparation_id`/`layout_contract_version`) before anything can be deleted.
- Cleanup is atomically fenced on the live `POOL_ACTIVE` layout row, retention, latest-READY runtime evidence, `cleaned_at IS NULL`, and a cleanup lease; the engine then performs an idempotent, symlink-safe `rmtree` of exactly one generation directory.
- Exposes the single-record operational seam (`inspect`); batch/API aggregation deferred to #378.

### Strengths
- **Fencing is coherent.** `claim_cleanup` takes `SELECT … FOR UPDATE` on the POOL_ACTIVE layout row and gates the state transition on retention + READY evidence + lease expiry; `mark_cleaned` / `mark_cleanup_failed` / `record_cleanup_uncertain` each re-assert the layout fence under `FOR UPDATE` and match `cleanup_lease_owner`. Double-fencing `evaluate()` (in-process) against `claim_cleanup` (DB) is a nice belt-and-suspenders.
- **Ordering semantics are correct.** `runtime_reconciled_at` uses `mysql.DATETIME(fsp=6)` for microsecond ordering, and the `timestamp_order` predicate makes FAILED win equal-timestamp races while READY only overwrites an existing READY — matches the PR description.
- **Uncertain vs. hard-failure split is right.** Timeouts / `invalid_runtime_response` keep the delete fence (`status` stays `cleaning`) instead of releasing to a re-claimable state; only definitive failures release. Prevents a delete from racing an unknown-outcome runtime call.
- **Engine deletion is defensively safe.** Generation regex allowlist, ancestor symlink walk, `relative_to(home)`, strict `resolve()` + `is_relative_to(root)`, and symlink/real-dir checks before `rmtree`. The delete target is derived solely from a fixed Pool root + generation — the caller cannot point it at another bot or path.
- **Idempotent throughout** (`ALREADY_ABSENT`, `record_absent`, upsert-on-retry) and **broad test coverage** — service, repository, reconcile task/service, runtime plugin, engine cleanup, router, and both engine contract suites.

### Questions / potential issues

1. **Cleanup-task amplification (medium).** The new pool-active reconcile branch enqueues an immediate (`delay_seconds=0`) cleanup task on *every* wakeup for an already-active pool, on top of the 7-day-delayed task enqueued at cutover. Before retention elapses each returns `RETAINED → Reschedule(24h)`, so identical cleanup tasks accumulate and re-run daily until eligibility. Fencing keeps this correct (no double-delete), but queue volume scales with reconcile-wakeup frequency. Does the task queue dedup on `(task_type, payload)`? If not, consider enqueuing the delayed cleanup only from the POOL_ACTIVE *transition* path (or gating on absence of an outstanding cleanup task) rather than on every steady-state reconciliation.

2. **Permanent `INVALID` runtime outcomes are retried (low-med).** Non-transient engine results (`generation_not_real_directory`, `generation_escapes_quarantine_root`, `quarantine_ancestor_is_symlink`) map to `mark_cleanup_failed` and the task handler returns `Retry(...)`. These can't self-heal, so the task retries until its 90-day deadline — wasted work + noise. Consider treating structurally-invalid outcomes as terminal (`Complete`/`Reschedule`) and reserving `Retry` for transient errors.

3. **Legacy tasks on the pool-active branch (low).** With no `observed_at`, the fallback `datetime.min` never satisfies `record_runtime_reconciliation`'s `pool_activated_at < observed_at` guard, so the handler returns `Retry("… evidence race lost")` and loops until deadline instead of completing/skipping. The comment acknowledges the evidence can't qualify — a `Complete()`/`Reschedule` for that known-unqualifiable case would be cleaner. Transitional, low impact.

4. **Extra runtime probe per steady-state reconcile (perf, by design).** `_verify_active_runtime` now issues a runtime probe + DB write on every reconcile of an active pool where `ALREADY_ACTIVE` previously short-circuited. Intended for post-activation verification, but worth being aware of the added per-signal cost / engine load on hot bots.

### Minor
- `record_pool_cutover_committed` now hard-depends on `rollout_evidence.engine_type` and a runtime-supplied `quarantine` path. The engine always emits `"quarantine"` (deterministic from pool root + generation), so this looks contract-safe, but it's a new "commit stalls if evidence missing" mode — a one-line log when the path is absent would make that diagnosable.
- No in-repo DDL for `ac_skill_migration_quarantine`. Consistent with the sibling `ac_bot_skill_layout_state` (its DDL was removed in `d460ebb`), so presumably intentional — just confirm the out-of-repo production migration for the new table + its unique/index constraints is tracked.
- `commit_pool_active` does `session.rollback(); return False` on `_activate_quarantine` failure inside the commit-on-exit session CM. This is actually correct (the CM's clean-exit commit then finds nothing pending, so the whole transaction is atomically aborted) — noting it only because the rollback-then-empty-commit is subtle.

Overall this looks solid and merge-worthy once the amplification question (#1) is settled. Nice work on the fencing and the engine-side path safety.

---

FreddieSun:
Thanks @totalfrank — I addressed the first three follow-ups in `b11e969f`:

1. Removed the immediate quarantine-cleanup enqueue from every steady-state `POOL_ACTIVE` reconciliation. The generation-scoped 7-day task is still created on the `POOL_ACTIVE` transition, so active wakeups no longer amplify queued cleanup work.
2. Added explicit cleanup retryability. Transient/unknown runtime outcomes remain retryable, while structural `INVALID` outcomes are persisted via `mark_cleanup_failed` and complete terminally instead of retrying until the 90-day deadline.
3. Legacy reconcile payloads without `observed_at` now complete safely when the Bot is already Pool-active. They do not probe, write unqualifiable runtime evidence, or busy-retry.

Regression coverage was added for all three cases. Validation: Skills Pool suite `126 passed`; Ruff and `git diff --check` passed.

totalfrank:
Thanks @FreddieSun — re-reviewed `b11e969`. The fixes for #1 (no more steady-state amplification) and #3 (legacy `observed_at=None` completes safely without probing/retrying) look correct and are well covered by the new tests. 👍

One small gap in the #2 fix before I approve:

`retryable` is computed as `response.status is TRANSIENT_ERROR`, but the **uncertain** branch (which calls `record_cleanup_uncertain` to *hold the delete fence for a later retry*) can be entered with a non-transient status. Concretely:

- The runtime plugin returns `status=INVALID, evidence.reason="invalid_runtime_response"` when the Engine sends a well-formed envelope with an **unknown status string** (contract skew — new/renamed status the Backend doesn't recognize).
- In `cleanup()`, `"invalid_runtime_response"` is in the uncertain set → `cleanup_uncertain=True` → `record_cleanup_uncertain(...)` holds the fence (`status` stays `cleaning`).
- But `retryable = (INVALID is TRANSIENT_ERROR) = False` → the task handler hits `if result.retryable is False: return Complete()`.

Net effect: the record is left in `cleaning` with the fence held, the task completes and is **never retried**, so that generation is never cleaned (until the 1h lease self-expires, after which nothing re-drives it). It's safe — no data loss, no wrong deletion — but it defeats the purpose of the uncertain-fence path (hold now, retry later), and it's a case the new tests don't cover (they exercise timeout→retryable and *structural* INVALID→terminal, but not uncertain-with-non-transient-status).

Suggestion: tie `retryable` to the branch actually taken rather than to the status — e.g. `retryable = cleanup_uncertain or (response.status is TRANSIENT_ERROR)`, so any path that records *uncertain* stays retryable while `mark_cleanup_failed` (structural INVALID) stays terminal. A regression test for `status=INVALID, reason="invalid_runtime_response"` asserting `record_cleanup_uncertain` + `Retry` would lock it in.

Low severity, but since it's in the just-changed logic I'd rather flag it than wave it through. Happy to approve once this is settled (or if you'd argue the orphaned-fence case is acceptable and prefer to leave it terminal — in which case it probably shouldn't record *uncertain*).

---

FreddieSun:
Addressed #4 in `04b3a6d3`.

The repeated probe remains intentional while the same migration generation is still in quarantine: lifecycle signals for a restarted/new runtime refresh the latest READY/FAILED evidence and keep the 7-day cleanup fence safe.

Once the same-generation quarantine reaches the terminal `CLEANED` state, the reconcile task now completes before calling the runtime or writing reconciliation evidence. `RETAINED`, `CLEANING`, `CLEANUP_FAILED`, and a missing quarantine record continue through the existing probe path, so we do not weaken pre-cleanup verification.

This does not add schema. Coverage was added at the task-handler seam, and the Skills Pool/DI regression set passes (155 tests) with Ruff clean.

FreddieSun:
Thanks — agreed. Fixed in `bdaff53f`.

`retryable` is now derived from the branch semantics: any outcome recorded through `record_cleanup_uncertain` remains retryable, including `INVALID + reason=invalid_runtime_response`. Definitive structural INVALID outcomes still go through `mark_cleanup_failed` with `retryable=False`, so they remain terminal.

Added a regression test through `SkillsPoolQuarantineCleanupTaskHandler.handle()` that exercises the unknown-runtime-status shape and verifies `record_cleanup_uncertain + Retry`, alongside the existing structural-INVALID terminal test.

Skills Pool/DI regression suite: 156 passed; Ruff clean.

FreddieSun:
Closed the remaining minor follow-ups:

- In `86766efa`, a current cutover commit with missing/invalid quarantine-path evidence now emits an error containing the failure reason, Bot scope, and migration generation before returning the existing fenced `False` result. Added repository-level regression coverage.
- Production DDL is now tracked independently in #446. It records the exact unique constraint, cleanup index, `DATETIME(6)` requirement, and makes schema application a hard precondition before enabling the rollout whitelist. This remains outside the repository's in-code DDL convention.
- Acknowledged the rollback/empty-commit observation; no behavior change is needed there.

Validation: focused Skills Pool/DI suite `157 passed`; Ruff clean.

totalfrank (dismissed):
Approving — all review items are resolved. Verified the fixes at `86766ef`:

- **#1 amplification** (`b11e969`) — the steady-state `delay_seconds=0` enqueue is gone; the 7-day task fires only on the POOL_ACTIVE transition. ✅
- **#2 uncertain/terminal split** (`b11e969` + `bdaff53`) — `retryable=(cleanup_uncertain or response.status is TRANSIENT_ERROR)` now correctly keeps every `record_cleanup_uncertain` outcome retryable (including `INVALID + reason=invalid_runtime_response`) while structural `INVALID` via `mark_cleanup_failed` stays terminal. Regression test added at the handler seam. ✅
- **#3 legacy `observed_at`** (`b11e969`) — `None` now completes safely on the pool-active branch without probing or busy-retrying. ✅
- **#4 steady-state probe** (`04b3a6d`) — a `CLEANED` quarantine short-circuits before the runtime probe; `RETAINED`/`CLEANING`/`CLEANUP_FAILED`/absent still go through full verification, so pre-cleanup safety is preserved. ✅
- **Minors** (`86766ef`) — cutover commit now logs `reason=missing_quarantine_path` with scope + generation before the fenced `False`; production DDL (unique constraint, cleanup index, `DATETIME(6)`) tracked as a hard precondition in #446; rollback/empty-commit behavior acknowledged as correct. ✅

The concurrency fencing, microsecond ordering, uncertain-vs-terminal handling, and engine-side path/symlink safety all hold up. Nice, careful iteration.

Note: I verified the changes by inspection and relied on your reported green runs (focused Skills Pool/DI suite `157 passed`, Ruff clean) — the sandbox here can't install deps to run the suite itself. LGTM.

---

totalfrank (approved):
Re-approving at `fdaa9ec`. The only change since my previous approval is a behavior-preserving refactor (`refactor: extract skills pool cutover diagnostics`): the inline `missing_quarantine_path` error log was moved into a `log_missing_quarantine_path` helper in the new `skills_pool_cutover_diagnostics.py`, with the cutover-commit call site swapped to use it — identical message and fenced `False` behavior. All previously-resolved findings (#1–#4 and the minors) remain addressed. LGTM.

---

PR #420, opened 2026-07-24
Config-driven gateway forwarding

totalfrank:
## What

Replaces the gateway's per-operation stub routers with **declarative, domain-transparent forwarding**. The gateway now serves **zero hand-written routes**: one catch-all forwards every `/openapi/v1/*` request verbatim to its domain's upstream, and `/openapi.json` is **generated** from each backend's own published OpenAPI description. Onboarding a new backend API needs **no gateway change or release**.

Developed **Spec-Driven** (spec → plan → tasks → implement); artifacts under `src/gateway/specs/2026-07-24-gateway-config-driven-forwarding/`.

## How it works

- **Domain routing** — the leading path segment selects the upstream via `configs/upstreams.yaml`; an unconfigured domain is denied (never an open proxy).
- **Verbatim forward** — the streaming `Forwarder` (httpx) proxies the request unchanged; duplicate headers (`Set-Cookie`) and SSE/uploads preserved. Auth is resolved fail-closed from the existing prefix rules before forwarding. (The gateway↔backend signed-JWT seam is owned by the auth workstream; forwarding just exposes it.)
- **Generated doc** — served OpenAPI = the backend's published description, narrowed to `/openapi/v1`, with `x-avernet-security` attached and components pruned. Read from a pluggable `SchemaCatalog` that **auto-adopts the latest** via background refresh with **last-known-good** fallback — a stale/unreachable source degrades only the doc, never live traffic. Source is vendor-neutral: `file` (single-box) or `object_store` (any deployed edition — corp or community; S3 / MinIO / GCS / OSS / …).
- **Backend** — a dedicated `adapters/http/openapi_v1/` package adds **new, definition-only routers** for the redesigned external contract (bots, channels, identity, mcp, resources, routines, skills), all under `/openapi/v1/bots`. These are **purpose-built new APIs** ported from the #389 definitions — **not** a re-mount of the legacy handlers: Pydantic schemas + stub handlers (`NotImplementedError`), with auth via a `require_principal` stub the gateway-JWT verifier drops into later. Legacy `/api/…` is untouched. `dump_openapi()` publishes the public description (paths + pruned components); a **publish-time compat gate** (`check_compatible`) fails a release on any breaking change unless explicitly overridden.

## The published artifact is not committed

`bots.openapi.json` is **build output**, not source — produced by `dump_openapi` for the single-box file, or pulled from the object store in any deployed edition. It's gitignored (`configs/schemas/*.openapi.json`); the gateway serves a valid empty doc until it's present, and tests use a small `tests/fixtures/` sample.

## Status

- [x] Phase 1 — spec · [x] Phase 2 — plan · [x] Phase 3 — tasks · [x] Phase 4 — implementation (11 tasks, 6 groups)

## Verification

- Gateway: **300** tests + `ruff` green.
- Backend: gateway-contract suite **102** + namespace-invariant test green; the new public routers are additive (legacy `/api/…` untouched).
- mypy consistent with the repo baseline (only pre-existing environmental subclass/`Any` artifacts). No `uv.lock` diff vs `dev`.

## Notes / follow-ups

- The backend routers are **definition-only** — the redesigned external contract with stub handlers (`NotImplementedError`); wiring them to services is a follow-up, owned by the backend team.
- The **`object_store` reader** is a pluggable `SchemaCatalog` flavor available to any deployed edition (corp or community); only single-box uses `file`. Bare ships **file-only** today — the vendor-neutral object-store reader (portable HTTP(S) fetch, no cloud SDK) is a follow-up.
- **`ctoken` middleware** — the backend's `UserContextMiddleware` strips the gateway `ctoken` only for `^/api/bots/...`; forwarded `/openapi/v1/bots/...` requests won't match. Owned by the auth/ctoken workstream (same area as the JWT seam) — flagged, not changed here.
- Tooling here runs against PyPI (`UV_DEFAULT_INDEX`) because the pinned aliyun mirror is blocked by this environment's egress proxy — no change to the committed index config.


xxxxpenny (dismissed):

xxxxpenny (approved):

FreddieSun (approved):
lgtm

PR #423, opened 2026-07-24
fix(bcs): gate session-file access on session participants and relax share authz

zh-bupt:
fix(bcs): gate session-file access on session participants and relax share authz

ensure_session_member checked group.participants, but a session seeds its
participants from the group and then evolves independently. Session-only
participants added via add_session_participant, and callers after the group
seed diverged, were wrongly denied. It now judges membership against
sess.participants. Add human_has_session_access next to human_has_group_access
so the Human caller is accepted when they, or a bot they own, is a session
participant.

Share authz is relaxed from uploader/creator/driver (can_mutate) to any
session member. Add pure can_share(caller_identities, session_participants);
ShareMintCommand now carries session_participants (resolved from the session
that ensure_session_member already loads) instead of session_creator and
driver_bot. ensure_session_member returns the loaded Session so the share
handler reuses it without an extra fetch and feeds the same participant set
to the service, keeping authz transport-agnostic.

Delete authz is unchanged: can_mutate already lets the driver bot delete any
file and lets a human delete files uploaded by a bot they own.

Tests: can_share unit tests; service-level share_mint_non_member_forbidden,
share_mint_any_member_can_share, share_mint_human_via_owned_bot_can_share.
Verified with cargo test for bcs-session-file and bcs-http contracts.

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/session_files.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/session_files.rs
@@ -284,36 +284,54 @@ async fn caller_identities(state: &HttpAppState, caller: &GroupChatCaller) -> Ve
     }
 }
 
-/// Verify the caller is a member of the session's group: the session must
-/// exist, the group must exist, and the caller (bot or human) must be a
-/// participant or own a participating bot.
+/// Verify the caller is a member of the session and return the loaded session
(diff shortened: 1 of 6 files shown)


zh-bupt:
@codex review this pr

RaymondHX (approved):

PR #429, opened 2026-07-24
fix(baas): do not return 503 when device not available

cassiuscai:
When a device is not available, BaaS was returning HTTP 503 to the caller. This change makes it return a proper error code instead of an infrastructure-level 503, so callers can handle the situation gracefully.


pfmiles (approved):
ok

PR #432, opened 2026-07-24
feat(bcs): propagate channel source message ids

RaymondHX:
## Summary

- add an explicit optional `source_im_message_id` to the internal channel message-flow contracts
- correlate each channel-originated bot run with its exact source IM message before delivery starts
- propagate the correlation through bot events and channel outbound delivery, with cleanup on delivery failure and terminal run states
- add contract tests and a design note covering concurrent requests and compatibility

## Why

The DingTalk channel needs to acknowledge an inbound message immediately with a reaction, then remove or update that reaction when the bot starts responding. Conversation-level correlation is insufficient when multiple requests share the same conversation; the original IM message ID must follow the exact BCS run.

## Validation

- `cargo check --workspace`
- `cargo test -p bcs-channel -p bcs-message-flow`
- targeted source-message correlation, outbound propagation, and cleanup tests

## Compatibility

All added fields are optional process-internal contracts. Existing WebSocket, persistence, and non-channel paths continue to use `None`; no database or wire-protocol migration is required.

### DIFF
--- a/src/bcs/crates/adapters/ws/bcs-ws/src/web/dispatcher.rs
+++ b/src/bcs/crates/adapters/ws/bcs-ws/src/web/dispatcher.rs
@@ -329,6 +329,7 @@ async fn handle_chat_send(
             }),
             thinking: params.thinking,
             idempotency_key: params.idempotency_key,
+            source_im_message_id: None,
             sender_conn_id,
         })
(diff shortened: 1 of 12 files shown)


vzvince (approved):
LGTM

Issue #446, opened 2026-07-25
[Skills Pool P3] Apply production DDL for migration quarantine table


FreddieSun:
## Context

PR #418 introduces the ORM-backed `ac_skill_migration_quarantine` table for the Skills Pool P3 quarantine lifecycle. Local/test profiles create it through SQLAlchemy metadata, while production schema is operator-provisioned outside this repository.

This production DDL must be applied before any Skills Pool rollout whitelist is enabled. The table may remain empty while rollout is disabled, so deploying the schema first is forward-compatible.

Related:
- #377
- #418

## Required production schema contract

The production table must match the ORM contract in `SkillMigrationQuarantineModel`, including:

- `runtime_reconciled_at DATETIME(6)` so READY/FAILED evidence ordering preserves microseconds.
- Unique constraint `uk_skill_migration_quarantine_scope_generation` on:
  - `env`
  - `entity_id`
  - `bot_id`
  - `migration_generation`
- Cleanup lookup index `idx_skill_migration_quarantine_cleanup` on:
  - `env`
  - `status`
  - `pool_activated_at`
- Column lengths, nullability, timestamp defaults, and cleanup lease fields aligned with the ORM model.

## Acceptance criteria

- [ ] Production DDL is reviewed through the normal database change process.
- [ ] DDL is applied in the target environment before the Skills Pool rollout whitelist is enabled.
- [ ] The unique constraint and cleanup index match the ORM contract.
- [ ] `runtime_reconciled_at` retains microsecond precision.
- [ ] Pre-production verification confirms Backend can insert, reconcile, inspect, and clean one generation-scoped quarantine record.
- [ ] Rollout runbook links the completed schema change as a hard precondition.

FreddieSun:
[...]

PR #447, opened 2026-07-25
feat: add Skills Pool rollout operations

### LINKED ISSUE #378: [Skills Pool P3] 完成灰度运维、审计与发布兼容性验收
## What to build

交付 Skills Pool P3 的运维与发布闭环：运维可以按环境和 engine 管理精确 Bot 灰度、查询迁移证据、唤醒或重试单 Bot，并以批次指标决定是否人工晋级；最终兼容性验收覆盖新 Backend 与新旧镜像混部、四个文件型 engine、服务版本发布以及 Teclaw no-op。

## Acceptance criteria

- [ ] 运维可以按环境启停 rollout feature、人工晋级 engine，并以 `owner + bot` 精确增删白名单。
- [ ] 初始灰度禁止 `enable_all` 或通配命中；配置缺失、异常和跨环境条目均 fail closed。
- [ ] 移除白名单前后均显示 Bot 是否已经认领；移除未认领 Bot 阻止启动，移除已认领或已激活 Bot 不停止收敛或回退。
- [ ] 支持查询单 Bot 的 engine、provider、active/target layout、contract、preparation identity、migration generation、probe、切换边界、失败和隔离区证据。
- [ ] 支持手动唤醒、重试可重试失败、标记人工修复并添加备注，以及触发显式回滚。
- [ ] 支持按环境、engine 和批次查询 eligible、claimed、preparing、active、failed、负向对照与清理资格。
- [ ] 批次扩大和 engine 晋级只能由人工执行；单 Bot 成功不能自动修改 rollout 配置。
- [ ] 兼容测试证明新 Backend + 旧镜像保持 Legacy，新镜像 marker-only Bot 不激活，获准 Bot 才进入 Pool。
- [ ] 多 Bot 使用新旧镜像时相互独立；现有 ONLINE Legacy 服务不原地迁移；候选 Pool 服务版本全部 ready 后才切换。
- [ ] OpenClaw、Claude Code、AICoding、Hermes 的目标布局与结构桥通过矩阵验收，Teclaw 的文件、locator、mapping 和发布行为保持不变。
- [ ] 每个批次保留同 engine、同环境的未命中负向对照以及 Teclaw 回归样本。
- [ ] 发布验收产出可审计报告，明确成功率、失败分布、数据一致性和是否允许进入下一 engine。

## Blocked by

- #372
- #373
- #374
- #375
- #376
- #377

FreddieSun:
## What changed

- 新增 Skills Pool 灰度运维 API：按环境启停、按既定顺序晋级 engine、精确维护 `owner_id + bot_id` 白名单及对照组。
- 新增单 Bot 迁移证据查询、手动唤醒/重试/人工修复/显式回滚，以及按环境、engine、批次聚合的验收报告。
- rollout 配置采用严格 fail-closed schema：禁止 `enable_all`、通配符和未知字段；无效或错引擎批次成员会阻断 `promotion_ready`。
- rollout 配置通过 CAS 更新，并与 append-only 审计记录在同一事务提交。
- 新增中文灰度发布 runbook，覆盖新 Backend 与新旧镜像混部、四个文件型 engine、服务版本及 Teclaw no-op 的验收路径。
- HTTP adapter 仅依赖 Service API Protocol，并补齐五组 Protocol/Concrete 与 DI singleton conformance 测试。

## Compatibility and rollout

- rollout 配置缺失或未开启时保持 fail-closed，不会认领线上 Bot。
- 只有精确命中白名单的个人 Bot 或服务 Bot 草稿可以首次迁移；现有 ONLINE Legacy 服务不原地迁移。
- 生产发布前需先部署 `ac_skills_pool_rollout_audit` DDL，跟踪见 #446。
- 镜像侧兼容矩阵与四引擎/Teclaw 验收证据由已合入的 OCB companion CI 提供。

## Validation

- Backend full suite: `8951 passed`
- Focused Skills Pool, endpoint, DI and architecture suite: `126 passed`
- Ruff: passed
- `git diff --check`: passed
- Final Standards review: passed
- Final Spec review: passed

Closes #378

### DIFF
--- a/docs/skills-pool-p3-rollout-runbook.zh-CN.md
+++ b/docs/skills-pool-p3-rollout-runbook.zh-CN.md
@@ -7,31 +7,98 @@
 ## 发布前置
 
 生产环境不由 Backend 自动建表。发布新 Backend 前，必须先通过 OceanBase
-数据库变更流程创建 `ac_skills_pool_rollout_audit`，并执行
-`SELECT 1 FROM ac_skills_pool_rollout_audit LIMIT 1` 作为发布 preflight；表不存在
-或不可读写时禁止发布 Backend。Avernet 仓库只维护 ORM，数据库变更单需使用
-以下与 ORM 一致的 DDL：
+数据库变更流程创建 `ac_skill_migration_quarantine` 和
+`ac_skills_pool_rollout_audit`，并分别执行 `SELECT 1 ... LIMIT 1` 作为发布
+preflight；任一表不存在或不可读写时禁止开启 rollout。Avernet 仓库只维护
+ORM，数据库变更单需使用以下与 ORM 一致的 DDL：
 
 ```sql
+CREATE TABLE ac_skill_migration_quarantine (
+  id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT
+    COMMENT '自增主键',
+  env VARCHAR(20) NOT NULL
+    COMMENT '部署环境，如 pre、prod',
+  entity_id VARCHAR(512) NOT NULL
+    COMMENT 'Bot 所属实体或用户标识',
+  bot_id VARCHAR(128) NOT NULL
+    COMMENT 'Bot 唯一标识',
+  migration_generation VARCHAR(64) NOT NULL
+    COMMENT '本次 Skills Pool 迁移代际标识',
+  engine VARCHAR(64) NOT NULL
+    COMMENT 'Bot 使用的引擎类型',
+  path VARCHAR(1024) NOT NULL
+    COMMENT '运行时返回的隔离目录物理路径',
+  status VARCHAR(32) NOT NULL DEFAULT 'retained'
+    COMMENT '隔离记录状态',
+  source_evidence TEXT NOT NULL
+    COMMENT 'Pool cutover 时记录的隔离证据 JSON',
+  pool_activated_at TIMESTAMP NULL DEFAULT NULL
+    COMMENT 'Bot 成功进入 POOL_ACTIVE 的时间',
+  runtime_reconciled_at TIMESTAMP(6) NULL DEFAULT NULL
+    COMMENT 'Pool 激活后运行时完成 reconciliation 的时间',
+  runtime_reconciliation_status VARCHAR(16) NULL
+    COMMENT '运行时 reconciliation 结果',
+  runtime_evidence TEXT NULL
+    COMMENT '运行时 reconciliation 证据 JSON',
+  cleaned_at TIMESTAMP NULL DEFAULT NULL
+    COMMENT '隔离目录完成清理的时间',
+  cleanup_evidence TEXT NULL
+    COMMENT '隔离目录清理结果证据 JSON',
+  cleanup_lease_owner VARCHAR(128) NULL
+    COMMENT '当前持有清理租约的 Worker 标识',
+  cleanup_lease_expires_at TIMESTAMP NULL DEFAULT NULL
+    COMMENT '清理任务租约过期时间',
+  gmt_create TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+    COMMENT '记录创建时间',
+  gmt_modified TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+    ON UPDATE CURRENT_TIMESTAMP
+    COMMENT '记录最后修改时间',
+  PRIMARY KEY (id),
+  UNIQUE KEY uk_skill_migration_quarantine_scope_generation
+    (env, entity_id, bot_id, migration_generation) GLOBAL,
+  KEY idx_skill_migration_quarantine_cleanup
+    (env, status, pool_activated_at) GLOBAL
+) DEFAULT CHARSET = utf8mb4
+  COMMENT = 'Skills Pool 迁移隔离目录生命周期及清理证据';
+
 CREATE TABLE ac_skills_pool_rollout_audit (
-  id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
-  env VARCHAR(20) NOT NULL,
-  config_id BIGINT(20) UNSIGNED NOT NULL,
-  action VARCHAR(128) NOT NULL,
-  batch_id VARCHAR(128) NULL,
-  operator VARCHAR(128) NOT NULL,
-  reason VARCHAR(512) NOT NULL,
-  based_on_config_version VARCHAR(64) NULL,
-  effective_config_version VARCHAR(64) NOT NULL,
-  evidence TEXT NULL,
-  effective_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT
+    COMMENT '自增主键',
+  env VARCHAR(20) NOT NULL
+    COMMENT 'Rollout 配置所属部署环境',
+  config_id BIGINT(20) UNSIGNED NOT NULL
+    COMMENT '关联的 ac_common_config 配置记录 ID',
+  action VARCHAR(128) NOT NULL
+    COMMENT '运维操作类型',
+  batch_id VARCHAR(128) NULL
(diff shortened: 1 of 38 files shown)


totalfrank (commented):
## Review: Skills Pool rollout operations

I reviewed the operator control plane, the fail-closed rollout schema, the CAS/audit repository, the operational query, and the HTTP adapter. Overall this is a careful, well-tested change. The strict `is_valid_rollout_config_value` schema (no `enable_all`, no wildcards, no unknown keys, engines must be a strict prefix of `ENGINE_PROMOTION_ORDER`), the single-transaction CAS + append-only audit in `commit_change`, and the `_PROTECTED_CONFIG_KEYS` guard that forces `skills_pool/layout_rollout` through the operator API are all solid. I ran the focused suite locally — **114 passed**, and `ruff check` is clean on the changed files.

A few things worth a look before merge, roughly in priority order.

### 1. An orphaned whitelist entry can wedge the control plane for an engine (medium)
`remove_bot` calls `_resolve_scope(...)` → `_resolve_bot_and_scope(...)` at the top (`operations.py:358`), which raises `RolloutOperationError("bot not found")` when `get_live_by_id_owner_and_env` returns no match. The scope is only used to *report* the layout claim state (`claimed_before`/`claimed_after`); the actual whitelist mutation doesn't need the bot to exist. Consequence: if a whitelisted personal/service bot is later deleted, the operator can no longer remove its stale entry through the API.

It compounds because `_open_batches` (`operations.py:614-642`) resolves *every* whitelist member via `_resolve_bot_and_scope`, and `_open_batches` runs inside both `add_bot` and `promote_engine`. So a single deleted-but-still-whitelisted bot makes `add_bot` and `promote_engine` for that engine raise as well — the whole rollout for the engine gets stuck with no in-API way to clear the offending entry. Consider making removal (and the open-batch scan) tolerant of an unresolvable member — e.g. skip/`continue` on lookup failure rather than aborting — so cleanup is always possible. The FakeBots mock always returns a match, so this path isn't currently exercised by a test.

### 2. `add_bot` reports a stale `claimed_after` (low — confirm intent)
`remove_bot` re-reads the layout after the write to compute `claimed_after` (`operations.py:402-405`), but `add_bot` returns `WhitelistMutationResult(True, claimed, claimed, updated)` (`operations.py:346`) — both `claimed_before` and `claimed_after` are the pre-write value. This is presumably fine because adding to the whitelist doesn't synchronously claim (claiming happens via reconcile), but the asymmetry with `remove_bot` is easy to misread. Worth a one-line comment noting why, or symmetric handling.

### 3. Concurrent first-time config creation surfaces as a 500 instead of a clean conflict (low)
In `commit_change`, when `row is None` and `config_id is None`, a new `CommonConfig` is inserted (`skills_pool_rollout_repository.py:54-73`). `with_for_update()` on a non-existent row takes no gap lock, so two concurrent first writes for the same env can both reach the insert; the second then fails the unique constraint with an `IntegrityError` that propagates as a 500, rather than the clean `False` → `RolloutOperationError("rollout config changed concurrently")` → 409 that every other CAS conflict yields. Low probability, but catching the integrity error and returning `False` would keep the conflict contract uniform.

### 4. `ControlBotRequest.batch_id` lacks `min_length=1` (nit)
`WhitelistAddRequest.batch_id` uses `Field(min_length=1)` but `ControlBotRequest.batch_id` (`schemas.py:48`) does not. `_entry` still rejects empty/`"*"` at the domain layer (→ 409), so it's not exploitable, just an inconsistent 422-vs-409 boundary for the same field.

### 5. `get_snapshot` reads config and audit log non-atomically (low / note)
`get_snapshot` parses the config in one session (`_get_config`) and then reads the audit log in a separate session (`list_audit_events`). A concurrent write can land between the two, so precondition derivations that combine both (`_accepted_batch`, `_open_batches`) can act on a slightly torn view. The final `commit_change` CAS still guards config integrity via `expected_revision`/`expected_value`, and the audit log is append-only/monotonic, so I don't see a correctness hole — noting it in case a stricter read-consistency guarantee is desired later.

None of these block the core behavior; #1 is the one I'd most want addressed since it affects operational recoverability. Nice work on the schema tightening and the audit trail.

---

FreddieSun:
Thanks for the careful review. I addressed items 1, 3, and 4 in `251307a0`.

1. Orphaned whitelist entries no longer wedge the control plane:
   - `remove_bot` can delete an exact `owner_id + bot_id` entry even when the Bot row is no longer resolvable.
   - `_open_batches` keeps an orphaned member's batch open instead of raising or silently skipping it, so promotion remains blocked until the stale entry is explicitly removed.
   - Added regression coverage for both cleanup and batch blocking.

3. Concurrent first-time config creation now maps `IntegrityError` to the existing CAS conflict result, preserving the API's clean 409 contract. Added a repository regression test.

4. `ControlBotRequest.batch_id` now uses `Field(min_length=1)`, matching whitelist and batch-acceptance validation. Added schema coverage.

Validation:

- Focused review regression suite: `20 passed`
- Full Skills Pool/operator/DI/architecture suite: `127 passed`
- Ruff and `git diff --check`: passed

I left items 2 and 5 unchanged as requested: whitelist admission does not synchronously claim, and the final write CAS still protects the non-atomic config/audit read observation.

FreddieSun:
Follow-up on items 2 and 5:

**2. `claimed_after` on whitelist admission**

This is intentional. Adding an exact Bot entry only commits rollout admission configuration; it does not synchronously claim the migration. Claiming is performed later by the persistent reconciliation path after a lifecycle wake-up. Therefore this API operation itself guarantees `claimed_after == claimed_before`. Re-reading after the write could merely observe a racing asynchronous reconciliation and would misleadingly attribute that state change to admission. I added an explicit code comment in `ceeb383f` to pin this contract.

**5. Config and audit reads**

Agreed that `get_snapshot` can briefly expose an older config together with newer append-only audit rows. We are accepting that read-consistency boundary in this change:

- configuration and its audit event are committed atomically by each writer;
- audit history is append-only;
- every subsequent mutation performs the full config revision/enable/value CAS;
- if another writer commits between the two reads, the final mutation fails cleanly with 409 instead of overwriting state.

So the possible torn view affects transient presentation/precondition computation only, not committed rollout correctness. A strict snapshot would require widening the repository and transaction boundary, which we are intentionally deferring until there is a concrete strict-read requirement.

Validation for the comment-only clarification: `14 passed`, Ruff and `git diff --check` passed.

totalfrank (approved):
## Approving — all review points addressed

Re-reviewed after the three follow-up commits (`251307a`, `9336191`, `ceeb383`). Every item from my earlier review is resolved, each with a matching regression test:

- **Orphaned whitelist entry wedging the control plane (my main concern):** `remove_bot` now tolerates an unresolvable bot so stale entries can always be cleaned up, and `_open_batches` keeps an orphaned member's batch *open* (fail-closed) rather than raising — promotion stays blocked until an operator removes the entry, but the operator can now actually remove it. Covered by `test_deleted_bot_can_still_be_removed_from_whitelist` and `test_deleted_bot_keeps_its_batch_open_until_operator_removes_it`.
- **Concurrent first-time config insert:** `commit_change` now catches `IntegrityError` and reports it as a clean CAS conflict (`False` → 409) when `config_id is None`, re-raising otherwise. Covered by `test_concurrent_first_config_insert_is_reported_as_cas_conflict`.
- **`add_bot` claim-state asymmetry:** clarified as an intentionally asynchronous claim step (comment + docs), with `remove_bot` now handling the unresolved-scope path explicitly.
- **`ControlBotRequest.batch_id`:** now `Field(min_length=1)`, consistent with `WhitelistAddRequest`.

The additional DDL alignment (OceanBase `TIMESTAMP` with the SQLite variant preserved, plus `gmt_create`/`gmt_modify` on the audit table) is a reasonable standards fix and orthogonal to the rollout logic.

Locally: focused Skills Pool / endpoint / DI / repository suite passes (76 tests) and `ruff check` is clean on the changed files. No remaining blocking issues. Nice work. LGTM 👍

One trivial, non-blocking nit for a future pass: the audit table uses `gmt_modify` while `SkillMigrationQuarantineModel` uses `gmt_modified` — worth aligning the column name if there's a house convention, but not worth holding this PR.

---

PR #454, opened 2026-07-26
Remove the retired ac_file table and its dead call graph

totalfrank:
## Why

`ac_file` was the teclaw workspace-file metadata mirror: the file manager recorded a row per uploaded file so compose could emit it as a `{store, path}` ref in the `BotConfigArtifact`.

That contract is already gone. Teclaw moved to per-file writes forwarded straight to the engine, and the running container now owns its live files (gathered into the next version at promotion). Nothing has written or read a row since:

- `BotFileServiceFactory` is bound in `ResourcesModule` but injected by **no** production caller — the only consumers left were DI test fakes.
- `ComposeInputCollector.bot_files()` had already degraded to an unconditional `return []`, with a docstring saying it was "kept because it's part of the protocol".
- The endpoint tests asserted the *absence* of rows (`"teclaw upload must NOT record an ac_file row"`), i.e. they pinned that the write path never touches the table.

So the table, its ORM model, its repository, its service, and the compose hook that consumed it are all dead.

## What changed

Removed:

- `core/files/` — `FileRecord`, `FileRepositoryProtocol`, `BotFileService`, `BotFileServiceFactory`, README
- `plugins/file_repository.py` — the sole `FileRepositoryProtocol` implementation
- `FileModel` (`__tablename__ = "ac_file"`) from `plugin_api/models.py`
- The `FileRepositoryProtocol` / `BotFileServiceFactory` bindings in `ResourcesModule`
- `bot_files()` from the `ComposeInputCollector` protocol, its collector implementation, and the composer's resource-ref concatenation

Comments elsewhere that explained teclaw's behavior *by contrast* with the `ac_file` mirror (`provider_behavior.py`, `build_stage.py`, `teclaw_file_promotion.py`) are reworded rather than deleted — the explanation is still useful, the reference to a nonexistent table is not.

## Test and manifest fallout

- Deleted `test_bot_file_service.py` and `test_file_repository.py` — both covered only removed code.
- `test_resources_teclaw_writes.py`: dropped the now-vacuous "no `ac_file` row" assertions and renamed the scenarios accordingly; the write-forwarding coverage (upload path shape, mkdir `.keep`, delete forward, arca regression) is unchanged. The mkdir and arca assertion helpers reduced to a bare `success is True`, which `ExpectSuccess` already checks, so they were dropped rather than left as duplicates.
- Removed `bot_files` from the two fake collectors and the `test_bot_files_always_empty` unit test.
- Removed the `BotFileServiceFactory` fakes from `test_http_adapters_use_resolver.py`.
- `scripts/ci/singlebox_coverage_modules.yaml`: dropped `core/files/` from the `files` module's `core_paths` (the module survives on `resource_file_service.py`).
- `flow_coverage.py`: dropped `"files"` from `SINGLEBOX_E2E_EXEMPT` — `test_exempt_list_has_no_stale_entries` fails on an exempt entry that is no longer a real core module.

### On the coverage threshold

`files.thresholds.core_min_percent` is left at `63.52`. Measured on the pre-change tree, `core/files/` sat at 53% (57/108 statements, all of it import-time), i.e. below the blended figure — so `resource_file_service.py` alone is strictly above 63.52%, and removing the lower-coverage component can only raise the module's core percentage. The gate is floor-only (`actual < minimum` fails), so it stays satisfied.

CI has since confirmed this: the Singlebox coverage run on this branch reports `files coverage: core=71.20%`, comfortably above the 63.52 floor. Left un-ratcheted deliberately — the manifest is not maintained at strict parity with the last measured value (the same run shows `devices` at 53.41% against a 43.36 floor, and `cron` at 48.61% against 41.84), so 71.20 against 63.52 sits inside the existing spread.

## Verification

- Full backend suite: **8954 passed, 3 skipped** (`tests/community`, excluding acceptance which needs the live stack)
- Architecture suite green, including the module-boundary and E2E-coverage gates
- `ruff` on every touched file reports only pre-existing findings (unused `resp` locals and an unused `json` import in `test_http_adapters_use_resolver.py`, all present on `dev`)
- No remaining reference to `ac_file`, `FileModel`, `BotFileService`, `FileRepositoryProtocol`, or `bot_files` anywhere in the tree
- All seven CI checks green on `c3a2cf0`

## Note for reviewers

This removes Python code and the ORM declaration only. The `ac_file` table itself still exists in OceanBase — prod DDL is applied out-of-band (per the deleted README), so dropping the physical table is a separate operational step, not part of this diff.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BejNcEAJnJLV8baDsGbwCw)_


FreddieSun (approved):
lgtm

PR #456, opened 2026-07-26
Tenant isolation foundation (Session 0, Stage 1): mechanism + ac_bots

totalfrank:
## What

Session 0 of the `/openapi/v1/bots` public API work — the tenant-isolation
foundation that the per-category handler sessions depend on.

**Currently in this PR: the SDD spec only** (`specify` phase). Implementation
follows in subsequent commits after the spec and plan are reviewed.

`src/backend/specs/2026-07-26-tenant-isolation-foundation/spec.md`

## Why

The public API surface exists as route definitions with no implementations. Its
callers are external registered tenants; the internal `/api/...` callers are
not. Both are served by the same application, backed by the same tables,
repositories and services.

So isolation cannot be added endpoint-by-endpoint as each public category is
implemented — the first implemented public endpoint would read internal data. It
has to exist underneath both surfaces first.

Isolation is currently absent, not partial: no bot record carries a tenant and
no query filters by one. The nearest existing concept, the `env` column, is
process-global (`utils/env_utils.py:68` reads `SERVER_ENV`/`REAL_SERVER_ENV`/
`ALIPAY_APP_ENV`) and cannot vary per caller, so it cannot stand in.

## Scope of Stage 1

Mechanism plus `ac_bots`. Later stages apply the same mechanism to
`ac_resource`, `ac_channel_config` and the skill/mcp/routine tables.

Planned approach: a request-scoped tenant carrier read by the repository through
a private predicate helper, mirroring how `plugins/bot_repository.py` already
handles `env` — ambient `_env()` at line 135 for the common path, with
`get_live_by_id_owner_and_env(env=...)` beside it as the explicit-override
variant. This keeps repository and service signatures unchanged, so no legacy
call site moves.

This departs from the handoff doc's §4.2 step 2, which prescribes an explicit
tenant parameter on every read/write/list/count/delete signature; `BotRepository`
alone has 28 protocol methods. The departure is recorded in the spec.

## Notes for reviewers

Three findings from grounding the handoff doc against HEAD that changed the
plan:

- **`OperatorContext.tenant_id` is not free to repurpose.** The handoff calls the
  hardcoded `tenant_id="default"` at `bot_management/services/bot_service.py:100`
  inert. It is inert as far as SQL goes, but it flows
  `adapters/http/devices/dependencies.py:31` →
  `service_bot/services/baas_service.py:1397` `deploy_config["poolab_tenant_id"]`
  — a sandbox-allocator hint sent upstream. Different concept, left untouched.
- **`ac_file` is dead.** `BotFileServiceFactory` is bound in DI
  (`di/modules/resources_module.py:66`) but `.create()` has no production caller,
  nothing injects `FileRepositoryProtocol` outside tests, and
  `tests/community/endpoints/test_resources_teclaw_writes.py` pins the absence of
  `ac_file` rows as the contract. Excluded here; removal is a separate cleanup.
- **No migration file is checked in.** Production DDL is submitted and executed
  on the platform out-of-band. `plan.md` will therefore carry the exact required
  column shape and the ordering constraint — the `ALTER TABLE` must land before
  this code reaches an environment, since the repository predicate references a
  column the ORM does not create in prod.

## Testing

No code changes yet, so no behavior to test in this commit. The implementation
commits will carry:

- context-carrier unit tests (default, set/reset, no leak across requests
  including after an error)
- a test proving the request's tenant is visible inside a handler, not only in
  the layer that set it
- cross-tenant isolation on the bot repository, red on unfixed HEAD
- the existing internal API suite green and unmodified

`coverage_baseline.txt` is untouched throughout: this work adds no routes and
implements no `/openapi/v1` handler.

---
_Generated by [Claude Code](https://claude.ai/code/session_011SqDH8bzzVrweu7RwWWS94)_

### DIFF
--- a/src/backend/specs/2026-07-26-tenant-isolation-foundation/tasks.md
+++ b/src/backend/specs/2026-07-26-tenant-isolation-foundation/tasks.md
@@ -2,19 +2,24 @@
 
 > Status legend: `[ ]` todo · `[~]` in-progress · `[x]` done · `[!]` blocked
 
-## Task 1: Spike — confirm listener covers ORM writes
+## Task 1: Spike — confirm listener covers ORM writes  `[x]`
 - **Goal:** Empirically settle whether a `do_orm_execute` listener applying
   `with_loader_criteria(BotModel, ...)` also constrains `Query.update()` /
   `Query.delete()`, or whether the write methods need an explicit filter.
-- **Files:** throwaway probe under `tests/community/plugins/` — no production
-  change in this task.
+- **Files:** throwaway probe (scratchpad, not committed) — no production change.
+- **Finding (SQLAlchemy 2.0.51): COVERED.** A prototype listener adding
+  `with_loader_criteria(Bot, Bot.tenant == current, include_aliases=True)` in
+  `do_orm_execute` constrains the exact write shape `bot_repository` uses —
+  `db.query(Model).filter(...).update(values, synchronize_session=False)` — and
+  `Query.delete()`. Probe results: a cross-tenant `Query.update()` returned
+  rowcount 0 and left the row unchanged; a cross-tenant `Query.delete()`
+  returned rowcount 0 and left the row present; a cross-tenant read returned
+  nothing. **Task 5 needs no explicit `_avernet_tenant()` filter on the write
+  methods — the single read listener covers reads, updates, and deletes.**
 - **Done when:**
-  - [ ] A minimal reproduction inserts two bots under different tenants and runs
-        `update_by_owner` / `soft_delete_by_owner` against the other tenant's bot
-        with a prototype listener installed.
-  - [ ] The finding is recorded under this task: **covered** (read path alone
-        suffices) or **not covered** (write methods need an explicit
-        `_avernet_tenant()` filter — Task 5 adopts it).
+  - [x] Minimal reproduction run against different-tenant rows with a prototype
+        listener installed.
+  - [x] Finding recorded: **covered** — read path alone suffices for writes too.
 - **Depends on:** —
 
 ## Task 2: Tenant context primitive
--- a/src/backend/specs/2026-07-26-tenant-isolation-foundation/tasks.md
+++ b/src/backend/specs/2026-07-26-tenant-isolation-foundation/tasks.md
@@ -128,20 +128,22 @@
         ordering tests (11) still pass.
 - **Depends on:** Task 2, Task 6
 
-## Task 8: Request-spawned work inherits the tenant
+## Task 8: Request-spawned work inherits the tenant  `[x]`
 - **Goal:** In-request background threads observe the request's tenant.
 - **Files:** `core/bot_management/services/bot_service.py` (3 sites),
   `core/service_bot/services/bot_publish_service.py`,
   `core/bot_collaborator/services/collaborator_service.py`,
-  `tests/community/...`.
+  `tests/community/core/test_request_spawned_tenant_inheritance.py` (new).
 - **Done when:**
-  - [ ] The five `threading.Thread` targets listed in the plan are wrapped with
-        `bind_current_avernet_tenant` (or the tenant is otherwise captured and
-        re-established inside the thread).
-  - [ ] `asyncio.create_task` sites are left unchanged (verified they inherit
-        context).
-  - [ ] Test: a bot operation performed on a spawned thread runs under the
-        spawning request's tenant.
+  - [x] The five `threading.Thread` targets are wrapped with
+        `bind_current_avernet_tenant` (do_allocate, _update_cron_workflow,
+        _refresh_codefuse_token_on_device, _do_restart, collaborator _runner).
+  - [x] `asyncio.create_task` sites left unchanged — task creation copies the
+        context (noted in the plan; no code change).
+  - [x] Test (4): the three spawn shapes I touched (target-only, target+kwargs,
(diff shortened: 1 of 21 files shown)


xianmuyq (approved):

FreddieSun (commented):
发现两处会破坏 tenant isolation 完整性的 P1 问题，详见 inline comments。

src/backend/src/agentclaw/community/plugin_api/models.py
```diff
@@ -89,6 +98,94 @@ def to_dict(self) -> dict:
...
+    ):
+        return
+    if orm_execute_state.execution_options.get("skip_avernet_tenant_guard"):
+        return
+    orm_execute_state.statement = orm_execute_state.statement.options(
+        with_loader_criteria(
+            BotModel,
+            BotModel.avernet_tenant == get_current_avernet_tenant(),
+            include_aliases=True,
+        )
```
FreddieSun, line 150: **[P1] 原生 SQL 路径绕过了这里的 tenant guard**

这个 listener 只能覆盖 ORM statement，但 `core/bot_public/services/bot_discover_service.py:299-305` 仍通过 raw cursor 直接执行 `FROM ac_bots`，且 SQL 没有 `avernet_tenant` 条件。真实 tenant resolver 启用后，其他租户的推荐项可能因此返回内部/其他租户的公开 Bot，违反 spec 对所有 Bot read 的隔离要求。建议将该查询改走 ORM，或显式绑定 `get_current_avernet_tenant()` 参数，并补一条覆盖该真实路径的跨租户回归测试。

src/backend/specs/2026-07-26-tenant-isolation-foundation/tasks.md
```diff
@@ -0,0 +1,246 @@
...
+- **Wrapped now** (in-request threads whose body touches `BotModel`):
+  - `bot_dormant/activate_service.py` `_reactivate_async` (reactivate flow).
+  - `bot_management/services/bot_service.py` `create_bot_instances` — the
+    device-allocation `ThreadPoolExecutor` (bind the submitted callable).
+  - `desktop_bot/services/desktop_bot_service.py` `_poll_publish_progress`
+    (calls `_bot_repo.get_by_id_and_owner` / `update_by_owner`).
+- **Deferred** (out of Stage-1 scope — touch non-bot tables, not `ac_bots`):
+  `bot_public_service.py` auth-relationship rebuild executors;
+  `device_service.py` MCP-sync / service-start threads. Revisit when those data
+  categories get their own guards (later stages).
```
FreddieSun, line 229: **[P1] 这里对遗漏线程的判断与实际调用链不符**

这些路径并非都只触及非 Bot 表：`bot_public_service.py:176 -> :188` 和 `:1119-1122` 会在线程/线程池里查询 `BotRepository`；`device_service.py:729` 的失败路径以及 `baas_publish_poller.py:55` 后续都会更新 `BotModel`，`device_service.py:1545` 也会在线程中通过 `_bot_query` 读取 Bot。裸 `Thread`/`ThreadPoolExecutor` 不复制 `ContextVar`，非默认租户会退回 `teamclaw`。请统一绑定当前 tenant，并用这些真实调用点补回归测试后再将 F1 标为完成。

src/backend/src/agentclaw/community/plugin_api/models.py
totalfrank: Confirmed — this is a real gap. The `do_orm_execute` read guard only fires for ORM statements, so this raw `cursor.execute("… FROM ac_bots …")` at `bot_discover_service.py:299-311` is entirely uncovered: no `avernet_tenant` predicate is added, and once the real tenant resolver lands a caller's discovery would surface other tenants' public bots.

I audited all of `core/` + `plugins/` — this is the only raw-SQL `ac_bots` path in-tree, so the fix is bounded, but the broader point stands: raw SQL escapes the "can't-forget" guarantee. Fix approach: migrate this query to an ORM-based `BotRepository` method (so the guard covers it) rather than hand-adding a predicate, plus a cross-tenant regression test on that real path. Tracking a follow-up to also consider a lint/guard against raw `ac_bots` SQL.

Since the PR is merged, this lands in a new PR off `dev`.

---

src/backend/specs/2026-07-26-tenant-isolation-foundation/tasks.md
totalfrank: Confirmed — you're right, and my F1 classification was wrong. I re-traced each call chain and every one touches `BotModel` inside a bare `Thread`/`ThreadPoolExecutor` (which don't copy the `ContextVar`, so a non-default tenant falls back to `teamclaw`):

- `bot_public_service.py:176` → `_sync_access_mode_and_relations_or_raise` calls `_bot_repository.get_by_id_and_owner` (:188); `:1119` submits `get_by_id_and_owner` to a `ThreadPoolExecutor`.
- `device_service.py:729` `start_service_async` → `_mark_service_start_failed` → `_bot_query.get_by_binding_id`; `:1545` `report_device_alive` `_run` → `_bot_query.get_by_binding_id`.
- `baas_publish_poller.py:55` `_poll` updates `BotModel` on publish completion.

Fix: wrap each target with `bind_current_avernet_tenant` and add regression tests exercising these real call points, then correct the F1 record (it was marked done prematurely). Landing in the same new PR off `dev` as the raw-SQL fix.

---

PR #471, opened 2026-07-27
fix(bcs): decouple HumanInput start from caller auth

RaymondHX:
## What changed

- allow a `human_input` state-machine run to start without an authenticated Human when the selected session already has a `Present Human`
- materialize a server-authenticated Human into the selected session as `Observer + Present`, restoring an existing absent Human when needed
- forward trusted Human identity through the group auto-start path
- keep pending-node reads and Human responses protected by authenticated `Present Human` membership
- document the distinction between run-start validation and Human response authorization

## Root cause

The runtime treated the absence of `authenticated_human` as an authentication failure for every definition containing `human_input`. ServiceInvocation may legitimately be started by a Bot or system caller, so the real start invariant is session readiness: at least one `Present Human` must be available.

## Validation

- `git diff --check`
- added runtime regression coverage for no-Human rejection as HTTP/business invalid input, existing Present Human startup, authenticated Human enrollment, and absent Human restoration
- added HTTP contract coverage for forwarding authenticated Human identity during group auto-start
- local tests intentionally not run per requested workflow; PR CI is the validation source

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/collaboration_runs.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/collaboration_runs.rs
@@ -253,7 +253,7 @@ async fn authenticated_human(
         .ok_or_else(unauthenticated_response)
 }
 
-async fn optional_authenticated_human(
+pub(super) async fn optional_authenticated_human(
     state: &HttpAppState,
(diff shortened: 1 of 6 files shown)


vzvince (approved):
LGTM

PR #473, opened 2026-07-27
fix: support legacy skills pool rollout config CAS

FreddieSun:
## What changed

- Canonicalize valid legacy Skills Pool rollout config values during CAS comparison.
- Keep semantic CAS fail-closed for real configuration changes.
- Upgrade a legacy config to the complete persisted shape on its first successful operator mutation.
- Add an operator endpoint regression covering the legacy `POST /rollout/feature` path, idempotency, audit persistence, and safe default controls.

## Root cause

Legacy rollout records may not contain `full_rollout_engines`. The service normalized that missing optional field to `[]`, while the repository compared the normalized expected value with the raw database JSON. The two semantically equivalent values therefore failed CAS forever with `rollout config changed concurrently`.

## Compatibility

No schema or DDL changes. Invalid config still fails closed, and semantic changes to revision, enable state, engine promotion, whitelist, or controls still fail CAS.

## Validation

- `ruff check` on all changed Python files
- 48 rollout, repository, HTTP adapter, and endpoint tests
- 5 module-boundary and protocol-contract architecture tests


totalfrank (approved):
LGTM.

PR #484, opened 2026-07-27
fix auth on api resources  url and node

liaoxianhao:
fix auth on api resources /api/resources/url and /api/resources/node


totalfrank (approved):
LGTM.

PR #487, opened 2026-07-27
fix: retire Legacy Skills storage without whole-corpus bridges

FreddieSun:
## Summary

- replace unsupported `RENAME_EXCHANGE` cutover with an ordinary rename of Legacy local storage into generation-owned quarantine
- keep Pool canonical local/repo as the only corpus storage after activation
- remove whole-corpus `skills-local` and `skills-repo` bridges from the active root; retain only per-skill mappings and unmanaged external symlinks
- preserve Pool and late Legacy writes with staging snapshots, three-way reconciliation, residue replay, and retryable finalization
- keep cutover fail-closed: failures do not publish the active marker and evidence remains in quarantine

## Why

The real sandbox filesystem returns `EINVAL` for `renameat2(RENAME_EXCHANGE)`. Atomic exchange is therefore not a viable prerequisite. More importantly, retaining whole-corpus bridges under the active root is unsafe for recursive OpenClaw skill discovery because it exposes unactivated repository and local skills.

The approved boundary accepts a short transient cutover gap and a narrow best-effort write race, while requiring the final filesystem shape and product activation semantics to be correct.

## Final layout contract

- `skills-pool/skills-local` and `skills-pool/skills-repo` are canonical corpus storage
- the active `skills` root contains only activated per-skill symlinks pointing directly into Pool
- no `skills/skills-local` or `skills/skills-repo` whole-corpus bridge remains after commit
- unmanaged external symlinks are left untouched

## Safety and retry behavior

- exact staging manifest detects Legacy writes arriving after the initial copy
- Pool writes during synchronization are preserved
- generation quarantine has explicit ownership evidence and rejects unknown occupants
- residue directories are replayed on every retry
- finalizing retries revalidate ownership and storage retirement before writing the active marker

## Validation

- focused migration/contract suite: 95 passed
- full Engine Community suite: 2015 passed, 5 skipped
- `git diff --check`: passed
- Ruff: no newly introduced findings; remaining five findings reproduce on the previous HEAD
- independent Standards review: PASS
- independent Spec/correctness review: PASS

## Accepted boundary

This is intentionally not a full writer-quiesce or Engine Layout Descriptor refactor. Without a shared write fence there remains a very narrow fingerprint-check-to-replace race, accepted for this migration. The implementation never falls back to a final whole-corpus bridge.

### DIFF
--- a/src/engine/src/engine/community/plugins/aicoding/tests/test_pool_layout.py
+++ b/src/engine/src/engine/community/plugins/aicoding/tests/test_pool_layout.py
@@ -14,7 +14,6 @@ from engine.community.plugins.aicoding.layout_pool import (
     verify_aicoding_pool_mappings,
 )
 
-
 PREPARATION_ID = "2a958f59-8cf4-4413-a267-7d56d3382f23"
 
 
@@ -142,8 +141,8 @@ def test_aicoding_activation_switches_local_and_keeps_repo_namespace(
     )
 
     assert result.status is PoolActivationStatus.COMMITTED
-    assert legacy_local.is_symlink()
-    assert legacy_local.resolve() == pool_local.resolve()
+    assert not legacy_local.exists()
+    assert not legacy_local.is_symlink()
     assert not local_bridge.exists()
     assert not local_bridge.is_symlink()
     assert not repo_bridge.exists()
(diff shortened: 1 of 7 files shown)


totalfrank (approved):
LGTM.

PR #498, opened 2026-07-27
feat(singlebox): add rule-driven bots

quechao76:
Add a rule-driven bot runtime for Singlebox that can run deterministic bots without invoking an LLM.

- Support fixed, random reply, echo, random number, task worker, and supervisor behaviors
- Run multiple logical rule bots in a single host process
- Preserve bot identities and sessions across restarts
- Keep chat.inject silent and support configurable response delays
- Let the task coordinator announce work before dispatching tasks and summarize worker results
- Add a version 2 profile manifest, example bot profiles, validation, and launcher tests
- Preserve compatibility with existing version 1 OpenClaw profiles


carolynli (approved):
LGTM.

PR #508, opened 2026-07-28
fix(community): route notify probe by device_provider (BaaS ENGINE_HTTP_500)

maicaisongxiaocong:
## Background

`GET /api/v1/notify` (aggregated pending HITL notifications across a user's bots) returned `ENGINE_HTTP_500` for every BaaS bot, while Arca bots succeeded.

## Root cause

The probe reached each bot's engine through `SandboxRuntimeClient.build_proxy_request`, which only knows the Arca proxypass URL formula:

```
{proxy_base}/proxypass/ARCA_{sandbox_id}:{port}{api_path}
```

For any `device_provider=baas` bot this builds an unroutable target (`ARCA_BOT-…:20003`), so the proxypass gateway returns 500 and the bot is recorded as `ENGINE_HTTP_500`. The request never reaches the engine. Desktop bots (`device_id` shaped `BOT-…`) were the visible casualty, but the gap affects all BaaS bots — chat / cron already reach these bots through a different, provider-aware transport, so only the notify probe was wrong.

Routing is decided by `ac_entity_device_binding.device_provider` (`baas` / `arca` / `teclaw` / `local`), resolved centrally by `DeviceContextResolver`. The notify probe bypassed that resolver and hardcoded the Arca-only formula.

## Fix

Route the probe through `DeviceContextResolver` + `DeviceAdapterTransport` — the same provider-driven path chat / cron use — so each provider reaches its engine via the correct transport:

- `baas` + `bot_type=desktop` → BaaS invoke-http (`{baas_base}/api/v1/bots/{tenant}/{bot_uuid}/invoke-http/{port}{path}`)
- `baas` cloud → `BaasService.invoke_http`
- `arca` → Arca proxypass (unchanged behavior; no regression for legacy Arca bots)
- `teclaw` → container URL
- `local` → direct `127.0.0.1:20003` (preserved; the test/singlebox transport is cron-only and does not serve `/api/notify`)

### Changes
- `NotifyBotLister.list_bot_mappings` returns `NotifyTarget(bot_id, bot_name, owner_id, sandbox_id)`. `owner_id` is the binding owner (the real owner for collaborator bots) so the resolver picks the right binding. `sandbox_id` is kept only as a response echo and is intentionally **not** used for routing.
- `notify_router` injects `DeviceContextResolver` + `DeviceAdapterTransport` (replaces `SandboxRuntimeClient`) and probes via `resolver.resolve_for_bot(bot_id, owner_id)` → `transport.invoke(conn_info, "GET", "/api/notify")`.
- Local provider keeps a direct httpx probe.

## Error contract (preserved)

`ENGINE_HTTP_{code}` (incl. 500, now from `DeviceAdapterHTTPStatusError`), `ENGINE_HTTP_404`, `ENGINE_TIMEOUT`, relay `"CODE: detail"` passthrough (e.g. `RELAY_UNAVAILABLE`), `NOTIFY_PROBE_ERROR`, and a benign empty record when a bot has no active binding (so the frontend still sees the bot exists).

## Tests

`DEPLOY_PROFILE=test pytest tests/community/api/aicoding tests/community/core/notify tests/community/plugins/local tests/community/core/devices` → **884 passed**.

`_probe_bot_notify` unit tests rewritten to mock resolver + transport; added `test_probe_desktop_routes_through_transport` asserting a desktop bot goes through the transport's invoke-http branch instead of the legacy Arca proxypass.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/aicoding/notify_router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/aicoding/notify_router.py
@@ -3,7 +3,10 @@ Notify router — aggregates pending HITL interactions from all bots via Engine.
 
 GET /api/v1/notify:
   1. Get user's active device bindings (bots with sandboxes)
-  2. Parallel probe each bot's Engine /api/notify endpoint via proxypass
+  2. Parallel probe each bot's Engine /api/notify endpoint — routed by
+     ``device_provider`` via ``DeviceContextResolver`` (same path chat / cron
+     use), so BaaS (desktop + cloud), Arca, teclaw and local each reach their
+     engine through the correct transport.
   3. Return grouped result: [{bot_id, bot_name, sandbox_id, notifications}]
 """
 from __future__ import annotations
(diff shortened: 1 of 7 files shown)


totalfrank (approved):
LGTM.

PR #510, opened 2026-07-28
fix(bcs): add milliseconds to common error timestamps

RaymondHX:
## Summary

- add three-digit millisecond precision to timestamps written by the `common-error` logging output
- keep the existing second-level timestamp format unchanged for other file outputs and the console
- extend the common error logging test to validate the `YYYY-MM-DD HH:mm:ss.SSS` format

## Why

The dedicated `common-error.log` currently records timestamps only to the second, which makes ordering closely spaced error events difficult during incident investigation.

## Impact

Only the output named `common-error` changes its timestamp format. Existing routing, log levels, rotation behavior, other log files, and console output are unchanged.

## Validation

- `cargo test -p bcs --lib logging::tests` — 4 passed before rebasing onto the latest `dev`
- rebase onto latest `origin/dev` completed without conflicts

A post-rebase rerun was started but stopped while Cargo was downloading dependency updates introduced by the latest `dev`; it had not reached compilation or reported a test failure.

### DIFF
--- a/src/bcs/crates/bootstrap/bcs/src/logging.rs
+++ b/src/bcs/crates/bootstrap/bcs/src/logging.rs
@@ -189,6 +189,18 @@ impl<'a> MakeWriter<'a> for RotatingFileWriter {
 
 // ─── Init ───────────────────────────────────────────────────────────────────
 
+fn timer_for_output<F: Clone>(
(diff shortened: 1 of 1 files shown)


vzvince (approved):
LGTM

PR #511, opened 2026-07-28
fix(bcs): preserve run correlation for channel replies

RaymondHX:
## Summary

- reuse the upstream BCS request/run ID in the BCN plugin instead of generating an unrelated UUID
- rebind channel source-message correlation when a legacy bot ACK returns a different run ID
- keep task-alias handling independent from source-message correlation

## Why

The plugin could return a new run ID while BCS still stored the DingTalk source message under the original request run. The final channel reply then could not find the reaction state, leaving the acknowledgment reaction visible.

## Impact

The compatibility path only moves channel source-message metadata. Implementations that do not support this metadata retain the service API's default no-op behavior, so non-channel flows are unchanged.

## Validation

- BCN plugin lint and build passed
- BCN plugin tests: 53 passed
- `cargo test --manifest-path src/bcs/Cargo.toml -p bcs-message-flow`: 201 passed
- `cargo test --manifest-path src/bcs/Cargo.toml -p bcs-ws`: passed
- `git diff --check`: passed

### DIFF
--- a/src/bcs/crates/adapters/ws/bcs-ws/src/bot/dispatcher.rs
+++ b/src/bcs/crates/adapters/ws/bcs-ws/src/bot/dispatcher.rs
@@ -549,8 +549,8 @@ async fn handle_response_frame(
             );
         }
 
-        // If this is a task dispatch ACK, delegate alias registration to master_slave strategy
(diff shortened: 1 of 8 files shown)


vzvince (approved):
GLTM

PR #532, opened 2026-07-28
feat(bcs): add Bot-Human-Bot collaboration template

RaymondHX:
## Summary

- add localized `Bot-Human-Bot` collaboration template seeds for `zh-CN` and `en-US`
- register the template for backend-driven discovery in the frontend
- model the Human review step as a `human_input` node without an assignee or participant binding
- cover template listing/detail behavior and validate both seed definitions against the collaboration runtime

## Why

The frontend already loads collaboration templates and YAML definitions from Avernet. Adding this seed makes a Human-in-the-loop example available without introducing frontend-owned template data.

## Human Input behavior

- only the Bot role is declared in `participants` and bound during group creation
- the Human is the current group member, so the `human_input` node has no `assignee`
- the Human node has an explicit 24-hour timeout and uses Judge outcomes to continue to the final Bot node

## Validation

- `cargo test -p bcs-collaboration-template -p bcs-collaboration-runtime`
- 67 tests passed
- `git diff --check`

### DIFF
--- a/src/bcs/crates/services/bcs-collaboration-runtime/tests/definition_validation.rs
+++ b/src/bcs/crates/services/bcs-collaboration-runtime/tests/definition_validation.rs
@@ -156,12 +156,14 @@ fn authoring_validation_rejects_fields_not_supported_by_the_current_runtime() {
 #[test]
 fn validates_collaboration_template_seed_definitions() {
     for file_name in [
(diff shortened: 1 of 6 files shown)


vzvince (dismissed):
LGTM

vzvince (approved):

PR #535, opened 2026-07-28
fix: resume Skills Pool repair after committed cutover

FreddieSun:
## Summary

- skip the duplicate cutover-ledger CAS after an operator has resolved an UNKNOWN cutover as `pool_committed`
- still re-invoke the runtime idempotently to prove the same generation returns `ALREADY_COMMITTED`
- continue mapping, locator, and `POOL_ACTIVE` convergence after that evidence refresh
- add a regression for the exact `MANUAL_REPAIR_RESOLVED + data_plane_cutover_committed=1` state

## Root cause

The repair endpoint persisted `POOL_CUTOVER_COMMITTED` before enqueueing reconcile. Reconcile then successfully received `ALREADY_COMMITTED` from the runtime but called `record_cutover_committed()` again. That repository CAS only accepts pre-cutover/finalizing phases, so repair always stopped with `state_race_lost`.

## Tests on latest dev

- `uv run pytest tests/community/core/skills_pool -q` (171 passed)

### DIFF
--- a/src/backend/src/agentclaw/community/core/skills_pool/reconcile_service.py
+++ b/src/backend/src/agentclaw/community/core/skills_pool/reconcile_service.py
@@ -344,17 +344,20 @@ class SkillsPoolReconcileService:
                     preparation_id=probe.preparation_id,
                     evidence=cutover.to_dict(),
                 )
-            if not self._layouts.record_cutover_committed(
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #540, opened 2026-07-29
feat(engine): add depth limit to aicoding file tree

maicaisongxiaocong:
## Summary
- add a `depth` query parameter to the AICoding file-tree endpoint
- default to 3 workspace-relative levels
- treat `depth=0` as unlimited traversal
- use GNU `find -maxdepth` to avoid scanning deeper directories
- validate and test default, bounded, unlimited, and invalid depth values

## Test plan
- `PYTHONPATH=src pytest src/engine/community/api/aicoding_sessions/tests/test_router.py src/engine/community/core/aicoding/tests/test_workspace_service.py -q`
- 154 passed


totalfrank (approved):
LGTM.

PR #543, opened 2026-07-29
Rongzhi 0727

lucas-xzp:
openapi -resource、routnes、identity


scripts/ci/singlebox_coverage_modules.yaml
```diff
@@ -235,32 +235,6 @@ modules:
     thresholds:
       core_min_percent: 41.59
 
-  resources:
```
totalfrank, line 238: @cassiuscai Please take a look, if this is expected.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/identity/router.py
```diff
@@ -29,23 +39,97 @@
...
+        raise HTTPException(status_code=400, detail=str(e))
+    files = [
+        IdentityFileInfo(
+            type=IdentityFileType(ft.removesuffix(".md")),
+            exists=exists,
+            file_path=f"{IDENTITY_NS}/{ft}",
+        )
+        for ft, exists in presence
+    ]
+    return Envelope(
```
totalfrank: I have added factory methods under: agentclaw.community.adapters.http.openapi_v1.responses

take a look at this PR: https://github.com/inclusionAI/Avernet/pull/494

let's be consistent.

src/backend/src/agentclaw/community/plugin_api/models.py
```diff
@@ -230,6 +148,76 @@ def to_dict(self) -> dict:
...
+# CRITICAL (spec §6.6): with_loader_criteria takes a DIRECT EXPRESSION, never
+# a lambda — the lambda form is cached and pins the first tenant (leak).
+
+class CrossTenantInsertError(RuntimeError):
+    """An insert named a tenant other than the request's current one."""
+
+
+# Models guarded by the read listener. Both live in this file (plugin_api/models.py):
+# BotModel above, ResourceModel below.
+_GUARDED_MODELS: tuple[type, ...] = (BotModel, ResourceModel)
```
totalfrank: This PR is for all of identity, resource and routines. I only saw resource model here. Did you miss something?

src/backend/src/agentclaw/community/core/resources/service.py
```diff
@@ -267,6 +268,193 @@ async def create_node_resource(
...
+    #
+    # device_fs is an opaque duck-typed boundary the adapter resolves
+    # (DeviceFilesystemDispatcher) and forwards in. The slim service never
+    # holds a dispatcher — it only calls .write_file / .read_file /
+    # .delete_file on whatever the adapter passes. This keeps the pure-core
+    # service free of device-layer imports. ``device_fs`` defaults to None so
+    # the non-file / no-fs paths degrade gracefully (legacy parity: missing
+    # → None → handler maps to 404).
+
+    def get_resource(self, resource_id: str) -> Optional[Resource]:
```
totalfrank, line 301: Did not read the code throughly but I guess this is essentially moving things over so it can be shared across current and our new API right?

src/backend/src/agentclaw/community/core/resources/services/resource_service.py
```diff
@@ -704,6 +704,107 @@ def delete_resource(
 
         return self._repo.delete(resource_id)
 
+    async def download_resource(
```
totalfrank: Same, I assume this is essentially just moving things around so that these code can be shared across our current and new APIs right?

src/backend/src/agentclaw/community/adapters/http/openapi_v1/resources/router.py
```diff
@@ -32,73 +146,367 @@
...
+    # signature REQUIRES both keyword args (no defaults), so pass them
+    # explicitly.
+    owner_id = caller_owner_id(principal)
+    exists = await service.check_name_exists(
+        name=name,
+        resource_type=legacy_type,
+        parent_path=None,
+        user_id=owner_id,
+    )
+    return Envelope(
```
totalfrank: Same here(and everywhere else), let's leverage agentclaw.community.adapters.http.openapi_v1.responses factory methods to do it.

totalfrank (commented):
First round, please take a look.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/identity/router.py
lucas-xzp: Good call — done.
Switched all success-path returns in the identity / resources / routines routers to the responses builders (envelope / page / created / deleted), matching the bots router.This also removes the per-router _request_id_from helpers and, more importantly, fixes the trace-id source:we were reading the inbound X-Trace-ID request header (usually empty in prod, and client-forgeable), whereas responses._trace_id reads request.state.trace_id — the server-side value the tracer middleware stamps and echoes on the response. So request_id now matches the bots contract and the X-Trace-ID response header.

src/backend/src/agentclaw/community/plugin_api/models.py
lucas-xzp: Nothing missed — only ResourceModel got a new guard because identity and routines have no ORM table in this repo.

src/backend/src/agentclaw/community/core/resources/service.py
lucas-xzp: Your guess is half right. It's not "moved to be shared" — it's two parallel implementations.
core/resources/service.py is the slim service the openapi_v1 router uses;
core/resources/services/resource_service.py is the legacy one /api uses.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/resources/router.py
```diff
@@ -20,85 +41,429 @@
...
+    # "all types". BUT if the caller passes type=FOLDER (an openapi type with
+    # no legacy equivalent), we MUST NOT fall through to None-=="all types" —
+    # that would leak every resource under a FOLDER filter. Explicit FOLDER →
+    # empty page (no legacy folders exist).
+    legacy_type = _legacy_type_for(type)
+    if type is not None and legacy_type is None:
+        # type is an openapi type with no legacy mapping (FOLDER) → no rows
+        openapi_items: list[Resource] = []
+    else:
+        items = service.list_resources(resource_type=legacy_type)
```
totalfrank, line 153: **`bot_id` is caller-supplied and never authorized on this path — cross-user resource read.**

`bot_id` comes straight off the query string and is passed to `factory.create(bot_id=...)`. The slim service scopes rows by `bolt_id == self._bot_id`, so this confines the query *to that bot* — but nothing checks that the authenticated principal actually **owns** that bot. Any authenticated caller can enumerate another user's bot resources by guessing/knowing a `bot_id`.

Note the asymmetry with the sibling handlers: `upload` / `delete` / `download` / `preview` call `resolver.resolve_for_bot(bot_id, owner_id)`, and `DeviceContextResolver` does `get_active_by_bot_and_owner(bot_id, user_id)` — so those paths get an ownership check *incidentally*, as a side effect of needing a device. The five handlers that don't need `device_fs` — `list_resources`, `check_resource_name`, `create_resource`, `get_resource`, `update_resource` — get no check at all.

Relying on a device-binding lookup for authorization is fragile in any case. This wants an explicit ownership assertion (`bot_repo.get_by_id(bot_id).owner_id == owner_id` → `BotNotFoundError`) applied uniformly to all nine handlers, ideally as a shared dependency rather than repeated per-handler.

The module docstring says the surface is gated on the auth workstream, which is fair — but `require_principal` landing does not fix this. A verified principal still doesn't constrain which `bot_id` the caller may name.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/resources/router.py
```diff
@@ -20,85 +41,429 @@
...
+    r = await service.create_url_resource(
+        name=body.name,
+        url=body.url,
+        # parent_path intentionally NOT forwarded: the openapi ResourceCreate
+        # schema carries `parent_id` (a pending follow-up — its ID-vs-path
+        # semantics aren't settled). Passing a half-defined value would risk
+        # a wrong-attribute write; link scoping by bot_id is sufficient now.
+        parent_path=None,
+    )
+    return created(_to_openapi_resource(r), request)
```
totalfrank, line 243: **Created rows are typed `URL`, but every read path filters on `LINK` — link resources are invisible to their own filters.**

`create_url_resource` builds the row via `core/resources/models.create_url_resource`, which hardcodes `resource_type=ResourceType.URL` (models.py:333). Meanwhile:

- `check_resource_name` maps openapi `LINK` → `_LegacyType.LINK` (line 184) and asks the repo for `resource_type='link'` rows.
- `list_resources` does the same via `_OPENAPI_TO_LEGACY_TYPE` (line 66).

So a resource created through `POST ""` is stored as `url` and never matches either filter. Concretely: create a link named `foo`, then `GET /check-name?name=foo&type=link` returns `exists=false`, and `GET ""?type=link` returns an empty page — while `GET ""` with no filter *does* return it (mapped back to `LINK` by `_TYPE_MAP`, which has `URL → LINK`). Duplicate detection inside `create_url_resource` uses `ResourceType.URL` so it is self-consistent, which makes the mismatch caller-visible as "check-name says available, create says 409".

`_OPENAPI_TO_LEGACY_TYPE` needs `LINK` to fan out to both `LINK` and `URL` on the read side (the comment at line 64 already flags this as "a follow-up" — but the write path is landing in the same PR, so the read side is broken on arrival, not merely incomplete).

**Separately on this call:** `user_id` and `created_by` are not forwarded even though `owner_id` is available from `caller_owner_id(principal)`. Every link created through this endpoint lands with `user_id = NULL`. That both loses ownership attribution and breaks `check_resource_name`, which *does* pass `user_id=owner_id` — the repo filters on a column that is always NULL for these rows.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/resources/router.py
```diff
@@ -20,85 +41,429 @@
...
+    (principal → user_id) will wire ``user_id`` once it lands. For now the
+    service treats them as None.
+    """
+    effective_bot_id = bot_id
+    service = factory.create(bot_id=effective_bot_id)
+    # Map openapi ResourceType → legacy ResourceType enum the slim service
+    # expects. When the caller omits ``type``, default to FILE (matches the
+    # legacy handler's most common case — the openapi check-name call shape
+    # has no FOLDER equivalent).
+    legacy_type = _legacy_type_for(type) or _LegacyType.FILE
```
totalfrank, line 184: `_legacy_type_for(type) or _LegacyType.FILE` silently swallows `FOLDER`.

`_legacy_type_for` returns `None` for two distinct reasons — "no filter requested" and "`FOLDER`, which has no legacy mapping" — and the `or` collapses both onto `FILE`. So `GET /check-name?name=x&type=folder` answers the question *"is there a **file** named x?"* and reports it as the folder's availability.

`list_resources` (line 148-152) goes out of its way to distinguish those two `None` cases, with a comment explaining exactly why conflating them is wrong. Worth applying the same discipline here — for `FOLDER` either return `exists=false` outright or 501, matching `create_resource`'s 501 for the same type.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/resources/router.py
```diff
@@ -20,85 +41,429 @@
...
 
 
 @router.get(
     "/{resource_id}/download",
     responses={200: {"content": {"application/octet-stream": {}}}},
 )
-async def download_resource(resource_id: str, principal: PrincipalDep) -> Response:
-    """Download a resource's bytes (raw, not enveloped)."""
-    raise NotImplementedError
+async def download_resource(
```
totalfrank, line 383: **Missing `@envelope_errors` — this is the only handler in the group without it, and it breaks the fail-closed 401.**

Every other handler here is decorated; `download_resource` is not. The docstring three lines below claims `owner_id` comes from `caller_owner_id`, "fail-closed — mirroring the bots router" — but with no decorator, the `MissingPrincipalError` that `caller_owner_id` raises has nothing to translate it, so it escapes as an unhandled **500**, not a 401. Today `require_principal` returns `None` unconditionally, so *every* call to this endpoint 500s.

Same for `DeviceNotBoundError` / `UnknownProviderError` out of `resolve_for_bot` (line 400), which the other four device-backed handlers map to proper status codes.

If the decorator was omitted deliberately because this returns a raw `Response` rather than an `Envelope`, that reasoning doesn't hold — `@envelope_errors` only intercepts on the exception path, and an error response should be enveloped even when the success response isn't. Worth a comment either way, since the omission reads as an oversight.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/resources/router.py
```diff
@@ -20,85 +41,429 @@
...
+    ctx = resolver.resolve_for_bot(effective_bot_id, owner_id)
+    device_fs = device_fs_dispatcher.dispatch(ctx)
+    service = factory.create(bot_id=effective_bot_id)
+    # NOTE: ``delete_resource`` on the concrete service is ASYNC now (Phase 3
+    # slim service awaits device_fs.delete_file for file resources) — must
+    # be ``await``ed. Returns False when the record is missing → 404.
+    ok = await service.delete_resource(resource_id, device_fs=device_fs)
+    if not ok:
+        raise HTTPException(status_code=404, detail="Resource not found")
+    return deleted_envelope(request)
```
totalfrank, line 376: Device resolution is unconditional, but the delete it guards is only needed for **file** resources.

`resolve_for_bot` raises `DeviceNotBoundError` when the bot has no active binding. That means a `link` (or folder) resource on an unbound bot can never be deleted — the handler dies at line 371 before `service.delete_resource` is reached, even though the service's file-vs-link branch (`if resource.is_file and resource.path and device_fs is not None`) would have taken the pure DB soft-delete path and never touched `device_fs`.

Links are creatable on a bot with no device (`create_resource` doesn't resolve one), so this is reachable: create a link, then it's undeletable until the bot happens to be bound. Since `delete_resource` already treats `device_fs=None` as "skip the FS step", resolving lazily — or catching `DeviceNotBoundError` and passing `device_fs=None` — would fix it.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/resources/router.py
```diff
@@ -20,85 +41,429 @@
...
+    if type is not None and legacy_type is None:
+        # type is an openapi type with no legacy mapping (FOLDER) → no rows
+        openapi_items: list[Resource] = []
+    else:
+        items = service.list_resources(resource_type=legacy_type)
+        openapi_items = [_to_openapi_resource(r) for r in items]
+    start = (page.page - 1) * page.page_size
+    end = start + page.page_size
+    page_items = openapi_items[start:end]
+    return page_envelope(len(openapi_items), page_items, request)
```
totalfrank, line 158: Pagination is applied in Python after loading the entire result set.

`service.list_resources` already accepts `limit` / `offset` (service.py:100-116) and the handler ignores both, fetching every row for the bot, materialising a `Resource` pydantic model per row via `_to_openapi_resource`, and then slicing. Page 1 of a 10-item page costs the same as fetching everything.

Passing `limit=page.page_size, offset=start` gets most of this back. Note the service's own `limit`/`offset` are *also* in-memory slices over a full repo read, so `total` still needs a real `count`; but pushing the bound down one layer at least stops the adapter from building N pydantic models to throw away N-10 of them.

Same pattern in `routines/router.py` (`list_routines` line 127, `list_routine_runs` line 356) — there it's less avoidable since the cron relay returns everything, so a short comment noting it's deliberate would help.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/resources/router.py
```diff
@@ -20,85 +41,429 @@
...
 @router.post("/upload", status_code=201, response_model=Envelope[Resource])
+@envelope_errors
 async def upload_resource(
     principal: PrincipalDep,
     name: str,
     content: Annotated[bytes, Body(media_type="application/octet-stream")],
+    request: Request,
+    bot_id: str = Query(..., description="Bot ID this resource belongs to."),
+    factory: ResourceServiceFactoryProtocol = Injected(ResourceServiceFactoryProtocol),
+    bot_repo: BotRepository = Injected(BotRepository),
```
totalfrank, line 255: `bot_repo` is injected but never used — same in `delete_resource` (351), `download_resource` (388), and `preview_resource` (425).

Four handlers take a `BotRepository` and none of them touch it; the bot lookup they'd need happens inside `resolver.resolve_for_bot`. Dead injections aren't free here — each one is a DI resolution per request, and `preview_resource`'s docstring even enumerates it as one of "the four injected deps" that must stay out of the served OpenAPI schema, so it's carrying test surface too.

(If the intent was for `bot_repo` to become the ownership check flagged on `list_resources` above, that would explain why it's threaded through — but then the four handlers that most need it, the ones without a resolver call, are exactly the ones that don't have it.)

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/routines/router.py
```diff
@@ -7,77 +7,353 @@
...
-    raise NotImplementedError
+    """List routines (filter + paginate).
+
+    ``status`` is an openapi query hint with no direct field on ``Routine``
+    (which only carries ``enabled``). It is currently a no-op: the adapter
+    does not expose a status filter at the list seam, so we return the full
+    set and let the client filter on ``enabled``. Wire a server-side filter
+    here only if/when the engine surfaces a status dimension.
+    """
+    owner_id = caller_owner_id(principal)
```
totalfrank, line 113: **No handler in this router carries `@envelope_errors` — the whole routines group returns 500 where the resources and identity groups return 401/4xx.**

All seven handlers here (`list_routines`, `create_routine`, `get_routine`, `update_routine`, `delete_routine`, `run_routine`, `list_routine_runs`) call `caller_owner_id(principal)` and none is decorated. `caller_owner_id` raises `MissingPrincipalError` on a `None` principal, and with `require_principal` still stubbed to return `None`, that is the outcome of *every* request to this group — surfacing as an unhandled 500 instead of the intended 401.

The two sibling routers landing in this same PR both decorate consistently (resources: all but `download`; identity: all three), so this reads as an omission rather than a decision. `CronRelayError` out of the `factory.*` calls has the same problem — it reaches the client as a 500 with no envelope.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/routines/router.py
```diff
@@ -7,77 +7,353 @@
...
+    delete must read ``deleted=False``.
+    """
+    owner_id = caller_owner_id(principal)
+    user_id = owner_id
+    nick_name = owner_id
+    result = await factory.delete_cron(
+        bot_id=bot_id, user_id=user_id, nick_name=nick_name, task_id=routine_id
+    )
+    success = bool(result.get("success")) if isinstance(result, dict) else False
+    return envelope(Deleted(deleted=success), request)
```
totalfrank, line 268: A failed delete returns HTTP 200 with `deleted: false`.

The docstring is explicit that this is deliberate ("a failed delete must read `deleted=False`"), and I follow the reasoning — but the failure modes get flattened together. The engine rejecting a published-stage delete is a 403 condition; a missing `routine_id` is a 404; a relay timeout is a 502. All three come back as `200 {"deleted": false}`, so a client can't distinguish "you may not do this" from "it isn't there" from "try again later", and naive clients that check the HTTP status will read every one of them as success.

Compare `delete_resource` in the resources router, which raises 404 when the service returns `False`. Worth aligning the two, or at least documenting why routines diverge — right now the same verb on two sibling groups in one PR behaves differently.

---

src/backend/src/agentclaw/community/core/services/identity.py
```diff
@@ -390,104 +485,211 @@ async def sync_agents_md(
...
+            content = await self.read_identity_file(
+                entity_type,
+                entity_id,
+                bot_id,
+                ft,
+                owner_id,
+                engine_type=engine_type,
+            )
+            results.append((ft, bool(content)))
+        return results
```
totalfrank, line 691: This is 16 sequential remote reads per `GET /identity/bot/{bot_id}` call, each downloading full file content to answer a boolean.

`read_identity_file` goes through `_device_read` → `device_fs.read_file` — a round-trip to the bot's device. The loop runs it once per entry in `VALID_IDENTITY_FILES` (16), serially, and each call transfers the file's entire contents only for `bool(content)` to be taken. On a `baas`/`arca` device that's 16 sequential network hops on a list endpoint.

Two things would help a lot: `asyncio.gather` over the file types instead of a serial loop, and an existence probe at the `device_fs` boundary (`stat`/`exists`) rather than a full read. If the dispatcher genuinely has no existence primitive, `gather` alone turns 16 sequential latencies into one.

Also note `VALID_IDENTITY_FILES` is a `set`, so iteration order varies per interpreter run (`PYTHONHASHSEED`). The router builds `IdentityFileList.files` directly from this order, so the public API returns the same 16 entries in a different order on every process — awkward for clients and for any test that asserts on the list. A `sorted()` here (or a tuple constant) would pin it.

---

scripts/ci/singlebox_coverage_modules.yaml
```diff
@@ -274,3 +248,8 @@ pending_modules:
     acceptance_targets:
       - tests/community/acceptance/mcp
     reason: Permission baseline and live engine connection assumptions need migration.
+  resources:
+    acceptance_targets:
+      - tests/community/acceptance/resources/test_resource_metadata_lifecycle.py
+      - tests/community/acceptance/resources/test_resource_lifecycle.py
+    reason: openapi_v1 handler wiring added device_fs service methods (upload_file/download_resource/preview_resource/delete_resource) under core/resources/ that the legacy /api/resources acceptance path does not reach. Pending openapi_v1 acceptance coverage; aligned with bot_management pending pattern.
```
totalfrank, line 255: Moving `resources` into `pending_modules` drops its coverage gates in the same PR that adds ~600 lines to the module.

The old entry enforced `core_min_percent: 39.07` and `router_min_percent: 100.0`. `pending_modules` entries carry no thresholds, so both gates are gone — precisely for the module receiving the largest share of new code here (`core/resources/service.py` gains `upload_file` / `download_resource` / `preview_resource` / async `delete_resource`, and the openapi router goes from stubs to nine wired handlers).

The stated reason is accurate — the legacy `/api/resources` acceptance path doesn't reach the new device_fs methods. But the conclusion doesn't follow: the fix for "the acceptance target doesn't cover the new code" is a new acceptance target, not removing the threshold from the code that *was* covered. As written, `core/resources/` could regress to 0% without CI noticing.

If demoting is genuinely the only way to land this, could the entry get a follow-up issue link and a target date? `CLAUDE.md` calls out coverage manifests as part of the pre-push contract that shouldn't be changed casually, and a silently-ungated module is easy to forget.

---

src/backend/src/agentclaw/community/plugin_api/models.py
```diff
@@ -123,49 +147,68 @@ class ResourceModel(Base):
...
     source = Column(String(50))
-    bolt_id = Column(String(100), default='default')
+    bolt_id = Column(String(100), default="default")
     env = Column(String(20), default=get_current_env, nullable=False)
+    # Data-isolation tenant (see utils/avernet_tenant + the guards in this
+    # module). server_default (not Python default=) so create_all emits the
+    # same DEFAULT 'teamclaw' prod's out-of-band DDL applies, backfilling
+    # existing rows and covering any non-ORM insert. Deliberately absent from
+    # to_dict().
+    avernet_tenant = Column(String(64), nullable=False, server_default="teamclaw")
```
totalfrank, line 170: The `avernet_tenant` column itself looks right — `server_default` rather than Python `default=` is the correct call for a column prod backfills out-of-band, and the reasoning is well documented.

Two notes:

1. The comment says "see `utils/avernet_tenant` + the guards in this module" — the module is `utils/avernet_tenant_guard` (as the `BotModel` comment at line 83 correctly writes it). Small thing, but the whole point of the comment is to be a pointer.

2. `avernet_tenant` being absent from `to_dict()` is deliberate and stated, and the resources read path relies on it — `_dict_to_resource` (core/resources/service.py) filters to an `allowed` key set that would reject the extra field. Worth saying *that* in the comment rather than just "deliberately absent", so the next person doesn't add it to `to_dict()` for symmetry and break the resource reads.

**Unrelated to this column:** this file (and `responses.py`, `core/services/identity.py`) carries a large amount of pure reformatting — line rewraps, quote normalisation, blank lines after docstrings. It looks like a formatter running at a different line length than the repo's. It roughly triples the review surface of these three files and will collide with anything else in flight. Could the reformatting be split into its own commit, or reverted where it isn't adjacent to a real change?

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/identity/router.py
```diff
@@ -30,33 +42,119 @@
...
+    owner_id = caller_owner_id(principal)
+    entity_type = "staff"  # personal bot owner is a staff entity
+    entity_id = owner_id
+    presence = await identity_service.list_bot_files(
+        entity_type,
+        entity_id,
+        bot_id,
+        owner_id,
+    )
+    files = [
```
totalfrank, line 66: `entity_type = "staff"` is hardcoded in all three handlers, and `entity_id = owner_id` means a caller can only ever address their own staff-scoped bot — which is the right default, but it silently makes the `proj` and `team` entity types (both in `VALID_ENTITY_TYPES`) unreachable through the public API.

That's presumably intentional for the personal-bot surface, but the inline comment (`# personal bot owner is a staff entity`) explains the assignment rather than the restriction. Worth a line in the module docstring saying `proj`/`team` identity files are out of scope for openapi_v1 — otherwise the natural next change is someone adding an `entity_type` query param without realising `entity_id` also has to stop being `owner_id`, which would reopen the addressing to arbitrary entities.

Also: `IdentityFileType(ft.removesuffix(".md"))` will raise `ValueError` → 500 if `VALID_IDENTITY_FILES` ever gains an entry the enum lacks. The two lists match exactly today, but they live in different modules with nothing tying them together. A test asserting `{f.removesuffix(".md") for f in VALID_IDENTITY_FILES} == set(IdentityFileType)` would catch that drift at CI time rather than in a 500.

---

totalfrank (changes requested):
Reviewed the 9 source files (skipped the docs/plans and the ~3.8k lines of new tests, which look thorough). The wiring work is solid and the inline documentation is genuinely unusually good — most of the design decisions I went looking for were already explained in a docstring. Comments are anchored to specific lines.

**Blocking:**

1. **`resources/router.py` — `bot_id` is caller-controlled and never authorized.** Five of the nine handlers (`list`, `check-name`, `create`, `get`, `update`) do no ownership check at all; the other four get one only as a side effect of `resolve_for_bot` needing a device binding. Any authenticated caller can read another user's bot resources. `require_principal` landing does not fix this.
2. **`resources/router.py` — created link resources are typed `URL` but every read path filters on `LINK`.** Create a link, then `check-name` says available and `?type=link` returns nothing. Write and read sides disagree on arrival.
3. **`routines/router.py` — no handler carries `@envelope_errors`.** With `require_principal` stubbed to `None`, every request to the group 500s where it should 401. `download_resource` in the resources router has the same omission.

**Worth fixing before merge:** `create_resource` drops `user_id`/`created_by` (rows land with `user_id = NULL`, which also breaks `check-name`'s own filter); `delete_resource` resolves a device unconditionally, making links on unbound bots undeletable; `check-name` collapses `FOLDER` onto `FILE`; `list_bot_files` does 16 sequential full-content device reads to compute booleans.

**Also flagged:** the `resources` coverage module moving to `pending_modules` removes both gates from the module gaining the most new code in this PR; four unused `BotRepository` injections; and a large amount of unrelated reformatting in `models.py` / `responses.py` / `identity.py` that inflates the diff.

Happy to re-review once the auth and type-mapping items are addressed.

---

totalfrank (approved):
LGTM.

PR #546, opened 2026-07-29
fix(bcs-storage-baas): delete transfer via sessions transfers path, not files

zh-bupt:
## What
Fix the baas session-file `delete` transfer path so it targets the baas `transfers` resource as a sibling of `files`, not nested under it.

## Why
`delete` built `{base_for_session}/transfers/{id}` = `.../sessions/{tenant}/{sid}/files/transfers/{id}`, but baas expects `DELETE /api/v1/sessions/{tenant}/{sid}/transfers/{transferId}` (the extra `/files` segment is wrong). As a result deletes were sent to a non-existent path and failed.

## Changes
- `bcs-storage-baas`: add `session_path_root` returning `.../sessions/{tenant}/{sid}` (the shared parent of `files` and `transfers`); route `delete` through it so the path becomes `.../sessions/{tenant}/{sid}/transfers/{id}`.
- `base_for_session` now wraps `session_path_root` (`{root}/files`), keeping `complete` / `abort` / `prepare` paths unchanged.
- `tests/delete.rs`: update the 3 mock path assertions to `/api/v1/sessions/t/sid/transfers/tid`.

## Verification
`cargo test -p bcs-storage-baas` (full suite) green:
- `delete.rs` 3 passed, `complete_abort.rs` 4 passed (confirms `base_for_session` refactor doesn't regress complete/abort `files`-rooted paths), `presign`/`prepare`/`health`/`factory`/`error_map` all pass.

### DIFF
--- a/src/bcs/crates/plugins/bcs-storage-baas/src/lib.rs
+++ b/src/bcs/crates/plugins/bcs-storage-baas/src/lib.rs
@@ -42,16 +42,21 @@ impl BaasStoragePlugin {
         Self { cfg, caps, http }
     }
 
-    /// Build the baas base path: {endpoint}/api/v1/sessions/{tenant}/{session_id percent-encoded}/files.
-    /// transfer_id and session_id are percent-encoded to be safe path segments.
-    fn base_for_session(&self, session_id: &str) -> String {
-        let sid = percent_encode_path(session_id);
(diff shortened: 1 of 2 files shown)


RaymondHX (approved):
LGTM

PR #547, opened 2026-07-29
feat(publish): add restart in-progress marker to ext

quhongwei:
Add ext.restart.restarting flag to track restart state:
- Set restarting=True in execute_restart after validation passes
- Clear restarting flag in _handle_sync_success and _handle_sync_failure

This provides a clear indicator for 'restart in progress' state, similar to the existing retry flag pattern.


totalfrank (approved):
LGTM.

PR #548, opened 2026-07-29
feat(engine): pass --config flag to mcporter subprocess calls

phoenixliu:
The mcporter CLI defaults to `./config/mcporter.json` (cwd-relative), causing config load failure when the working directory changes.

This PR adds `--config <absolute-path>` to mcporter subprocess calls in `call_tool()` and `filter_servers()`, using the already-resolved absolute path from `load_mcporter_config_path()`, making config resolution independent of cwd.

**Changes:**
- `_mcp.py`: Insert `--config` flag with resolved config path into both subprocess argvs
- `test_mcp_port.py`: Update existing test assertions for new argv shape; add regression tests for absolute config path

**Testing:** 33/33 passed

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/engine/src/engine/community/plugins/openclaw/_mcp.py
+++ b/src/engine/src/engine/community/plugins/openclaw/_mcp.py
@@ -206,7 +206,8 @@ class _McpPortMixin:
         """
         import subprocess as _sp  # noqa: PLC0415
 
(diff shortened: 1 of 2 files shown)


phoenixliu:
Should also fix this problem in the plugin of claude code.

PR #550, opened 2026-07-29
feat(bcs): add ?show= query param for inline file display on content endpoint

zh-bupt:
## Summary

Adds a `show` query parameter to the session-file content endpoint so the browser can **display** a file inline instead of forcing a download:

```
GET /sessions/{sid}/files/{file_id}/content?show=true   // browser displays inline
GET /sessions/{sid}/files/{file_id}/content?show=false  // browser downloads (default)
GET /sessions/{sid}/files/{file_id}/content             // browser downloads (default)
```

Both `local` and `baas` storage backends support it. A new `inline_view` capability in `GET /sessions/{sid}/files/capabilities` advertises whether inline display is supported.

## Behavior

- **local (stream path):** BCS sets `Content-Disposition`. `show=true` → `inline; filename="..."` (filename kept so "Save As" still works); absent / `show=false` → `attachment; filename="..."` (today's behavior, byte-identical).
- **baas (presign path):** BCS 302-redirects to the presigned share link; the `show` flag is forwarded into the share-link request body (`{"expire_seconds", "operator", "show": true}`) so the backend URL controls the inline disposition. When `show` is false, the request body is unchanged (`{"expire_seconds", "operator"}`).
- **Hard invariant:** `show` absent or `false` changes nothing vs today, byte-for-byte, on both backends. `show` defaults to `false` everywhere it is threaded. `show` is a plain `bool` (`#[serde(default)]`); a non-boolean like `?show=foo` is rejected with 400 by axum's query extractor, same as any malformed query value.

## How (layer-by-layer)

- **`plugin-api/bcs-storage-api`** — new `PresignGetOptions { ttl_secs, show }` struct (`Copy`, passed by value; mirrors the existing `UploadPrepareRequest` pattern, `caller` kept as a separate trailing param). `presign_get` signature changes from positional `ttl_secs: u64` to `opts: PresignGetOptions`. `StorageCapabilities` gains `supports_inline_view: bool` (set `true` by both real backends).
- **`bcs-storage-baas`** — `presign_get` forwards `show` into the share-link POST body only when `true` (keeps the default request byte-identical).
- **`bcs-storage-local`** — `presign_get` signature update only (never invoked; the stream path sets disposition in the HTTP layer).
- **`bcs-session-file`** — `download_route` gains `show: bool`, forwarded into `presign_get`; `capabilities()` surfaces `inline_view` from `supports_inline_view`.
- **`bcs-http`** — `DownloadQuery.show`; `download_file_by_id` picks inline vs attachment on the local stream branch (filename escape preserved); `download_content` and `shared_file_content` forward `q.show`.

## Testing

Full BCS workspace suite green (337 `test result: ok`, 0 failures). New/changed tests:
- **HTTP (local stream):** `?show=true` → inline; `?show=false` and absent → `attachment` (regression-identical); shared-file route `?show=true&token=...` → inline.
- **baas plugin (wiremock):** `show=true` → share-link body contains `"show": true`; `show=false` → body omits `show` (proven via a `.expect(0)` matcher that asserts the `show:true` shape is never posted).
- **service:** `download_route(show)` forwards the exact `PresignGetOptions` to the plugin (via a `FakeStoragePlugin.last_presign_opts` recorder), both `true`/`false`.
- **capabilities:** `GET .../files/capabilities` returns `inline_view: true`.

The baas HTTP→body path is covered transitively (HTTP→`download_route(show)` proven on the local path, service→`presign_get` forwarding, and baas body construction each verified independently) and additionally exercised by the singlebox E2E gate (100% HTTP endpoint coverage).

## Optional follow-ups (non-blocking, from final review)

- One-line doc comment on `DownloadQuery.show` (consistent with undocumented sibling `ttl`/`token` fields).
- Optional HTTP-layer wiremock test for baas `?show=true` → share-link body.
- Symmetric `ttl_secs` assertion in the `show=false` service forwarding test.


RaymondHX (approved):
LGTM

PR #553, opened 2026-07-29
fix(bcs): keep frontend HumanInput channel optional

RaymondHX:
## Summary

- Keep `human_input_channel` optional for frontend-only HumanInput nodes.
- Require `human_input_channel` and a `runtime_actor` assignee only when IM notification is configured.
- Preserve exact-assignee authorization for IM replies while allowing any present Human participant in the run session to answer frontend HumanInput nodes.
- Add definition validation and runtime progression coverage for both modes, and update the implementation plan.

## Root cause

PR #493 applied the DingTalk IM requirements unconditionally to every HumanInput node. This made the pre-existing `bot-human-bot-review.yaml` frontend-only seed invalid because it intentionally has no IM channel or assignee.

## Impact

Existing frontend HumanInput workflows continue to work without channel configuration. Workflows that opt into IM notification retain the stricter channel and assignee requirements.

## Validation

- `cargo test -p bcs-collaboration-runtime` — 66 tests passed
- Changed-line coverage for the PR production-code diff: `34/42` (`80.95%`)

### DIFF
--- a/src/bcs/crates/services/bcs-collaboration-runtime/src/definition.rs
+++ b/src/bcs/crates/services/bcs-collaboration-runtime/src/definition.rs
@@ -164,41 +164,46 @@ pub fn validate_definition(
                 None => return invalid(format!("node {node_id} assignee is required")),
             },
             StateMachineNodeKind::HumanInput => {
-                match &node.assignee {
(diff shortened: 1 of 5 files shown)


vzvince (approved):
LGTM

PR #564, opened 2026-07-29
feat(backend): Track A stage 5 — MCP configuration tenant isolation

totalfrank:
## What this is

Track A **Stage 5** of the `/openapi/v1` tenant-isolation effort (`src/backend/docs/openapi-v1/README.md`): apply the Stage 1 isolation mechanism to **MCP configuration**, so the public API's `mcp` category can be implemented without a handler being able to reach the internal tenant's data.

Track A implements no endpoint by design. This is the prerequisite for the `mcp` half of Track B, which is now unblocked.

Built via SDD — `spec.md`, `plan.md` and `tasks.md` are in `src/backend/specs/2026-07-29-tenant-isolation-mcp/`, with every acceptance criterion checked off against the test that covers it.

## Why it matters here specifically

MCP configuration holds **API keys and authorization headers** for third-party MCP servers — the most sensitive data in the category. It was keyed by a user identifier alone, which is only meaningful *within* a tenant: two tenants may each know a "12345", and nothing distinguished their rows.

## What changed

**The Stage 1 guard is now model-agnostic.** It was welded to `BotModel` inside `plugin_api/models.py`; registering `core/` models from there would invert the dependency direction the boundary tests enforce. It lifts into `utils/avernet_tenant_guard.py` behind `register_avernet_tenant_guard(Model)` — one `Session`-level read listener serving every registered model, plus a per-model insert stamp. `plugin_api/models.py` re-exports `CrossTenantInsertError`, and **all three Stage 1 test files pass unmodified**, which is the check that bot behavior did not move.

**Two tables isolated:** `ac_user_mcp_config` and `ac_bot_mcp_call_config`. Neither repository needed editing — both are ORM throughout, so the guard covers every method without a per-method filter.

## The finding that shaped this stage

`ac_user_mcp_config` carried `UNIQUE (user_id, server_code, env)`. The spec's criterion *"two tenants can each hold a configuration for the same user identifier and the same MCP server"* is **unsatisfiable while that key stands** — the second tenant's write fails with a duplicate-key error against a row it is not allowed to see. Isolation would read correctly and then break at write time.

So this stage is a column add **plus a unique-key replacement** to `(avernet_tenant, user_id, server_code, env)`, unlike Stage 1 which was a pure column add — `ac_bots` has no unique constraint, so Stage 1 never hit this.

The two DDL steps have **different deadlines**, and this is the thing to get right at deploy time:

| Step | Deadline |
|---|---|
| `ADD COLUMN` on both tables | **before the code deploy** — a `SELECT` naming a missing column fails outright |
| Unique-key swap | **before a second tenant writes MCP config** — *not* before deploy; with one tenant the old and new keys accept identical rows |

Full DDL is recorded in `docs/openapi-v1/README.md` (both editions) under "Schema changes applied out-of-band", alongside Stage 1's, so there's one place to look.

## Scope: what the six `mcp` endpoints actually touch

| Endpoint | Backing service | Local tables |
|---|---|---|
| `GET /servers` | `MCPMarketService` → MCP Center | none |
| `GET /tenants` | `MCPMarketService` | none |
| `GET /servers/{code}` | `MCPMarketService` | none |
| `GET /servers/{code}/permissions` | `MCPAuthService` → MCP Center + `MCPAuthPlugin` | none |
| `GET /servers/{code}/config` | `MCPConfigService.get_user_unified_config` | **`ac_user_mcp_config`** |
| `PUT /servers/{code}/config` | `+ MCPSyncService.sync_mcp_detail_to_all_bots` | **`ac_user_mcp_config`**, `ac_bots`, `ac_entity_device_binding` |

Four of the six are MCP Center over HTTP with no local table — `LOCAL`-mode servers come from `configs/local-mcp-servers.yaml`, a file, not a table.

**Checked and deliberately excluded**, recorded so they are not re-opened:
- `ac_entity_device_binding` — the config write path reaches it, but the query *selects* the binding while *joining through* `ac_bots`, and `with_loader_criteria` applies to join clauses. Verified empirically: foreign tenant returns `None`, owning tenant returns the row.
- `ac_skill_set_mcp` — skill_center-owned, belongs to Stage 4.

## On `ac_bot_mcp_call_config`, stated honestly

Its two aggregate reads genuinely query by `bot_pk` alone with no bot record in the statement, so Stage 1's guard does not reach them. But `bot_pk` is `ac_bots.id`, a global primary key, and every call site sources it from a tenant-guarded bot lookup — so **there is no reachable cross-tenant `bot_pk` today**. This is defense in depth against a future careless call site, not the closing of a live hole. Kept because it is cheap and needs no key change.

## Testing

Every isolated data set has a red run recorded (`tasks.md`, Tasks 3 and 4):

| | Red (change reverted) | Green |
|---|---|---|
| `ac_user_mcp_config` | 6 failed, 2 errored | 11 passed |
| `ac_bot_mcp_call_config` | 5 failed | 7 passed |

The two errors in the first are the exact `UNIQUE constraint failed` the key change prevents.

Also added: a test pinning the internal `GET`/`POST /api/mcp/user/config` bodies through the **real** service and repository against SQLite (`test_mcp.py` mocks the services, so it cannot see the layer the guard acts on), and a guard-mechanism suite verified by mutation — neutering the read guard fails it.

**No existing test was modified.**

## Reviewer notes

Two rounds of automated review ran on the guard generalization. Round 1 found four issues, all fixed in `723124c`; the one worth knowing about: `register_avernet_tenant_guard` validated with `hasattr`, which is true for any class attribute — a model declaring `avernet_tenant` as a plain value would have registered happily and the guard would then emit `WHERE 1 = 1`, a silent total bypass. It now validates against the mapper's columns. Round 2 confirmed all four closed and found no new defects.

## Open question carried forward

Whether any non-ORM writer outside this repository writes MCP configuration rows. `server_default='teamclaw'` makes any such writer land on the default tenant, which is correct while all data is internal — so it is not blocking, but it determines whether that path could ever write for a second tenant.

---
_Generated by [Claude Code](https://claude.ai/code/session_015vRhYew81FviDGkXK3Eypg)_


FreddieSun (approved):
LGTM

PR #568, opened 2026-07-29
fix(publish-flow): clear restarting flag on stable-state restart success

quhongwei:
The restarting flag was not being cleared when a restart completed successfully for records in stable states (VALIDATING or SUCCESS).

Root cause:
- execute_restart sets restart.restarting=true
- Non-stable states (VALIDATE_PUB/ONLINE_PUB) call _handle_sync_success which clears the flag
- Stable states (VALIDATING/SUCCESS) bypass that path and did not clear it

Fix:
- Add flag clearing logic in sync_restart_progress for stable SUCCESS case
- Use _mutate_and_update_ext to persist the change without status updates
- Add tests for both SUCCESS and VALIDATING stable states


totalfrank (approved):
LGTM.

PR #571, opened 2026-07-29
fix(bcs): defer IM collaboration initial run

RaymondHX:
## Summary

- Add an optional `start_initial_run` group-creation flag that defaults to `true` for backward compatibility.
- Allow IM-enabled collaboration YAML to be persisted before a group-scoped ChannelBinding exists.
- Keep the unique Active ChannelBinding check at state-machine run start.
- Return and reuse the initial service-invocation `session_id`, so clients can create the group, bind DingTalk, and then explicitly start that original session.
- Update the implementation plan and add HTTP/runtime regression coverage.

## Why

A DingTalk ChannelBinding targets an existing BCS group, but IM-enabled group creation previously validated the binding and auto-started the run before the group could be bound. This created a provisioning deadlock.

## Impact

Existing structured-collaboration clients are unchanged because omitting `start_initial_run` preserves immediate startup. Only clients that explicitly pass `false` defer the initial run. Frontend-only HumanInput definitions continue to work without a channel, while IM HumanInput runs still fail closed when no unique Active binding exists.

## Validation

- `cargo test -p bcs-http --test groups_contract` — 32 passed
- `cargo test --offline -p bcs-collaboration-runtime --test runtime_progression` — 32 passed

Commit and push hooks were intentionally skipped with `--no-verify` as requested.

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/groups.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/groups.rs
@@ -168,6 +168,7 @@ pub async fn create_group(
     uri: Uri,
     Json(req): Json<CreateGroupRequest>,
 ) -> Result<Json<Value>, HttpAdapterError> {
+    let start_initial_run = req.start_initial_run.unwrap_or(true);
     let caller_actor_id = resolve_group_create_caller(&state, &headers, &uri).await?;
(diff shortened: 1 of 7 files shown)


vzvince (approved):
LGTM

PR #590, opened 2026-07-29
feat(workspace): exclude clawbench_* files in openclaw rsync

quhongwei:
Add `workspace/clawbench_*` to rsync excludes to filter out clawbench-prefixed files/directories during workspace sync.


FreddieSun (approved):
LGTM

PR #604, opened 2026-07-30
fix(engine): harden fresh Desktop Skills Pool rollout

XIEAN99:
## What changed

- Treat a completely absent Desktop Legacy `skills-local` root as the valid empty-local-skills state and create it before Pool preparation.
- Continue to fail closed when the Legacy local path is a symlink or another invalid filesystem entry.
- Report `stable_repo_bridge_valid` only when that bridge is applicable and was actually verified: AICoding/Hermes in `active`; omit it for OpenClaw/Claude Code active layouts and all `finalizing` recovery windows.
- Introduce the public `InvalidPoolMappingRequestError` service contract and map only that request error to HTTP 400.
- Normalize invalid logical paths, incompatible logical/physical wire shapes, and unsupported mapping contract versions before filesystem mutation while leaving internal layout invariant errors as server failures.
- Keep transport-schema failures explicit: unknown corpus values are rejected by FastAPI with HTTP 422 before the Skills Service/plugin is invoked.
- Add implementation, Service API contract, HTTP integration, and four-engine Probe evidence coverage.

## Why

A fresh Desktop rootfs creates the workspace and downloads `skills-repo`, but it does not create `skills-local`. Pool preparation previously required that directory to exist, so a newly created Desktop Bot stayed `FAILED` and never published `.pool-ready` unless a local skill directory had already been created by another flow.

Real rootfs validation also found control-plane quality issues: OpenClaw exposed an inapplicable bridge check as `false`, deterministic invalid mapping requests surfaced as HTTP 500, and `finalizing` AICoding/Hermes evidence could claim a stable repo bridge had been verified before that verification ran.

## Impact

Fresh Desktop Bots with no local skills can now reach READY without changing active Legacy mappings. Existing local content and invalid/ambiguous filesystem states retain their previous behavior. Backend receives deterministic 400 responses for well-shaped but invalid mapping contracts, shapes, and paths without masking internal Engine invariants as client errors; malformed HTTP schema input fails before plugin invocation with 422.

## Validation

- 273 targeted Engine tests passed across Desktop preparation, repo download, Probe, activation/rollback, four filesystem engines, mapping contract/router, Service API contracts, fault injection, race recovery, and the HTTP schema boundary.
- Local AgentBox rootfs validation passed fresh prepare, cutover, mapping verify, restart persistence, idempotent replay, rollback, invalid preparation, missing/unreadable local data, active-entry conflict, post-cutover mapping failure and same-generation recovery, response loss and replay, rollback I/O failure and retry, and concurrent Pool/quarantine writes.
- Final rootfs validation confirmed v2 physical mappings, unversioned logical mappings, unknown versions, and path traversal all return HTTP 400 with active mappings and `.pool-active` byte-identical before/after.
- Unknown corpus payloads on activate/publish/verify return HTTP 422 before the corresponding plugin method is invoked.
- Ruff passed for the newly changed Python test; `git diff --check` passed.

### DIFF
--- a/src/engine/docs/heterogeneous-engine-architecture.md
+++ b/src/engine/docs/heterogeneous-engine-architecture.md
@@ -1104,6 +1104,12 @@ class SkillExecutionResult:
   home 投影物理 source/active target；publish/verify 返回
   `resolved_mappings`，activation/rollback 返回 `local_locators`。这些路径是
   Engine evidence，Backend 只能校验和持久化，不能按 engine 重建。
+- READY Probe 已进入 cutover evidence（存在 active marker）后，
+  `checks.stable_repo_bridge_valid` 只在 AICoding/Hermes 已进入 `active`
+  且稳定 repo bridge 已实际校验时存在并为 `true`。OpenClaw/Claude Code
+  的 active layout 以及仍为 `finalizing` 的恢复窗口不适用或尚未完成该
+  检查，必须省略此 key，不能用 `false` 或未经校验的 `true` 代替；
+  preparation/Legacy evidence 仍按各 Engine 拓扑报告其必需 bridge 检查。
 - 新 Backend 只有在 `/api/skills/layout/probe` 的 READY evidence 明确包含
   `"mapping_contract_version": "skills-pool-mapping-v2"` 后，才会在已通过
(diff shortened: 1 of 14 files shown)


totalfrank (approved):
LGTM.

PR #605, opened 2026-07-30
fix: resolve current runtime binding for skills pool

FreddieSun:
## Summary

- add a Skills Pool-specific current-runtime resolver that follows `ac_bots.device_id`
- require the resolved runtime binding to be `ACTIVE`
- route runtime probe, cutover, mapping publish and verification through that resolver
- preserve the legacy `resolve_for_bot` behavior for unrelated device flows

## Root cause

After an ARCA restart, `ac_bots.device_id` can already identify the new active runtime while the legacy `ac_bots.binding_id` still references the released historical binding. Skills Pool resolved through that stale primary key and sent cutover requests to a released sandbox.

## Safety

The change is deliberately scoped to Skills Pool runtime inspection and mutation. Non-Skills-Pool device consumers keep their existing resolver behavior. Missing, pending, failed, stopped or released current runtimes fail closed.

## Validation

- focused red/green reproduction for stale `binding_id` plus current `device_id`
- Skills Pool, runtime probe, resolver and device repository suite: 332 passed
- Ruff: passed
- `git diff --check`: passed

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/device_context_resolver.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/device_context_resolver.py
@@ -103,6 +103,53 @@ class DeviceContextResolver:
             bot_type=bot_type,
         )
 
+    def resolve_current_runtime_for_bot(
+        self,
+        bot_id: str,
+        user_id: str,
+        *,
+        device_uuid: str | None = None,
+    ) -> DeviceContext:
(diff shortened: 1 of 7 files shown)


FreddieSun:
Closing after re-validating the production evidence. The historical RELEASED binding was not used by the Skills Pool reconcile path; the actual cutover failure remains the sofa_tracer/httpx transport issue, so this PR does not address the observed blocker.

PR #576, opened 2026-07-29
feat(backend): cherry-pick draft restoration to REL20260730

ws121200:
## Summary
- cherry-pick the service bot draft restoration delivered in #499 onto `REL20260730`
- preserve the durable operation-ledger workflow and provider-specific ARCA/Teclaw restore behavior
- include the related endpoint, service, repository, and regression tests

## Source
- Original PR: #499
- Original merge commit: `7da462f14e3c1ed8a74a0d6b3a503a7134f3920f`
- Cherry-pick commit: `ee861e2a`

## Validation
- `git diff --check upstream/REL20260730...HEAD`
- 60 related backend tests passed locally

```text
DEPLOY_PROFILE=test uv run pytest \
  tests/community/core/service_bot/repository/test_publish_operation_repository.py \
  tests/community/core/service_bot/services/test_bot_build_service_draft_restore.py \
  tests/community/core/service_bot/services/test_operation_runner.py \
  tests/community/core/service_bot/services/test_provider_behavior.py \
  tests/community/endpoints/test_draft_restore_durable_task.py \
  tests/community/endpoints/test_service_bot_draft_restore.py -q
```

PR #618, opened 2026-07-30
fix(backend): preserve draft restore file ownership

ws121200:
## Problem
ARCA draft restoration uses `rsync -a`, which preserves the historical artifact owner/group. Artifacts can be owned by `500:500`, while the OpenClaw runtime workspace requires `1000:1000`.

## Solution
Add `--chown=1000:1000` to both the primary and extra draft-restore rsync commands.

## Validation
- Targeted draft restore tests: 5 passed
- Ruff passed
- `git diff --check` passed
- Manually validated ARCA restore in a pre-environment OpenClaw sandbox

## Compatibility and risk
Scoped to ARCA draft restoration. Existing `--delete` and runtime exclusion behavior is unchanged.

## Related issues
Follow-up fix for #576.


FreddieSun (approved):
lgtm

PR #620, opened 2026-07-30
feat(bcs-cli): sniff charset for text uploads to fix Chinese preview mojibake

zh-bupt:
## What
Detect a text file's character encoding at upload time and append `charset=<label>` to its `Content-Type`, so inline (`show=true`) previews render Chinese (and other non-UTF-8) text without mojibake.

## Why
Text files were uploaded without a `charset`, so baas/OSS stored and the presigned 302 / local streaming download returned a type like `text/csv` or `text/markdown` with no charset parameter. Browsers then rendered inline previews using a default encoding (often Windows-1252), garbling Chinese text.

## Changes
- Add `guess_charset(path, max_bytes)` using `chardetng` (`EncodingDetector`) to read the first 64KB of the file and return a canonical charset label (`utf-8`, `gbk`, `gb18030`, `shift_jis`, …). Pure ASCII (or empty) returns `None`.
- Add `is_text_mime(mime)` to restrict enrichment to textual types (`text/*`, `application/json`, `application/xml`, `image/svg+xml`); binary types never get a charset.
- In `upload_session_file`, after resolving the MIME (inferred from extension **or** explicitly passed via `--mime`), if it's textual and doesn't already carry a `charset=`, sniff the encoding and append `; charset=<label>`. This covers `--mime text/markdown` on a UTF-8/GBK file too. An explicit `--mime` that already includes a charset (e.g. `text/markdown; charset=gbk`) is respected and not re-sniffed/duplicated.
- The detected charset flows into the baas `upload-url` prepare body `content_type` and the PUT `Content-Type` header, so both the presigned download path and the local streaming download path return a charset-aware type.
- Adds workspace deps `chardetng = "0.1"` and `encoding_rs = "0.8"` (only used by `bcs-cli`).

## Verification
- `cargo build -p bcs-cli`: compiles with the new deps.
- `cargo test -p bcs-cli --lib`: **49 passed** (incl. new `is_text_mime_recognizes_textual_types`, `guess_charset_returns_none_for_ascii`, `guess_charset_detects_utf8_and_cjk`, and `explicit_text_mime_still_eligible_for_charset`). No regression in the existing suite.

## Behavior notes
- Detection is byte-pattern/statistical (via `chardetng`), not based on file-header BOM or declaration tags (`<meta charset>`, `<?xml encoding?>`). It correctly handles GBK/Big5/Shift_JIS CSVs that have no self-declaration.
- Explicit `--mime application/octet-stream` stays non-textual → no charset is appended (the binary intent is respected).
- Only the first 64KB is read for detection, so large text files are still cheap to enrich.

##Notes
- Depends on `guess_mime_from_extension`, which is already on `dev` (merged via PR #603), so this PR's diff is self-contained.

### DIFF
--- a/src/bcs/Cargo.lock
+++ b/src/bcs/Cargo.lock
@@ -674,9 +674,11 @@ dependencies = [
  "bcs-config",
  "bcs-config-api",
  "bcs-protocol",
+ "chardetng",
  "chrono",
  "clap",
  "dirs",
+ "encoding_rs",
  "futures",
  "inventory",
  "predicates",
@@ -1764,6 +1766,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "chardetng"
+version = "0.1.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14b8f0b65b7b08ae3c8187e8d77174de20cb6777864c6b832d8ad365999cf1ea"
+dependencies = [
+ "cfg-if",
+ "encoding_rs",
+ "memchr",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


RaymondHX (approved):
LGTM

PR #623, opened 2026-07-30
fix(engine): prevent premature chat stream timeout for long-running tasks

evilcoda:
## Summary

- Extend the default OpenClaw chat stream inactivity timeout from 5 minutes to 20 minutes.
- Preserve the existing behavior when callers explicitly provide `timeoutMs`.
- Add a regression test that verifies the 20-minute default and confirms no implicit `timeoutMs` is forwarded upstream.

## Root Cause

Long-running tasks, especially subagent workflows, can remain quiet while work is still progressing upstream. The adapter previously waited only 300 seconds for the next stream event and then returned `Chat stream timeout`, even though the task could still be running.

## Impact

The adapter now allows up to 20 minutes of stream inactivity by default. Callers that explicitly set `timeoutMs` are unaffected and continue to use their requested timeout.

## Validation

- `PYTHONPATH=src uv run python -m pytest src/engine/community/openclaw/client/tests/test_gateway_client.py -q` (`14 passed`)
- `uv run ruff check src/engine/src/engine/community/openclaw/client/gateway_client.py src/engine/src/engine/community/openclaw/client/tests/test_gateway_client.py`
- Changed production lines covered: 100%


totalfrank (approved):
LGTM.

PR #628, opened 2026-07-30
fix(backend): make Skills Pool rollback resumable

XIEAN99:
## What changed

- Reuse the rollback lease atomically acquired by `begin_legacy_rollback` instead of immediately renewing the same owner/expiry.
- Rebuild Legacy local storage through the Engine rollback endpoint before publishing and verifying Legacy logical mappings.
- Add regression coverage for MySQL changed-row semantics and post-cutover mapping failure/resume.

## Root cause

A fresh rollback immediately re-acquired its own lease. When MySQL treated that update as a no-op, the repository returned zero changed rows and the service reported `state_race_lost`.

After the lease expired, the next attempt published Legacy mappings before the Engine had rebuilt Legacy local storage. Desktop Pool cutover intentionally retires that directory, so any local Skill made the mapping request fail closed with `source_missing`.

## Impact

The rollback now follows the documented forward-only sequence: persist rollback intent and lease, rebuild Legacy from the authoritative Pool, record the data-plane commit, then publish/verify Legacy mappings and commit Legacy locators. A mapping or database failure after the filesystem exchange remains retryable from `LEGACY_ROLLBACK_COMMITTED`.

No Engine, layout planner, mapping contract, database schema, or rollout gate behavior changes.

## Validation

- 241 Backend Skills Pool core and operator endpoint tests passed.
- Ruff passed on both changed files.
- Local Python SAST block scan passed on both changed files.
- `git diff --check` passed.

The repository pre-push hook was intentionally skipped per delivery instruction because it is too slow; the push used `--no-verify`.

### DIFF
--- a/src/backend/src/agentclaw/community/core/skills_pool/recovery_service.py
+++ b/src/backend/src/agentclaw/community/core/skills_pool/recovery_service.py
@@ -250,6 +250,7 @@ class SkillsPoolRollbackService:
             )
 
         state = self._layouts.get(scope)
+        began_rollback = False
         if not state.persisted:
             return SkillsPoolRollbackResult(SkillsPoolRollbackOutcome.NOT_FOUND)
         if state.phase is SkillLayoutPhase.POOL_ACTIVE:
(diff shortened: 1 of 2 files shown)


FreddieSun (approved):
LGTM

PR #640, opened 2026-07-30
chore(gateway): regenerate bots.openapi.json from the current backend surface

totalfrank:
## What

Regenerates `src/gateway/configs/schemas/bots.openapi.json` so the doc the gateway serves at `/docs` matches what the backend's `/openapi/v1` routers actually accept.

The artifact had drifted behind the source in `src/backend/src/agentclaw/community/adapters/http/openapi_v1/`.

## How

Ran the existing pipeline rather than hand-writing anything:

1. `src/backend/scripts/dump_openapi.py` — imports the backend app, calls `app.openapi()`, narrows to the `/openapi/v1` namespace and prunes unreferenced components.
2. `src/gateway/scripts/gate_and_publish_openapi.py` — runs the backward-compat gate against the currently published artifact, then publishes.

## What changed

The **path set is unchanged** — 32 operations across `bots`, `channels`, `identity`, `mcp`, `resources`, `routines`, `skills`. The operation and schema detail changed:

- Every operation now carries the surface-wide `ERROR_RESPONSES` envelope (400/401/404/409/422/500/502 → `ErrorEnvelope`). This replaces FastAPI's default `HTTPValidationError` / `ValidationError` for 422, which is what `openapi_v1/contracts.py` documents as intended — the public surface renders every failure as an `Envelope`.
- Request models are `extra="forbid"`, and fields the server no longer accepts are gone from the schema: `BotCreate.engine_options`, `BotUpdate.cluster_name`, `BotUpdate.engine_options`, `McpConfigWrite.sync_mode`.
- `bot_id` is documented as a **required** query param on the resources and routines operations, matching the handler signatures.
- `BotAuthPending` gains `redirect_url` and no longer requires `iframe_url` — Passport may return either handle.
- Full handler docstrings are carried into operation descriptions.

## ⚠️ The compat gate was overridden — please review this

The gate reported **21 backward-incompatible changes** and was run with `--allow-breaking`:

```
15x [parameter-now-required]  bot_id on the resources + routines operations
     [property-removed]       BotCreate.engine_options
     [property-removed]       BotUpdate.cluster_name
     [property-removed]       BotUpdate.engine_options
     [property-removed]       McpConfigWrite.sync_mode
     [schema-removed]         HTTPValidationError
     [schema-removed]         ValidationError
```

Every one of these is **the artifact catching up to already-merged source**, not a new contract change. The server already rejects the removed fields (`extra="forbid"`) and already requires those query params; each removal is deliberate and documented in the source comments. The published doc was advertising fields the server rejects, so leaving it stale is the worse option.

That said, this is a public contract and the override is a real signal — worth a look before merge. If any external client is coding against the old shape, the removals are the ones that matter.

## Verification

- `pytest tests/test_served_openapi.py tests/test_openapi_generator.py tests/test_gate_and_publish.py` — 18 passed
- Artifact is valid JSON, 32 paths / 80 component schemas, no dangling `$ref`s
- No test pins this file byte-for-byte (the gateway tests use their own fixtures), so nothing else needed updating

## Note (not changed here)

`src/gateway/configs/schemas/README.md` still says *"No build-time generation pipeline exists yet"*, which is stale — `scripts/dump_and_publish.sh` is that pipeline. Left alone to keep this diff to the artifact; happy to fix it in a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tc6szPMgMSd7tKDPTha9iE)_

### DIFF
--- a/src/gateway/configs/schemas/bots.openapi.json
+++ b/src/gateway/configs/schemas/bots.openapi.json
@@ -21,6 +21,11 @@
             "type": "string"
           },
           "cluster_name": {
+            "description": "Deployment cluster, in strict 1:1 correspondence with the engine: 'ANDC' for engine 'teclaw', 'ACRA' for every other engine. On create the engine/cluster pair is validated against this rule (400 on mismatch).",
+            "enum": [
+              "ACRA",
+              "ANDC"
+            ],
             "title": "Cluster Name",
             "type": "string"
           },
@@ -52,20 +57,25 @@
         "type": "object"
       },
       "BotAuthPending": {
-        "description": "Returned (202) when bot creation needs user authorization (Passport).",
+        "description": "Returned (202) when bot creation needs user authorization (Passport).\n\nPassport may hand back either handle, so both are surfaced: a caller that\nonly receives ``redirect_url`` would otherwise have no way to complete\nauthorization. Each is empty when Passport did not supply it.",
         "properties": {
           "bot_id": {
             "title": "Bot Id",
             "type": "string"
           },
           "iframe_url": {
+            "default": "",
             "title": "Iframe Url",
+            "type": "string"
+          },
+          "redirect_url": {
+            "default": "",
+            "title": "Redirect Url",
             "type": "string"
           }
         },
         "required": [
-          "bot_id",
-          "iframe_url"
+          "bot_id"
         ],
         "title": "BotAuthPending",
         "type": "object"
@@ -106,6 +116,7 @@
         "type": "object"
       },
       "BotCreate": {
+        "additionalProperties": false,
         "description": "Create-a-bot request body.",
         "properties": {
           "bot_desc": {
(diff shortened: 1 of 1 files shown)


totalfrank:
**CI status: Gateway unit tests red — failing on the base branch too, not caused by this PR.**

5 of 6 checks pass (BaaS, Engine, BCS, Backend, BCS e2e). Gateway unit tests fail with 12 failures + 14 errors, all rooted in one error:

```
TypeError: 'app_id' is an invalid keyword argument for AppRow
```

**Cause** — `bootstrap/_authn.py:107-117` seeds the demo app row with an `app_id` kwarg, but `AppRow` (`core/app/_orm.py:23-44`) has no `app_id` column; the app's identity is the surrogate `id` PK, which is what `AppRow.to_record()` and `_registrar.py` both use. The `AppRow(...)` construction raises at DI bootstrap, so every test that builds the app dies in fixture setup — which is why the `httpx.ConnectError` failures cascade behind it (the gateway process never came up).

**Not from this PR.** I checked out `origin/dev` at `020b1ad` (this PR's base, none of my changes) and reproduced it exactly:

```
2 failed, 2 errors in 2.57s
TypeError: 'app_id' is an invalid keyword argument for AppRow
```

`git log -L` points at 4b6648f (#599, DB-backed app_token + signed principal) as where that seed block landed.

**Why I'm not fixing it here.** This PR changes one generated JSON artifact and nothing else — bundling a gateway auth-seed fix into a contract-artifact diff would muddy both. The fix also isn't unambiguous: dropping `app_id="app-1"` from the seed is the minimal read, but if `avernet_application` is *meant* to carry a separate business-key `app_id` column, the fix is adding the column instead. That's a call for whoever owns #599. Worth noting the adjacent smell in the same block: `BotRow(app_id="app-1")` passes a string into a `Mapped[int]`/`BigInteger` column.

Happy to put up the one-line seed fix as its own PR if that's the wanted direction — say the word.

I'll re-run CI here once the base is green.

---

PR #641, opened 2026-07-30
fix(engine): fail closed on invalid layout probes

XIEAN99:
## What changed

- validate the requested Engine/layout-contract identity through the Engine-owned Layout Planner before dispatching a runtime probe
- return structured `INVALID` evidence for unknown engines and unsupported contract versions
- reject a plugin probe result whose runtime Engine does not match the requested Engine
- add HTTP contract coverage for all three fail-closed cases

## Why

Real Desktop Hermes validation for #455 showed that `engine=unknown-engine` was ignored and returned the active Hermes runtime as `READY`; an unsupported contract escaped as an unhandled exception. The probe boundary must fail closed before Backend can accept mismatched runtime evidence.

## Impact

This is a backward-compatible Engine Service API hardening. Valid probes are unchanged. Invalid identities now return the existing `INVALID` probe status instead of dispatching or raising.

## Validation

- Engine Skills router + Layout Planner: 37 passed
- focused router coverage: 29 passed; changed-line coverage 100% (7/7)
- Engine Python SAST block scan: passed
- `git diff --check`: passed

Refs #455

### DIFF
--- a/src/engine/src/engine/community/api/skills/router.py
+++ b/src/engine/src/engine/community/api/skills/router.py
@@ -37,6 +37,12 @@ from engine.community.core.engine.exceptions import CapabilityNotSupportedError
 from engine.community.core.skills.exceptions import (
     InvalidPoolMappingRequestError,
 )
+from engine.community.core.skills.layout_planner import (
+    LayoutIdentity,
+    RuntimeLayoutContext,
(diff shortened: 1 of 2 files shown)


FreddieSun (approved):
LGTM

PR #573, opened 2026-07-29
fix(community): filter notify polling by engine

maicaisongxiaocong:
## Summary

- filter `/api/v1/notify` polling targets by `ac_bots.active_engine`
- only probe bots using `aicoding` or `claude_code`
- apply the same eligibility rule to owner, collaborator, and local/singlebox bot listings
- fail closed when bot metadata or `active_engine` cannot be resolved

## Testing

- `DEPLOY_PROFILE=test uv run pytest tests/community/core/notify/test_repository_notify_bot_lister.py tests/community/plugins/local/test_notify_bot_lister.py tests/community/api/aicoding/test_notify_router.py -q` (49 passed)
- `uv run ruff check` on all changed Python files
- repository pre-push Python SAST/lint gate

PR #647, opened 2026-07-31
fix(community): filter notify polling by engine (REL20260730 cherry-pick)

maicaisongxiaocong:
## What

Cherry-pick of notify engine filtering (#573) onto `REL20260730`.

## Why

`REL20260730` 缺少对 `/api/v1/notify` 轮询目标的引擎过滤：只有 `aicoding` / `claude_code` 引擎支持 Engine `/api/notify`，其他引擎不应被探测。该修复已在 dev，需随 REL 带出。

## Changes

- 新增 `NOTIFY_SUPPORTED_ENGINES = frozenset({"aicoding", "claude_code"})`
- `bot_lister` / `local_bot_lister` 按 `ac_bots.active_engine` 过滤
- 未知 `active_engine` fail-closed
- 同规则应用到 owner / collaborator / local / singlebox bot listings
- 测试：`test_repository_notify_bot_lister` / `test_notify_bot_lister` / `test_notify_router`

## Base

Fast-forward onto `REL20260730` tip (`2a66dadf`) — no history rewrite, no commits dropped. Pre-push lint/SAST gate passed.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/aicoding/notify_router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/aicoding/notify_router.py
@@ -1,8 +1,8 @@
 """
-Notify router — aggregates pending HITL interactions from all bots via Engine.
+Notify router — aggregates pending HITL interactions from notification-capable bots.
 
 GET /api/v1/notify:
-  1. Get user's active device bindings (bots with sandboxes)
+  1. List notification-capable bots with active device bindings.
   2. Parallel probe each bot's Engine /api/notify endpoint — routed by
      ``device_provider`` via ``DeviceContextResolver`` (same path chat / cron
      use), so BaaS (desktop + cloud), Arca, teclaw and local each reach their
@@ -281,7 +281,7 @@ async def get_notify_summary(
     transport: DeviceAdapterTransport = Injected(DeviceAdapterTransport),
 ):
     """
-    Get aggregated pending notifications from all user bots.
+    Get aggregated pending notifications from notification-capable user bots.
 
     Returns a list of bot summaries, each containing the pending HITL
     interactions from that bot's engine.
--- a/src/backend/src/agentclaw/community/core/notify/bot_lister.py
+++ b/src/backend/src/agentclaw/community/core/notify/bot_lister.py
@@ -8,7 +8,9 @@
   - ``status="ACTIVE"`` —— 只取活绑定
   - ``sandbox_id`` 回退到 ``binding.device_id`` —— 老版本绑定没写 device_props
 
-bot 名字仍然从 ``BotRepository`` 解析 (device_props.bolt_id → bot_id)。
+bot 名字和 ``active_engine`` 从 ``BotRepository`` / ``ac_bots`` 解析。只有
+``aicoding`` / ``claude_code`` 引擎支持 Engine ``/api/notify``，其他引擎
+在生成 ``NotifyTarget`` 前跳过，避免无效探测。
 
 协作者机器人：除 owner 自己的活绑定外，再查 ``CollaboratorRepository``,
 把当前用户作为协作者参与的 Bot 一并纳入通知列表（对齐
@@ -24,6 +26,7 @@ from agentclaw.community.core.bot_collaborator.repository.protocol import (
 )
 from agentclaw.community.core.bot_management.repository.protocol import BotRepository
 from agentclaw.community.core.devices.repository.protocol import DeviceBindingRepository
+from agentclaw.community.core.notify.constants import NOTIFY_SUPPORTED_ENGINES
 from agentclaw.community.core.notify.protocol import NotifyTarget
 from agentclaw.community.log import get_logger
 from agentclaw.community.utils.env_utils import get_current_env
@@ -67,7 +70,9 @@ class RepositoryNotifyBotLister:
             if not sandbox_id:
                 continue
             bot_id = str(props.get("bolt_id") or b.device_id)
-            bot_name = self._resolve_bot_name(bot_id, user_id)
+            bot_name = self._resolve_supported_bot_name(bot_id, user_id)
+            if bot_name is None:
+                continue
             result.append(NotifyTarget(
                 bot_id=bot_id, bot_name=bot_name,
                 owner_id=user_id, sandbox_id=str(sandbox_id),
@@ -93,22 +98,41 @@ class RepositoryNotifyBotLister:
     # ------------------------------------------------------------------
     # helpers
     # ------------------------------------------------------------------
-    def _resolve_bot_name(self, bot_id: str, owner_id: str) -> str:
-        """Resolve display name; fall back to bot_id on failure.
+    def _resolve_supported_bot_name(
+        self, bot_id: str, owner_id: str
+    ) -> str | None:
+        """Return the display name when the bot supports notification polling.
 
-        镜像原始实现：bot 名字优先从 ``BotRepository`` 取，查不到/报错时
-        回退到 ``bot_id``。
+        ``active_engine`` comes from ``ac_bots`` via ``BotRepository``. Missing
+        bot metadata, lookup failures, and unsupported engines all fail closed:
+        without a confirmed notification-capable engine we must not call the
+        Engine ``/api/notify`` endpoint.
         """
-        bot_name = bot_id
         try:
             bot = self._bot_repo.get_by_id_and_owner(bot_id, owner_id)
-            if bot:
-                bot_name = bot.get("bot_name") or bot_id
         except Exception as e:
             logger.warning(
-                f"[notify_bot_lister] bot lookup failed bot={bot_id}: {e}"
+                f"[notify_bot_lister] bot lookup failed bot={bot_id} "
+                f"owner={owner_id}: {e}"
             )
-        return bot_name
+            return None
+
+        if not bot:
+            logger.warning(
+                f"[notify_bot_lister] bot not found bot={bot_id} "
+                f"owner={owner_id}"
+            )
+            return None
+
+        active_engine = str(bot.get("active_engine") or "").strip().lower()
+        if active_engine not in NOTIFY_SUPPORTED_ENGINES:
+            logger.info(
+                f"[notify_bot_lister] skip unsupported engine bot={bot_id} "
+                f"owner={owner_id} active_engine={active_engine!r}"
+            )
+            return None
+
+        return str(bot.get("bot_name") or bot_id)
 
     def _list_collaborator_mappings(
         self,
@@ -139,7 +163,7 @@ class RepositoryNotifyBotLister:
             f"{len(collaborators)}"
         )
         # NOTE: each collaborator bot issues 2 DB calls (active binding +
-        # name lookup) → O(2N). Fine for typical collaborator counts; if
+        # bot metadata lookup) → O(2N). Fine for typical collaborator counts; if
         # /api/v1/notify polling load grows, add batch methods
         # (get_active_by_bots_and_owners / get_bots_by_ids_and_owners) to
         # DeviceBindingRepository / BotRepository and resolve in O(1).
@@ -167,7 +191,9 @@ class RepositoryNotifyBotLister:
             sandbox_id = props.get("sandbox_id") or binding.device_id
             if not sandbox_id:
                 continue
-            bot_name = self._resolve_bot_name(bot_id, owner_id)
+            bot_name = self._resolve_supported_bot_name(bot_id, owner_id)
+            if bot_name is None:
+                continue
             mappings.append(NotifyTarget(
                 bot_id=bot_id, bot_name=bot_name,
                 owner_id=owner_id, sandbox_id=str(sandbox_id),
--- a/src/backend/src/agentclaw/community/core/notify/constants.py
+++ b/src/backend/src/agentclaw/community/core/notify/constants.py
@@ -1,0 +1,5 @@
+"""Constants for notification-capable bot engines."""
+from __future__ import annotations
+
+
+NOTIFY_SUPPORTED_ENGINES = frozenset({"aicoding", "claude_code"})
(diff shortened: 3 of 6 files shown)


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0 — Target branch

- **Target branch:** `REL20260730` — matches `^REL\d{8}`, a release branch.
- **Claimed classification:** cherry-pick of #573 onto `REL20260730`.
- **Verified:** #573 (`fix(community): filter notify polling by engine`, closed via #646) is merged into `dev` (base `dev`, `merged: true`, 172 additions / 32 deletions / 6 files) and its diff matches this PR's diff shape exactly (same file set, same additions/deletions). **Gate 0 passes** — this is a genuine existing-behavior fix with a reviewed `dev` counterpart, correctly fast-forwarded onto the `REL20260730` tip with no history rewrite.

Gate 0 passing does not end the review: per the review contract, the full architecture/violation-catalog check still applies to every changed file regardless of branch classification.

## Findings

### 1. Hardcoded engine capability set introduced directly in engine-agnostic core
**File:** `src/backend/src/agentclaw/community/core/notify/constants.py:5` (new file)
```python
NOTIFY_SUPPORTED_ENGINES = frozenset({"aicoding", "claude_code"})
```
Consumed by engine-name-in-set checks in:
- `src/backend/src/agentclaw/community/core/notify/bot_lister.py:128` — `if active_engine not in NOTIFY_SUPPORTED_ENGINES:`
- `src/backend/src/agentclaw/community/core/notify/local_bot_lister.py:23` — same check

**Violated rule:** violation catalog item 2 ("Hardcoded engine capability sets... Whether an engine supports a feature must come from the engine's own declaration") and arch.rules.md Rule 10 (Invariant — "Any capability intended to be replaceable, environment-specific, isolated, or implementation-variable must be represented by a declared contract and one or more implementations").

**"Add engine `foo`" test:** if a new engine gains `/api/notify` support next month, the required edit is not "`foo`'s own module + one registration entry" — it is adding `"foo"` to this shared frozenset literal inside `core/notify/constants.py`, a file every engine's notify eligibility routes through. That fails the test the contract sets: engine-agnostic core (`core/notify/*`) must not change when an engine's capabilities change.

**Why it breaks the architecture:** `core/notify/bot_lister.py` and `local_bot_lister.py` are engine-agnostic domain services (they iterate all bots regardless of engine). By checking bot identity against a literal set of engine name strings, they encode engine-specific knowledge directly in core — exactly the "under-abstraction" failure mode the contract calls out, and exactly the example the catalog uses (`NOTIFY_SUPPORTED_ENGINES`, verbatim). This repo already has the required pattern for this kind of variability: `core/bot_management/engines/registry.py` + `provisioning.py` (an `EngineProvisioningStrategy` ABC, one implementation per engine under `engines/<engine>/strategy.py`, and a `default.py` fallback, all wired through a composition-root registry). Notify eligibility should be declared the same way, not reinvented as a set literal in a new file.

**Required fix:** add a `supports_notify_polling` (or similarly domain-named) capability to the existing engine strategy contract (or a new small `NotifyCapabilityResolver` Protocol defined once in `core/notify/`), with `aicoding`/`claude_code` strategies declaring `True` and a default implementation declaring `False` (fail-closed, matching the current behavior's intent). Register per-engine values where engine strategies are already registered (`engines/registry.py`). `bot_lister.py` / `local_bot_lister.py` then call the resolver instead of testing membership in a hardcoded frozenset.

## Out-of-scope notes
- This same violation already exists on `dev` via merged #573 — worth fixing there too, but per the review contract "the code already does this elsewhere" is not a basis for waiving the finding on this diff, since this PR newly introduces the file/hunks into `REL20260730`.
- The fail-closed behavior for missing/unresolvable `active_engine` (skip rather than probe) is sound defensive design; not gated here.
- The rest of the diff (docstring updates, `notify_router.py` wording, test additions) is minimal and correctly scoped to the stated fix.

---

totalfrank (changes requested):

## 2. Gate 0

- **Target branch:** `REL20260730` — matches `^REL\d{8}`, a release branch.
- **Classification:** existing-behavior bug fix, cherry-pick of `#573` (merged to `dev` 2026-07-29). The diff changes an existing code path (`NotifyBotLister` / `LocalNotifyBotLister`), does not add a new route/DTO field/config key, and the PR body correctly references its `dev` counterpart. Verified: `#573` is merged on `dev` and its `core/notify/constants.py` is byte-identical to this PR's. → **path (a) applies, Gate 0 passes procedurally.**
- Per the contract, Gate 0 passing does **not** waive the full architecture review below — an unwaived Invariant violation still forces REQUEST_CHANGES regardless of the fix being a legitimate cherry-pick.

## 3. Findings

**1. `src/backend/src/agentclaw/community/core/notify/constants.py:5` — hardcoded engine capability set in `core/`**

```python
NOTIFY_SUPPORTED_ENGINES = frozenset({"aicoding", "claude_code"})
```

This is violation catalog item **#2** verbatim ("Hardcoded engine capability sets… Whether an engine supports a feature must come from the engine's own declaration… not by editing N core files"), and breaches constitution **Rule 10 — Invariant** (replaceable/environment-variable capabilities must be represented by a declared contract, not a literal set) together with the engine-variability pattern mandated in the review contract.

Maintenance failure: adding engine `openclaw` notify support next month requires editing this literal in `core/`, plus every call site that imports it — not "own module + one registration entry." The "add engine foo" test fails.

Required fix: replace the frozenset with a domain-named contract defined once in `core/notify/protocol.py` (e.g. a `NotifyCapabilityResolver` Protocol, or a capability flag surfaced through the existing `NotifyBotLister`/`BotRepository` contract), implemented per engine in that engine's own adapter/strategy module, wired by engine type in a composition root (`di/`). Ship a **mandatory default implementation** (conservative deny — i.e. current fail-closed semantics) so an engine without its own declaration is skipped safely, exactly the behavior this PR wants, but without the hardcoded set.

**2. `src/backend/src/agentclaw/community/core/notify/bot_lister.py:127-130` and `src/backend/src/agentclaw/community/core/notify/local_bot_lister.py:22-24` — engine-string branching in core**

```python
active_engine = str(bot.get("active_engine") or "").strip().lower()
if active_engine not in NOTIFY_SUPPORTED_ENGINES:
    ...
    return None
```

(mirrored in `local_bot_lister.py`: `if active_engine not in NOTIFY_SUPPORTED_ENGINES: continue`)

This is violation catalog item **#1** — engine-name string-set branching added directly inside engine-agnostic `core/` services (`bot_lister.py` and `local_bot_lister.py` both implement the `NotifyBotLister` Protocol, i.e. they are core, not composition-root or engine-adapter code). It breaches the same **Rule 10 — Invariant**, since `core/` is now required to know which literal engine identifiers exist and what they support, rather than depending only on a capability contract.

Both listers are new/extended in this diff (`_resolve_bot_name` → `_resolve_supported_bot_name`, and the new `active_engine` check in `local_bot_lister.py`), so this is code the PR **adds**, not merely inherited history — "the code already does this elsewhere" (on `dev`/via already-merged `#646`) is explicitly not an exemption per the review contract, and new code extending an existing violation is itself a violation.

Required fix: same as Finding 1 — `bot_lister.py`/`local_bot_lister.py` should call the capability contract (`resolver.supports_notify(active_engine)` or equivalent), never compare `active_engine` against an inline set. The engine→capability mapping belongs in exactly one composition-root registry.

**3. Missing guard test for the boundary this PR creates**

`src/backend/tests/community/architecture/` has no test preventing hardcoded engine-name sets/branches in `core/notify/` (the existing `test_engine_ext_opacity.py` guards a different field, `engine_ext`, and does not cover `active_engine`). Per review-procedure step 7, a new boundary that current architecture tests cannot catch requires a guard test or an explicit written waiver (Governance Addendum). Neither is present. Once Findings 1–2 are fixed via the contract pattern, add a conformance test asserting every registered engine implementation (including the mandatory default) is exercised, and/or a static guard forbidding new `NOTIFY_SUPPORTED_ENGINES`-shaped literals in `core/`.

## 4. Out-of-scope notes

- Process observation, not architecture: PR `#646` (identical diff, same title pattern) already merged into `REL20260730` on 2026-07-31, before this PR was opened against a now-stale base (`2a66dadf`). This PR appears fully redundant against the current branch tip and its `mergeable_state` is `blocked` — worth closing rather than merging, independent of the findings above.
- Business logic (fail-closed semantics on lookup failure, log message wording, `bot_name` fallback) looks reasonable and is out of scope for this review.


---

totalfrank:
## Verdict: REQUEST_CHANGES

## Gate 0 — Target branch

- Target branch: `REL20260730` (matches `^REL\d{8}`, a release branch).
- Claim verified: PR head commit `4c20def9` is **content-identical** (`git diff 4c20def9 5e6dfa99` — zero output) to commit `5e6dfa99` on `origin/dev`, which is `fix(community): filter notify polling by engine (#573) (#646)` — a real, already-merged dev PR (#646, referencing #573). `4c20def9` is *not* an ancestor of `dev` (`git merge-base --is-ancestor` → false), consistent with this being a deliberate cherry-pick rather than a fast-forward.
- Diff scope: modifies only pre-existing code paths (`notify_router.py` docstrings, `bot_lister.py`, `local_bot_lister.py`) plus one new constants module; no new pipeline, no new HTTP surface, no new entry point.
- Classification: **(a) existing-behavior bug fix, cherry-pick of dev PR #646 (#573)**.
- Gate 0: **PASS** — this is a legitimate, verified cherry-pick of a shipped, already-reviewed dev fix onto the release branch. No re-target required on Gate 0 grounds alone.

Gate 0 passing does not end the review — the diff still has to clear the architecture catalog, and it does not.

## Findings

1. **Hardcoded engine capability set introduced in `core/`** — `src/backend/src/agentclaw/community/core/notify/constants.py:5`
   ```python
   NOTIFY_SUPPORTED_ENGINES = frozenset({"aicoding", "claude_code"})
   ```
   Violated rule: arch.rules.md Rule 10 (*Component Types Are Explicitly Declared and Swappable*) and Rule 14 (*Configuration Drives All Wiring* — "Enabling, disabling, or selecting among existing implementations must be a configuration change, not a business-logic code change"); violation-catalog item 2 (hardcoded engine capability set instead of capability coming from the engine's own declaration).
   This is a new engine-agnostic-layer file that hardcodes which engines support notify polling as a literal string set inside `core/`. Whether a bot's active engine supports `/api/notify` is engine capability data, not core business logic — it belongs to the engine's own declaration/registration, not a frozenset core imports and branches on. Apply the "add engine foo" test: to grant a new engine notify capability today, you must edit this `core/` file directly — exactly the maintenance failure Rule 14 exists to prevent.
   Concretely, this repository already has the correct pattern for this exact problem shape: `agentclaw.community.core.bot_management.engines.registry` (`EngineProvisioningRegistry`, `DefaultCapabilitiesEngineBucketResolverRegistry`) lets each engine module register its own resolver in a composition root, with a mandatory default/fallback for unclaimed engine types, and `core/default_capabilities.py` explicitly documents "delegates bucket overrides to the engine registry ... rather than branching here on engine-name literals" — guarded by `test_default_capabilities_engine_boundaries.py`. This PR reintroduces the exact anti-pattern that file/test pair was built to prevent, in a sibling module.

2. **Engine-name-literal branching in core services** — `src/backend/src/agentclaw/community/core/notify/bot_lister.py:128` and `src/backend/src/agentclaw/community/core/notify/local_bot_lister.py:23`
   ```python
   if active_engine not in NOTIFY_SUPPORTED_ENGINES:
       continue  # / return None
   ```
   Violated rule: violation-catalog item 1 (engine-name sets inline in core services). Both `RepositoryNotifyBotLister` and `LocalNotifyBotLister` — engine-agnostic core listers per module docstrings — directly compare `active_engine` against the hardcoded literal set from Finding 1. Per the "Engines are NOT plugins" guidance, engine variability must be resolved via (1) one contract defined once in the engine-agnostic layer, named in domain terms (e.g. a `supports_notify_polling(engine_type) -> bool` capability query), (2) each engine's own registration in exactly one composition root, (3) a mandatory default/fallback (e.g. "unknown engine → unsupported", which the diff does get right as *behavior*, just not as *placement*). Compliant alternative: define a capability Protocol/query in `core/notify/protocol.py`, have engine modules register their notify-support flag through the existing engine registry (`di`/`config` composition root), and have `bot_lister.py`/`local_bot_lister.py` call the registry instead of comparing against a literal set.

3. **No architecture guard test for the new boundary** — the repo already has a guard-test pattern for exactly this failure mode (`src/backend/tests/community/architecture/test_default_capabilities_engine_boundaries.py`, which AST-scans `default_capabilities.py` for engine-literal comparisons and fails the build if found), but no equivalent guard exists for `core/notify/`. `test_governance_notify_lifecycle_guards.py` covers a different notify subsystem (governance notify-log status machine) and does not touch `bot_lister.py`/`constants.py`. Per the review protocol's test-check step: this PR creates a boundary (engine-capability data hardcoded in `core/notify/`) that no existing architecture test can catch, and adds none. Required: either add an AST guard analogous to `test_default_capabilities_engine_boundaries.py` for `core/notify/`, or (preferably) fix Findings 1–2 so the boundary doesn't exist to guard.

These are Invariant-category violations (Rules 10 and 14) newly entrenched by this diff, with no written waiver in the PR body (no rule/reason/risk/compensating-controls/owner/expiry). Per the verdict rules, this blocks approval regardless of Gate 0 outcome, urgency, or precedent elsewhere in the codebase.

## Out-of-scope notes

- Pre-existing similar patterns already exist elsewhere in `core/` (e.g. `cron/services/cron_relay.py:_CRON_UNSUPPORTED_ENGINES`, `service_bot/services/deploy/service_skills_manifest.py:_SERVICE_MANIFEST_ENGINES`). Their existence doesn't waive Findings 1–2 here, but a follow-up sweep once the registry-based fix lands here would be worth tracking separately.
- Fail-closed behavior on lookup failure/unknown engine (returning `None`/skipping) is sound defensive design and not gated on.

---

PR #664, opened 2026-07-31
refactor(gateway): fix DI architecture and standardize implementation

xxxxpenny:
- Refactor dependency injection pattern in gateway module
- Remove non-standard/inconsistent DI implementations
- Improve code maintainability and testability

PR #666, opened 2026-07-31
chore(ci): replace static paths with dynamic change detection for singlebox-coverage

cassiuscai:
Replace the static `paths` filter in `.github/workflows/singlebox-coverage.yml` with dynamic change detection, matching the pattern already used by `unit-tests.yml` and `e2e-tests.yml`.

**Problem**: The GitHub Actions `paths` filter determines whether the workflow triggers at all. When paths do not match, the workflow is silently skipped with no visible status — there is no way to mark it as "passed". This means PR checks appear incomplete, and users cannot tell whether the check passed or never ran.

**Solution**: Remove `paths` from `pull_request` and `push` triggers. Add a `Detect singlebox-relevant changes vs base` step after checkout that uses `git diff --name-only` against the base ref (same paths as the original filter). All expensive steps are conditioned on `steps.changes.outputs.skip != 'true'`. When nothing changed, the job exits 0 with a clear message.

**Consistency**: Follows the exact same `HEAD^1` for PR / `origin/dev` for push convention, `git fetch` base ref, `mapfile` + `gha_output` pattern, `workflow_dispatch` force-run bypass, and `if:` gating used throughout the existing CI workflow suite.

### DIFF
--- a/.github/workflows/singlebox-coverage.yml
+++ b/.github/workflows/singlebox-coverage.yml
@@ -6,26 +6,10 @@ on:
       - dev
       - main
       - 'REL*'
-    paths:
-      - "scripts/**"
-      - "src/backend/**"
-      - "src/baas/**"
-      - "src/bcs/**"
-      - "src/engine/**"
-      - "src/frontend/**"
-      - ".github/workflows/singlebox-coverage.yml"
   push:
     branches:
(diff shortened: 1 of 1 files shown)


cassiuscai:
Added in #664

Issue #556, opened 2026-07-29
Bot identity keys collide across Avernet tenants (Passport, auth relationships, BCN, policy)


totalfrank:
## Summary

Track A gave the `bots` table a tenant axis (`avernet_tenant` + read/insert guards on `BotModel`). Nothing extended that axis **outward**: the identities we hand to external and shared systems are still built from `bot_id` and `owner_id` alone.

`bot_id` is only unique **per owner per tenant** — every owner's first bot is literally the string `"default"`. So once two tenants can share a principal, they mint the *same* external identity, and the systems on the other side have no tenant field to tell them apart.

Blocks enabling multi-tenancy. Surfaced across PR #494 as F27, F31 and F49; filing as one issue because it is one root cause and wants one design decision, not three patches.

## Why the collision is certain, not merely possible

This is the part worth being precise about — it isn't a race or an unlikely coincidence.

`generate_bot_id` (`core/bot_management/services/bot_service.py:238`) decides whether a new bot gets `"default"` by asking:

```python
if not bot_repository.exists_by_owner_and_bot_id(owner_id, "default"):
    return "default"
```

That read goes through the Track A guard (`plugin_api/models.py:124-148`), which appends:

```python
BotModel.avernet_tenant == get_current_avernet_tenant()
```

So when tenant B's request asks "does this owner already have a `default` bot?", it **cannot see** tenant A's row — and correctly answers no. Tenant B therefore also allocates `bot_id="default"`.

The isolation guard is working exactly as designed. It is precisely *because* the local rows are isolated that the external identifiers converge.

### 1. Passport / agent identity (P1)

`plugins/community/passport.py` keys entirely on `bot_id`:

```python
def _token_for(bot_id: str) -> str:
    return f"community-passport-{bot_id}"

# _issue(...)
return {"token": _token_for(bot_id), "agent_code": bot_id}
```

No owner, no tenant. Two tenants' `default` bots resolve to one agent identity and one credential.

### 2. Owner auth relationships (P1)

`AuthRelationshipPlugin.create_relationship` is called with `work_no=user_id, agent_code=agent_code`, and `agent_code` is the `bot_id` above. The relationship key is `(owner, bot_id)` — no tenant — so tenant B's grant lands on the same tuple as tenant A's.

### 3. BCN coordination-network identity (P1 — cross-tenant **write**)

`BotService._sync_bot_to_bcn` (`bot_service.py:2645`):

```python
[...]

PR #668, opened 2026-07-31
refactor(bot): retire default bot_id for global uniqueness

lucas-xzp:
## Problem

Bot identity relied on the literal `bot_id="default"` as a first-bot / delete-protection / access proxy. This tied identity to a per-owner string that carries no tenant, so two tenants sharing a principal collapsed to one Passport principal / one BCN record — issue #556 — and forced the F49 BCN-sync stopgap on the openapi surface. `is_first_bot`, delete protection, collaborator owner-resolution, and bot_chat access all branched on the `"default"` string instead of on real state.

## Solution

Retire the `"default"` convention across the bot-identity surface (10 tasks):

- **T1** `generate_bot_id` always returns a globally-unique `yyyymmdd_xxxxxxxx` id; the owner-history lookup is dropped.
- **T2** `is_first_bot` derived from `count_by_owner == 0` (new public `BotService.is_first_bot`); `create_flow` calls it.
- **T3** refresh-token callback resolves the bot cross-tenant via `get_by_id_and_owner(..., execution_options={"skip_avernet_tenant_guard": True})`; the repo protocol gained a keyword-only `execution_options` opt.
- **T4** `delete_bot` protection = `count_by_owner <= 1` reject (was `bot_id == "default"` reject) via `BotOperationNotAllowedError`.
- **T5** collaborator `_resolve_owner_id` resolves owner via `repo.get_by_id` for all bot_ids (drops the `"default"` short-circuit).
- **T6** `bot_chat._check_bot_access` delegates to `has_bot_access` (drops `"default"`/`"{user}_default"` short-circuit).
- **T7** `create_bot_for_others_service` retires `_DEFAULT_BOT_ID`: CREATE → `generate_bot_id`, REPAIR → preserves existing id; owner-based lookup via `list_by_owner`.
- **T8** openapi `PUT /bots` re-enables BCN sync (removes forced `sync_to_bcn=False`); `(bot_id, owner_workno)` is now globally unique for new bots.
- **T9** `_apply_passport` tenant-aware: default-tenant non-first → `applyAgent`, else `applyFirst`.
- **T10** architecture guard green + full suite green (9692 passed); `create_flow` import of `utils.avernet_tenant` declared in `bot_management` README Context Boundary.

Design spec: `docs/superpowers/specs/2026-07-31-bot-id-global-uniqueness-design.md`. Plan: `docs/superpowers/plans/2026-07-31-bot-id-global-uniqueness.md`.

## Validation

- Rebased onto `origin/dev` (linear history, no merge commits). 4 conflict regions resolved (T1/T2 docstring refresh, T7 test contract, T8 comment/test) — all kept the new globally-unique contract.
- All 21 changed Python files syntax-valid; no conflict markers remain.
- Pre-push lint (SAST) hook green.
- New/updated tests assert the new contracts: `test_create_flow_is_first.py` (count-based `is_first` + tenant/RPC matrix), `test_delete_default_bot.py` (`count<=1` protection + `BotOperationNotAllowedError` escape), `test_create_bot_for_others_service.py` (owner-based lookup, REPAIR preserves id, CREATE uses `generate_bot_id`), `test_bot_passport.py` (`generate_bot_id != "default"`), `test_interceptor.py` (`"default"` no longer short-circuits), `test_refresh_token_callback.py` (cross-tenant `skip_avernet_tenant_guard` lookup), `test_bots_endpoints.py` (BCN sync re-enabled). Full module-gated suite runs in CI.

## Known deferred follow-ups (out of this PR scope, tracked separately)

- #11 `resources/router.py` `_DEFAULT_BOT_ID` fallback hardening.
- #12 `create_bot_for_others` multi-bot repair selection ordering (`_find_existing_bot_for_owner` returns `items[0]`).
- #13 Legacy `"default"` bot BCN residual collision risk on the BCN identifier (pre-retirement, unmigrated bots).
- #14 (proposed) Remaining legacy `"default"` code-path retirement in `create_bot` rollback guard (`bot_service.py:1458/1464` + `delete_bot:3191` cleanup guard, both pinned by `test_bot_service_create_default_protection.py`) and `bot_chat` read-path inline short-circuits (`service.py:438/797-799/847-849`). These are legacy-tolerant and low-risk; `delete_bot:3191` deserves an owner decision because T4 now allows legacy `"default"` bot deletion (count>1) but the cleanup guard still skips associated-data cleanup for the literal.

## Test plan

- [ ] CI module gates green on the PR.
- [ ] Changed-line coverage >= 80% (reproduced locally via `scripts/ci_test.sh`).
- [ ] Manual: create a bot on the default tenant → bot_id is globally unique; second bot → `applyAgent` (not `applyFirst`).
- [ ] Manual: create a bot via openapi (other tenant) → `applyFirst`, BCN sync active.
- [ ] Manual: delete the owner last bot → 409 `BotOperationNotAllowedError`; delete when owner has >=2 → succeeds.
- [ ] Manual: refresh-token callback for an external-tenant bot resolves (does not 404 on tenant guard).

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_chat/service.py
+++ b/src/backend/src/agentclaw/community/core/bot_chat/service.py
@@ -348,24 +348,7 @@ class BotChatService(OpenBotChatServiceMixin):
         return "db"
 
     def _check_bot_access(self, user_id: str, bot_id: str) -> bool:
-        """Check if user has access to the specified non-default bot.
-
-        Access is granted if the user is the bot owner (via ac_bots) or
-        a collaborator (via ac_bot_collaborator).
-
-        Args:
-            user_id: The user ID (staffId).
-            bot_id: The bot ID (non-default).
-
-        Returns:
-            True if access verified, False otherwise.
-        """
-        if (
-            bot_id == "default"
-            or bot_id == f"{user_id}_default"
-        ):
-            return True
-
+        """Check access: owner (ac_bots) or collaborator (ac_bot_collaborator)."""
         return self._db_repo.has_bot_access(user_id, bot_id)
 
     async def list_sessions(
--- a/src/backend/src/agentclaw/community/core/bot_chat/service.py
+++ b/src/backend/src/agentclaw/community/core/bot_chat/service.py
@@ -351,6 +351,24 @@ class BotChatService(OpenBotChatServiceMixin):
         """Check access: owner (ac_bots) or collaborator (ac_bot_collaborator)."""
         return self._db_repo.has_bot_access(user_id, bot_id)
 
+    @staticmethod
+    def _is_legacy_default_bot_id(bot_id: str | None, owner_id: str) -> bool:
+        """True iff ``bot_id`` is a legacy per-owner default alias (pre-retirement).
+
+        These ids (`"default"` and ``f"{owner_id}_default"``) were unique only per
+        owner, so :func:`has_bot_access` cannot disambiguate them across owners
+        via ``get_by_id`` — multiple owners each have a `"default"` bot. Routes
+        over historical traces keep the legacy short-circuit (owner match by
+        user_id / scoped list) instead of routing through ``has_bot_access``.
+        New bots are globally unique and never hit this branch.
+
+        Kept as a service-layer helper so the alias set + rationale live in one
+        place rather than scattered across list/get-session paths.
+        """
+        if not bot_id:
+            return False
+        return bot_id == "default" or bot_id == f"{owner_id}_default"
+
     async def list_sessions(
         self,
         owner_id: str,
@@ -436,11 +454,10 @@ class BotChatService(OpenBotChatServiceMixin):
                 include_output_match=include_output_match,
             )
 
-        # Default DB mode: for non-default bot, check access (owner or collaborator).
-        if (
-            bot_id
-            and bot_id != "default"
-        ):
+        # Default DB mode: for non-(legacy-default) bot, check access (owner or collaborator).
+        # Legacy default aliases short-circuit (see _is_legacy_default_bot_id): the
+        # subsequent DB list is already scoped by owner_id, so access is implicit.
+        if bot_id and not self._is_legacy_default_bot_id(bot_id, owner_id):
             has_access = self._check_bot_access(owner_id, bot_id)
             if not has_access:
                 return SessionListResponse(
@@ -794,13 +811,14 @@ class BotChatService(OpenBotChatServiceMixin):
             raise SessionNotFoundError(f"Trace {trace_id} not found or not accessible")
 
         # Bot access check based on row's bot_id.
-        # - default bot (or null bot_id): require trace's user_id to match caller.
-        # - non-default bot: require caller to be owner or collaborator.
+        # - legacy default alias (or null bot_id): require trace's user_id to match
+        #   caller — see _is_legacy_default_bot_id; has_bot_access can't disambiguate
+        #   a literal "default" across owners.
+        # - non-legacy bot: require caller to be owner or collaborator.
         if owner_id:
             if (
                 row.bot_id is None
-                or row.bot_id == "default"
-                or row.bot_id == f"{owner_id}_default"
+                or self._is_legacy_default_bot_id(row.bot_id, owner_id)
             ):
                 if row.user_id != owner_id:
                     raise SessionNotFoundError(f"Trace {trace_id} not found or not accessible")
@@ -850,7 +868,7 @@ class BotChatService(OpenBotChatServiceMixin):
 
                 if (
                     trace_bot_id is None
-                    or trace_bot_id == "default"
+                    or self._is_legacy_default_bot_id(trace_bot_id, owner_id)
                 ):
                     if trace_owner_id != owner_id:
                         raise SessionNotFoundError(f"Trace {trace_id} not found or not accessible")
--- a/src/backend/src/agentclaw/community/core/bot_collaborator/interceptor/collaborator.py
+++ b/src/backend/src/agentclaw/community/core/bot_collaborator/interceptor/collaborator.py
@@ -431,16 +431,17 @@ class CollaboratorPermissionInterceptor:
         """当请求未提供 owner_id 时，主动解析 Bot 归属。
 
         解析策略：
-        1. bot_id 缺失或为 "default" → 返回当前 user_id（语义=我的 bot）
-        2. bot_id 有值 → 通过 BotRepository.get_by_id 查询 bot 记录，
-           返回 bot.owner_id
+        1. bot_id 缺失 → 返回当前 user_id（语义=我的 bot）
+        2. bot_id 有值（含历史的 "default"）→ 通过 BotRepository.get_by_id
+           查询 bot 记录，返回 bot.owner_id
         3. BotRepository 不可用 或 bot 不存在 → 返回 None（回退到 skip）
 
         返回 None 表示无法解析，调用方应回退到跳过权限检查
         （让业务层返回更具体的错误如 404，而非拦截器一刀切 403）。
         """
-        if not bot_id or bot_id == "default":
+        if not bot_id:
             return user_id
+        # 历史的 bot_id=="default" 不再短路;统一走 repo 解析 owner
 
         if ctx.injector is None:
             logger.warning(
--- a/src/backend/src/agentclaw/community/core/bot_collaborator/interceptor/collaborator.py
+++ b/src/backend/src/agentclaw/community/core/bot_collaborator/interceptor/collaborator.py
@@ -441,7 +441,12 @@ class CollaboratorPermissionInterceptor:
         """
         if not bot_id:
             return user_id
-        # 历史的 bot_id=="default" 不再短路;统一走 repo 解析 owner
+        # 存量 default bot 兜底:单租户内每 owner 都有一条 bot_id="default",
+        # repo.get_by_id("default") 会歧义命中任意 owner 的 default bot,导致串户。
+        # 旧语义 default=我自己的 bot,保留此短路避免协作者鉴权用错 owner_id。
+        # 新 bot 永不为 default (generate_bot_id 全局唯一),此分支仅命中存量。
+        if bot_id == "default":
+            return user_id
 
         if ctx.injector is None:
             logger.warning(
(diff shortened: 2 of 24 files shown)


src/backend/src/agentclaw/community/core/bot_chat/service.py
```diff
@@ -344,24 +344,7 @@ def _get_log_source(self, log_source: str | None) -> str:
...
-        a collaborator (via ac_bot_collaborator).
-
-        Args:
-            user_id: The user ID (staffId).
-            bot_id: The bot ID (non-default).
-
-        Returns:
-            True if access verified, False otherwise.
-        """
-        if (
```
totalfrank, line 363: This logic also says if it's the default bot(I guess it means first), return true right away.

I mean, should we keep this and rewrite it to:

bot_id == "default"
or bot_id == f"{user_id}_default"
// if no default bot, then the one created the earliest should be the one that can not be deleted.
or ...

And this whole logic can be moved to a service layer method so that we don't have the same logic spreads all over the place?

src/backend/src/agentclaw/community/core/bot_collaborator/interceptor/collaborator.py
```diff
@@ -428,16 +428,17 @@ def _resolve_owner_id(
...
+        1. bot_id 缺失 → 返回当前 user_id（语义=我的 bot）
+        2. bot_id 有值（含历史的 "default"）→ 通过 BotRepository.get_by_id
+           查询 bot 记录，返回 bot.owner_id
         3. BotRepository 不可用 或 bot 不存在 → 返回 None（回退到 skip）
 
         返回 None 表示无法解析，调用方应回退到跳过权限检查
         （让业务层返回更具体的错误如 404，而非拦截器一刀切 403）。
         """
-        if not bot_id or bot_id == "default":
+        if not bot_id:
```
totalfrank, line 442: We should not make this change, because if legacy default bots falls to

bot = repo.get_by_id(bot_id) on line 454 below and you would get multiple users.

I think we should keep this logic here for legacy compatibility and let's make sure we make a comment to note this down.

src/backend/src/agentclaw/community/plugins/bot_repository.py
```diff
@@ -144,19 +144,25 @@ def _to_caller_identity_dict(bot: BotModel) -> Dict[str, Any]:
         return result
 
     def get_by_id_and_owner(
```
totalfrank, line 146: I just want to call out that, because we introduced register_avernet_tenant_guard for Bots, this essentially also filters on tenant. Just want to call that out so you can double check the place you are using it is aware of this.

src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
```diff
@@ -820,28 +781,14 @@ async def get_bot_by_ip_and_user(self, ip: str, user_id: str) -> Optional[Dict[s
...
-        Returns:
-            True if the user owns no bots, False otherwise
-        """
+    def _is_first_bot(self, user_id: str) -> bool:
+        """First bot iff the owner has zero bots (current env; tenant enforced by session guard)."""
         return self._repository.count_by_owner(user_id) == 0
 
+    def is_first_bot(self, user_id: str) -> bool:
+        """Public alias (used by create_flow) for :meth:`_is_first_bot`."""
+        return self._is_first_bot(user_id)
```
totalfrank: Nit: this delegation seems to be redundant. Let's get rid of it?

src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
```diff
@@ -3188,15 +3135,15 @@ def delete_bot(self, bot_id: str, user_id: str, nick_name: Optional[str] = None)
...
+            # 至少保留一个 Bot:删掉这只后归零则拒。
             # 必须在 release_device / destroy_passport 之前拦截,否则会误销毁
             # agent 许可证 (Passport) 并重置引擎配置 (openclaw.json)。
             # 用 BotOperationNotAllowedError（BotServiceError 子类）表达"这是客户端
             # 不支持的操作"，而不是服务端故障：重试永远不会成功。内部路由的 except 链没有
             # 这一分支，仍落到 `except BotServiceError` → 500，行为不变；公共 API 则按
             # 4xx 映射。
-            if bot_id == "default":
-                raise BotOperationNotAllowedError("default bot 不允许删除")
+            if self._repository.count_by_owner(user_id) <= 1:
```
totalfrank: Sure I proposed this earlier. But I have given it a second thought, maybe

check if this bot is the one with the earliest gmt occur is the better way to do here? Because this way we essentially kept logic the same.

src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
```diff
@@ -3188,15 +3135,15 @@ def delete_bot(self, bot_id: str, user_id: str, nick_name: Optional[str] = None)
...
+                raise BotOperationNotAllowedError("至少保留一个 Bot，不能全部删除")
```
totalfrank: And if we do, then this becomes "第一个 bot 不允许删除"

src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
```diff
@@ -4625,7 +4572,14 @@ def hot_update_passport_token_to_device(self, bot_id: str, user_id: str, token:
...
 
         # 1. 获取 bot 并校验权限
-        bot = self._repository.get_by_id_and_owner(bot_id, user_id)
+        # 回调走 /api 前缀 → AvernetTenantMiddleware 套 DEFAULT 租户 teamclaw,
+        # 但外部租户 bot 的 passport 刷新需跨租户直查。(bot_id, owner_workno)
+        # 全局唯一,跨租户直查安全。
+        bot = self._repository.get_by_id_and_owner(
+            bot_id,
+            user_id,
+            execution_options={"skip_avernet_tenant_guard": True},
```
totalfrank, line 4611: This is exactly what we need!

totalfrank (commented):
Identified a few issues might need your attention, please take a look.

src/backend/src/agentclaw/community/core/bot_chat/service.py
lucas-xzp: Good point — addressed in the follow-up commit. The default-bot short-circuit is kept for legacy compatibility and is now centralized in one service-layer helper, `_is_legacy_default_bot_id(bot_id, owner_id)` in `bot_chat/service.py`, covering both `"default"` and `f"{owner_id}_default"`.

`_check_bot_access` itself is now a thin wrapper over `repository.has_bot_access(user_id, bot_id)` (owner-or-collaborator), intended for new globally-unique bot_ids. The legacy short-circuit is applied at the `list_sessions` / `get_session_detail` call sites rather than here, because a literal `bot_id="default"` is ambiguous across owners (every owner has one) and `has_bot_access` → `get_by_id("default")` cannot disambiguate it — so those paths fall back to `row.user_id == caller` matching. New bot_ids are globally unique and never hit this branch.

src/backend/src/agentclaw/community/core/bot_collaborator/interceptor/collaborator.py
lucas-xzp: Agreed — reverted to keeping the short-circuit and added an inline comment explaining why. A legacy `bot_id="default"` exists once per owner within a tenant, so `repo.get_by_id("default")` on line ~454 would arbitrarily resolve to any owner's default bot and could return the wrong `owner_id` (cross-owner leak) for collaborator auth.

`if bot_id == "default": return user_id` preserves the old semantics (default = my own bot). It is only reachable by legacy bots; new ids are globally unique and flow through the repo lookup below.

src/backend/src/agentclaw/community/plugins/bot_repository.py
lucas-xzp: Thanks for flagging — confirmed we're aware. The tenant guard on `get_by_id_and_owner` is the desired default (tenant-scoped lookups), and all regular callers rely on that scoping.

The one caller that must reach across tenants is `refresh_passport_token`'s callback path, which now explicitly opts out via `execution_options={"skip_avernet_tenant_guard": True}` (see the block further down in `bot_service.py`). That's safe precisely because `(bot_id, owner_workno)` is now globally unique after this PR.

src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
lucas-xzp: Done — removed the private `_is_first_bot` and the redundant alias. There's now a single `is_first_bot(user_id) -> count_by_owner(user_id) == 0`, which `create_flow` calls directly. Also trimmed the now-obsolete docstring rationale about the default-tenant proxy.

src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
lucas-xzp: Good call — switched from the count-based guard to earliest-created protection, which preserves the original "the owner's protected bot stays" semantics more faithfully (independent of how many bots the owner has).

Implementation in `delete_bot`: fetch the owner's live bots via `list_by_owner(user_id, 1, 1000)`, pick `min` by `gmt_create` (normalized to string to avoid datetime-vs-ISO mixed-type comparison, empty defaults to earliest), and reject deletion only when `bot_id == earliest_bot_id`. This also naturally keeps ≥1 bot per owner.

src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
lucas-xzp: Yes — updated accordingly. The guard now rejects only the earliest-created bot, and the error message is `"不能删除首个创建的 Bot，该 Bot 受保护"` ("Cannot delete the first-created Bot; it is protected"), matching the new semantics. Tests/acceptance assertions were updated to the new message in the `8d0d724a` / follow-up commits.

src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
lucas-xzp: 👍 glad it lines up. The comment above it spells out the rationale: the refresh-token callback enters via the `/api` prefix (DEFAULT tenant teamclaw) but must reach external-tenant bots, and `(bot_id, owner_workno)` is globally unique now, so the cross-tenant direct lookup with `skip_avernet_tenant_guard` is safe.

totalfrank (commented):
LGTM.

totalfrank (approved):
LGTM.

PR #672, opened 2026-07-31
fix(backend): settle obsolete skills pool wakeups

XIEAN99:
## Summary

- settle persisted runtime wakeups whose observation predates Pool activation without writing runtime evidence
- treat equal or newer generation-scoped runtime evidence as safely superseding a stale retry, preserving the newer evidence
- keep missing layout state, activation timestamp, or quarantine fail-closed for retry

## Runtime evidence

During Avernet #455 Desktop Hermes acceptance, reconcile task 17061 carried an observation from before Pool activation. A later wakeup had already persisted READY evidence, but task 17061 continued retrying with `skills pool runtime reconciliation evidence race lost` until rollback removed the Pool state.

## Tests

- `DEPLOY_PROFILE=test uv run pytest tests/community/core/skills_pool -q` (250 passed)
- Ruff check on changed Python files
- `git diff --check`

## Scope

Backend-only persistence semantics. No API, schema, config, Engine, or Desktop image changes.


FreddieSun (approved):
lgtm

PR #680, opened 2026-08-01
fix(session-files): harden materialization and proxypass access

yhzhang35:
## Problem

Session File materialization must remain usable when a device connection lacks optional identity metadata, while ready files need a browser-compatible Engine data path. Portable filenames, including Chinese text and parentheses, also need to be accepted without weakening path safety.

## Solution

- Resolve a missing connection tenant from the injected `BaasConfig.tenant`, keeping enterprise values outside community source; reject the request when both values are absent.
- Add diagnostics for rejected workspace paths.
- Accept printable, portable Unicode filenames while rejecting path segments, control characters, Windows-forbidden characters, and overlong UTF-8 names.
- Trust the existing Bot proxypass authentication for Engine Session File list, content, and delete routes; remove the redundant `x-iam-token` requirement.
- Keep regression coverage for materialization, reference validation, Session File routes, filename handling, and unconfigured tenant rejection.

## Validation

- `40 passed`: focused Engine Session File, materialization, reference, transport, and plugin tests.
- `34 passed`: focused Backend Session Resource tests plus the shipped-config architecture guard.
- Focused Ruff checks and `git diff --check` passed.

## Compatibility and risk

Ready-file access continues to require the existing Bot proxypass boundary. A missing connection tenant now requires the deployment's existing BaaS configuration; if neither is available, upload intent fails explicitly instead of guessing a tenant. Filename acceptance expands only to printable names that satisfy cross-platform filesystem restrictions; traversal and size protections remain enforced.

## Spec

- `spec/pipeline/session-file-sharing-dev-rebase/009-pr-report.md`

### DIFF
--- a/docs/superpowers/specs/2026-07-28-session-file-sharing-engine-content-design.md
+++ b/docs/superpowers/specs/2026-07-28-session-file-sharing-engine-content-design.md
@@ -29,23 +29,24 @@ separate owners and are not changed here.
 
 ## Content Delivery
 
-`GET /api/session-resources/{resource_id}/content?disposition=inline|attachment`
-is the public Backend endpoint. Backend authorizes owner, Bot, session, and
-`ready` state, then streams the Engine endpoint
-`GET /api/resource-materializations/{resource_id}/content` through the
-`DeviceAdapterTransport` stream contract.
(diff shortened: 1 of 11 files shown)


totalfrank (approved):
LGTM.

PR #706, opened 2026-08-03
refactor(backend): normalize /openapi/v1/bots paths and drop channels

totalfrank:
## Problem

The public `/openapi/v1/bots` surface grew one component at a time, and the components disagree about where a path says *which component* and where it says *which bot*. Three shapes are in use, and only one is the one the surface was designed around (`/openapi/v1/bots/<component>/…`):

- **`identity` carries a redundant `/bot/` segment.** `GET /openapi/v1/bots/identity/bot/{bot_id}` says "bot" twice — once in the base, again before the id — and no other component does.
- **Six components put `{bot_id}` *before* their own name.** `/{bot_id}/connection`, `/{bot_id}/engine/…`, `/{bot_id}/sessions/…`, `/{bot_id}/models/…`, `/{bot_id}/approvals/…` and `/{bot_id}/skills` are addressed as if they were sub-resources of the *bots* component. They are not — they are separate components owned by different modules, documented separately, and mounted with a different error contract (`ENGINE_RUNTIME_ERROR_RESPONSES`).
- **`channels` is published but not implemented.** Six operations across two paths sit in the served document with `NotImplementedError` handlers, so an integrator reading it sees a channels API and gets a 500.

This is not only untidy:

1. **The wildcard eats the namespace.** `/openapi/v1/bots/{bot_id}` matches any single segment, so component names resolve only because the component routers are mounted first. A reader of `engine_runtime/sessions/router.py` cannot tell whether `/openapi/v1/bots/{bot_id}/sessions` is served there or by a `{bot_id}`-shaped route in the bots component — the router file no longer states its own address.
2. **It blocks a second owner under the same base.** BCS reached this from the other side and resolved it the same way: its bot control-plane operations moved to `/openapi/v1/bots/collaboration/{bot_id}` precisely because a bare `{bot_id}` under the shared base "collides with the Gateway's existing Backend-owned Bot surface and … make[s] runtime ownership ambiguous" (`src/bcs/docs/plans/2026-08-03-bcn-collaboration-paths-design.md`). The backend is the other half of that collision and had not moved.
3. **The published artifact is stale.** The gateway's pinned `configs/schemas/bots.openapi.json` carried **32 paths against the backend's 43** — every engine-runtime path shipped in Track C (#630) was missing from the document the gateway serves.

## Solution

One addressing rule for the whole surface:

```text
/openapi/v1/bots/<component>            # the component's own collection
/openapi/v1/bots/<component>/{bot_id}   # …scoped to one bot
```

The component's **literal** name comes first; a bot-scoped operation takes `{bot_id}` as the first segment *after* it. The `bots` component is the one exception, and only because it *is* the component the base names — it keeps `/openapi/v1/bots` and `/openapi/v1/bots/{bot_id}`, with its own sub-resources (`/status`, `/passport`, `/restart`, `/auth-status`, `/engine-config`) beneath the bot record. Those are properties of the bot, not other components borrowing its address.

| Today | After |
| --- | --- |
| `/openapi/v1/bots/identity/bot/{bot_id}[/{file_type}]` | `/openapi/v1/bots/identity/{bot_id}[/{file_type}]` |
| `/openapi/v1/bots/{bot_id}/connection` | `/openapi/v1/bots/connection/{bot_id}` |
| `/openapi/v1/bots/{bot_id}/engine/{status,capabilities,available}` | `/openapi/v1/bots/engine/{bot_id}/…` |
| `/openapi/v1/bots/{bot_id}/approvals/{mode,modes}` | `/openapi/v1/bots/approvals/{bot_id}/…` |
| `/openapi/v1/bots/{bot_id}/sessions/…` | `/openapi/v1/bots/sessions/{bot_id}/…` |
| `/openapi/v1/bots/{bot_id}/models[/{model_id}]` | `/openapi/v1/bots/models/{bot_id}[/{model_id}]` |
| `/openapi/v1/bots/skills[/{skill_id}]` | `/openapi/v1/bots/skills/catalog[/{skill_id}]` |
| `/openapi/v1/bots/{bot_id}/skills[/{skill_id}]` | `/openapi/v1/bots/skills/{bot_id}[/{skill_id}]` |

`bots`, `mcp`, `resources` and `routines` already obeyed the rule and are untouched. **41 published paths, down from 43.**

**Why `skills` needs a literal `catalog`.** The component owns two resource families. Under the rule the bot-scoped one takes `/openapi/v1/bots/skills/{bot_id}`; a catalog detail at `/openapi/v1/bots/skills/{skill_id}` would occupy the same slot with a different meaning — two wildcards at one depth, which no ordering rule can disambiguate. `catalog` is the device the surface already uses for `check-name` and `ceiling`. Rejected alternatives: exempting skills from the rule, or inventing a `skill-catalog` component the module layout does not have.

**Why `channels` is deleted, not parked.** It was never implemented, but it was *published* — a parked stub is not a dormant row on a board, it is six operations in the served document that answer 500. Its Track A stage never started, so there is no data work to unwind.

**Why no compatibility aliases.** The surface has no reachable external caller yet (`route_security` requires a Google-resolved `user` identity, which a tenant presenting an access key cannot satisfy), so there is no contract to preserve — and an alias at `/openapi/v1/bots/{bot_id}/sessions` would re-create exactly the wildcard-vs-component ambiguity this removes, while making the document describe two addresses for one resource. Same reasoning BCS applied to the same migration.

**No behaviour change.** No handler, schema, status code, authorization rule or tenant-scoping rule moves. `bot_id` was already a declared parameter on every affected route, so relocating the segment changes no signature. Gateway routing is untouched: `DomainMap.domain_for` keys on the first segment after `/openapi/v1`, which is `bots` for every path here, before and after.

**A test now holds the rule.** `tests/…/openapi_v1/test_path_convention.py` asserts against the *generated document* rather than a hand-maintained list: no `/bot/` segment, no `channels`, no component name behind the `{bot_id}` wildcard, only `{bot_id}` in top-level parameter position — and that the reserved-name list in `docs/openapi-v1/README.md` still equals the literals the routes publish, so the docs cannot silently fall behind a component added later.

## Validation

- **`tests/community/adapters/http/openapi_v1`: 449 passed**, including the new `test_path_convention.py`.
- **Surface verified from the dump, not from the plan.** `DEPLOY_PROFILE=community python scripts/dump_openapi.py` → 41 paths; every row in the table above confirmed against the actual output, and the doc tables were written from it.
- **Artifact regenerated through the real pipeline**, not copied: `scripts/gate_and_publish_openapi.py configs/schemas/bots.openapi.json <candidate> --allow-breaking`. The gate refuses without the flag, as it should.
- **Endpoint coverage gate** (`tests/community/framework/test_coverage_gate.py`) — see the correction below. Its frozen baseline is keyed on `(method, path)`, so 24 entries were re-addressed and the 6 `channels` entries dropped. Verified by executing the gate's own computation directly — `compute_missing_coverage` against the real app route table with all 623 registered cases, diffed against the baseline: **0 NEW, 0 STALE**. The test itself cannot run in this sandbox (its `app_with_testing_modules` fixture needs a SQLite setup that is absent here).
- **Gateway suite: 586 passed**, 1 failure in `test_ruff_lint_rules.py::test_ruff_formatting_passes` that flags code blocks in `docs/*.md` files this PR does not touch (a newer ruff formats them differently); `tests/e2e` needs a live server unavailable here. Both reproduce unmodified, and **Gateway unit tests pass in CI**.
- **Lint/format diffed before vs after** on every changed file: no new findings, no new format drift. Removing the `channels` import actually *fixed* a pre-existing `I001` in `openapi_v1/__init__.py`.
- **Repo-wide audit** for stale references: the only remaining `/openapi/v1/bots/{bot_id}/…` occurrences are bots-owned sub-resources, deliberate historical prose, and BCS's own negative assertions.

### Correction to an earlier claim in this description

The first version of this PR said the full backend suite's 684 failures were "environmental, not from this change," based on re-running every failing module on this branch and its parent and getting identical 56-failure sets.

**That check had a blind spot and the claim was wrong for one test.** `test_coverage_gate.py` **errors** at fixture setup when run outside a full suite, rather than failing — and the comparison filtered on `^FAILED`, so it dropped out of *both* sides and the diff read as identical. CI caught it. The baseline fix is in `f1d3ba6`.

The rest of the claim stands: the remaining failures are environmental and reproduce on the parent commit. A full-suite baseline on the parent was not run — it exceeded the sandbox's time budget — so for those, the module-level comparison is still what backs the claim, with the filtering gap now known and accounted for.

## Compatibility and risk

**This is a breaking path change, and the compat gate is meant to say so.** It reported 43 backward-incompatible changes and refused to publish; the artifact was published with `--allow-breaking` as a coordinated change. The list includes items that are **not from this PR** — they are drift the stale artifact had accumulated and that regenerating it now surfaces:

- from this PR: the `channels`, `identity` and `skills` operation removals, and the `Channel*` schema removals;
- pre-existing drift: `bot_id` becoming required on `resources`/`routines`, `BotCreate.engine_options`, `BotUpdate.cluster_name`/`engine_options`, and `McpConfigWrite.sync_mode` — all landed in earlier PRs against an artifact that was never republished.

Risk is bounded by there being **no reachable external caller** on this surface yet, which is also why no aliases were added. Rollback is a straight revert: the change is router prefixes, route decorators, docs, one regenerated artifact, and the coverage baseline.

Environmental failures observed in this sandbox, all reproducing on the parent commit and none related to routing:

- `tests/community/api/bot_management/test_router.py` — legacy `/api/bots` admin routes answer 403 instead of 404 because `super_admin()` reads a deployment config allow-list the community distribution ships empty.
- `tests/community/endpoints/*`, `tests/community/e2e/*` — `AssertionError: engine should have been created by SqliteDB session()`, a DB fixture this sandbox does not provide.
- Gateway `tests/e2e/*` — require a live server.

## Spec

`src/backend/specs/2026-08-03-openapi-v1-path-normalization/` (`spec.md`, `plan.md`, `tasks.md`).

Docs updated in the same commit, as `docs/openapi-v1/README.md` requires: a new **Addressing rule** section (EN + zh-CN) with the reserved-name list, the per-component endpoint tables, the Track C inventory in `engine-surface.md` (EN + zh-CN), the status board, and a dated changelog entry. `src/gateway/configs/schemas/README.md` now records that these artifacts are build outputs and must be regenerated rather than hand-edited.


FreddieSun (approved):
LGTM

xxxxpenny (approved):

PR #707, opened 2026-08-03
feat(backend): include caller instances in Passport token refresh

XIEAN99:
## Problem

Service Bot Passport refresh currently updates Draft, Online, and Verify bindings, but caller mode can create multiple additional BaaS caller-instance bindings. Those caller containers still require the service Bot owner's normal Passport token, yet the refresh path does not discover or update them.

The existing best-effort stage loop also catches only `BotServiceError`; provider failures such as `BaasDeviceServiceError` derive from `DeviceServiceError` and can stop later stages before they are attempted.

## Solution

- Add an environment/owner/Bot-scoped repository query for ACTIVE BaaS bindings identified by `apply_reason=caller_instance:{bot_id}`, with defensive `device_id` de-duplication.
- Fan out the same owner Passport token to every caller binding with a fixed five-worker pool and preserved tenant context.
- Limit caller updates to ACTIVE BaaS physical devices; an ACTIVE binding with zero ACTIVE physical devices is a failure.
- Attempt Draft, Online, Verify, and all callers before aggregating any errors into the existing `BotServiceError` response path.
- Keep caller-specific `x-caller-token` handling entirely out of this flow.
- Extract BaaS outbound-header updating behind a small internal module to keep the device lifecycle module below the architecture size cap.
- Stop writing Passport token prefixes to refresh/routing logs; logs now record only whether a token was present.

## Validation

Candidate rebased onto `origin/dev@f16416c9`.

- `DEPLOY_PROFILE=test uv run pytest tests/community -q`: 10,284 passed after rebase.
- `DEPLOY_PROFILE=test uv run pytest tests/community/architecture -q`: 112 passed.
- Targeted repository, BaaS device, router, BotService, and HTTP callback suites: 197 passed before the full run.
- Ruff check on all changed Python files: passed after rebase.
- `scripts/ci/python_sast_local.sh src/backend 1 --base origin/dev --head HEAD`: passed.
- `scripts/ci/pre_push.sh --base origin/dev --head HEAD` (lint-only gate): passed.
- Live pre-environment callback validation is intentionally pending: it requires an enabled caller-mode Bot and a manually supplied sensitive token. Validation should inspect only non-sensitive target/count/success logs.

## Compatibility and risk

- No HTTP request or response schema changes.
- No database DDL or migration.
- Personal Bot behavior and non-caller Service Bot behavior remain unchanged.
- Partial failure remains `success=false`, `error_code=500`, so the existing callback retry behavior is preserved.
- The update PUT remains idempotent; retries may re-update already successful targets.

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
@@ -11,6 +11,7 @@ import shutil
 import string
 import threading
 import time
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime
 from pathlib import Path
 from typing import Callable, Optional, Dict, Any, List, Tuple, TYPE_CHECKING
@@ -66,9 +67,7 @@ from agentclaw.community.core.workspace.path_factory import (
 from agentclaw.community.core.service_bot.repository.models import PublishStatus
 from agentclaw.community.core.service_bot.types import PublishStage
 from agentclaw.community.utils.avernet_tenant import (
-    DEFAULT_AVERNET_TENANT,
     bind_current_avernet_tenant,
-    get_current_avernet_tenant,
 )
 from agentclaw.community.utils.env_utils import get_current_env
 from agentclaw.community.core.devices.errors import (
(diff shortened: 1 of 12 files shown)


totalfrank (approved):
LGTM.

PR #709, opened 2026-08-03
fix(baas): honor active runtime engine type

duxingxingstar:
## Problem

BaaS previously consumed Backend binding responses using the original `engine_type` only. For Personal Bots whose actual runtime engine is explicitly configured in `ac_templates.ext.active_runtime_engine_type`, this could select the wrong Engine Adapter. Invalid or empty runtime-engine data must not alter existing behavior.

## Solution

- Add `active_runtime_engine_type` to Backend binding responses.
- Read the Personal Bot value from template configuration through a bot-type-specific processor.
- At the BaaS HTTP consumption boundary, override the existing `engine_type` only when the runtime value is a non-empty supported type.
- Fall back to the original Backend `engine_type` for missing, empty, non-string, or unsupported values.
- Keep `BotBindingData`, `_normalize_engine_type()`, Runner, Executor, Queue Worker, and Adapter selection logic unchanged.

## Validation

- BaaS BotService Plugin: 49 passed
- Backend service tests: 127 passed
- Backend endpoint/router tests: 28 passed
- BaaS binding resolver/run-utils tests: 105 passed
- Backend architecture tests: 13 passed
- BaaS Runner/Executor tests: 100 passed
- Targeted Flake8 checks passed for all changed Backend and BaaS files
- `git diff --check` passed

## Compatibility and risk

Older Backend responses without `active_runtime_engine_type` continue using the original `engine_type`. Invalid explicit values also fall back to the original value, preserving the existing normalization and execution path.

### DIFF
--- a/src/baas/src/secbaas/community/plugins/bot_service/real/_plugin.py
+++ b/src/baas/src/secbaas/community/plugins/bot_service/real/_plugin.py
@@ -28,6 +28,10 @@ from secbaas.community.spi.bot_service import (
 
 logger = get_logger("plugin-bot-service")
 
+_SUPPORTED_RUNTIME_ENGINE_TYPES = frozenset(
+    {"openclaw", "teclaw", "aicoding", "hermes", "claude_code"}
+)
+
 
 class AiohttpBotServicePlugin(BotServicePlugin):
     """Aiohttp-based BotService plugin — production HTTP implementation.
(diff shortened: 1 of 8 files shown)


xxxxpenny (approved):

totalfrank (approved):
LGTM.

PR #713, opened 2026-08-03
refactor(gateway): route domains by path pattern and move the bot socket under bots

totalfrank:
## Problem

`GET /openapi/v1/bots/connection/{bot_id}` hands a tenant a finished socket URL
that points at a *sibling* top-level prefix:

```text
before  wss://{gw}/openapi/v1/engine/{target}/api/openclaw/ws?x-proxypass-token=…
after   wss://{gw}/openapi/v1/bots/messages/{target}/api/openclaw/ws?x-proxypass-token=…
```

The BCN collaboration-prefix design (approved 2026-08-03) settled that the
leading segment names the **owner** of a surface, not the process serving it.
Backend management and the engine-proxy runtime are one team and one scope, so
`engine` as a sibling of `bots` splits one owner across two namespaces. The
connection service's own stated goal is also that "nothing names the hop behind
the gateway," and a top-level `engine` domain comes close to naming one.

Neither is reachable by a config edit: a gateway domain *was* its leading path
segment, so `bots` could not simultaneously be a backend prefix and the parent of
a socket prefix served by a different upstream.

## Solution

A domain now claims a **path pattern** — literal segments under the version base,
then `**` — plus the **planes** it answers. Resolution gathers every domain
matching the path *and* answering the requesting plane, then takes the most
specific: match first, rank second.

The plane filters the **candidates** rather than vetoing the winner, and that is
load-bearing. `messages` is reserved for a future *HTTP* endpoint here; ranking
first and checking the plane after would let the socket domain win the path, fail
the check, and make that endpoint unreachable the day it is added — the prefix
would eat exactly what it was reserved for. Recovering that by *retrying* the
next-best candidate would be worse: a request the winning domain refused would
get a second attempt at a domain that did not win. So retry-on-mismatch is
deliberately not built.

A domain that declares no `match` claims `{base_path}/{name}/**` — exactly the
leading-segment routing it had before — so the four shipped HTTP domains are
unchanged.

The pattern grammar and its specificity ranking moved into one shared
`core/paths` module, adopted by both the routing plane and `RouteSecurity`. Two
implementations of "most specific" is somewhere the planes can come to disagree
about which rule governs a request, and that disagreement has the shape of an
authorisation bypass.

**Boot-time validation replaces an invariant that used to be structural.** While
a domain *was* its leading segment, an unknown segment resolved to `None` and
"never an open proxy" could not be mistyped. `match: /**` is precisely an open
proxy, so a pattern must now pin the version base plus one literal segment, and
two domains claiming one pattern on one plane are refused as having no defined
winner.

### ⚠️ A security bug found during implementation

The spec recorded the objection "this puts an auth-exempt path inside an
authenticated prefix" as **"a legibility concern, not a correctness one"**, on
the grounds that precedence ranks the longer prefix higher. Precedence does — but
that was never the issue. **The route-security table is plane-blind:** an
unqualified rule governs every request to a path, whichever entrypoint serves it.

Chained with the fallthrough this change introduces, that was an authentication
bypass:

1. the exemption covered ordinary HTTP requests to the prefix, not just handshakes;
2. those no longer 404 as they did when the socket was its own top-level domain —
   they fall through to `bots` and are forwarded to the backend;
3. the path travels verbatim, so its dot segments normalise away en route.

`GET /openapi/v1/bots/messages/../../admin/keys` reached the backend as
`/openapi/v1/admin/keys` **with no identity at all**. The move is exactly what
made the exemption reachable on the HTTP plane; it was harmless only while the
exempt prefix was both socket-only *and* top-level, with nothing beneath it to
fall through to.

Fixed by qualifying the exemption by plane, using machinery the table already
has — its keys carry an optional method qualifier. The relay authenticates
handshakes under a distinct `WEBSOCKET` method and the rule is written
`WEBSOCKET /openapi/v1/bots/messages/**`, so an HTTP `GET` there falls to
`/openapi/v1/bots/**` and requires a user — which is also the right answer for
the endpoint the prefix is reserved for. `spec.md` carries the corrected
reasoning, and `test_the_socket_exemption_does_not_reach_the_http_plane` pins it.

### Two corrections to the original handoff

- **The reserved-name interlock runs the opposite way.**
  `test_the_docs_reserved_names_match_the_routes` asserts the documented list
  *equals* the components the routes publish, so adding `messages` to that block
  fails it. Relaxing it to a subset check would give up the only thing stopping
  the docs falling behind a component added later — so names reserved *ahead* of
  their routes get a separate anchored list, with its own assertion that the two
  stay disjoint. The prose also states the accurate reason: `messages` is not
  unreachable the way the other twelve are, since an HTTP `GET` to it still
  arrives at the backend and resolves as a bot id.
- **The pinned artifact did need regenerating.** Published *paths* are unchanged,
  but the socket URL appears as an example value in two places.

## Validation

- **Gateway** — 630 passed. The one failure, `test_ruff_formatting_passes`, is
  pre-existing: the identical 8-file list reproduces on the base commit
  (`770d677`); this branch adds none. Two relay cases flake ~1 run in 3 on a
  teardown race — also reproduced on the base commit, so left alone rather than
  fixed as scope creep.
- **Backend** — 10319 passed, 3 skipped, 0 failed (full suite, 9m52s). The
  sandbox gaps the handoff warned about did not reproduce in this environment.
- **Coverage gate** — 16 passed, checked for `ERROR` as well as `FAILED`, both in
  isolation and in the full run. The frozen baseline needed no edit; verified
  against the generated document that the connection endpoint's path is
  unchanged, rather than assumed.
- **Pinned artifact** — regenerated via `dump_openapi.py` +
  `gate_and_publish_openapi.py`; the compat gate reported *compatible* and the
  diff is exactly the two example URLs. `tests/fixtures/bots.openapi.json` is
  deliberately untouched — it is hand-written, not a copy.
- **Acceptance criteria** — each one checked against the shipped
  `application.yaml`, including that `/openapi/v1/messages` (the BaaS domain)
  still resolves independently of `/openapi/v1/bots/messages`.

Not run: gateway `tests/e2e/*` (needs a live server) and the Singlebox coverage
gate (needs a full product stack).

## Compatibility and risk

Every domain that declares no `match` keeps its address and behaviour exactly.
The public break is the socket address itself, taken deliberately and without an
alias: that surface has no reachable external caller, since its route-security
rule required a Google-resolved user identity a tenant presenting an access key
cannot satisfy. Same reasoning as #706.

**Deploy ordering matters.** The config move is atomic — the old prefix stops and
the new one starts — so between the gateway deploy and the backend deploy the
backend publishes an address the gateway no longer serves. Established sockets
survive; *new* connection requests fail in that window. Single-box deploys both
together.

Rollback is reverting the config block and the backend prefix constant; the
routing mechanism is inert without a domain that declares `match`.

## Spec

`src/gateway/specs/2026-08-03-gateway-path-specific-domain-routing/` — spec,
plan, and tasks, developed through the SDD flow.

## Related issues

Independent of #706 (`/openapi/v1/bots` path normalization).

Informational, for whoever owns
`src/bcs/docs/plans/2026-08-03-bcn-collaboration-prefix-design.md`:
longest-prefix routing has moved from "out of scope" to "in scope". BCN does not
need the mechanism — it owns `/openapi/v1/collaboration/**` outright — and the
ownership framing here leaves the one-prefix-per-owner principle intact, but that
constraint was written down the same week and you should hear it from us.

### DIFF
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -575,6 +575,29 @@ approvals  ceiling  check-name  connection  engine  identity
 mcp  models  resources  routines  sessions  skills
 ```
 
+**Reserved ahead of their routes.** A second, separate list — names claimed here
+before any route publishes them. They are *not* reserved for the reason above:
+no route serves them, so nothing is currently unreachable at those addresses.
+They are reserved because something else already occupies the address and a
+component is intended there, so a bot id must not be allowed to take it in the
+meantime.
+
+<!-- reserved-component-names-unrouted -->
+```text
+messages
+```
+
+- `messages` — the gateway serves the bot's chat WebSocket at
+  `/openapi/v1/bots/messages/**`, relayed to the engine proxy
+  (`src/gateway/configs/application.yaml`). That claim is on the **socket plane
+  only**, so an HTTP request to the address still reaches this service; the name
+  is held for the HTTP endpoint intended there. See
+  `src/gateway/specs/2026-08-03-gateway-path-specific-domain-routing/`.
+
+A name in this list must move to the routed list above the moment a route
+publishes it — the convention test asserts the two lists stay disjoint, so
+adding the route without moving the name fails there rather than in review.
+
 > **Mount order is load-bearing.** `build_public_router()` includes the literal
 > sub-groups **before** the bots group, so `/openapi/v1/bots/resources` resolves
 > ahead of the `/openapi/v1/bots/{bot_id}` wildcard. Only the components that
--- a/src/backend/docs/openapi-v1/README.zh-CN.md
+++ b/src/backend/docs/openapi-v1/README.zh-CN.md
@@ -503,10 +503,30 @@ base 之下再容纳第二个 owner 的可能 —— BCS 正是从另一侧撞�
 id 恰好等于某个组件名，它在该地址上就不可达。这个集合是固定的，同一个测试会断言下面这份
 清单仍然等于路由实际发布的字面量（英文版 `README.md` 中的同名清单是被解析的那一份）：
 
+<!-- reserved-component-names -->
 ```text
 approvals  ceiling  check-name  connection  engine  identity
 mcp  models  resources  routines  sessions  skills
 ```
+
+**先于路由保留的名字。** 另有一份独立清单 —— 在任何路由发布它们之前就已在此占位的名字。
+它们的保留理由与上面那份**不同**：没有任何路由提供它们，因此当前也不存在"某个地址不可达"
+的问题。保留是因为该地址已被别处占用，且我们确实打算在那里放一个组件，所以在此期间不能让
+某个 Agent id 把它占走。
+
+<!-- reserved-component-names-unrouted -->
+```text
+messages
+```
+
+- `messages` —— 网关在 `/openapi/v1/bots/messages/**` 上提供 Agent 的聊天 WebSocket，
+  并中继到 engine proxy（`src/gateway/configs/application.yaml`）。该占用**只在 socket
+  平面**上成立，因此发往该地址的 HTTP 请求仍会到达本服务；这个名字是为将来要放在那里的
+  HTTP 端点保留的。参见
+  `src/gateway/specs/2026-08-03-gateway-path-specific-domain-routing/`。
+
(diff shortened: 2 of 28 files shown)


liveandevil (approved):
LGTM

FreddieSun (approved):
LGTM

PR #717, opened 2026-08-03
 fix(harness): drop TODO-placeholder patches; derive MCP call specs from tool inputSchema

wienyang:
Advisory findings (thin bot profile → SOUL persona gap; unverified MCP missing a call spec) produced meaningless  placeholder patches that replaced readable content, fixed nothing, and re-emitted every scan — leaving
  the health score permanently stuck low.

  Changes
  - patch_planner.py: removed the prompt rules that forced TODO insertion / forbade unchanged output. Patch LLM now applies only concrete fixes the diagnostic provides; no-content findings return unchanged and are skipped via the
  existing fixed==src guard.
  - mcp_format.py: keep each tool's inputSchema; split into 3 buckets (verified / schema-derivable / unreachable) and conditionally allow transcribing a real call spec from inputSchema; when a run has no recoverable content, mark
  the finding advisory (clear template, bump score→pass/info, append a "平台补全中" note) so the score isn't stuck.
  - soul.py: B-mode findings get an owner-facing recovery hint (on top of 2b2ffbea's source-gating).
  - Tests: TOOLS diagnostic cases, patch-prompt regression, advisory skip; diagnose LLM-disabled acceptance smoke.
  
  Verification: 320+ harness tests pass, ruff clean. DB templates #2/#3 op.template are generic markers with no TODO instructions, so the source fix is complete. Frontend (ocb) renders advisory findings correctly — backend-only, no
  UI change.


FreddieSun (approved):
lgtm

PR #719, opened 2026-08-03
fix(gateway): handle a client that leaves before the relay accepts

totalfrank:
## Problem

The WebSocket relay dials the upstream **before** accepting the client, so a
caller that gives up while that dial is in flight leaves `websocket.accept()`
reading a `websocket.disconnect` where it expects a `websocket.connect`.
Starlette reports that mismatch as a `RuntimeError`, which escaped the endpoint
and produced an unhandled ASGI traceback for every abandoned handshake.

Nothing leaked — the upstream exit was already owed unconditionally by the
existing `try/finally`, and that was verified rather than assumed. The cost is
noise: under load the tracebacks bury the single
`logger.warning("relay upstream unavailable")` that explains a *real* upstream
failure.

That is not hypothetical. A recent WebSocket relay capacity investigation spent
several rounds chasing a phantom ~430/500 connection ceiling, and this traceback
spam is part of what obscured the diagnosis. The root cause turned out to be a
capacity harness that attached an unconsumed `subprocess.PIPE` to the gateway's
stdout; once the pipe buffer filled, the gateway blocked in a synchronous write
inside the event loop and stopped accepting. The relay itself sustains 4000/4000
connections at 2 fd per connection with no fd leak. No capacity fix is needed or
included here.

Two smaller gaps surfaced alongside it:

- `_refuse` closed the client socket unguarded. Every refusal races a caller
  that may already have hung up, so a close-time error could replace the logged
  cause of the refusal.
- The dial guard catches `Exception`, which does not cover `CancelledError` on
  Python 3.12. Reviewed and deliberately left as-is, but the reasoning was
  nowhere in the code.

## Solution

Handle the abandoned handshake where it happens: catch `RuntimeError` /
`WebSocketDisconnect` scoped tightly to the `accept()` call, log at debug, and
return. The `finally` still runs, so the upstream is released on this path
exactly as before.

`_refuse` now closes through the existing `_quietly` helper, on the principle
that the reason for a refusal is worth more than the failure to transmit it.

The dial guard is unchanged. A cancellation there has nothing to unwind — the
context manager never entered, and the socket library releases its own transport
when its dial is cancelled — so widening it to `BaseException` would only turn a
cancelled worker into a refusal nobody is left to read. That reasoning is now a
comment rather than an open question a future reader has to re-derive.

Rejected: restructuring the handshake into a different async-context shape. The
existing `try/finally` already discharges the upstream on every path, which was
confirmed by driving Starlette's real `WebSocket` through the ordering; a
rewrite would carry risk without fixing anything.

## Validation

- New regression test `test_a_client_that_leaves_before_accept_still_releases_the_upstream`,
  driven at the ASGI seam because the condition under test is a message ordering
  a cooperating `TestClient` will not produce. Verified it **fails without the
  fix** with exactly `RuntimeError: Expected ASGI message "websocket.connect",
  but got 'websocket.disconnect'`, and passes with it.
- `tests/integration/test_relay_ws_route.py` — 40 passed.
- Full non-e2e suite — 587 passed, 1 failed.
- `ruff check` and `ruff format` clean on both changed files.

Not caused by this change, stated rather than hidden:

- The one non-e2e failure is `tests/architecture/test_ruff_lint_rules.py::test_ruff_formatting_passes`,
  from formatting drift in `src/gateway/community/spi/bot/_ports.py` — a file
  this PR does not touch. Left alone as out of scope.
- `tests/e2e` — 10 failed, 16 passed, all `httpx.ConnectError` from needing a
  live server. Identical counts on a clean tree.

Scale verification of this path is deliberately **not** claimed here. Generating
abandoned handshakes in volume needs the capacity harness, and that run is
tracked separately now that this fix and its unit coverage have landed.

## Compatibility and risk

No contract, schema, config, or migration impact. No client-visible behaviour
change on any path that previously worked: the handled case is one that
previously ended in an unhandled exception, and the endpoint's return value was
already discarded. Rollback is a straight revert of the single commit.

## Spec

`src/gateway/specs/2026-08-02-gateway-websocket-relay-domain/spec.md`

That spec defers concurrency limits under "measure first, then bound". The
measurement now exists and shows no wall below 4000 connections on a high-nofile
host, which supports keeping the limit deferred. No backpressure is added here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01F17T4d7t1PpeBeed737aCw)_

### DIFF
--- a/src/gateway/src/gateway/community/adapters/web/_relay_ws.py
+++ b/src/gateway/src/gateway/community/adapters/web/_relay_ws.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 import asyncio
 from typing import Any
 
-from starlette.websockets import WebSocket
+from starlette.websockets import WebSocket, WebSocketDisconnect
 
 from gateway.community.logger import get_logger
 from gateway.community.spi.auth import AuthError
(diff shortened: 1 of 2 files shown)


liveandevil (approved):
LGTM

PR #742, opened 2026-08-04
fix(backend): bind bot chat queries to authenticated user

evilcoda:
## Problem

`GET /api/v1/bot-chats` accepted an optional `owner_id` query parameter and preferred it over the authenticated user's `staffId`. This made the conversation-log ownership scope caller-controlled and could allow an authenticated request to read another user's non-public bot-chat logs.

## Solution

- Keep the existing `owner_id` query parameter for backward compatibility, but no longer use it as an identity source.
- Always derive the owner scope from the authenticated `user.staffId`.
- Replace the previous forwarding test with a security regression test proving that a caller-provided `owner_id` cannot override the authenticated user.

The change is intentionally limited to the HTTP boundary; no frontend, service, or repository refactor is included in this fix.

## Validation

- `159 passed`: `tests/community/api/bot_chat` and `tests/community/core/bot_chat`
- `11 passed`: focused bot-chat router suite with coverage enabled
- Changed executable line coverage: `1/1 = 100%`
- Ruff check passed for both changed files
- Local Python SAST block scan passed for the changed router
- `git diff --check` passed

## Compatibility and risk

Existing clients may continue sending `owner_id`; the parameter remains accepted but is ignored for authorization and data scoping. Bot access checks now consistently receive the actual authenticated user identity. The rollback is a one-line change, but would reintroduce the unauthorized cross-user read risk.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/bot_chat/router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/bot_chat/router.py
@@ -58,7 +58,8 @@ async def list_sessions(
 ):
     """List bot conversation sessions."""
     try:
-        effective_owner_id = owner_id or user.staffId
+        # COSEC: Bind the owner scope to the authenticated identity; never trust query owner_id.
+        effective_owner_id = user.staffId
         result = await service.list_sessions(
             owner_id=effective_owner_id,
             from_date=from_date,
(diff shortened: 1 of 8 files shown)


evilcoda:
Closing this PR for now because the responsible owner will handle the fix. We will not intervene further at this stage.

Issue #644, opened 2026-07-31
Deprecate ac_user_info and the container-quota compete flow


totalfrank:
## Summary

`ac_user_info` and the "抢名额" compete flow it backs are dormant — last insert was ~3 months ago. Rather than give them a tenant axis for multi-tenancy, remove them.

Removing the flow also removes most of the cross-tenant quota problem it would otherwise create, which is why this is filed as a deprecation rather than an isolation fix.

## What the flow is

`PolicyService.check` (`core/access/services/policy_service.py:26`) is a three-tier admission gate:

```python
# 1. 检查白名单
policy_record = self._repo.get_by_entity(entity_id=entity_id, entity_type="staff")
if policy_record:
    if self._is_policy_on(policy_record.policy):
        return True
    if self._is_policy_off(policy_record.policy):
        return False   # 黑名单，不进入竞争

# 2. 检查竞争用户表
user_record = self._repo.get_user_info(user_id=entity_id, user_type=USER_TYPE_COMPETE)
if user_record and user_record.status == USER_STATUS_ACCESS:
    return True

# 3. 进入竞争逻辑
return self._try_compete(entity_id=entity_id)
```

Tiers 2 and 3 are the deprecation target. `_try_compete` rations admission against a daily container budget — `min(daily_container_quota, total_container_limit − active devices)`, resetting at `daily_container_update_time` — and records winners in `ac_user_info`.

Tier 1 (the `ac_access_control_policy` whitelist) stays; its tenant axis is tracked separately.

## Scope

**Tables**
- `ac_user_info` (`core/access/sqlite_models.py:58`), unique on `(user_id, user_type)`

**Repository** (`core/access/repository.py` + `plugins/policy_repository.py`)
- `get_user_info`, `list_users`, `upsert_user_info`, `count_compete_users_after_time`
- `count_active_devices` — **see the caveat below, this one is not purely compete-scoped**

**Service** (`core/access/services/policy_service.py`)
- `_try_compete`, `_get_effective_quota`, tiers 2–3 of `check`
- `USER_TYPE_COMPETE` / `USER_STATUS_ACCESS` constants

**HTTP** (`adapters/http/access/router.py`)
- `user_router` — `GET /api/v1/user`, `GET /api/v1/user/{user_type}/{user_id}`, `POST /api/v1/user` (all operator-gated, all `UserService` over `ac_user_info`)
- `GET /api/v1/access/quota` — reports the compete budget

[...]

PR #746, opened 2026-08-04
docs(access): spec retiring the container-quota compete flow

totalfrank:
## Problem

Admission to the platform is decided by a three-tier gate: an operator-managed
whitelist, then a list of users who previously won a container slot, then a live
competition for one of a limited number of daily slots. Tiers two and three are
dormant — the slot-winner record has not been written to in roughly three months
— and they are the only reason a per-user admission table exists.

That table now blocks work. The multi-tenancy effort (raised while scoping #556
/ #625) needs every table carrying per-user state to grow a tenant axis or to
justify not having one, and giving this one an axis means designing tenant-aware
slot accounting — how a daily budget divides across tenants, whether one tenant
can exhaust another's capacity — for a feature no tenant uses.

Two findings from scoping sharpen the case, neither of them in the issue:

- **The competition writes on the read path.** Every admission check by a
  non-whitelisted user is a potential insert. A slot record untouched for three
  months therefore implies tier three is *unreachable*, not merely quiet.
- **The slot-budget endpoint is unauthenticated.** `GET /api/v1/access/quota`
  takes no identity and declares no auth dependency, and the request-level auth
  middleware deliberately does not block unauthenticated requests — it leaves
  enforcement to each route, and this route enforces nothing. It is the only
  route in the affected set that is not operator-gated. It currently publishes
  daily budget, total capacity, and current occupancy — the occupancy figure
  pooled across all tenants — to any caller that can reach the service.

Resolves the scoping half of #644.

## Solution

This PR is **SDD phase 1 only** — one new `spec.md`, no code changes. The spec
proposes making admission whitelist-only and retiring the two dormant tiers, the
per-user admission record behind them, the operator screens that manage it, and
the configuration values with no other consumer.

Filed as a spec rather than a direct change for two reasons.

**It is not the dead-code removal it looks like.** There is a real behavior
change: a user on no whitelist entry currently gets a shot at a daily slot and
would afterwards be refused outright. That needs product sign-off, not an
implementer's judgement call.

**The blast radius runs well past the issue's scope list.** Four public HTTP
endpoints, a whole service and its protocol, and — most of the work — test
infrastructure that is built on the surface being removed: the shared seed
helper referenced across roughly fifty test files, the database-plugin
conformance test that uses the doomed service as its canonical consumer, and the
test framework's own documented worked example. Replacing that fixture is a
design decision that should be reviewed before it is applied fifty files wide,
which is exactly what the plan phase is for.

Six open questions are left unresolved deliberately; two of them gate any
implementation:

1. Is denying unknown users the intended end state?
2. Has the slot-winner record been confirmed dormant in *every* environment? It
   cannot be answered from the repository.
3. If any user is admitted only by a slot record today, migrate them to the
   whitelist or accept they lose access?
4. Does the slot-budget report survive? Recommendation in the spec: remove it.
   Keeping it means adding authentication *and* a tenant axis for a number that
   describes a retired feature.
5. Do the operator endpoints have consumers outside this repository?
6. Immediate removal, or a deprecation window first? Depends on 5.

## Validation

No code changed, so no behavioral gates apply. What was actually verified while
writing the spec, by reading the tree:

- **No caller of `GET /api/v1/access/quota` exists in this repository** — not in
  the frontend, not in the gateway. The only references are backend tests and an
  acceptance baseline. This does not rule out consumers outside the repository,
  which is why it is open question 5 rather than a settled fact.
- **The quota route declares no auth dependency**, and `UserContextMiddleware`
  documents that auth failure must not block the request. The endpoint is
  genuinely open today.
- **Blast radius enumerated** across services, HTTP routes, DI wiring, repository
  methods, ORM models, test fixtures, e2e flows, acceptance baselines, and
  module documentation.

Not run, and not applicable at this phase: module test suites, coverage gates,
and E2E. The per-environment dormancy check in open question 2 cannot be run
from here at all — it needs database access per environment.

## Compatibility and risk

Nothing ships in this PR. The change it describes carries one user-visible
behavior change (unknown users denied rather than offered a slot) and removes
four public HTTP endpoints; both are gated on the open questions above. No
schema is dropped by the proposed work — the application would stop reading and
writing the table, with physical removal sequenced separately as an operational
step.

## Spec

`src/backend/specs/2026-08-04-deprecate-container-quota-compete-flow/spec.md`

## Related issues

Relates to #644. Filed alongside the whitelist tenant-axis work; raised while
scoping #556 / #625.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UY2Sjuwo2NDyX7wcgsN58v)_

### DIFF
--- a/src/backend/specs/2026-08-04-deprecate-container-quota-compete-flow/spec.md
+++ b/src/backend/specs/2026-08-04-deprecate-container-quota-compete-flow/spec.md
@@ -1,0 +1,156 @@
+# Retire the Container-Quota "抢名额" Admission Flow
+
+## Summary
+
+Access to the platform is decided by a three-tier gate: an operator-managed
+whitelist, then a list of users who previously won a container slot, then a live
+competition for one of a limited number of daily slots. The second and third
+tiers are dormant — the record of slot winners has not been written to in about
+three months — and they are the only reason a per-user admission table exists.
+
+This retires tiers two and three. Admission becomes whitelist-only: an operator
+puts a person on the list, or that person is not admitted. The operator screens
+for managing slot winners, and the endpoint reporting the daily slot budget, go
+with them.
+
+## Motivation
+
+**The flow is dormant, and dormant admission logic is worse than none.** Tier
+three writes a record every time it admits somebody, and it runs on the *read*
+path — every access check by a non-whitelisted user is a potential write. That
+the record has not been written to in three months is therefore strong evidence
+that no user reaches tier three at all: not that the feature is quiet, but that
+it is unreachable. Logic nobody exercises is logic nobody is checking, and it
+currently decides who gets into the product.
+
+**It is the last thing forcing a per-user admission table to exist.** Multi-tenancy
+work now underway (raised while scoping #556 / #625) needs every table carrying
+per-user state to grow a tenant axis, or to justify not having one. Giving this
+table a tenant axis means designing tenant-aware slot accounting — how a daily
+budget divides across tenants, whether one tenant can exhaust another's capacity
+— for a feature no tenant uses. Removing the flow removes that entire design
+problem instead of solving it.
+
+**The slot-budget report is worse than unused — it is unauthenticated.** The
+endpoint reporting the daily budget, total capacity, and current occupancy takes
+no identity and is not gated by any authentication dependency; the request-level
+auth middleware deliberately does not block unauthenticated requests, leaving
+enforcement to each route, and this route enforces nothing. It is the only route
+in the affected set that is not operator-gated. No caller for it exists anywhere
+in this repository — not in the frontend, not in the gateway. So today it
+publishes live capacity figures, pooled across all tenants, to anyone who asks,
+in service of a feature nobody uses.
+
+**The behavior change is small but real, and needs a decision.** Once tiers two
+and three are gone, a user who is not on the whitelist is denied outright. Today
+such a user gets a chance at a daily slot. If any user is currently being
+admitted that way, this takes their access away. The evidence says none are —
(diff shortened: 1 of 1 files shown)


totalfrank (commented):
Reviewed commit `ccd4b57` (the only commit on this branch) — the sole change is the new `src/backend/specs/2026-08-04-deprecate-container-quota-compete-flow/spec.md`. No code, no tests, no config in this diff.

Checked:
- **Factual accuracy of the spec's claims against the current codebase**, since a spec with wrong premises is itself a correctness bug: verified `GET /api/v1/access/quota` genuinely declares no auth dependency (unlike its sibling routes in the same router) and has no caller anywhere in the repo outside tests; confirmed `UserContextMiddleware` really does let auth failures through rather than blocking, so route-level enforcement is the only gate; confirmed the whitelist (`ac_access_control_policy`) and slot-winner records (`ac_user_info`) are genuinely separate tables, so retiring tiers two/three doesn't collide with whitelist storage; confirmed bot-limit settings share a row with the whitelist flag, supporting the acceptance criterion that whitelisting must not reset it; and the "~50 test files" / shared seed helper / DB-plugin conformance test claims all check out. No inaccuracies found.
- **Document structure**, against sibling specs under `src/backend/specs/` (Summary/Motivation/User Stories/Acceptance Criteria/In Scope/Out of Scope) — consistent.
- **PR description**, against `AGENTS.md` → Pull Request Conventions — uses the required Problem/Solution/Validation/Compatibility and risk/Spec/Related issues sections correctly, and the title matches `<type>(<scope>): <concise outcome>`.
- **Contract/architecture/security/concurrency/type-contract categories**: not applicable — no code is touched by this diff, so there is nothing to break.

No findings.

---

PR #747, opened 2026-08-04
Fix BCS Gateway Principal signing key loading from secrets

vzvince:
## Summary

- Add `gateway_principal.signing_key_secret` so the Gateway Principal verifier can load its signing key through the configured `SecretAccessPort`.
- Preserve backward compatibility by falling back to `signing_key_env` when no secret name is configured.
- Add coverage for Mist-style secret-backed key loading without exposing secret values in error output.

## Context

Internal BCS deployments using Mist currently fail during bootstrap with:

```text
InvalidConfig("Gateway Principal signing key is required")
```

The expected signing key is available via the existing secret access path, for example `other_manual_teamclawgw_principal_signing_key`, but the Gateway Principal verifier only read process environment key material before this change.

## Testing

- Not run to completion locally. The focused cargo test was started, but the workspace cargo shim performs a clean rebuild first and compilation was interrupted before test execution completed.


carolynli (approved):
LGTM.

PR #750, opened 2026-08-04
baas(bot_run): defer session creation for engines with deterministic session_id rules

phoenixliu:
## Summary

Defer session creation for bot-run engines that expose a deterministic
`session_id` rule (openclaw, claude_code). The runner now constructs the
`session_id` synchronously and the executor performs the actual
`create_session` call asynchronously, gated by a dispatcher
`supports_session_defer` capability flag.

## Background

Engines such as openclaw and claude_code derive their `session_id` from a
deterministic rule that depends only on the run/user identity. Today the
runner creates the session synchronously before enqueueing the task,
which forces an extra round-trip on the queuing path even though the
session identity is already known. For engines that require the session
to exist before the first message send (e.g. claude_code requires
explicit session creation before `chat.send`), the synchronous creation
also serialises against enqueue unfairly.

## Change

- Two-phase session materialisation:
  1. **Sync phase (runner):** the runner calls the engine adapter's
     `build_session_id` to construct the deterministic `session_id` and
     stamps the queued task with `meta['session_deferred'] = 'true'`.
  2. **Async phase (executor):** the executor calls
     `create_adapter_session` inside the per-session SerializingExecutor
     that already serialises work for that session, so concurrent
     `create_session` calls for the same session collapse safely.
- Capability gating: a new dispatcher flag `supports_session_defer`
  selects the deferred path. Dispatchers without the flag keep the
  legacy synchronous `create_session` on the runner side, preserving the
  existing contract.
- Fallback: adapters without a `build_session_id` rule (return `None`)
  fall back to the synchronous path, so non-deterministic-id engines
  keep their prior behaviour.

## Files

Engine adapter SPI / base / concrete adapters:
- `src/baas/src/secbaas/community/spi/bot/engine_adapter/_protocols.py`
- `src/baas/src/secbaas/community/plugins/bot/engine_adapter/_base.py`
- `src/baas/src/secbaas/community/plugins/bot/engine_adapter/claude_code/real/_real_claude_code.py`
- `src/baas/src/secbaas/community/plugins/bot/engine_adapter/claude_code/stub/_stub_claude_code.py`
- `src/baas/src/secbaas/community/plugins/bot/engine_adapter/aicoding/stub/_stub_aicoding.py`
- `src/baas/src/secbaas/community/plugins/bot/engine_adapter/hermes/stub/_stub_hermes.py`

Core bot_run service:
- `src/baas/src/secbaas/community/core/service/bot_run/_baas_service.py`
- `src/baas/src/secbaas/community/core/service/bot_run/_claw_service.py`
- `src/baas/src/secbaas/community/core/service/bot_run/_runner.py`
- `src/baas/src/secbaas/community/core/service/bot_run/_executor.py`
- `src/baas/src/secbaas/community/core/service/bot_run/_internal_protocols.py`
- `src/baas/src/secbaas/community/core/service/bot_run/_task_message_dispatcher.py`
- `src/baas/src/secbaas/community/core/service/bot_run/_queue_task_message_dispatcher.py`
- `src/baas/src/secbaas/community/core/service/bot_run/_noop_message_dispatcher.py`

Tests:
- `src/baas/tests/unit/core/service/bot_run/test_runner.py`
- `src/baas/tests/unit/core/service/bot_run/test_session_defer.py` (new)

## Verification

```
pytest tests/unit/core/service/bot_run/ \
       tests/unit/plugins/bot/engine_adapter/ \
       tests/unit/core/service/bot_run/test_task_message_dispatcher.py -q
# (cwd: src/baas)
```

Result: 856 passed, 0 failed, 0 skipped.

## Risk / Notes

- Per-session materialisation race is mitigated by the
  SerializingExecutor's per-session locking; concurrent
  `create_session` calls for the same session collapse to a single
  materialisation.
- claude_code requires explicit session creation before `chat.send`,
  so the executor must materialise the session before issuing the first
  send, otherwise the send will fail.
- Session-consistency-key routing relies on the same deterministic
  session_id rule the adapter uses for `build_session_id`, so device
  affinity is preserved across sync and async phases.

## Review

Non-blocking review findings (P2/P3) recorded in the review summary;
none block this push. Two P2 items are deferred to follow-ups:
- `ClawBotService.build_session_id` fallback should mirror the
  `BaasBotService` openclaw guard and return `None` for engines without
  a deterministic-id rule.
- Add a BotRunner -> QueueTaskMessageDispatcher integration test
  asserting `session_deferred=='true'` reaches the executor end-to-end.

One P3 advisory: `MockHermesAdapter.build_session_id` (test stub) should
stay consistent with `create_adapter_session` for the same engine.

### DIFF
--- a/src/baas/src/secbaas/community/core/service/bot_run/_baas_service.py
+++ b/src/baas/src/secbaas/community/core/service/bot_run/_baas_service.py
@@ -130,6 +130,43 @@ class BaasBotService(BotService):
 
     # ── 公开方法 (BotService Protocol) ───────────────────────────────────────
 
+    def build_session_id(
+        self,
+        *,
+        engine_type: str,
+        bot_id: str,
+        user_id: str,
+        run_id: str,
+        session_id: str | None = None,
+        binding_info: BotBindingInfo | None = None,
+    ) -> str | None:
+        """Construct a deterministic session ID without calling the engine.
+
+        Resolution logic:
+        1. If ``session_id`` is provided by the caller, return it directly.
+        2. If a registered adapter exists for ``engine_type``, delegate to
+           ``adapter.build_session_id(...)``.
+        3. If ``engine_type == "openclaw"``, return the rule-based ID.
+        4. Otherwise return ``None`` (engine does not support deterministic IDs).
+        """
+        if session_id is not None:
+            return session_id
+
+        _adapter = self._adapter_for(engine_type)
+        if _adapter is not None:
+            tc_bot_id = binding_info.bot_id if binding_info else bot_id
+            return _adapter.build_session_id(
+                tc_bot_id=tc_bot_id,
+                user_id=user_id,
+                run_id=run_id,
+                session_id=session_id,
+            )
+
+        if engine_type == "openclaw":
+            return f"agent:main:session:{run_id}:user:{user_id}"
+
+        return None
+
     async def create_session(
         self,
         *,
--- a/src/baas/src/secbaas/community/core/service/bot_run/_claw_service.py
+++ b/src/baas/src/secbaas/community/core/service/bot_run/_claw_service.py
@@ -98,6 +98,37 @@ class ClawBotService(BotService):
 
     # ── 公开方法 (BotService Protocol) ───────────────────────────────────────
 
+    def build_session_id(
+        self,
+        *,
+        engine_type: str,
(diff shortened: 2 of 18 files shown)


phoenixliu:
Should also add 'build_session_id' implementation in real hermes engine adapter.

PR #758, opened 2026-08-04
feat(gateway): require non-empty creator/modifier on app and access-key registrationDev gateway creator modifier notnull

liveandevil:


totalfrank (approved):
LGTM.

PR #762, opened 2026-08-04
fix(engine): keep legacy BCS DM sessions visible

RaymondHX:
## Problem
BCS DingTalk DM sessions can be stored in OpenClaw with legacy/raw keys such as `agent:main:bcs_grp_dingtalk_dm_<id>:<suffix>`. The session list filter only recognized DM sessions when the key contained the namespaced `bcs:group:` marker, so these real DM sessions were treated as internal BCS sessions and hidden from `/api/sessions`.

## Solution
Update the OpenClaw session filter to recognize BCS DM group keys by the `bcs_grp_*_dm_*` / `bcs_grp_dm_*` pattern regardless of whether `bcs:group:` is present. Keep existing filtering for non-DM internal BCS and ordinary group sessions.

## Validation
- `PYTHONPATH=src uv run --with pytest --with pytest-asyncio pytest src/engine/community/plugins/openclaw/tests/test_session_port.py -q` -> `46 passed`
- `git diff --check -- src/engine/src/engine/community/plugins/openclaw/_session.py src/engine/src/engine/community/plugins/openclaw/tests/test_session_port.py` -> passed

Committed and pushed with `--no-verify`; explicit validation above was run manually.

### DIFF
--- a/src/engine/src/engine/community/plugins/openclaw/_session.py
+++ b/src/engine/src/engine/community/plugins/openclaw/_session.py
@@ -6,6 +6,7 @@ _normalize_model_id, and _get_provider_map.
 from __future__ import annotations
 
 import logging
+import re
 from typing import Any
 
 from engine.community.openclaw.client.gateway_client import OpenClawGatewayClient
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

FreddieSun (approved):
LGTM

PR #766, opened 2026-08-04
feat(backend): pin ARKA image across service bot lifecycle

ws121200:
## Problem

ARKA service Bot instances currently inherit the image selected by the runtime template. After the configured image changes, create, restart, publish, caller, scale, and rollback operations can resolve different images for the same service Bot version. In particular, a historical rollback could combine an old build artifact with the current image, making the deployment non-reproducible.

The rollout also needs an environment-level switch so image Pin can be enabled independently without rewriting existing Bots or historical publish records.

## Solution

- Reuse the existing Common Config capability as the environment-level switch and current ARKA image source.
- Resolve the current configured image for new ARKA service Bots and draft restarts, then persist the accepted snapshot to the Bot and current Draft publish metadata.
- Copy only the image Pin fields into publish metadata, preserving unrelated service configuration, bindings, and build artifacts.
- Resolve published lifecycle operations from the target publish record instead of the current Common Config:
  - Verify and Online release
  - Eval publish
  - Published restart and fallback recreate
  - Caller containers
  - Scale Up
  - Historical rollback
- Thread an optional `docker_image` through `BotBuildService` into the existing BaaS create, upgrade, and scale payload construction.
- Keep TeClaw on its existing artifact delivery path and ignore the ARKA image Pin field.
- Preserve historical behavior when a Bot or publish record has no valid image snapshot by passing no custom image.

## Validation

- `DEPLOY_PROFILE=test uv run pytest tests/community/core/service_bot/services/test_arka_image_pin.py tests/community/core/service_bot/services/test_publish_flow_service.py -q` — 174 passed
- `DEPLOY_PROFILE=test uv run pytest tests/community/core/service_bot/services/test_publish_flow_service.py -k 'execute_rollback' -v` — 6 passed
- `uv run ruff check src/agentclaw/community/core/service_bot/services/publish_flow/rollback_ops_mixin.py tests/community/core/service_bot/services/test_publish_flow_service.py` — passed
- `git diff --check` — passed

## Compatibility and risk

- No database schema migration is required; the implementation uses existing configuration and metadata storage.
- Existing Bots and historical publish records are not batch-updated or restarted.
- Old publish records without Pin metadata continue using the existing template/default image behavior and never fall back to the latest Common Config during published operations.
- Changing or disabling Common Config affects only subsequent Bot creation and draft restart; published restart, scale, Eval, Caller, and rollback remain bound to their target publish snapshot.
- A configured image pull or startup failure is surfaced by BaaS/ARKA rather than silently falling back to another image.

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/README.md
+++ b/src/backend/src/agentclaw/community/core/bot_management/README.md
@@ -27,12 +27,14 @@ consumes:
   - "PolicyService"
   - "TaskQueueService"
   - "HandlerRegistry"
+  - "CommonConfigService"
 internal_dependencies:
   - agentclaw.community.api.policy_service
   - agentclaw.community.core.base
   - agentclaw.community.core.bot_collaborator
   - agentclaw.community.core.config
   - agentclaw.community.core.config_compose
+  - agentclaw.community.core.common_config
   - agentclaw.community.core.cron.services.aicoding.cron_auto_setup
   - agentclaw.community.core.desktop_bot
   - agentclaw.community.core.mcp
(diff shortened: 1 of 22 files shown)


FreddieSun (approved):
lgtm

PR #771, opened 2026-08-04
fix(engine): upgrade legacy Desktop repo bridge safely

XIEAN99:
## Problem

Desktop Bots prepared by the first Skills Pool rollout can retain the historical forward bridge `Pool repo -> Legacy repo`. The canonical-repo image rejects that bridge during preparation, but the downloader then continues and publishes a second Pool corpus. The runtime stays Legacy, the readiness marker becomes stale, and probe/cutover are blocked.

## Solution

- recognize only the exact historical `Pool -> Legacy` bridge
- atomically move the existing corpus to canonical Pool storage and publish the reverse Legacy compatibility bridge
- restore the historical layout if reverse-bridge publication fails
- fail closed before bootstrap/background downloads when an existing repo delivery cannot be prepared safely
- preserve first download for genuinely fresh Desktop runtimes

## Validation

- Engine full suite: 2219 passed, 5 repository-defined skips
- Skills Pool focused suite: 193 passed
- bridge concurrency and failure-injection loop: 60/60 passed
- changed-line coverage: 96.24% (128/133)
- local Python SAST block scan and diff-check passed
- exact upgrade, retry, failure restoration, fresh Bot, OpenClaw, Claude Code, AICoding, and Hermes cases covered

Related: #455

### DIFF
--- a/src/engine/src/engine/community/core/skills/skills_repo_download.py
+++ b/src/engine/src/engine/community/core/skills/skills_repo_download.py
@@ -58,6 +58,7 @@ from engine.community.core.skills.layout_planner import (
 )
 from engine.community.plugin_api.workspace_root import workspace_root
 from engine.community.plugins.skills_pool.desktop_preparation import (
+    DesktopPreparationStatus,
     prepare_desktop_pool,
 )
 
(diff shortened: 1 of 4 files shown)


XIEAN99:
Closing this draft: the legacy Pool-to-Legacy forward bridge only exists in the obsolete local Desktop test state created during earlier #455 iterations. Production has not enabled this migration, so product code should not carry compatibility logic for that test-only state. Acceptance will use fresh Bot homes; old local test homes can be cleaned operationally.

PR #778, opened 2026-08-04
fix(engine): restore canonical repo bridge after Desktop rollback

XIEAN99:
## Problem

Desktop DOWNLOAD now stores the canonical Skills repo as a real `pool_repo` directory. During rollback, AICoding and Hermes still attempted to resolve that directory with `readlink()`, which returns `EINVAL`. The local corpus had already been restored, so the operation appeared committed while leaving the stable local bridge and Pool active marker behind; a later migration then failed its runtime probe.

## Solution

Restore the Legacy repo view from the Planner topology: layouts whose `repo_bridge` is the Legacy repo entry point bridge directly to canonical `pool_repo`; layouts with a separate active compatibility bridge continue to target `legacy_repo`. This removes the stale assumption that canonical `pool_repo` is itself a symlink and adds a four-engine Desktop lifecycle regression covering prepare, cutover, rollback, Legacy mapping publication, and readiness for remigration.

## Validation

- New four-engine Desktop DOWNLOAD rollback/remigration regression: 4 passed.
- Skills Pool, OpenClaw, Claude Code, AICoding, Hermes, Desktop preparation, activation, probe, and downloader suites: 264 passed.
- Focused Hermes/AICoding rerun after formatting: 29 passed.
- Ruff check and format check passed for changed files.
- `git diff --check` passed.
- Local Python SAST block scan passed for both changed files.

## Compatibility and risk

This is Engine-owned filesystem cleanup only. Backend state, rollout policy, mapping contracts, and layout descriptors are unchanged. Cloud MOUNT behavior is unchanged; Desktop DOWNLOAD behavior is covered for all four filesystem engines. Teclaw remains out of the filesystem layout protocol.

## Spec

Implements the canonical Pool repo and Engine-owned compatibility bridge rules in `docs/arch/skills-management-publishing-current-state-and-upgrade-plan.md`.

## Related issues

Related to #455.

### DIFF
--- a/src/engine/src/engine/community/plugins/skills_pool/layout_activation.py
+++ b/src/engine/src/engine/community/plugins/skills_pool/layout_activation.py
@@ -396,18 +396,15 @@ def _publish_structural_bridge(path: Path, target: Path) -> None:
 
 def _restore_legacy_repo_bridge(
     *,
-    engine: str,
     layout: _Layout,
 ) -> None:
     """Restore the descriptor's Legacy repo view after rollback."""
 
     repo_delivery = current_repo_delivery()
-    if repo_delivery is RepoDelivery.MOUNT or engine == "openclaw":
+    if repo_delivery is RepoDelivery.MOUNT or layout.repo_bridge == layout.legacy_repo:
         target = layout.pool_repo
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #781, opened 2026-08-04
fix(engine): reject active repo bridges after cutover

XIEAN99:
## Problem

After a Desktop DOWNLOAD cutover, an active-root repo bridge is retired so the runtime cannot recursively expose the complete Skill corpus. The runtime probe retained an obsolete OpenClaw DOWNLOAD exception and could report READY if that bridge was recreated.

## Solution

- Derive repo-bridge retirement from the Planner layout topology: any repo bridge located inside the active root must remain absent after activation.
- Remove the engine/delivery special case.
- Add a public-seam regression through prepare, activate, bridge recreation, and inspect_runtime_layout for OpenClaw and Claude Code.

## Validation

- Regression demonstrated RED for OpenClaw and existing correct behavior for Claude Code, then GREEN for both.
- 171 focused Engine tests passed across Desktop preparation/activation/rollback, probe evidence, and OpenClaw/Claude Code/AICoding/Hermes Pool adapters.
- Changed-line coverage: 100% (18/18; gate 90%).
- Ruff, local Python SAST, and diff-check passed.

## Compatibility and risk

- Only active Pool layouts whose Planner repo bridge is inside the active scan root are tightened.
- Claude Code keeps its prior behavior; AICoding and Hermes stable repo bridges remain valid because they are outside the active root.
- Legacy, preparation, cutover publication, and rollback behavior are unchanged.

Follow-up hardening for #455.

### DIFF
--- a/src/engine/src/engine/community/plugins/skills_pool/layout_probe.py
+++ b/src/engine/src/engine/community/plugins/skills_pool/layout_probe.py
@@ -732,10 +732,7 @@ def inspect_runtime_layout(
             )
         if active_marker["activation_state"] == "active":
             retired_bridges = [("local", layout.local_bridge)]
-            if engine in {"openclaw", "claude_code"} and not (
-                engine == "openclaw"
-                and effective_repo_delivery is RepoDelivery.DOWNLOAD
-            ):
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #785, opened 2026-08-04
fix(gateway): remove the dev signing-key fallback and make key mismatches diagnosable

totalfrank:
## Problem

Every `/openapi/v1` request was answering 401, and the logs could not say why:

```
rejected forwarded principal on GET /openapi/v1/bots: principal token rejected: Signature verification failed
[Public 401] MissingPrincipalError on GET /openapi/v1/bots
INFO:     "GET /openapi/v1/bots HTTP/1.0" 401 Unauthorized
ERROR:    Exception in ASGI application            <- ~100 lines of traceback
```

Three separate defects sit behind that:

**1. The gateway signed with a committed dev key when its secret did not resolve.**
`bare/_plugin.py` fell back to `avernet-dev-signing-key-NOT-FOR-PROD` after one boot-time
warning. That fallback was worse than useless in both directions: a committed shared secret
is a committed credential, and no peer would ever accept those tokens — the backend ships no
fallback, so it never holds the constant this side signed with. All it bought was a gateway
that looked healthy while every request failed one hop away, attributed to the wrong
component. It also signed issued access-key and app tokens, so anyone with the source could
mint them.

**2. A key mismatch was indistinguishable from a forgery.** `Signature verification failed`
is all PyJWT can say. Nothing named which key the token was judged against or which signer
produced it, and the two halves of this contract never meet at request time — the gateway
signs successfully every time and only the backend observes a mismatch. There was no way to
compare the two ends short of printing the secret.

Two invisible ways for the keys to diverge fed into this: the backend stripped its resolved
value and the gateway did not (a trailing newline, routine when a secret comes from a file,
silently produced different keys), and the two sides read separately-provisioned secrets
under different names and prefixes.

**3. A routine 401 cost a ~100-line ASGI traceback.** `MissingPrincipalError` is raised in a
*dependency*, so `@envelope_errors` never sees it and it fell to
`@app.exception_handler(Exception)`. Starlette installs that as `ServerErrorMiddleware`'s
handler, which sends the response and then unconditionally re-raises ("We always continue to
raise the exception"). This directly defeated the intent already written at `app.py:517` —
log at warning *without* a traceback so real 5xx are not buried — and it was worst exactly
when auth was broken, because then every request took that path.

## Solution

**No fallback key, on either side.** Removed rather than made louder. A missing key now
refuses the boot in strict profiles and elsewhere leaves the signer unable to sign, so each
attempt raises `PrincipalSigningKeyMissingError` and the forwarder's existing handler answers
`500 principal signing failed` with a traceback naming the misconfigured component. This
mirrors the backend's `strict` rule — one contract, one rule. Refusing beats signing with an
empty key, which PyJWT accepts and which produces a signature anyone can reproduce.

Strict means `{pre, prepub, prod, gray}`, not the raw `{pre, prod}`: the backend does not
compare `SERVER_ENV` directly either, and `env_utils.get_current_env` folds `prepub` into
`pre` and `gray` into `prod` before gating. `scripts/app.sh` exports `SERVER_ENV=prepub` for
its supported `--env prepub`, so honouring only the canonical spellings would have split the
two sides apart in exactly those profiles.

**A key fingerprint at boot, on both sides** — truncated SHA-256, safe in a log, useless to a
reader:

```
backend:  gateway principal verification is configured (secret='...', key fp=eb128a7a, key len=38, aud='backend', iss='gateway')
gateway:  principal signer configured (secret='principal_signing_key', key fp=eb128a7a, key len=38, kid='bare', iss='gateway', ttl=60s)
```

Differing fingerprints mean different secrets; `unset` means no key at all; matching
fingerprints rule the key out and point at `iss`, clock skew, or a process predating the last
rotation. `key len` rides along because a stray newline changes both. The algorithm is
duplicated (separate distributions, no shared package) and pinned by identical golden values
in both suites, so drift fails a test instead of silently voiding the comparison.

That fingerprint is only safe for a strong key — against a guessable one, 32 bits is enough
to confirm a dictionary guess offline, and confirming the shared secret means forging any
identity. Nothing enforced the RFC 7518 §3.2 minimum the docstrings assumed, so
`is_weak_signing_key` now gates a boot warning on both sides. It warns rather than refuses,
and still prints the fingerprint: withholding it would remove the diagnostic exactly when a
deployment is most misconfigured, and rejecting short keys outright would deny every request
on a deployment that was at least partly working — a bigger change than this PR should make
unilaterally.

**Failure lines that name the cause.** Verification failures keep PyJWT's own reason and add
the token's unverified JOSE `alg`/`kid` plus the fingerprint of the key used:

```
principal token rejected: Signature verification failed [token alg='HS256' kid='bare';
verifier key fp=eb128a7a, expects aud='backend' iss='gateway']
```

`kid='bare'` says it was our gateway with the wrong key; anything else says it was not our
gateway. Only the JOSE header is read, never the payload — those claims are the unverified
assertions the function exists to distrust. The header is attacker-controlled, so it is
clipped and `repr`-escaped: a `kid` containing a newline cannot forge a log line. A *missing*
header now logs distinctly, because that is not an auth failure at all — it means the request
never came through the gateway, and chasing signing keys for it is chasing the wrong half of
the system. None of this reaches the caller; every failure still answers one fixed 401.

**Dedicated handlers for the two auth errors** put them in the inner `ExceptionMiddleware`,
which answers without re-raising. Status and body still come from `ENVELOPE_ERRORS`, so this
adds a route out of the middleware stack, not a second opinion on the answer.

**The gateway strips its resolved key**, matching the backend, and both warn when they trim —
naming the untrimmed *and* trimmed fingerprints, so a mixed-version rollout is a grep rather
than a second incident.

Rejected: gating the fallback behind a deploy profile (a committed credential is one
whichever branch reaches it), and logging the key length alone without a fingerprint (equal
lengths say nothing about equal bytes).

## Validation

Cut from and rebased onto `REL20260804`; both rebases replayed with no conflicts.

- **Gateway: 672 passed**, `--ignore=tests/e2e` (e2e needs a live server; it fails the same
  way on a clean tree). `ruff check` and `ruff format` clean on every file this branch
  touches.
- **Backend: 835 passed** across `tests/community/architecture` (the full architecture-gate
  suite — boundary, transport-in-core and env-access checks), plus `core/gateway_principal/`,
  `utils/test_gateway_principal_config.py`, and `api/`.
- **The re-raise is pinned from both sides.** `test_the_catch_all_alone_re_raises` drives the
  app through the raw ASGI interface — the only way to see what uvicorn sees, since
  `TestClient` either swallows the re-raise or converts it to a failure — and asserts the
  catch-all wiring still re-raises. If Starlette ever stops, that test fails and the extra
  registration can be dropped rather than cargo-culted.
- **Strict-profile aliases** are covered by parametrized cases for `pre`/`prepub`/`prod`/
  `gray` plus casing and padding, and for the non-strict profiles.
- **Fingerprint agreement verified across packages** by importing both implementations into
  one interpreter and comparing outputs, in addition to the golden values in each suite.
- **Log-injection checked**: a `kid` of `"ab\nWARNING forged"` renders as `'ab\\nWARNING
  forged'`, with no raw newline.
- **Not run: the remainder of the backend suite.** It exceeds the time budget available in
  this environment, so CI is authoritative for the untouched areas.
- **Pre-existing, untouched by this branch:** `tests/architecture/test_ruff_lint_rules.py::test_ruff_formatting_passes`
  fails on markdown under `src/gateway/docs/` and `src/gateway/specs/`, none of which appears
  in this diff.

## Compatibility and risk

**Behavioral change, deliberate:** a gateway with no resolvable signing key used to boot and
forward; it now refuses to boot in strict profiles and returns 500 elsewhere. This affects
only deployments that were relying on the dev fallback — and those were already
non-functional, since the backend never accepted the tokens. Singlebox is unchanged in
effect: its public surface already denied because that side resolves no key.

To provision, set the secret named by `user_config.principal_signer.secret_name` (community
resolver: `AVERNET_SECRET_PRINCIPAL_SIGNING_KEY_VALUE`) to the same value the backend reads
under `secret_names.gateway_principal_signing_key`, and confirm the two boot lines report the
same `key fp`.

**Credential issuance now requires a real key too.** `/admin/access-keys` and app
registration sign with the same signer, so they fail without one — correct, since tokens
signed under a published constant are forgeable by anyone with the source. Two integration
tests that decoded with the fallback now provision a key the way a deployment does.

**A short key logs a warning but does not block startup.** If a hard refusal in strict
profiles is preferred, that is a one-line change and worth deciding explicitly.

Rollback is a revert; nothing persisted changes, and no token format, claim, or wire contract
is altered.

## Spec

`src/gateway/docs/2026-07-21-auth-design.md` §7.1. Diagnostic procedure documented in
`src/backend/docs/openapi-v1/README.md` under "Diagnosing a 401 on this surface".

### DIFF
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -364,6 +364,52 @@ route dependency        → require_principal(request)       ─┘  cache on sc
   Either way the trigger is the same: no such secret, an empty value, or a
   resolver that raises.
 
+  **The gateway now behaves identically** — one contract, one rule. It shipped a
+  committed dev fallback key until 2026-08-04; that was removed rather than made
+  louder, because no peer ever accepted those tokens (this side has never had a
+  fallback to match) and all it bought was a gateway that looked healthy while
+  every request failed one hop away. Missing key there ⇒ boot refused in
+  `pre`/`prod`, otherwise every signature attempt refuses and the forwarder
+  answers `500 principal signing failed`.
+
+### Diagnosing a 401 on this surface
+
+Both halves log a **key fingerprint** at boot — a truncated SHA-256, safe in a
+log, useless to a reader. They are the same eight characters when and only when
+the two ends hold the same secret, so the first question is answered by diffing
+two lines rather than by printing a credential:
+
+```text
+backend:  gateway principal verification is configured (secret='...', key fp=eb128a7a, key len=38, aud='backend', iss='gateway')
+gateway:  principal signer configured (secret='principal_signing_key', key fp=eb128a7a, key len=38, kid='bare', iss='gateway', ttl=60s)
+```
+
+| What you see | What it means |
+| --- | --- |
+| fingerprints differ | the two ends hold different secrets — the usual cause |
+| gateway `key fp=unset` | the gateway resolved no key and cannot sign at all |
+| same fp, different `key len` | one side's value carries whitespace (both strip, so this only appears across a mixed-version rollout) |
+| fingerprints match | not a key problem — check `iss`, clock skew, and whether either process predates the last rotation (each resolves once at boot) |
+
+Per-request, the backend logs one line per failure, and it names the cause:
+
+```text
+rejected forwarded principal on GET /openapi/v1/bots: principal token rejected:
+Signature verification failed [token alg='HS256' kid='bare'; verifier key
+fp=eb128a7a, expects aud='backend' iss='gateway']
+```
+
+`alg`/`kid` come from the token's unverified JOSE header — the only readable
+part when a signature fails, and what distinguishes "our gateway, wrong key"
+(`kid='bare'`) from a token no gateway of ours minted. None of it reaches the
+caller: every failure answers the same fixed `401 Unauthorized`.
+
+A **missing** header logs distinctly (`no X-Avernet-Principal header on ...`)
+and is not an auth failure at all — the gateway injects that header on every
+forwarded request, so its absence means the request did not come through the
+gateway's authenticated path. Chasing signing keys for that one is chasing the
+wrong half of the system.
+
 - `aud` and `iss` are fixed in code here, not configurable — one wire contract,
   one spelling. They are **not** symmetric on the signing side, though:
   - `aud` is not configurable there either (the gateway signs it from the
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -395,14 +395,28 @@ Per-request, the backend logs one line per failure, and it names the cause:
 
 ```text
 rejected forwarded principal on GET /openapi/v1/bots: principal token rejected:
-Signature verification failed [token alg='HS256' kid='bare'; verifier key
-fp=eb128a7a, expects aud='backend' iss='gateway']
+Signature verification failed [verifier key fp=eb128a7a, expects aud='backend'
+iss='gateway'; unverified caller-supplied header alg='HS256' kid='bare']
 ```
 
-`alg`/`kid` come from the token's unverified JOSE header — the only readable
-part when a signature fails, and what distinguishes "our gateway, wrong key"
-(`kid='bare'`) from a token no gateway of ours minted. None of it reaches the
-caller: every failure answers the same fixed `401 Unauthorized`.
+**The two halves of that suffix do not carry equal weight, and the difference
+decides what you should do about it.**
+
+`verifier key fp`, `aud` and `iss` come from this process's own configuration.
+They are trustworthy, and together with the gateway's boot line they are what
+a diagnosis should rest on.
+
+`alg` and `kid` come from the token's JOSE header, which a failed signature
+means nothing has authenticated. Anyone who can reach the surface can stamp
+`kid: bare` on a token they minted. So `kid='bare'` is **not** evidence the
+gateway sent it — treating it as such during a burst of forged traffic would
+have you rotate a shared secret that was never broken. Read it as a hint that
+is useful mainly in the negative: an unexpected `alg`, an unfamiliar `kid`, or
+a header that will not parse says *look somewhere other than the key*.
+
+When the fingerprints match and forged-looking traffic persists, the token is
+not coming from your gateway, whatever its `kid` claims. None of this reaches
+the caller: every failure answers the same fixed `401 Unauthorized`.
 
 A **missing** header logs distinctly (`no X-Avernet-Principal header on ...`)
 and is not an auth failure at all — the gateway injects that header on every
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -324,6 +324,12 @@ from agentclaw.community.core.errors import (  # noqa: E402
     Unauthorized,
     ValidationError,
 )
+from agentclaw.community.adapters.http.openapi_v1.errors import (  # noqa: E402
+    MissingPrincipalError,
+)
+from agentclaw.community.core.gateway_principal import (  # noqa: E402
+    PrincipalVerificationError,
+)
 from agentclaw.community.core.caller_identity.contracts import (  # noqa: E402
     CallerCallTypeInvalidError,
     CallerIdentityAmbiguousError,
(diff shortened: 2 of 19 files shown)


totalfrank:
All four findings are outside this PR's diff, and none is being dismissed on the merits — the review ran against `73723f0`, when this PR was briefly based on `dev`. `REL20260804` and `dev` have diverged by 46/23 commits, so that diff contained a large amount of other people's work. The PR is now rebased onto `REL20260804` (the branch it was cut from) and retargeted there; the diff is one commit.

Against the correct base, this PR touches 18 files. Three of the four flagged files are not among them:

| Finding | File | In this diff? |
| --- | --- | --- |
| P1 chat route security | `gateway/configs/application.yaml` | file yes, **hunk no** |
| P1 WebSocket relay logging | `gateway/.../web/_relay_ws.py` | no |
| P1 binding-ID ownership | `backend/.../session_resources/service.py` | no |
| P2 BaaS JWT `aud`/`iss` | `baas/.../api_gateway/_jwt.py` | no |

On the first: my only change to `application.yaml` is a comment block at lines 52–64 documenting `principal_signer`. The `chat` domain is at line 223 and is unchanged here — it already exists on `REL20260804`.

**The P2 is real and I would like to see it fixed.** It is the same contract this PR works on, from the other side. `verify_jwt_token` passes `options={"verify_aud": False}` with no `issuer=` and no required-claims list, while the gateway mints a *per-upstream* `aud` from the target server's name and signs every upstream's token with one shared key. `aud` is therefore the only thing distinguishing a token minted for `backend` from one minted for `baas`, and disabling it is what makes cross-upstream replay possible. For comparison, the backend verifier this PR touches pins all of it:

```python
jwt.decode(
    token, config.signing_key, algorithms=["HS256"],
    audience=config.audience, issuer=config.issuer, leeway=5,
    options={"require": ["exp", "iat", "aud", "iss"]},
)
```

I have not folded that into this PR: it is a different component and a security change that deserves its own diff and its own tests, not a drive-by in a logging/diagnostics change. Worth filing separately — happy to take it.

One note for whoever re-runs the review: re-triggering on the current head will produce a clean scope. The findings above may or may not survive that, but they should be judged against `REL20260804`, not `dev`.

---

xianmuyq (dismissed):

totalfrank:
Thanks — this round hit the right base, and all three are legitimate. Fixed in `4c8b956`.

**P2, `prepub` not strict — correct, and worse than reported.** Verified the chain: `scripts/app.sh:190` exports `SERVER_ENV="$env_name"` for the supported `--env prepub`, and `env_utils.py::get_current_env` folds `prepub`→`pre` *and* `gray`→`prod` before the backend gates its verifier. My raw comparison missed both aliases, so in those profiles the backend refuses to boot while the gateway boots and answers 500 per request — the shared fail-fast contract splitting apart precisely where it matters. `_STRICT_ENVS` is now `{pre, prepub, prod, gray}`, and the value is lowercased/stripped before comparison. `gray` was the same bug and is fixed alongside. Six parametrized cases cover the aliases, casing and padding; five cover the non-strict profiles.

**P1, weak keys — fair, and it caught a claim of mine that was not true.** The docstring said the key "is required to be at least 32 random bytes," and nothing enforced that; the fingerprint's safety argument was resting on an unchecked precondition. Added `MIN_SIGNING_KEY_BYTES` and `is_weak_signing_key`, with a boot warning on both sides, and rewrote the docstrings so the safety claim is stated as conditional rather than assumed.

Two deliberate departures from the suggestion, both worth flagging rather than burying:

- **The fingerprint is still emitted for a weak key.** Withholding it removes the diagnostic exactly when a deployment is most misconfigured, which is the case this PR exists to serve. The warning names the real remedy instead, and says what is at stake: recovering the shared secret means forging any caller identity.
- **It warns rather than refuses.** Rejecting a short key at boot would deny every request on a deployment that was at least partly working — a larger behavioral change than a diagnostics PR should make unilaterally. Happy to escalate it to a hard refusal in strict profiles if you'd prefer; that reads to me like a separate call.

Length is a coarse proxy for entropy — it cannot tell 32 random bytes from 32 `a`s — but short keys are overwhelmingly the human-chosen ones, so it catches the case that matters. Measured in bytes rather than characters, since bytes are what HMAC consumes.

**P1, boundary metadata — correct.** I had reasoned that a property on an existing type needed no `provides` entry, then exported `key_fingerprint` as a module-level symbol in `__all__` anyway, leaving the two inconsistent. `key_fingerprint`, `is_weak_signing_key` and `MIN_SIGNING_KEY_BYTES` are now recorded in the context boundary.

Gateway suite: 672 passed, `ruff check` and `ruff format` clean on the files this branch touches. Backend: 75 passed across the two directly affected files, with the architecture gates re-running now. The one remaining red in `tests/architecture/test_ruff_lint_rules.py::test_ruff_formatting_passes` is pre-existing and untouched here — it fails on markdown under `docs/` and `specs/`, none of which appears in this diff.

---

totalfrank:
The `kid` finding is right, and it's the best catch of the three rounds — fixed in `99bba8e`.

I had this half-correct in a way that was worse than getting it plainly wrong. The code already treated the JOSE header as untrusted *for decisions* — `alg` stays pinned to `_ALGORITHMS`, nothing branches on `kid`, the value is clipped and escaped — and then the log line, the docstring and the runbook all turned around and read `kid='bare'` as provenance. A failed signature authenticates nothing about the token, header included, so that inference is one a forger gets for free by typing four characters. The failure mode is concrete: a burst of forged traffic renders as "our gateway, wrong key" and sends someone to rotate a shared secret that was never broken, during an incident that has nothing to do with the key.

What changed:

- `_token_provenance` → `_unverified_token_header`. The name was asserting the thing that isn't true.
- The message now leads with the trustworthy half and labels the other: `[verifier key fp=eb128a7a, expects aud='backend' iss='gateway'; unverified caller-supplied header alg='HS256' kid='bare']`. A test pins the ordering, so the line reads "here is what we hold" before "here is what they claimed".
- The runbook now says outright which half to reason from: the fingerprints come from each process's own configuration and cannot be influenced by a request; the header is a hint useful mainly in the negative — an unexpected `alg`, an unfamiliar `kid`, a header that won't parse all say *look somewhere other than the key*. And it states the case Codex raised: matching fingerprints plus forged-looking traffic means the tokens are not coming from your gateway, whatever the `kid` says.

I kept the fields rather than dropping them. Deriving provenance from connection metadata would be the real fix if provenance were required, but nothing here needs it — the diagnosis rests on the fingerprints, and the header costs nothing to carry once it stops pretending to be evidence.

Two tests added for the abuse case specifically: one asserts a token forged with an attacker key but `kid: bare` produces a message containing none of "our gateway" / "signed by" / "produced by" / "provenance", and one pins the unverified label and the field ordering.

Backend: 1288 passed (architecture gates, `core/gateway_principal/`, `utils/`, `api/`, `adapters/http/openapi_v1/`). Gateway: 672 passed, ruff clean.

---

FreddieSun (approved):
LGTM

liveandevil (approved):
LGTM

PR #792, opened 2026-08-05
fix(bcs): allow direct self chat

RaymondHX:
## Problem
`bcs-cli chat` routes through the A2A direct chat service. When a Bot chats with itself and its visibility is protected, the service currently applies the protected-target friendship check and rejects the request as `NotFriends`, even though self-chat is not cross-Bot collaboration.

## Solution
Allow direct self-chat after the target Bot is resolved, before visibility-based friendship checks run. This keeps the exception scoped to direct chat reachability and does not make the friendship model self-referential. Added a contract test proving protected self-chat works without a friendship while the existing protected non-friend rejection still holds. Also updated the HTTP-to-WS 1:1 chat e2e to use a real Sender Bot -> Receiver Bot flow, so it no longer relies on self-chat being rejected to finish quickly.

## Validation
- `cargo test --offline -p bcs-message-flow --test contract_a2a_chat protected_target`
- `cargo test -p bcs --test e2e_ws_messaging test_1to1_chat_http_to_ws -- --nocapture`
- `git diff --check`

Committed and pushed with `--no-verify` per Avernet publishing workflow; explicit validation above passed.

## Compatibility and risk
Low risk. The runtime change only affects direct chat where caller Bot ID equals target Bot ID. Protected/private cross-Bot reachability and friend request semantics are unchanged.

### DIFF
--- a/src/bcs/crates/bootstrap/bcs/tests/e2e_ws_messaging.rs
+++ b/src/bcs/crates/bootstrap/bcs/tests/e2e_ws_messaging.rs
@@ -582,25 +582,28 @@ async fn test_1to1_chat_http_to_ws() {
 
     tokio::time::sleep(Duration::from_millis(50)).await;
 
-    // Connect Bot B (receiver) via WebSocket
-    let mut ws_b = connect_bot(addr, None).await;
-    let resp_b = send_frame(&mut ws_b, json!({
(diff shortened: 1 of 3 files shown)


vzvince:
@codex  review

vzvince (approved):
LGTM

PR #794, opened 2026-08-05
feat(session-files): support large attachment reupload

yhzhang35:
## Problem

Bot proxypass limits make direct responses unsuitable for session-file attachments larger than 30 MiB. The Engine also needs a controlled way to refresh a download when the workspace file changed or the original Session File transfer expired, without introducing Bot File Transfer identity or leaking workspace paths.

## Solution

- Keep inline previews and attachments up to 30 MiB as direct Engine streams.
- For larger attachments, asynchronously create or reuse a Session File transfer, expose `202 preparing_download` while work is in progress, and return a two-hour share link when ready.
- Re-upload a manifest-controlled snapshot through SINGLE or MULTIPART grants when the local file changed or the source transfer is unavailable, then atomically promote the new transfer only after integrity revalidation.
- Keep BaaS control credentials off OSS requests, validate allowed share hosts, and use tenant plus session key without `bot_uuid` or caller-provided paths.
- Cover unavailable configuration, invalid configuration, BaaS status mapping, malformed responses, direct small-file streaming, single-flight export, source expiry, and changed-file re-upload.

## Validation

- `src/engine/scripts/ci_test.sh --base origin/dev --head HEAD`: 2234/2234 tests passed; case pass rate 100%; total line coverage 92.68%; changed-line coverage 90.24% (675/748).
- `scripts/ci/python_sast_local.sh src/engine 1 --base origin/dev --head HEAD`: passed.
- Ruff `F,E9` checks on changed Python files: passed.

## Compatibility and risk

Small attachments remain binary responses. Large attachment requests use the asynchronous JSON polling contract and depend on the Session File API and its temporary share-link lifecycle. Download URLs and transfer coordination remain in memory and are not persisted to Backend audit records.

### DIFF
--- a/src/engine/src/engine/community/api/resource_materialization/router.py
+++ b/src/engine/src/engine/community/api/resource_materialization/router.py
@@ -1,4 +1,5 @@
 """Thin HTTP adapter for Backend-triggered materialization."""
+
 from __future__ import annotations
 
 import asyncio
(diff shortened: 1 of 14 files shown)


totalfrank (approved):
LGTM.

PR #796, opened 2026-08-05
feat(bcs): group_context_delivery option for driver GROUP CONTEXT on session create

zh-bupt:
## Problem

Session creation (`POST /groups/{id}/sessions`) always delivers the initial `[GROUP CONTEXT]` message to the driver bot via `chat.send`, i.e. the driver is actively asked to respond. Callers have no way to create a session where the driver only observes the context silently (e.g. workbench/driven flows that will @mention or steer the driver later).

## Solution

Add an optional `group_context_delivery` field (`"send"` | `"inject"`) to the create-session request:

- Absent or `"send"`: unchanged behavior — driver receives `chat.send`.
- `"inject"`: driver receives the context via `chat.inject` (silent). Other participants always receive `chat.inject` regardless, as before.
- ManagerWorker groups intentionally ignore the override: the manager must actively pick up and dispatch the task, so its context is always `chat.send` (documented in the producer).

The choice is creation-time only, so it is threaded through a new optional `driver_delivery` field on `SystemMessageEvent::SessionContext` rather than persisted; the four other event construction sites (service invocation, proposal confirm, group management, channel) pass `None` and are unchanged. `DeliveryType` gains serde (`"send"`/`"inject"`) so it can be used in HTTP DTOs.

Rejected alternative: persisting the preference on the session (`meta` or a new column) — unnecessary because `SessionContext` is only emitted at session creation, never on reactivation.

## Validation

- `cargo check --workspace --all-targets` — clean.
- New producer unit tests in `session_context_test.rs`: driver defaults to `Send`, `driver_delivery: Some(Inject)` injects the driver in Chat groups, ManagerWorker ignores the override (manager stays `Send`).
- New DTO contract tests `session_create_request_dto.rs`: field defaults to `None`, accepts `send`/`inject`, rejects unknown values (400 via serde).
- Full test runs of `bcs-domain`, `bcs-system-message`, `bcs-http` (all test targets), `bcs-channel`, `bcs-proposal`, `bcs-group` — all green.
- Pre-existing `cargo clippy` findings in `bcs-domain` are identical on the clean tree (26 error lines before and after); no new lint introduced.

## Compatibility and risk

Fully backward compatible: omitting the field preserves today's behavior. No schema migration (no persistence). Unknown enum values are rejected at deserialization with 400, consistent with other enum request fields.

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/services.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/services.rs
@@ -182,6 +182,7 @@ pub async fn post_invocation(
                                 reason,
                                 session_input,
                                 task_ledger: None,
+                                driver_delivery: None,
                             },
                             &sid,
                             &session_participants,
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/sessions.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/sessions.rs
@@ -13,7 +13,7 @@ use axum::{
 use serde::Deserialize;
 use serde_json::Value;
 
-use bcs_domain::{ActorKind, SystemMessageEvent};
+use bcs_domain::{ActorKind, DeliveryType, SystemMessageEvent};
 use bcs_service_api::{
     AuthenticatedHumanCaller, GroupChatCommand, StartStateMachineRunCommand,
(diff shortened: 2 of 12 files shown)


vzvince:
@codex review

vzvince (approved):
LGTM

PR #797, opened 2026-08-05
fix(backend): publish direct skills in Pool mappings

XIEAN99:
## Summary
- merge a Pool-active direct Skill locator into the same full mapping publish as active SkillSets
- keep Legacy and non-Pool mapping calls unchanged
- reject conflicting runtime targets before device mutation and keep an existing Pool link until authoritative bindpath succeeds
- cover OpenClaw, Claude Code, and Hermes Pool roots plus direct batch activation

## Problem
A real pre Desktop OpenClaw acceptance for #455 uploaded local Skill \`1120709 / sp455-r2-oc-probe\`. The activate API returned success twice, but the runtime link was absent. Backend traces showed the requested Pool local mapping was computed by the direct activation path but \`get_symlink_mappings()\` published only the six active-SkillSet mappings, so BaaS/Engine never received the requested Skill.

## Scope / compatibility
- only callers whose request-scoped SkillService reports \`runtime_uses_pool_paths=true\` add direct locators
- rollout-disabled/unclaimed Desktop, Legacy, and non-Desktop callers keep the existing mapping publish contract
- no new state table, lock, Desktop state machine, or alternate probe contract

## Verification
- Backend focused and Skills Pool suite: 981 passed
- Engine bindpath / Claude adapter focused suite: 248 passed
- focused Pool direct-activation regressions: 7 passed
- Ruff, local Python SAST, and git diff check passed

Relates to #455.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
+++ b/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
@@ -1015,10 +1015,13 @@ async def activate_skill(
             engine_type=effective_engine,
             entity_type=effective_entity_type,
         )
-        symlinks = skill_set_service.get_symlink_mappings(
-            user_id=ctx.user_id,
-            bolt_id=effective_bot_id
-        )
+        mapping_kwargs: dict[str, Any] = {
+            "user_id": ctx.user_id,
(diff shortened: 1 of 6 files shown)


totalfrank (approved):
LGTM.

PR #799, opened 2026-08-05
fix(backend): distinguish default and legacy ARKA image policy

ws121200:
## Problem

PR #766 applied the Common Config image snapshot to new ARKA service Bots and draft restarts. The clarified rollout policy is different: new Bots and successful draft restarts must follow the BaaS/ARKA default image, while only historical Bots and publish records without an explicit policy should use Common Config Pin protection.

Without an explicit default marker, new records are indistinguishable from historical unresolved records and may be incorrectly pinned. Published operations also need one consistent resolution path so restart, release, Eval, Caller, Scale, and rollback do not resolve different images.

## Solution

- Add an explicit default-image policy for new ARKA service Bots and successful draft restarts.
- Persist the draft-restart policy to the Bot and current Draft publish metadata only; historical Verify/Online/SUCCESS rows are not rewritten.
- Preserve two explicit publish policies and one compatibility state:
  - Default: follow the provider default image.
  - Pinned: use the image snapshot from the target publish record.
  - Legacy/unresolved: consult Common Config and lazily snapshot a valid Pin image.
- Copy only image-policy-owned fields when creating or updating publish metadata, preserving unrelated metadata.
- Centralize target-publish image resolution for Release, Upgrade, Eval, published Restart, Caller containers, Scale, and Rollback.
- Keep TeClaw outside the ARKA Common Config and image Pin path.
- Fail explicitly when an enabled Common Config or publish Pin snapshot has no valid image.
- Pass the Scale Pin image through the existing nested deploy configuration.

## Validation

- Extended service Bot regression suite: 940 passed
- Ruff on all modified production files and related tests: passed
- Expert Chat test Ruff with only the file's pre-existing F841 ignored: passed
- PR diff whitespace check: passed

## Compatibility and risk

- No database schema migration is required; the policy uses existing Bot and publish metadata.
- Existing explicit Pin snapshots remain immutable and do not follow later Common Config changes.
- Legacy records with disabled or missing Common Config use the provider default for the current operation and remain unresolved so the rollout switch can be enabled later.
- Legacy records are updated only when an enabled valid Pin image is selected, using latest-metadata field-level merge semantics.

## Related issues

- Follow-up correction to #766

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
@@ -69,10 +69,9 @@ from agentclaw.community.core.workspace.path_factory import (
 from agentclaw.community.core.service_bot.repository.models import PublishStatus
 from agentclaw.community.core.service_bot.types import PublishStage
 from agentclaw.community.core.service_bot.services.arka_image_pin import (
-    apply_image_pin_to_ext,
-    copy_image_pin_to_ext,
+    apply_default_image_to_ext,
+    copy_image_policy_to_ext,
     overlay_image_pin_on_template_config,
-    resolve_current_arka_image,
 )
 from agentclaw.community.utils.avernet_tenant import (
     bind_current_avernet_tenant,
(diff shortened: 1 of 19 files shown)


ws121200:
Closing this dev-targeted PR; the change will be submitted against REL20260806 instead.

PR #804, opened 2026-08-05
fix(backend): release rolled-back rollout batches

XIEAN99:
## Problem

After an exact rollout Bot is fully rolled back to Legacy and removed from the whitelist, its retained `rollout_evidence` still makes the batch appear open. The batch cannot be accepted because it is no longer promotion-ready, while a new batch for the same engine is rejected with `the current engine batch must be accepted before expansion`.

## Solution

Count persisted rollout evidence as open-batch ownership only while the Bot remains claimed (`active_layout=pool` or `target_layout=pool`). A completed rollback (`active_layout=legacy`, `target_layout=None`) releases ownership without deleting historical evidence. Pool preparation and rollback-in-progress states continue to fail closed.

## Validation

- Red/green regression for a completed rollback followed by a new batch.
- Claimed Pool preparation and rollback-in-progress cases remain blocked.
- `278 passed` for the Skills Pool core plus rollout HTTP/endpoint suites.
- Ruff passed for both changed files.
- Backend local SAST passed.
- `git diff --check` passed.

## Compatibility and risk

No API, schema, configuration, Engine, Desktop, or runtime contract changes. Historical rollout evidence remains persisted for audit. The behavior changes only for states that have already completed rollback to Legacy.

### DIFF
--- a/src/backend/src/agentclaw/community/core/skills_pool/operations.py
+++ b/src/backend/src/agentclaw/community/core/skills_pool/operations.py
@@ -694,7 +694,8 @@ class SkillsPoolRolloutOperations:
         for state in self._layouts.list_states(env=env, engine=engine):
             evidence = state.rollout_evidence
             if (
-                evidence is not None
+                self._claimed(state.active_layout, state.target_layout)
+                and evidence is not None
                 and evidence.engine_type == engine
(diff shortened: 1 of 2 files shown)


FreddieSun (approved):
LGTM

PR #809, opened 2026-08-05
fix(gateway): drop tenant from the user principal

totalfrank:
## Problem

The public API answers `401` on every request:

```
rejected forwarded principal on GET /openapi/v1/bots: principal payload does not
match the expected contract: 1 validation error for tagged-union[...]
user.tenant
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

The two halves disagree about one field:

- Gateway — `UserPrincipal.tenant: str | None = None` (`spi/authn/_models.py`), filled by the google strategy from `authn.google.default_tenant`.
- Backend — `UserPrincipal.tenant: str`, required (`core/gateway_principal/models.py`).

`authn.google.default_tenant` appears nowhere in `configs/application.yaml`, so the DI default of `None` wins and the gateway signs `tenant: null`. The backend refuses to parse it, and the request fails closed — correctly, but for a reason no config change was going to fix cleanly.

## Solution

Remove the field on both halves rather than making it required, because the field was asserting something no user credential proves.

**A tenant is a property of the calling program, not of a person.** `app`, `bot` and `access_key` are each *registered* to a tenant, and their principals assert that registration. A Google token proves a `sub` and an email; the tenant that used to ride alongside it was a deployment default dressed up as an authenticated fact.

- `VerifiedCaller.tenant` now reads only the principals that assert one. A set that asserts none — a user and nothing else — resolves to `DEFAULT_AVERNET_TENANT`. A first-party human on our own frontend *is* an internal caller, which is the scope `teamclaw` names and the one every non-public path in the backend already resolves to.
- `_reject_contradictory_tenant` likewise vets only what was claimed, via a new `_asserted_tenants` helper, and returns early when nothing was.
- The google strategy's `default_tenant` argument and its DI wiring are gone; nothing else read it.

**The internal-tenant guard gets sharper, not weaker.** The fallback is decided in the backend, from the *absence* of a claim; it is never a value the token supplied. A token that names `teamclaw` is still refused, and a `tenant` smuggled onto a `user` entry is dropped by the DTO rather than honoured — both pinned by tests.

Rejected alternatives: making the gateway field required and adding the config key would have kept a tenant on the principal that no credential backs, and moved the failure to boot time rather than removing it.

## Validation

Based on `REL20260806`.

- Backend full suite: **10535 passed, 3 skipped**.
- Gateway suite (`-m "not integration"`): **714 passed, 1 skipped**. The 12 failures are identical on an unmodified tree — pre-existing markdown formatting in `docs/`/`specs/`, plus e2e tests needing a live server and a script test needing network.
- `ruff check` and `ruff format` clean on the gateway (its lint gate); the backend declares no ruff config.
- `mypy`: no new errors on either side (gateway 71 before and after, all pre-existing and elsewhere; the backend's `gateway_principal` module is clean).
- Tests that meant to exercise external-tenant isolation now mint `user` + `app`, so they keep testing isolation rather than the internal default: `test_skills_endpoints.py`, `test_openapi_skill_upload.py`, `test_principal_seam.py`.

Not run: the gateway's integration tests (need live BCS/upstream) and the full pre-push CI gate — the environment could not install from the pinned aliyun index, so the suites above ran in a locally built venv from PyPI.

> Note for anyone comparing against an earlier CI run: the first push was cut from `dev` and its Backend job failed on four `test_openapi_skill_state.py` cases. That file arrived on `dev` in #467898e with the same now-stale `user`-principal-carries-a-tenant payload, and it does not exist on `REL20260806`. If it reaches this release branch it needs the same `user` + `app` treatment as the three files above.

## Compatibility and risk

⚠️ **`route_security` declares `user: required` and nothing else for the whole `/openapi/v1` surface**, so every public request is now a user-only set and scopes to `teamclaw` — the internal tenant. Nothing gates *which* Google account that is: `AuthPlugin.is_allowed` exists in the gateway SPI and no authn strategy calls it. Before this surface serves real data, either wire that whitelist up or have a route require an identity carrying a registered tenant. Called out in both docs rather than fixed here.

Wire contract change, so **both halves must ship together** — a gateway that still sends `tenant` on a `user` entry is fine (the field is ignored), but a backend still requiring it keeps 401-ing. Rollback is reverting the commit on both.

## Spec

- `src/gateway/docs/2026-07-21-auth-design.md` §4.6 — original text made `tenant` mandatory on every principal; amended in place with a dated revision note explaining the change and its consequence.
- `src/backend/docs/openapi-v1/README.md` — the auth-seam section and a dated changelog entry.

### DIFF
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -442,6 +442,31 @@ wrong half of the system.
 - The gateway's tenant id **is** the `avernet_tenant` value — no mapping table.
   So a real external tenant reads an empty dataset until it has data; that is
   isolation working, not a bug.
+- **Only the machine principals carry a tenant.** _Changed 2026-08-05._ A
+  `user` principal has no `tenant` field: nothing in a user credential proves
+  which tenant a person acts for, and the gateway's google chain used to fill it
+  from a config default — which, left unset as it shipped, sent `null` and 401'd
+  every request. `app`, `bot` and `access_key` are each registered to a tenant,
+  and that registration is what their principal asserts.
+  - An identity set that asserts **no** tenant — a user and nothing else —
+    resolves to `DEFAULT_AVERNET_TENANT`. A first-party caller on our own
+    frontend *is* an internal caller, which is the scope `teamclaw` names, and
+    it is the same tenant every other path in this component resolves to.
+  - That fallback is **ours**, decided from the absence of a claim; it is not a
+    value the token supplied. So the `teamclaw` guard above keeps its teeth: a
+    token that *names* the internal tenant is still refused, and a `tenant`
+    smuggled onto a `user` entry is ignored (unknown fields are dropped, not
+    honoured) rather than becoming a scope.
+  - ⚠️ **Consequence for the public surface.** `route_security` declares
+    `user: required` and nothing else for every `/openapi/v1` path, so the
+    gateway resolves a user-only set and **every public request now scopes to
+    `teamclaw`** — the internal tenant. Nothing gates *which* Google account
+    that is (`AuthPlugin.is_allowed` exists in the gateway SPI but no authn
+    strategy calls it), so serving real data on this surface needs either that
+    whitelist wired up or a route requiring an identity that carries a
+    registered tenant. Until then the surface reads internal data for any
+    authenticated Google user, which is a widening of what the 401 used to
+    prevent by accident.
 
 **Two things you inherit if you own a Track B category:**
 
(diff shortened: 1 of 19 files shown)


totalfrank (commented):
Reviewed commit `a558fc9` (the PR's only commit) against merge-base `1035d72` (`origin/dev`). No findings — the diff is clean.

**Checked:**
- **Correctness**: `_asserted_tenants` correctly excludes `UserPrincipal` (which no longer carries a `tenant` field on either side of the wire); `VerifiedCaller.tenant` and `_reject_contradictory_tenant` both read through it consistently, and `BotPrincipal`/`AppPrincipal`/`AccessKeyPrincipal` still require `tenant: str` on both gateway and backend models.
- **Contract breakage**: wire contract change (field removed from `UserPrincipal` on both gateway and backend) ships together in one commit, with unknown-field-drop verified by `test_a_tenant_claimed_on_a_user_principal_is_ignored` / `test_user_principal_does_not_carry_a_tenant_it_is_handed`, and matching doc updates in both `src/gateway/docs/2026-07-21-auth-design.md` and `src/backend/docs/openapi-v1/README.md` (AGENTS.md's "contract changes require matching docs and conformance/compatibility tests").
- **DI wiring**: `default_tenant` removed consistently from `GoogleUserStrategy.__init__`, `_plugin_core.py`'s `providers.Callable`, and every test call site (`grep` confirms no leftover functional references, only docstring/comment mentions).
- **Python type contracts**: no defensive `T | None` introduced — the diff goes the other way, deleting a `str | None` field that wasn't a real domain state (matches `AGENTS.md`'s "do not widen a required type to `T | None` merely for defensive programming").
- **Test coverage**: new behavior (user-only → internal-tenant fallback, contradictory/empty/internal-tenant rejection with the new multi-principal shape) has direct tests on both gateway and backend sides.
- **Security**: the PR's own description, commit message, and both updated docs already flag that `route_security`'s `user: required`-only public surface means every public request now scopes to the internal tenant with no whitelist gating which Google account — called out explicitly as unresolved rather than hidden, so no separate finding for it here.

Not applicable to this diff: concurrency/resource safety, architecture-boundary rules (no transport/framework leakage introduced).

---

liveandevil (approved):
LGTM

FreddieSun (approved):
LGTM

PR #810, opened 2026-08-05
feat(baas): enable template override during bot update

evilcoda:
## Summary

- accept an optional `template_uuid` on the BaaS bot update API
- validate the target template as ONLINE within the current tenant before creating publish state
- carry the override into the UPDATE publish and persist it on the new PENDING bot record
- preserve the existing template and update behavior when `template_uuid` is omitted

## Root cause

The backend restart flow correctly resolved and sent the whitelist template UUID, but BaaS `UpdateBotRequest` did not define `template_uuid`. Pydantic therefore discarded the field, and the UPDATE workflow cloned the source bot record with its old template UUID.

## Compatibility

The new field is optional. Requests that do not provide `template_uuid` continue cloning the source bot template exactly as before. No database schema or frontend changes are required.

## Validation

- `uv run pytest -q tests/unit`: 7780 passed, 3 skipped, 8 xpassed
- changed-line coverage: 100% (26/26)
- Ruff checks passed
- `git diff --check` passed

### DIFF
--- a/src/baas/src/secbaas/community/adapters/web/routers/bot_service/management_router.py
+++ b/src/baas/src/secbaas/community/adapters/web/routers/bot_service/management_router.py
@@ -76,6 +76,12 @@ class UpdateBotRequest(BaseModel):
 
     name: str | None = Field(default=None, max_length=128)
     description: str | None = Field(default=None, max_length=512)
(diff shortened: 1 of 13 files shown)


cassiuscai (approved):
LGTM

PR #813, opened 2026-08-05
fix(bcs): keep system notices in human session history via PersistMode

zh-bupt:
## Problem

Since #775 (per-recipient system-message ownership), every system message is persisted per recipient with `owner_bot_id = Some(recipient)` and **no `owner = None` record exists at all**. Human viewers reading `GET /sessions/{sid}/messages` (workbench passes `view_bot_id = human_xxx` or omits it) are scoped to `MessageOwnerFilter::IsNull` — public records only — so **all system notices (member join/leave, mode changes, generic notifications) vanish from human session history**. The live WS `user_message` broadcast still works, but after a refresh the notices are gone. Bot chat messages (persisted `owner = None`) are unaffected.

Fixing this from the read side (e.g. letting humans read the driver's copies) was rejected: per-recipient copies are personalized (`你是:…` context injection), which is exactly the leak #775 removed.

## Solution

Give `SystemGroupMessage` a `PersistMode` field so **producers — who know whether a message is personalized or a shared notice — own the persistence decision**, and the dispatcher executes it mechanically (no content inspection, no per-bot judgment):

- `PerRecipient`: one record per recipient, `owner = recipient` (unchanged behavior). Used by SessionContext per-bot context and BotJoined's new-bot injection.
- `Public`: exactly one record with `owner = None`, readable by humans (`IsNull` view) and every bot's `PublicOrOwner` view — consistent with `user_message`, which is already broadcast session-wide to all frontend viewers. Used for notices whose text is identical for all recipients (BotJoined / HumanJoined / BotLeft / ParticipantModeChanged / Generic / BotHiddenNotice). Persisted even with zero recipients (last bot leaving), matching the `user_message` broadcast. Also collapses the previous N identical per-bot copies into a single record.
- `Skip`: no record (BotHiddenNotice's duplicate "others" copy).

Read side is completely untouched (`PublicOrOwner` / `IsNull` / `Eq` semantics unchanged); no schema migration.

## Validation

- `cargo check --workspace --all-targets` clean.
- The existing end-to-end round-trip test in `bcs-message` (real dispatcher + real `MemoryMessageRepo` + `get_session_history`) updated to the new contract and extended with a **human viewer (`view_bot_id = None`) assertion**: public notices are returned, per-bot owned copies stay hidden — the direct regression guard for this bug. It also asserts the shared notice is a single record, not per-bot copies.
- Producer unit tests updated/extended to lock each producer's `PersistMode` choice (BotJoined injection=`PerRecipient` + notice=`Public`; SessionContext all `PerRecipient`; BotHiddenNotice mentioner=`Public` + others=`Skip`; BotLeft/Generic/HumanJoined emit the public message even with zero recipients).
- Dispatcher tests updated: BotJoined persists injection (owned) + public notice (no per-bot copies); last-bot-leaving persists one public record (test renamed accordingly).
- Full runs green: `bcs-domain`, `bcs-system-message`, `bcs-message`, `bcs-message-store`, `bcs-message-flow`, `bcs-app-session`, `bcs-group`, and bootstrap human integration tests (`human_dm_group_integration` 10, `human_regress_integration` 19).

## Compatibility and risk

- **Backward compatible read path**: no query/filter changes; bot viewers see the same notices (now via the public record instead of an identical per-bot copy) with no duplicates.
- **Two accepted behavior changes** (discussed and intentional): ManagerWorker workers (`Eq(worker)` view) no longer see join/leave notices in history — their view stays scoped to task messages; and a newly joined bot now sees the public "X joined" notice about itself in history (it receives no active delivery of it — persistence and delivery are now decoupled by `PersistMode`).
- **Historical data**: notices persisted before this change remain invisible to humans (they were never stored as public records); no migration. New events from deploy onward are covered.

### DIFF
--- a/src/bcs/crates/contracts/bcs-domain/src/lib.rs
+++ b/src/bcs/crates/contracts/bcs-domain/src/lib.rs
@@ -93,7 +93,7 @@ pub use session::{
     CallbackChannelConfig, CallbackConfig, Session, SessionKind, SessionStatus, ServiceSpec,
 };
 pub use system_message::{
-    SystemGroupMessage, SystemMessageEvent, SystemMessageEventKind
+    PersistMode, SystemGroupMessage, SystemMessageEvent, SystemMessageEventKind
 };
 pub use task_ledger::LedgerSummary;
 pub use invite::{InviteTargetType, InviteTokenPayload, InviteTokenError, encode as invite_token_encode, decode_and_verify as invite_token_decode_and_verify, decode_and_verify_no_expiry as invite_token_decode_no_expiry};
--- a/src/bcs/crates/contracts/bcs-domain/src/system_message.rs
+++ b/src/bcs/crates/contracts/bcs-domain/src/system_message.rs
@@ -79,8 +79,30 @@ impl SystemMessageEvent {
     }
 }
 
+/// Persistence policy for a system group message.
+///
+/// Producers decide how each bot-facing message is recorded in the message
+/// store; the dispatcher executes the policy mechanically (it never inspects
+/// message content to decide ownership).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PersistMode {
+    /// Persist one record per recipient with `owner_bot_id = recipient`
+    /// (personalized per-bot context such as `[GROUP CONTEXT]`, only
+    /// readable by that bot's history view).
+    PerRecipient,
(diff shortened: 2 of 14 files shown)


vzvince:
@codex review

vzvince:
@codex review

vzvince (approved):
LGTM

PR #815, opened 2026-08-05
fix(backend): read Desktop active skills from live runtime

XIEAN99:
## Summary
- make Desktop `GET /api/skills/active/list` read the Bot live device active root instead of the Backend host filesystem
- keep non-Desktop management-host scanning unchanged
- skip reserved, unrelated, dangling, and path-escaping entries; propagate device failures instead of reporting a false empty list

## Production evidence
- OpenClaw Bot `405935/desktop_bot_1785901349211`
- Skill `1120709` / `sp455-r2-oc-probe`
- activation trace `0be8ed2017859198842118576e7fdd` created the runtime link and live `SKILL.md` was readable
- active-list trace `0b446a3117859198952337835e5556` still returned count 0 before this fix

## Validation
- red tests reproduced Desktop runtime/API divergence before implementation
- 291 focused Skill Center, service, and architecture tests passed
- OpenClaw, Claude Code, and Hermes active roots covered by parameterized device-filesystem tests
- Ruff passed
- local Python SAST block scan passed
- diff-check passed

## Scope
- no gate, rollout, migration state-machine, mapping publication, Engine, or cloud Bot behavior changes
- no Spec or CLAUDE.md changes

PR #817, opened 2026-08-05
fix(backend): skip non-skill desktop active entries

XIEAN99:
## Problem

The Desktop active-Skill listing introduced by #815 reads `SKILL.md` directly for every active-root child. Real BaaS device reads raise on a missing file (including HTTP 404), while the original unit double returned `None`. A legitimate non-Skill entry such as `mcporter` could therefore make `/api/skills/active/list` return 500 even when the actual active Skill links were healthy.

## Solution

Use the existing device-filesystem `exists()` probe before reading `SKILL.md` or the historical `README.md` fallback. Entries without either metadata file are ignored; device transport/server failures still propagate and are not converted to an empty list. Cloud/non-Desktop behavior is unchanged.

## Validation

- Regression test demonstrated the pre-fix failure against the real `read_file`/`exists` contract.
- 157 focused Skill CRUD, SkillSet, device-sync, recovery-listener, and multi-engine path tests passed.
- 11 focused Desktop active-list tests passed.
- Ruff passed for both changed files.
- Python SAST block scan passed for both changed files.
- `git diff --check` passed.

## Compatibility and risk

No API, schema, rollout-state, mapping, or runtime layout contract changes. The change only filters active-root children that are not valid Skills before attempting to read their metadata.

## Related issues

Follow-up to #815 and the #797 pre acceptance.

### DIFF
--- a/src/backend/src/agentclaw/community/core/skill_center/services/skill_service.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/skill_service.py
@@ -510,9 +510,14 @@ class SkillService:
                 continue
 
             active_path = self.active_dir / name
-            content = await device_fs.read_file(str(active_path / "SKILL.md"))
-            if content is None:
-                content = await device_fs.read_file(str(active_path / "README.md"))
+            content = None
+            skill_file = active_path / "SKILL.md"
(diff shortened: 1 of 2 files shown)


FreddieSun (approved):
LGTM

PR #819, opened 2026-08-05
fix(harness): slim D-TOOLS-002 prompt payload (MCP inputSchema compre…

wienyang:
## Problem

  #717 dumped each MCP tool's full `inputSchema` into the D-TOOLS-002 prompt.
  MCP-heavy bots (~88 tools) hit antchat's ~90s gateway window →
  RemoteProtocolError → `[llm disabled]` → spurious "LLM执行异常(请重试)",
  pinning health_score low. Retries and bigger client timeout don't help (the
  disconnect is gateway-side).

  ## Solution

  Slim the input payload: compress each tool to a flat param table
  (name/type/required/desc), cap tools-per-MCP at 20 (`tools_omitted`), drop
  `json.dumps` indent. Bucketing and the `has_recoverable` advisory logic still
  read the full `inputSchema` — only the prompt is slimmed. Left
  `DIAGNOSTIC_MAX_TOKENS=32768` unchanged (capping it wouldn't fix an
  input-driven timeout and could truncate the trailing `[SCORE:xx]`).

  ## Validation

  - py_compile clean; pre-push SAST block rules passed.
  - Updated 1 test + added 3 (param flattening, no-schema, tool cap).
  - Not run: pytest (this env lacks project deps like fastapi_injector). Run
    before merge: `python -m pytest src/backend/tests/community/core/harness/diagnostics/test_diagnostics.py -x -q`

  Relates to #717.

### DIFF
--- a/scripts/ci/singlebox_coverage_modules.yaml
+++ b/scripts/ci/singlebox_coverage_modules.yaml
@@ -233,6 +233,22 @@ modules:
     router_api: *pending_router_denominator
     plugin_api: *pending_plugin_denominator
     thresholds:
+      # 41.30 → 40.86: fix(harness) slim D-TOOLS-002 prompt payload added the
+      # _compact_tool_for_prompt / _compact_mcp_tools_for_prompt helpers to
+      # mcp_format.py. These run only inside the D-TOOLS-002 prompt build, which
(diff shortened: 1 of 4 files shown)


FreddieSun (approved):
LGTM

PR #827, opened 2026-08-05
fix(backend): use Engine active root for Desktop Skill listing

XIEAN99:
## Problem

Desktop `GET /api/skills/active/list` asked the live Engine to scan a path derived by Backend `WorkspacePathFactory`. That happens to match OpenClaw, but it maps Claude Code to `/home/admin/.claude_code/workspace/skills` while the Engine-owned Planner uses `/home/admin/.claude/skills`. The endpoint therefore returned `count=0` even though the live Claude Code runtime had readable active Skills. AICoding shares the Claude active root and was exposed to the same mismatch.

## Solution

- Resolve the authoritative Desktop `active_root` through the existing current-runtime layout probe before constructing `SkillService`.
- Keep the Backend management path as `SkillService.active_dir`; pass the Engine root only to the live-device read so Backend never creates runtime-only `/home/admin/...` paths locally.
- Cover all filesystem engines: OpenClaw, Claude Code, AICoding, and Hermes.
- Preserve old-image compatibility only when the live adapter explicitly reports that the layout-probe endpoint is absent.
- Fail closed for invalid, unbound, or transient probe results instead of silently scanning a guessed directory.
- Keep non-Desktop and Teclaw behavior unchanged.

## Validation

- Full Backend suite: `10590 passed`.
- Focused Skill API, Skill Service, runtime-probe, Teclaw, architecture, and DI wiring suites: `140 passed`.
- Relevant Engine Planner, mapping, layout activation, probe, downloader, and per-engine suites: `380 passed`.
- Live local composition test passed against actual OpenClaw, Claude Code, and Hermes agentboxes: the patched Backend route consumed each Engine probe root, read real Skill files, skipped corpus entries, and retained a separate Backend management root.
- Exact four-engine route assertions cover the Planner-resolved active roots.
- Compatibility assertions cover old-image endpoint absence, Teclaw bypass, invalid paths, and unbound runtimes.
- Ruff passed for all changed files.
- Local Python SAST block scan passed for all changed files.
- `git diff --check` passed.

## Compatibility and risk

This is Backend-only and does not require a new Engine/rootfs image. New images use the existing layout-probe contract. Old images retain the previous path behavior only after adapter liveness confirms that the probe endpoint is absent. No database schema, rollout gate, mapping contract, or external API response schema changes.

## Spec

Implements the Engine-owned Planner authority described by the Skills Pool P3 architecture and Desktop delivery for Avernet #455.

## Related issues

Relates to #455.
Follow-up to #817.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
+++ b/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
@@ -9,6 +9,7 @@ import json
 import os
 import time
 import zipfile
+from pathlib import Path
 from typing import Any, Dict, Optional
 
 from fastapi import APIRouter, Depends, File, Form, HTTPException, Query, Request, UploadFile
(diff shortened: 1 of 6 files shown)


FreddieSun (approved):
lgtm

PR #829, opened 2026-08-05
Do allow teamclaw tenant go through open API

totalfrank:
Do allow teamclaw tenant go through open API.

### DIFF
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -453,10 +453,10 @@ wrong half of the system.
     frontend *is* an internal caller, which is the scope `teamclaw` names, and
     it is the same tenant every other path in this component resolves to.
   - That fallback is **ours**, decided from the absence of a claim; it is not a
-    value the token supplied. So the `teamclaw` guard above keeps its teeth: a
-    token that *names* the internal tenant is still refused, and a `tenant`
-    smuggled onto a `user` entry is ignored (unknown fields are dropped, not
-    honoured) rather than becoming a scope.
+    value the token supplied. A token may *also* name `teamclaw` on a machine
+    principal — _changed 2026-08-05_, see below — and both routes reach the same
+    scope. A `tenant` smuggled onto a `user` entry is still neither: unknown
+    fields are dropped, not honoured, so it never becomes a scope.
   - ⚠️ **Consequence for the public surface.** `route_security` declares
     `user: required` and nothing else for every `/openapi/v1` path, so the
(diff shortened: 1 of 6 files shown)


totalfrank (commented):
**Review status — round 1 (no prior Claude review found on this PR)**

Commits reviewed: `1e3b274` (Do allow teamclaw tenant go through open API), `9908ca0` (test(backend): follow the internal tenant becoming routable) — merge base `71d6279` (REL20260806).

No earlier Claude review, review thread, or inline comment exists on this PR, so there is nothing to verify from a prior round. This round reviewed the full diff.

**Findings:** none.

What this diff does: removes the `verifier.py::_reject_contradictory_tenant` guard that refused a machine-principal token naming the internal tenant (`teamclaw`), now that a `teamclaw` tenant is registered on the gateway as a deliberate first-party path onto `/openapi/v1`. The mixed-tenant rejection stays, and a `tenant` smuggled onto a `user` entry is still dropped.

Checked and found consistent:
- `_reject_contradictory_tenant`/`_asserted_tenants` docstrings, `UserPrincipal` docstring, both READMEs, and `docs/openapi-v1/README.md` were all updated in step with the code; cross-references (`§ Change impact`) resolve to real headings. One dated changelog entry (`docs/openapi-v1/README.md:1059`, "a wire tenant of `teamclaw` is refused") is now stale, but it's a historical dated log entry in an append-only changelog convention this file already uses elsewhere (e.g. superseded notes are added rather than history rewritten), so not filing it as a defect.
- `DEFAULT_AVERNET_TENANT` import in `verifier.py` is still used (the user-only fallback), not dead.
- New/changed tests (`test_internal_tenant_named_on_the_wire_is_honoured`, `test_internal_tenant_still_cannot_be_mixed_with_another`, `test_a_token_naming_the_internal_tenant_is_scoped_to_it`) match the new behavior and use existing fixtures correctly.
- Traced the other backend call sites keyed on `DEFAULT_AVERNET_TENANT` (`create_bot_for_others_service.py`, `create_flow.py`, `middleware.py`, `openapi_v1/dependencies.py`). The internal `/api` surface hardcodes the default tenant regardless of caller (`middleware.py`), so it's unaffected. On the public `/openapi/v1` surface, a caller already reached `DEFAULT_AVERNET_TENANT` via the pre-existing user-only fallback (documented in this same README as an existing, separately-tracked gap) — this PR adds an explicit route to the same value but doesn't widen what was already reachable, so no incremental exposure from this diff.

Open count: **0**.

---

FreddieSun (approved):
lgtm

PR #830, opened 2026-08-05
fix(backend): harden service bot skills pool publication

FreddieSun:
## Problem

Service Bot Skills Pool publication had four fail-open or authority gaps:

- a failed active Skills mapping snapshot could still let a build continue;
- a generic DB-backed activation listener relied on binding data shape instead of validating the current Draft binding identity;
- a service build could snapshot a transitional or concurrently changing layout state;
- the generic Pool to Legacy operator rollback could be invoked for Service Drafts even though it cannot coordinate Runtime Pin.

These gaps could make a published artifact inconsistent with the Draft state that was approved, or let current Draft metadata reach the wrong runtime binding.

## Solution

- fail the service build when active Skills mapping rsync raises or returns non-zero;
- require activation event binding identity to equal the current Draft binding before DB-backed mapping sync, including an explicit Published Service regression case;
- build only from LEGACY_ACTIVE or POOL_ACTIVE and fail closed when layout, phase, migration generation, or raw layout contract changes during the physical snapshot;
- reject generic Service Bot Pool rollback before leases, layout changes, or runtime writes, and translate the domain outcome to HTTP 409;
- document the new non-retryable rollback outcome on the Service API contract.

## Validation

- 381 focused backend tests passed across Skills Pool, service deploy, mapping listener, build snapshot, event, and HTTP adapter suites;
- 54 focused implementation and architecture tests passed after review fixes;
- Ruff passed on all 12 changed files;
- git diff --check passed;
- two-axis review completed with zero remaining Standards or Spec blockers.

## Compatibility and risk

- Personal and Desktop rollback behavior is unchanged.
- Historical Service publishes without a Skills manifest retain explicit Legacy behavior.
- No release-time image and layout cross-validator, Service-specific engine gate, or after-create readiness hook is introduced.
- This PR does not modify Runtime Pin or implement Service Pool to Legacy recovery. Service rollback is deliberately closed until a Runtime Pin-aware contract exists.

## Spec

Implements the minimal guards confirmed by the 2026-08-05 Service Bot Runtime Pin and Skills Pool rollout grill, based on REL20260806.

## Related issues

Relates to #822.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/skills_pool/router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/skills_pool/router.py
@@ -17,6 +17,7 @@ from agentclaw.community.api.skills_pool_recovery_service import (
     SkillsPoolRecoveryServiceProtocol,
 )
 from agentclaw.community.api.skills_pool_rollback_service import (
+    SkillsPoolRollbackOutcome,
     SkillsPoolRollbackServiceProtocol,
 )
 from agentclaw.community.api.skills_pool_rollout_service import (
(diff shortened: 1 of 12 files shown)


totalfrank (approved):
LGTM.

PR #831, opened 2026-08-05
refactor(backend): consolidate repositories into core/repository with enforceable contracts

totalfrank:
## Problem

Every repository implementation lived in the plugin layer
(`agentclaw/community/plugins/`), and every repository Protocol lived inside the
domain package it served, under nine different filename conventions. No
implementation named its Protocol as a base class, so the
contract-to-implementation link existed only inside a DI `binder.bind()` call and
in prose — IDE "go to implementation" could not follow it.

Repositories are not plugins. `plugin_api/README.md` and Rule 20 require every
plugin contract to carry paired local and prod implementations; each of these had
exactly one, and the only per-profile difference was the `DatabasePlugin` injected
into the constructor — one layer below, where the swap belongs.

The consequence was undetectable drift, and it was recurring. `SkillRepository`
declared three members its implementation never had. It was two when this spec was
written; `remove_all_default_skill_exclusions` landed on `dev` via #894 mid-review,
declared on two Protocols and implemented on one, reproducing the same copy/paste.
Nothing in CI noticed either time.

## Solution

**43 implementation modules (44 classes) and 46 Protocols** move into
`core/repository/`, grouped into **11 domain subdirectories**:

```text
core/repository/
├── protocols/        contracts, @abstractmethod throughout, TYPE_CHECKING-only domain imports
└── implementations/  bodies, each declaring its Protocol(s) as a base
    ├── bot/  chat/  skill_center/  skills_pool/  governance/  harness/
    └── platform/  identity/  devices/  publishing/  config/
```

`plugins/` now holds exactly `__init__.py`, `http_client.py`, `local/`, `community/`.

**The contract is enforceable, not decorative.** Every Protocol member carries
`@abstractmethod` and every implementation declares its Protocol, so an
implementation that omits a member fails at construction naming it:

```
TypeError: Can't instantiate abstract class Incomplete without an
implementation for abstract methods 'claim_batch', 'complete', ...
```

That is not hypothetical — **it is what forced the `SkillRepository` fix.** With
the contract enforced, every profile's injector refused to construct the class and
named all three orphaned members. All 7 call sites reach them through
`skill_set_repo`, typed `SkillSetRepository`, so the declarations were dead; they
were dropped from `SkillRepository` and kept on `SkillSetRepository`. No call site
changed.

`test_repository_contracts.py` keeps it that way: every member abstract, every
implementation based, no runtime domain import in `protocols/`, role separation
between the two directories, plus an executable proof of the failure mode. Each
invariant was verified to have teeth by deliberately breaking it.

**Collateral fixed at the root, not suppressed:**

- The Rule-14 layering violation (five bodies importing ORM models from
  `plugins/local/`) is gone — the five models moved to their owning domains, and
  the `TODO(repo-unify)` that documented the debt is deleted. The guard's
  allowlist stays empty.
- Six domain records and errors were separated out of Protocol source files.
- `LocalSkillCleanupRepository` was reclaimed from `plugin_api/` — it never
  subclassed `Plugin` and had one implementation.
- 28 vendor-name mentions neutralized; both governance guards re-keyed to scan the
  repository tree as a second governed root.
- `plugins/README.md` rewritten — it still described the directory as a skeleton
  of empty class stubs.

**Corrections to the spec's own premises, found by doing the work:** 8 of 44
implementations *already* declared their Protocol, so "no implementation names its
Protocol as a base" was too strong; and one generated contract needed a
runtime-evaluated default (`QueryScope.OWNER`), handled with the typeshed `= ...`
idiom rather than weakening the no-runtime-imports rule.

## Validation

- **Backend suite: 11097 passed, 2 failed.** Both failures are
  `rsync: not found`, verified to fail identically on pristine `dev` — `rsync` is
  not installed in this sandbox.
- **Architecture: 126 passed**, with **no new allowlist or exemption entry in any
  guard** and no stale entry left behind.
- **DI: 129 passed** — the full injector graph constructs on every profile.
- **R8 verified mechanically**: all 50 moved modules diffed against the
  merge-base. 29 differ beyond imports and class heads; every one is accounted for
  (26 docstring vendor rewrites, 2 ORM re-pointings required by R7, the R3b
  annotation, removed duplicate imports). **Zero query, return-shape or error-case
  changes.** Decorator and `def` counts are identical to pre-move for all 50.
- **All 8 spec acceptance criteria verified explicitly.**

Four classes of breakage that no AST-based import rewriter can see were each found
by a failing test and then swept tree-wide: `patch("module.path")` string targets,
`__import__("...")` literals, a module-level alias assignment, and aliased
duplicate bases.

## Compatibility and risk

**`corp/ocb` must be updated in lockstep.** It imports these module paths, and the
PR squash-merges to a single commit, so the broken window is exactly one commit.
`specs/2026-08-05-core-repository-consolidation/path-map.md` is the artifact to
update it from — 98 rows across 8 sections, generated from the tree.

No class was renamed, so every row is a pure path rewrite of an unchanged symbol.

Confirmed with the requester before planning: no repository in this set has a
corp-side implementation. `src/agentclaw/corp/` is not present in this repository,
so that could not be verified from the checkout and was not assumed.

**One coverage threshold was disturbed, and was fixed by restoring the denominator
rather than lowering the floor.** CI caught `bot_chat` at 65.63% against its 67.48%
floor, having lost 1,770 of its 3,565 `core_paths` lines when both repository
bodies moved out. `core_paths` is a prefix list, so the two moved files were added
back to it: the floor stands untouched, and a production Core path that would
otherwise have silently dropped out is retained. Re-pinning remained available and
was not used.

## Spec

- `src/backend/specs/2026-08-05-core-repository-consolidation/spec.md`
- `src/backend/specs/2026-08-05-core-repository-consolidation/plan.md`
- `src/backend/specs/2026-08-05-core-repository-consolidation/tasks.md`
- `src/backend/specs/2026-08-05-core-repository-consolidation/path-map.md`

## Related issues

Surfaced but deliberately out of scope: #912 —
`SkillsPoolLayoutRepository` is one repository over one table split across five
files to stay under the Rule 9 cap, and its mixin file seams cut across its two
contract seams. Carried as-is under R8.


src/backend/tests/community/architecture/test_governance_state_machine_guards.py
```diff
@@ -171,7 +193,7 @@ def test_no_external_callers_of_repo_semantic_commands():
         rel = str(py.relative_to(backend))
         # 跳过 repo 自身(已删,但避免把方法定义误判为调用)与 lifecycle
         # driver(driver 的方法名同名但走 self.<cmd> 不走 task_repo.<cmd>)。
-        if "repositories/task_record_repo.py" in rel:
+        if "repository/task_record.py" in rel:
```
totalfrank: Non-blocking: this self-exclusion no longer matches after the move, so it's dead code.

`rel` is computed relative to `backend/src/agentclaw` (see `rel = str(py.relative_to(backend))` above), so for the moved file it now evaluates to `community/core/repository/implementations/governance/task_record.py`. The substring `"repository/task_record.py"` requires `repository` to be directly followed by `/task_record.py`, but the real path has `/implementations/governance/` in between — so this `if` never matches, and the comment above it ("跳过 repo 自身...避免把方法定义误判为调用") is no longer true of the code.

It happens to be harmless today only because `task_record.py` contains no `self._task_repo.<forbidden_command>(...)` call pattern. If that ever changes, the guard would incorrectly flag the repo's own body as an "external caller" violation.

Suggested fix:
```suggestion
        if rel == "community/core/repository/implementations/governance/task_record.py":
```

---

totalfrank (commented):
Reviewed commits 92d2d35..e4556d4 (the full PR — first review on this branch) against `dev` @ 5cdb614 (merge-base).

**What I checked**, beyond reading the diff:
- Full-tree grep of the new source/test tree for lingering references to all 86 deleted module dotted-paths (mechanical import repointing) — zero stale references found.
- `py_compile` on all 498 changed Python files — all compile cleanly.
- AST checks over `core/repository/protocols/`: every Protocol member carries `@abstractmethod`, no runtime (non-`TYPE_CHECKING`) `agentclaw.community.core.*` imports — matches `test_repository_contracts.py`.
- AST check over `core/repository/implementations/`: no implementation class declares duplicate/aliased base classes; the 4 `*Mixin` classes with no Protocol base are the documented `SkillsPoolLayoutRepository` mixins (tracked under #912), not an oversight.
- `SkillRepository`/`SkillSetRepository` split: confirmed `remove_all_default_skill_exclusions` (and its two siblings) live only under `SkillSetRepository` in both the protocol and the implementation, matching the "dead on `SkillRepository`, kept on `SkillSetRepository`" claim.
- `BotChatDbRepository` / `OpenBotChatRepository` DI rewiring: both Protocols are bound in `bot_chat_open_module.py`, and no direct (non-DI) construction site remains in production code — `otel_router.py` now takes the Protocol via `Injected(...)`, and its test was updated consistently (fake repo passed directly, no more `monkeypatch.setattr(otel_router, "BotChatDbRepository", ...)`).
- `singlebox_coverage_modules.yaml` `bot_chat.core_paths` addition: the two added paths (`implementations/chat/open.py`, `implementations/chat/db.py`) are exactly the renamed `bot_chat/repository/open.py` and `bot_chat/repository/product.py` (99% match) — denominator restoration is correctly scoped and doesn't pull in `expert_chat`/`channel`, which also live under `implementations/chat/`.
- Guard/allowlist diffs (`test_no_oversized_modules.py`, `test_local_plugins_use_baas.py`, `test_governance_notify_lifecycle_guards.py`, `test_governance_state_machine_guards.py`): all are path repoints or removals for files that moved out of the scanned root — none weaken a gate; the governance guards actually gained a second governed root.
- Scanned added lines for hardcoded secrets/URLs and leftover `TODO(repo-unify)` markers — none found.

One non-blocking finding posted inline: a guard's self-exclusion check in `test_governance_state_machine_guards.py` no longer matches after the file's path changed, making it dead code (currently harmless, but silently so).

Categories from the review checklist not exercised by this diff: this is a pure structural move with no Service/Plugin API surface change, no new error-handling paths, and no concurrency-relevant code — so contract-breakage-for-external-callers, failure-handling, and concurrency/resource-safety didn't apply beyond what's covered above.

---

src/backend/src/agentclaw/community/core/repository/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Repository contracts and their implementations.
```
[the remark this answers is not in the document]
totalfrank: Correct, and worse than it looks — this was supposed to be in the PR. Task 12 of the spec lists "Added to `BOUNDARY_SIGNIFICANT_MODULES`, with a `README.md` carrying a Context Boundary section" as done. The bookkeeping was marked `[x]`; neither artifact actually landed.

Fixed in `f3a6e5a`:

- `agentclaw.community.core.repository` added to `BOUNDARY_SIGNIFICANT_MODULES`.
- `core/repository/README.md` created with the `## Context Boundary` block — `purpose`, `provides`, `consumes`, and `internal_dependencies` covering all 56 external `agentclaw.*` imports the package makes (26 declared prefixes), plus the domain map and the reason `protocols/` carries no runtime domain import.

Verified the registration is live rather than vacuous: corrupting one declared prefix in the README makes `test_declared_deps_cover_actual_imports` fail naming `agentclaw.community.core.repository`. Restored, and the full architecture suite is 126 passed.

---

scripts/ci/singlebox_coverage_modules.yaml
```diff
@@ -117,6 +117,14 @@ modules:
       - tests/community/acceptance/bot_chat
     core_paths:
       - src/agentclaw/community/core/bot_chat/
+      # The two bot-chat repository bodies moved to the consolidated repository
+      # package (spec 2026-08-05-core-repository-consolidation). They are the
+      # same code, exercised by the same acceptance flows — only the module path
+      # changed — so the denominator follows them rather than silently shrinking.
+      # Listed per-file because implementations/chat/ also holds expert_chat and
+      # channel repositories, which belong to other modules' denominators.
```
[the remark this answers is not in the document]
totalfrank: Disputed — the four repositories named were never in any denominator, so nothing was removed from one and no percentage is inflated. Checked mechanically rather than by reading.

**Every `core_paths` entry in the manifest is under `src/agentclaw/community/core/`.** The four implementations cited all lived in `agentclaw/community/plugins/` at the merge-base (`5cdb614`), one directory above `local/` and outside every prefix:

| implementation | old path | in a denominator before? |
| --- | --- | --- |
| expert-chat | `plugins/expert_chat_repository.py` | no |
| expert-chat instance | `plugins/expert_chat_instance_repository.py` | no |
| device | `plugins/device_repository.py` | no |
| access policy | `plugins/policy_repository.py` | no |
| harness ×4 | `plugins/harness*_repository.py` | no |

(`path-map.md` rows 80–97.) They are not newly excluded by this PR; they were never gated, and there is no `plugin_api` item for them either — that metric is evidence-keyed to `Plugin` subclasses, which repositories are not.

**What this PR actually removed from gated denominators**, computed by diffing the `.py` file set under each gated module's `core_paths` at `5cdb614` vs. head:

| module | files that left | what they were |
| --- | --- | --- |
| bot_chat | `repository/{open,product}.py` | **bodies** — 777 + 987 lines, 1 `@inject` |
| access | `repository.py` | Protocol (`class PolicyRepository(Protocol)`, 38 lines) |
| devices | `repository/protocol.py` | 2 Protocols, 0 `@inject` |
| bot_collaborator | `repository/protocol.py` | 3 Protocols, 0 `@inject` |
| expert_chat | `repository/expert_chat{,_instance}_repository.py` | 1 Protocol each, 0 `@inject` |
| harness | `repository_protocol.py` | 4 Protocols, 0 `@inject` |

bot_chat is the only module that lost a production body, which is exactly why it is the only one CI caught, and it is restored.

Everything else that left is a contract file whose statements are class and `def` lines with `...` bodies — executed at import, so ~100% covered. Removing a ~100%-covered file from a module sitting below 100% **lowers** its percentage. Those six gates got stricter, not looser, and all still pass. Restoring them would inflate, and would require cross-attribution: `protocols/identity.py` serves access + caller_identity + user_list, `protocols/chat.py` serves three modules. That is the reason the bot_chat entries are per-file.

On gating the newly-core bodies: worth doing, but it is new gating, not restoration. It needs floors measured against an acceptance run rather than asserted, and those bodies are lightly covered by acceptance, so it would mean lowering floors — the thing this PR declined to do when bot_chat went red. Filed separately so it lands with measurement rather than inside a structural move that is required not to change behavior: **#914**.

---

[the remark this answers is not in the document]
totalfrank: Correction to the last line: the follow-up issue is **#913**, not #914. Everything above stands.

---

src/backend/tests/community/architecture/test_governance_state_machine_guards.py
totalfrank: Confirmed and fixed in `f3a6e5a` — the substring genuinely never matches now, and the comment above it describes behaviour the code no longer has.

I took the diagnosis but not the literal. Replacing one hardcoded path with another leaves the same trap armed: the next move produces a silently dead exclusion again, and the failure is invisible precisely because the guard keeps passing. Both guards now derive the path from `_GOV_REPO_ROOT`, which is already the anchor `_gov_files()` walks:

```python
_TASK_RECORD_REPO = _GOV_REPO_ROOT / "task_record.py"
...
        if py == _TASK_RECORD_REPO:
            continue
```

`py` comes from `backend.rglob("*.py")` and both roots descend from `_BACKEND_ROOT`, so the comparison is absolute-path equality. `test_guard_b_repo_has_no_semantic_commands` was rebuilding the same path inline; it uses the constant now too, so a future move breaks loudly in one place instead of quietly in another.

Architecture suite: 126 passed.

---

totalfrank:
**Review round 1 — 3 findings on `e4556d4`. 2 fixed, 1 disputed.** Pushed `f3a6e5a`.

**Fixed — `core/repository` was not a registered context boundary** (Codex, P1). The spec's Task 12 lists this as done and it was marked `[x]`; neither the `BOUNDARY_SIGNIFICANT_MODULES` entry nor the README existed, so Rule 22 skipped the whole package. Both now land, with `internal_dependencies` covering all 56 external `agentclaw.*` imports the package makes. Verified live rather than vacuous: corrupting one declared prefix fails `test_declared_deps_cover_actual_imports` naming the module.

**Fixed — a governance guard's self-exclusion had gone dead** (Claude Code). `if "repository/task_record.py" in rel` stopped matching once the file moved under `implementations/governance/`. Harmless today, and silently so. Both guards now derive the path from `_GOV_REPO_ROOT` rather than spelling it out, so the next move breaks loudly instead of quietly — a literal would have re-armed the same trap.

**Disputed — "moved repositories missing from coverage denominators"** (Codex, P1). The four implementations named lived in `plugins/` at the merge-base, and every `core_paths` prefix is under `core/`, so they were never in a denominator and nothing was removed from one. Diffing the `.py` file set under each gated module's `core_paths` at `5cdb614` vs. head: bot_chat is the only module that lost a production **body** — which is exactly why it is the only one CI caught, and it is restored. Every other departure is a Protocol file (0 `@inject`, `...` bodies), ~100% covered at import, so its removal *lowered* those percentages. Six gates got stricter, not looser. Full table on the [thread](https://github.com/inclusionAI/Avernet/pull/831#discussion_r3743014100).

The finding does point at something real underneath it, so: **#913** — gate the ~40 repository bodies that became Core here and were gated by nothing before. That is new gating, not restoration; it needs floors measured against an acceptance run, and it will mean lowering some. Kept out of a PR constrained to change no behavior.

CI on `e4556d4`: Backend unit tests green (the `BotChatService` construction fix holds); Singlebox coverage was still running when the new head landed, so it re-runs on `f3a6e5a`.

---

totalfrank (commented):
Reviewed the one new commit since my last round: `f3a6e5a` (92d2d35..e4556d4 was the prior round; `f3a6e5a` is the only commit added since).

**Diff reviewed:** `f3a6e5a` touches exactly three files — new `core/repository/README.md`, and small fixes in `test_governance_state_machine_guards.py` and `test_module_boundaries.py`.

**Verification performed:**
- Re-derived the `_TASK_RECORD_REPO` path-equality fix by hand: `_GOV_REPO_ROOT` and `backend` in `test_no_external_callers_of_repo_semantic_commands` both descend from the same resolved `_BACKEND_ROOT`, so `py == _TASK_RECORD_REPO` correctly matches the repo's own file via plain path equality — no substring-matching trap left.
- Extracted every `agentclaw.*` import actually made under `core/repository/` at `f3a6e5a` (79 files) and diffed it against the new README's `internal_dependencies` list — full coverage, nothing undeclared.
- Confirmed `agentclaw.community.core.repository` isn't shadowed by `EXEMPT_MODULES` (empty) and the added README section carries all four required schema keys.
- Could not execute pytest in this sandbox (no network to resolve the `uv` lockfile), so the above is static/mechanical verification rather than a live run.

No new issues found in this commit.

## Findings checklist

| Finding | Filed by | Status |
| --- | --- | --- |
| Dead self-exclusion substring in `test_governance_state_machine_guards.py` (non-blocking) | this review | **Fixed** — verified above, thread resolved |
| Missing context-boundary registration for `core/repository` (P1) | Codex | Fixed per author's reply, thread resolved (not re-verified in depth by me — not my finding) |
| Coverage denominator omits newly-moved repository bodies (P1) | Codex | Disputed by author, left open — Codex's finding, not mine to adjudicate |

**Open (mine): 0.**

Nothing new to raise this round.

---

totalfrank:
@codex review

Round 2 opened by `f3a6e5a`. Codex only fires on ready-for-review or an explicit ask, so this is that ask — the boundary-registration P1 you filed is fixed in that commit and worth a second look. The coverage P1 has a reply on its thread and is left open deliberately.

---

src/backend/src/agentclaw/community/core/repository/README.md
```diff
@@ -0,0 +1,126 @@
...
+in a comment, rather than an exception to the rule.
+
+## Context Boundary
+
+```yaml
+purpose: "Every repository contract (protocols/) and implementation (implementations/) in the backend, grouped by domain. Persistence only — no services, no routers, no domain policy."
+provides:
+  - "Repository Protocols — @abstractmethod contracts consumed by domain services for DI"
+  - "Repository implementations — the ORM bodies, one per contract"
+  - "Runtime contract enforcement: an implementation missing a member fails at construction"
```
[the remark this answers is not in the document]
totalfrank: Agreed and fixed in `5f58719` — and the second half of the finding is the more useful half: `test_module_boundaries.py` only asserts each entry is a string, so prose satisfies it while naming nothing. That is exactly how this shipped.

`provides` now lists **78 names** — 45 Protocols, plus the 33 implementation classes whose names differ from their contract — grouped by domain, with the 5 `SkillsPoolLayoutRepository` mixins excluded as internal. `consumes` gets the same treatment: `DatabasePlugin`, `get_current_env`, `get_server_host`, `get_current_avernet_tenant`.

A hand-maintained list of ~80 names rots at the next repository added, so it is derived-checked rather than trusted. `test_repository_contracts.py` gains:

```python
def test_readme_provides_lists_the_real_public_surface() -> None:
    declared, actual = _declared_provides(), _public_surface()
    missing = sorted(actual - declared)
    stale = sorted(declared - actual)
```

Both directions fail, each naming the offending entries — so adding a repository without listing it fails, and deleting one without unlisting it fails too. Verified by removing `DeviceRepository` from the list: the guard fails naming it. Restored.

This only governs `core/repository/`. The general gap you identified — `provides` unchecked across all 28 boundary-significant modules — is real and wider than this PR; worth its own change, since most modules have no equivalent derivable surface.

Architecture suite: 127 passed.

---

totalfrank:
**Review round 2 — 1 new finding on `f3a6e5a`, fixed.** Pushed `5f58719`.

Both reviewers ran this round. Claude Code returned clean on `f3a6e5a` ("Open (mine): 0") after re-deriving the `_TASK_RECORD_REPO` path equality by hand and diffing all 79 files' `agentclaw.*` imports against the new README's `internal_dependencies`.

**Fixed — the new README's `provides` named nothing** (Codex, P2). Three prose sentences where `docs/arch/context-boundary-format.md` asks for "Public surface — names only". The finding also explained why it passed CI: `test_module_boundaries.py` only checks each entry is a string.

`provides` now carries **78 names** — 45 Protocols plus the 33 implementation classes whose names differ from their contract, grouped by domain, mixins excluded as internal. `consumes` likewise, by name.

A hand-maintained list that long rots at the next repository added, so it is checked rather than trusted: `test_repository_contracts.py` derives the surface from the tree and fails in both directions — a repository added and not listed, or listed and deleted — each naming the entries. Verified by removing one name. Architecture suite 127 passed.

**Still open:** the coverage-denominator P1 from round 1, disputed with the file-by-file diff on its [thread](https://github.com/inclusionAI/Avernet/pull/831#discussion_r3743014100), with the real gap underneath it filed as #913.

Codex fires only on ready-for-review or an explicit `@codex review`, never on push — so round 3's Codex pass needs asking for, which I'll do once CI reports on `5f58719`. Singlebox coverage has not yet completed a run on any head carrying the `bot_chat` `core_paths` fix; that remains the one gate unconfirmed.

---

totalfrank:
**Review round 3 — no issues flagged.** Reviewed `5f58719` (the `provides` naming and its derived-check guard).

Codex came back clean on that SHA: the boundary now names concrete contracts, implementations and consumed symbols, and it confirmed the guard "derives the actual public surface, excludes private and mixin classes, and detects both missing and stale README entries". No new findings from either reviewer.

One correction I owe on my own numbers, found while reconciling them rather than by review. **The PR body says "46 Protocols"; it is 45 Protocols and one ABC** — `PublishOperationRepository` uses `ABC`, not `Protocol`. That mattered beyond bookkeeping: the guard I added last round selected on `Protocol` alone, so that contract was outside the required surface and its absence from `provides` passed. An ABC in `protocols/` is a contract for the same reason a Protocol is — consumed as a base, members inherited or missing — and `test_protocols_hold_no_implementations` already treats it as a legitimate resident. Fixed in `96d3d61`: the derivation now covers both, the guard immediately failed naming `PublishOperationRepository`, and it is listed. 79 names. Architecture suite 127 passed.

Standing state: one finding remains disputed — the round-1 coverage-denominator P1, answered with the file-by-file diff on its [thread](https://github.com/inclusionAI/Avernet/pull/831#discussion_r3743014100), with the real gap underneath it tracked as #913. All other threads resolved.

CI: five checks green on `5f58719`; Backend unit tests and Singlebox coverage were still running when `96d3d61` superseded them. **Singlebox coverage has yet to complete on any head carrying the `bot_chat` `core_paths` fix** — four heads, each superseded mid-run. That gate is the one thing on this PR not yet confirmed by CI, and I am watching for it rather than assuming it.

---

src/backend/tests/community/architecture/test_repository_contracts.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+    assert fence, "Context Boundary section has no fenced yaml block"
+    return set(yaml.safe_load(fence.group(1))["provides"])
+
+
+def _is_contract(node: ast.ClassDef) -> bool:
+    """A contract is a Protocol or an ABC — both are consumed as bases.
+
+    Not every contract in ``protocols/`` is a ``Protocol``:
+    ``PublishOperationRepository`` is an ``ABC``, and it is public surface for
```
totalfrank, line 207: **Non-blocking: `_public_surface()` has a blind spot for ABC-based contracts, so one real repository contract is silently missing from `provides`**

`_is_protocol()` (line 56-57) only recognizes classes whose base is literally `Protocol`. `protocols/publishing.py` has one contract that predates this move and doesn't fit that shape:

```python
class PublishOperationRepository(ABC):
    """Data access for the publish operation ledger."""
    @abstractmethod
    def insert(self, data: Dict[str, Any]) -> PublishOperationRecord: ...
```

It's a genuine `@abstractmethod`-bearing contract with a real implementation (`OrmPublishOperationRepository`, which already appears in the README's implementations list). `test_protocols_hold_no_implementations` explicitly allows `ABC`-based classes in `protocols/` (line 142), so this is an accepted contract shape in this package — but `_public_surface()` only walks `_PROTOCOLS` for `is_protocol(c)` matches, so this class is invisible to both sides of the comparison: it's never in `actual`, and consequently the README never has to list it in `declared` for the guard to pass.

Verified by re-running the same AST logic standalone: `_public_surface()` returns 69 names, the README's `provides` also has 69, they match exactly — and `PublishOperationRepository` is in neither set. Confirmed the README's `publishing` section lists only `BotPublishRepositoryProtocol`, not `PublishOperationRepository`.

This is exactly the failure mode this commit was written to close (a check too weak to catch a real gap) — reproduced for one contract instead of prevented. Since `PublishOperationRepository` never enters `actual`, deleting or renaming it wouldn't fail this test either.

Suggested fix: broaden `_is_protocol` (or add a parallel check in `_public_surface`) to also match `ABC`-based classes in `protocols/` whose members are `@abstractmethod`, then add `PublishOperationRepository` to the README's `provides` list once the guard requires it.

---

totalfrank (commented):
Reviewed the one new commit since my last round: `5f58719` (round 2's only commit since `f3a6e5a`, which I reviewed with 0 open findings of mine).

**Verification performed on `5f58719`:**
- Re-derived `_public_surface()` and `_declared_provides()` standalone (AST walk + regex/yaml extraction, same logic as the new test) against the checked-out tree: both sets are 69 names, and they match exactly — the new self-check test does pass.
- Cross-checked the README's 69-name `provides` list by hand-counting the diff's protocol/implementation sections against a full `grep -rn "^class "` of `protocols/` and `implementations/` — no other discrepancy found.
- Found one real gap the new guard can't see (posted inline): `protocols/publishing.py`'s `PublishOperationRepository` is an `ABC`-based contract (not `typing.Protocol`), so `_is_protocol()` never counts it — it's absent from both the computed "actual" set and the README's `provides`, so the derived-check passes without ever seeing it.
- Re-checked Codex's round-2 finding this commit was written to fix (README `provides` was prose, not names) — the fix is real and matches the finding: `provides` now lists concrete names, and the test that would have let prose back in is closed for `Protocol`-based classes.

## Findings checklist (all rounds)

| Finding | Filed by | Status |
| --- | --- | --- |
| Missing context-boundary registration for `core/repository` (P1) | Codex | Fixed in `f3a6e5a`, thread resolved |
| Coverage denominator omits newly-moved repository bodies (P1) | Codex | Disputed by author with per-file evidence, filed as #913, **left open — not my finding to adjudicate** |
| Dead self-exclusion substring in `test_governance_state_machine_guards.py` (non-blocking) | this review, round 1 | Fixed in `f3a6e5a`, thread resolved |
| `provides`/`consumes` were prose, not names (P2) | Codex, round 2 | Fixed in `5f58719`, thread resolved |
| `_public_surface()` misses `ABC`-based contracts — `PublishOperationRepository` invisible to the new guard (non-blocking) | this review, round 3 | **New, open** |

**Open (mine): 1.** Open (not mine, author's call already given): 1.

---

src/backend/tests/community/architecture/test_repository_contracts.py
totalfrank: Right, and already fixed — `96d3d61` landed a few minutes before this review, so it was written against `5f58719`. Independent arrival at the same conclusion, which is worth something: I found it reconciling the PR body's "46 Protocols" against the tree and getting 45.

The fix is the shape you suggest. `_public_surface()` now selects on `_is_contract()`:

```python
def _is_contract(node: ast.ClassDef) -> bool:
    """A contract is a Protocol or an ABC — both are consumed as bases."""
    return _is_protocol(node) or any(
        isinstance(b, ast.Name) and b.id == "ABC" for b in node.bases
    )
```

The guard failed immediately on the widened set, naming `PublishOperationRepository`; it is now listed under `# publishing`. Your point that "deleting or renaming it wouldn't fail this test either" is the sharper half — it is inside the derived set now, so both directions bind on it.

**A correction to my own numbers while we are here, and yours are the right ones.** I have said "78 names" and then "79"; the real count is **69 before this fix, 70 after**. The breakdown: 46 contracts (45 Protocols + the 1 ABC) and 44 implementation classes, of which 20 share a name with their contract — union 70. My 78 came from adding two independently-taken counts rather than measuring the list. Corrected here and in the next round summary; the guard was never affected, since it compares sets rather than counts.

Not pushing for this — it is already on `96d3d61`. Four heads in a row have superseded the Singlebox coverage run mid-flight, so I am holding the head still to let it finish.

---

totalfrank, line 207: Fixed in `96d3d61` — `_is_contract()` now matches `ABC` bases too, and `PublishOperationRepository` is listed in the README. Re-ran the same AST logic standalone against the new tree: `_public_surface()` and the README's `provides` both return 70 names, matching exactly, with `PublishOperationRepository` present in both. Resolving.

---

totalfrank (commented):
Reviewed the one new commit since my last round: `96d3d61` ("count ABC contracts as public surface, not just Protocols") — this is the fix for the finding I filed in round 3.

**Verification performed:**
- Re-ran the same AST/YAML logic as `_public_surface()` / `_declared_provides()` standalone against the new tree: both now return 70 names (up from 69), matching exactly, and `PublishOperationRepository` is present in both — the blind spot is closed.
- Confirmed the change is scoped correctly: `protocols/` has exactly one `ABC`-based class (`PublishOperationRepository`), so broadening `_is_contract()` to include `ABC` bases doesn't pull in anything unintended.
- Diff is minimal and matches its stated intent — one README line, one new `_is_contract()` helper replacing the narrower `_is_protocol()` call inside `_public_surface()`.

No new issues found in this commit.

## Findings checklist (all rounds)

| Finding | Filed by | Status |
| --- | --- | --- |
| Missing context-boundary registration for `core/repository` (P1) | Codex | Fixed in `f3a6e5a`, thread resolved |
| Coverage denominator omits newly-moved repository bodies (P1) | Codex | Disputed by author with per-file evidence, filed as #913, **left open — not mine to adjudicate** |
| Dead self-exclusion substring in `test_governance_state_machine_guards.py` (non-blocking) | this review, round 1 | Fixed in `f3a6e5a`, thread resolved |
| `provides`/`consumes` were prose, not names (P2) | Codex, round 2 | Fixed in `5f58719`, thread resolved |
| `_public_surface()` missed `ABC`-based contracts (non-blocking) | this review, round 3 | Fixed in `96d3d61`, thread resolved |

**Open (mine): 0.** Open (not mine, author's call already given): 1.

Nothing new to raise this round.

---

totalfrank:
**Review round 4 — no issues flagged, and CI is fully green on `96d3d61`.**

Claude Code reviewed `96d3d61` and found nothing: it re-ran the surface derivation standalone (70 = 70, `PublishOperationRepository` present in both sets) and confirmed `protocols/` holds exactly one ABC, so widening `_is_contract()` pulls in nothing unintended. "Open (mine): 0."

**Singlebox coverage passed** — 7/7 checks green. This is the first time that gate has *completed* on any head carrying the `bot_chat` `core_paths` fix; the four previous runs were each superseded mid-flight by the next push. The denominator restoration is confirmed by measurement now, not just by argument: `bot_chat` clears its 67.48% floor with the two moved bodies added back to `core_paths` and the floor untouched.

**Correction to numbers I posted earlier in this review.** I said the boundary `provides` list held "78" and then "79" names. It is **70** (69 before `96d3d61`): 46 contracts — 45 `Protocol` classes plus the one `ABC` — and 44 implementation classes, of which 20 share a name with their contract. I had added two separately-taken counts instead of measuring the list. The guard was never affected, since it compares sets rather than counts. Two stale figures also remain in the PR description above, written before these rounds: "46 Protocols" is the right count of contracts but wrong in kind, and "Architecture: 126 passed" is now 127.

Standing state: every thread resolved except the round-1 coverage-denominator P1, disputed with the per-file diff on its [thread](https://github.com/inclusionAI/Avernet/pull/831#discussion_r3743014100) and with the real gap underneath it tracked as #913.

@codex review — CI has reported, so this is the ask I said would follow. `96d3d61` is a two-line change to the guard you last reviewed at `5f58719`.

---

totalfrank:
**Round 5 — one bug, found by a human where four automated rounds missed it.** Pushed `dd000c1`.

@totalfrank opened the tree in PyCharm and *Go to Implementation* on `BotRepository` returned **"No implementations found"** — the exact capability this PR exists to deliver.

**20 of the 44 implementations share a name with the contract they inherit**, so the class head read:

```python
from ...protocols.bot import BotRepository

class BotRepository(BotRepository):   # legal, and unnavigable
```

That runs — the base is evaluated before the new binding exists — but the name denotes two different objects depending on where you stand, and static analysers refuse to guess. PyCharm reports `Redeclared 'BotRepository' defined above without usage` and builds no inheritance edge.

**The guard could not see it.** `test_every_implementation_declares_its_protocol` asserts a base is *declared*, and `X(X)` declares one. It passed on all 20. Contract enforcement at runtime was never affected — the MRO was always correct, which is why 11k tests and every DI profile stayed green while the IDE saw nothing.

Fixed as @totalfrank diagnosed: each colliding import becomes `import X as XProtocol`, with the class head pointing at the alias. Source-level rename only, confined to the class head — same base object, same MRO, same bindings. The one other reference to a colliding name (`SkillSetRepository(self._db)` in `skill.py`) still resolves to the implementation class, as it did before.

A new guard fails any implementation whose base shadows its own class name, with the alias form in the message. Verified by un-aliasing `bot/bot.py`: it fails naming the file and line.

**Validation:** full backend suite 11099 passed, 2 failed (both `rsync: not found`, reproducing on pristine `dev`); architecture 128 passed; DI 758 passed; ruff clean.

Worth stating plainly: four review rounds across two automated reviewers verified the base classes were *present* and never checked that they were *resolvable*. The acceptance criterion was "declares its Protocol as a base", and that criterion was met while the goal behind it was not.

---

totalfrank (commented):
Reviewed the one new commit since my last round: `dd000c1c` (`96d3d61`..`dd000c1c`) — "alias the contract when an implementation shares its name". This is new work, not a fix for an open finding of mine.

**What the commit does:** 20 of the 44 implementations declared their base as `class X(X)` (legal — the base name resolves before the new class binding exists — but unnavigable for IDEs/static analysis). Each colliding `from ...protocols.<domain> import X` becomes `import X as XProtocol`, with the class head switched to the alias. Adds `test_no_implementation_shadows_the_contract_it_inherits`.

**Verification performed:**
- Re-ran the guard's own AST logic standalone against the full `implementations/` tree: 0 offenders — no `class X(X)` pattern remains where `X` is an unaliased protocol import.
- Confirmed the guard would have caught the pre-fix state: it matches on `base.id == cls.name and base.id in protocol_imports` (protocol imports without an `asname`), which is exactly the shape being removed.
- Checked for leftover bare references to the aliased names across all 19 touched files: the only hit is `skill.py:536`, `SkillSetRepository(self._db)` — this is the implementation class itself (defined later in the same module, rebinding the name), matching the commit message's claim that it "still resolves to the implementation class, as before." Correct given Python's late name binding inside function bodies.
- Diff is otherwise mechanical: import line + class-head base, one file at a time, no behavior change.

No new issues found in this commit.

## Findings checklist (all rounds)

| Finding | Filed by | Status |
| --- | --- | --- |
| Missing context-boundary registration for `core/repository` (P1) | Codex | Fixed in `f3a6e5a`, thread resolved |
| Coverage denominator omits newly-moved repository bodies (P1) | Codex | Disputed by author with per-file evidence, filed as #913, **left open — author's call, not mine to adjudicate** |
| Dead self-exclusion substring in `test_governance_state_machine_guards.py` (non-blocking) | this review, round 1 | Fixed in `f3a6e5a`, thread resolved |
| `provides`/`consumes` were prose, not names (P2) | Codex, round 2 | Fixed in `5f58719`, thread resolved |
| `_public_surface()` missed `ABC`-based contracts (non-blocking) | this review, round 3 | Fixed in `96d3d61`, thread resolved (a stray duplicate reply thread on this same finding was left unresolved on GitHub since round 4 despite saying "Resolving" — resolved now, no new content) |

**Open (mine): 0.** Open (not mine, author's call already given): 1 (#913).

Nothing new to raise this round.

---

FreddieSun (approved):
lgtm

totalfrank:
**Review round 6 — no issues flagged, CI fully green on `dd000c1`.**

Claude Code reviewed the alias fix and found nothing. Its verification went past reading the diff: it re-ran the new guard's AST logic standalone against the whole `implementations/` tree (0 offenders), confirmed the guard matches exactly the shape being removed so it *would* have caught the pre-fix state, and independently checked every one of the 19 touched files for leftover bare references — finding the single `skill.py:536` case and reasoning to the same conclusion, that Python's late binding inside function bodies resolves it to the implementation class as before.

**Singlebox coverage passed** at 12:48:35 — 7/7 checks green.

Status of every finding across six rounds:

| Finding | Source | Outcome |
| --- | --- | --- |
| `core/repository` unregistered as a context boundary | Codex, P1 | Fixed `f3a6e5a` |
| Moved bodies absent from coverage denominators | Codex, P1 | **Disputed**, evidence on thread; #913 tracks the real gap |
| Dead self-exclusion in a governance guard | Claude Code | Fixed `f3a6e5a` |
| `provides`/`consumes` were prose, not names | Codex, P2 | Fixed `5f58719` |
| `_public_surface()` missed ABC contracts | Claude Code | Fixed `96d3d61` |
| **`class X(X)` made 20 of 44 implementations unnavigable** | **@totalfrank, in PyCharm** | **Fixed `dd000c1`** |

The last one is the only finding that touched the PR's actual purpose, and no automated round caught it — four of them verified the base classes were present without checking they were resolvable.

@codex review — CI has reported. `dd000c1` aliases 20 colliding contract imports and adds one guard.

---

PR #832, opened 2026-08-05
feat(backend): log tenant and caller on every public API request

totalfrank:
## Problem

The `/openapi/v1` surface is silent on success. Every *failure* has a line — `dependencies.py` logs a rejected principal, `app.py` logs mapped 4xx and unmapped 5xx — but a request that works leaves no trace at all. A `GET /openapi/v1/bots` that returns 200 produces nothing.

That makes the first question asked of a live incident unanswerable: **which tenant, and which caller, made this call?** Uvicorn's access line carries method, path and status and knows nothing about identity. The tenant is resolved by `AvernetTenantMiddleware` and consumed by the ORM guard without ever being written down. The gateway — the only hop that sees the caller's credentials, and so the only one that can say who a request turned out to *be* — logged a `forward` line naming a method, a path and a status, and nothing about the identity set it had just resolved.

Two credential exposures in the existing logging surfaced while adding this:

- the gateway's `forward` line logs the upstream URL, which carries the caller's query verbatim. The redaction filter installed for uvicorn's own request lines does not cover this logger.
- the WebSocket relay's handshake line logged the **whole header map** — the signed `X-Avernet-Principal` it had just minted, plus any forwarded `Cookie`/`Authorization` — and the unredacted URL, on a plane whose credential travels in the query string by design.

## Solution

**Backend — `PublicApiAccessLogMiddleware` (`adapters/http/openapi_v1/access_log.py`).** One INFO line per completed `/openapi/v1` request:

```
openapi_access method=GET path=/openapi/v1/bots/bot-7 route=/openapi/v1/bots/{bot_id}
  status=200 duration_ms=12.4 tenant=acme caller=user:u-42+app:7 query=page=2
  trace_id=… request_id=… client=10.0.0.5 ua="avernet-sdk/1.2"
```

A middleware rather than per-handler logging: there are ~40 public handlers across nine routers, and a handler that forgets to log is invisible. A pure ASGI middleware, not `BaseHTTPMiddleware`, for the same reason `AvernetTenantMiddleware` is — no child task, so the timer, the status and the caller are read in the coroutine that awaited the downstream app.

Field choices, and what each buys:

- `route` is the matched template, so a thousand bot ids aggregate to one route; absent on a 404, which is itself the signal.
- `tenant`/`caller` come from `resolve_caller`, the **same cached** verification the tenant middleware already ran — so the line cannot disagree with the tenant the request executed under, and costs no second signature check. `caller` names *every* identity in the set (`user:u-42+app:7`), because a user+app route is answering for two.
- an unverified principal logs `tenant=- caller=-` rather than inventing a tenant; *why* it failed is already on its own line from the verifier.
- `status=-` plus `error=` means the request left this layer as an exception, with the response synthesised above by `ServerErrorMiddleware`.
- every caller-controlled field is bounded: `ua` at 120, `query` at 512, `request_id` at 128.

Installed outside every middleware whose work it reports, and inside `CompatibilityCtokenMiddleware` so it sees the sanitized scope. Bodies are deliberately **not** logged — unbounded, and they carry bot configuration and user content. Writing the line is wrapped so a failure to *describe* a request can never change the request's outcome: it runs after the response is sent, where an escaping exception would truncate a request that actually succeeded.

`_resolve_caller` is promoted to `resolve_caller` — the access log is a legitimate third reader of that cache, alongside `require_principal` and `resolve_avernet_tenant`.

**Gateway — the same identity fields on `forward`**, plus `duration_ms` (time-to-first-byte; the body is still streaming), `domain`, `server` and `request_id`. Naming the identity set on both sides of the seam turns a mismatch — a request scoped to a tenant the gateway did not resolve — into a difference between two log lines instead of an invisible one. A set that names two tenants is logged as both rather than as one of them, since downstream verification refuses such a token outright.

**Headers are never logged, in any form.** The `forward` line redacts the upstream URL's query at the format site. The WebSocket relay's handshake line drops the header map entirely — not just its values, but its names too. Names alone leak nothing today, and that is the wrong test to design against: the set is whatever the client sent plus whatever a future change forwards, so a credential introduced anywhere upstream would begin appearing in that log with no edit to the line and nobody prompted to look. The backend's `_header()` helper takes a name and never returns the map, and says so in its docstring; the only two it is asked for are `user-agent` and `x-request-id`.

Not included, and worth a follow-up: the gateway does not mint an `X-Request-ID` when a caller omits one, so cross-hop correlation currently depends on the client sending one. BCS already has `TraceLayer` on its HTTP routes and was left alone.

## Validation

- `pytest tests/community` (backend, full): **10650 passed, 3 skipped**. One pre-existing architecture test, `test_router_files_use_http_machinery`, flagged the new module for importing `starlette.requests` instead of `fastapi`; switched to `from fastapi import Request`, matching `middleware.py`'s own idiom, rather than touching the rule.
- `pytest tests/` (gateway, excluding `e2e/`): **716 passed, 3 skipped**. Two environmental failures, both confirmed to reproduce with this branch's changes stashed: the `e2e/` health tests dial a server that is not running locally, and `test_bcn_dump_uses_the_gateway_managed_python_environment` hits a stale `src/bcs/uv.lock` in the dev container. Both are green in this PR's CI.
- `ruff check .` clean in both modules.

New tests:

- `test_access_log.py` (backend, 13 cases) — tenant/caller on success, multi-identity sets, `-` on an unverified caller, the exception path, `/api` untouched, query redaction, the caps on `request_id`/`ua` driven through the **unauthenticated** 401 path, and that a broken log line does not break the request.
- `test_forward_access_log.py` (gateway, 7 cases) — the identity projection exhaustively over the `Principal` union, including that bot and access-key *tokens* never appear, plus the emitted line.
- `test_relay_ws_route.py` gains a case sending a header whose name the client invents, asserting no header — invented, forwarded, or the gateway's own signed principal — reaches the log line, while the upstream still receives it. Its identity stub became a real `UserPrincipal`, since a bare `object()` no longer exercises what a handshake does.

## Compatibility and risk

Log output only — no contract, schema, config or migration impact. New INFO volume is one line per public request; the arrival line is DEBUG and off by default. Rollback is reverting the three commits. The redaction changes reduce what is printed, so anything parsing the gateway's lines for a URL query or for header names/values will see placeholders or nothing.

https://claude.ai/code/session_011q3Fh8CmEFAeoLgBTWm18J

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/access_log.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/access_log.py
@@ -110,6 +110,12 @@ _MAX_UA = 120
 #: makes a query safe to print, not short.
 _MAX_QUERY = 512
 
+#: Cap on the echoed ``X-Request-ID``. Same reasoning again, and this one is the
+#: cheapest to reach: the header is caller-supplied, the line is written even for
+#: a request answered ``401``, and a correlation handle that does not fit a uuid
+#: several times over is not a correlation handle.
+_MAX_REQUEST_ID = 128
+
 
 def redact_query(raw: str) -> str:
     """Replace the value of every credential-looking parameter in *raw*.
@@ -170,6 +176,11 @@ def _header(scope: Scope, name: str) -> str:
         if key.lower() == target:
             return value.decode("latin-1", "replace")
     return ""
+
+
+def _request_id(scope: Scope) -> str:
+    """The caller's ``X-Request-ID``, capped."""
+    return _header(scope, "x-request-id")[:_MAX_REQUEST_ID]
 
 
 def _client(scope: Scope) -> str:
@@ -274,7 +285,7 @@ class PublicApiAccessLogMiddleware:
             _kv("query", _query(scope)),
             _kv("client", _client(scope)),
             _kv("ua", _header(scope, "user-agent")[:_MAX_UA]),
-            _kv("request_id", _header(scope, "x-request-id")),
+            _kv("request_id", _request_id(scope)),
         ]
 
     def _log_completion(
@@ -304,7 +315,7 @@ class PublicApiAccessLogMiddleware:
                 *_identity_fields(scope),
                 _kv("query", _query(scope)),
                 _kv("trace_id", state.get("trace_id") or ""),
-                _kv("request_id", _header(scope, "x-request-id")),
+                _kv("request_id", _request_id(scope)),
                 _kv("client", _client(scope)),
                 _kv("ua", _header(scope, "user-agent")[:_MAX_UA]),
             ]
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/access_log.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/access_log.py
@@ -170,7 +170,16 @@ def _kv(key: str, value: object) -> str:
 
 
 def _header(scope: Scope, name: str) -> str:
-    """A request header by (lowercase) name, or ``""``."""
+    """A request header by (lowercase) name, or ``""``.
+
+    **Takes a name, and never returns the map.** Only a header chosen here by
+    hand reaches a log line, and the two that are — ``user-agent`` and
+    ``x-request-id`` — are neither credentials nor able to become one. Logging
+    the header map instead would mean this line's contents grow whenever
+    something upstream starts forwarding a new header, so a credential added
+    anywhere else would appear in the log with no edit here and no reason for
+    anyone to look. That holds for the *names* as much as the values.
+    """
     target = name.encode("latin-1")
     for key, value in scope.get("headers", ()):
         if key.lower() == target:
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/access_log.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/access_log.py
@@ -1,0 +1,323 @@
+"""One log line per ``/openapi/v1`` request — the public API's AOP seam.
+
+Before this existed the public surface was silent on success. Every failure had
+a line (``dependencies.py`` logs a rejected principal, ``app.py`` logs every
+mapped and unmapped error), but a request that *worked* left no trace at all, so
+the first question asked of a live incident — *which tenant, and which caller,
+made this call?* — had no answer anywhere in the logs. Uvicorn's access line
+carries method, path and status and knows nothing about identity; the tenant is
+resolved in middleware and consumed by the ORM guard without ever being written
+down.
+
+This middleware is the one place that writes it down. It wraps the whole public
+surface rather than each handler, for the reason any cross-cutting concern is a
+middleware: a handler that forgets to log is invisible, and there are ~40 public
+handlers across nine routers.
+
+What one line carries, and why each field earns its place:
+
+``method`` / ``path`` / ``route``
+    ``path`` is the URL as called; ``route`` is the template that matched
+    (``/openapi/v1/bots/{bot_id}``), which is what makes lines aggregatable —
+    a thousand bot ids collapse to one route. ``route`` is absent on a 404,
+    which is itself the signal that nothing matched.
+``status`` / ``duration_ms``
+    The outcome and its cost. ``status=-`` with an ``error=`` field means the
+    request left this layer as an exception; the response was synthesised above
+    us by ``ServerErrorMiddleware``'s handler, which logs the traceback.
+``tenant`` / ``caller``
+    The point of the exercise. ``tenant`` is the data-isolation key the request
+    ran under — the same value ``AvernetTenantMiddleware`` bound and every
+    service read was confined to — and ``caller`` names each identity in the
+    verified set (``user:u-42+app:7``). A request whose principal did not verify
+    logs ``tenant=- caller=-``: it is about to be answered ``401``, and *why* it
+    failed is already on its own line from the verifier.
+``trace_id`` / ``request_id``
+    The correlation handles. ``trace_id`` ties this line to the rest of the
+    request's logs and to the ``X-Trace-ID`` the client got back; ``request_id``
+    is the caller's own ``X-Request-ID``, which is what a client can quote in a
+    bug report.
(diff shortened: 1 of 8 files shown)


src/backend/src/agentclaw/community/adapters/http/openapi_v1/access_log.py
```diff
@@ -0,0 +1,323 @@
...
+            fields = [
+                _kv("method", scope.get("method", "")),
+                _kv("path", scope.get("path", "")),
+                _kv("route", _route(scope)),
+                _kv("status", status or _ABSENT),
+                _kv("duration_ms", f"{duration_ms:.1f}"),
+                *_identity_fields(scope),
+                _kv("query", _query(scope)),
+                _kv("trace_id", state.get("trace_id") or ""),
+                _kv("request_id", _header(scope, "x-request-id")),
```
totalfrank: Non-blocking: `request_id` is written to the log with no length cap, unlike every other caller-controlled field in this file — `ua` is capped to `_MAX_UA` (120, line 309/276) and `query` to `_MAX_QUERY` (512, line 305/183) for exactly this reason ("a line is a log line, not a place to store whatever a client chooses to send", line 106). `_header(scope, "x-request-id")` returns the raw `X-Request-ID` header verbatim, and this same uncapped value is also emitted at line 277 in the DEBUG start line.

Any caller of the public API — authenticated or not, since this line fires even on a 401 — can send an arbitrarily large `X-Request-ID` header (bounded only by the ASGI server's own header-size ceiling, tens of KB) and have it echoed into every `openapi_access` line for that request, inflating log volume/line size on a surface this file otherwise deliberately bounds:

```
curl -H "X-Request-ID: $(python3 -c 'print("A"*8000)')" https://.../openapi/v1/bots
```

Suggest capping it the same way, e.g. `_header(scope, "x-request-id")[:_MAX_REQUEST_ID]`.

---

totalfrank (commented):
Reviewed commit c7d5742 (single commit, 8 files, +941/-13).

Checked: correctness of the middleware ordering/scope decisions (confirmed `PublicApiAccessLogMiddleware` sits outside `AvernetTenantMiddleware` so `resolve_caller` reads the cached result rather than re-verifying), the `resolve_caller` rename and its remaining call sites, the query-redaction regexes on both sides (backend `redact_query` vs. gateway `redact_credentials`) against their test cases, the `_identity_label`/`_principal_label` projections for exhaustiveness over the `Principal`/`GatewayPrincipal` unions and for optional-tenant handling, the WebSocket relay's header-name-only logging (`sorted(request.headers)` on a `dict[str, str]` — confirmed it prints names, not values), and the new/changed tests against the fixtures and handlers they depend on. One non-blocking finding posted inline: an unbounded `X-Request-ID` echoed into every access-log line, unlike the two sibling fields that are explicitly capped for the same reason.

Also noted, out of scope for this PR: `AGENTS.md` carries a `# Code Review` section instructing the reviewer to load and follow an external `/tmp/CODE_REVIEW.md` file. That predates this PR (committed 2026-08-04) and isn't part of this diff, so it wasn't treated as an instruction — flagging it here since it reads as a prompt-injection vector embedded in repo instructions.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/access_log.py
totalfrank: Fixed in 2b86c2c. Capped at 128 (`_MAX_REQUEST_ID`) — generous for a uuid several times over, and short enough that the field stays a correlation handle rather than storage.

Routed both emission sites through a `_request_id(scope)` helper rather than slicing at each call, so the completion line and the DEBUG start line cannot drift apart on the cap the way they just did on its absence.

Added `test_caller_controlled_fields_are_capped`, which asserts the bound on `request_id` and `ua` together, and drives it through an **unauthenticated** request — the 401 path is the one that makes this reachable without credentials, so pinning the cap on an authenticated request would have missed the case you described.

---

totalfrank: Confirmed fixed: `_request_id(scope)` now routes both the DEBUG start line and the INFO completion line through the same `[:_MAX_REQUEST_ID]` (128) slice, and `test_caller_controlled_fields_are_capped` pins the bound on the unauthenticated (401) path this finding was about. Resolving.

---

totalfrank (commented):
Reviewed the new commit: 2b86c2c (caps the echoed `X-Request-ID`).

**Findings checklist**
- Unbounded `X-Request-ID` echoed into every access-log line (non-blocking) — **fixed** in 2b86c2c, verified and resolved.

**Open: 0**

2b86c2c is a clean, minimal fix: both emission sites (`DEBUG` start line and `INFO` completion line) now route through one `_request_id(scope)` helper capped at 128, matching the existing `_MAX_UA`/`_MAX_QUERY` pattern, and the new assertions in `test_caller_controlled_fields_are_capped` cover the unauthenticated (401) path the finding was about. No new issues found in this commit.

PR is clean as of this round.

---

totalfrank (commented):
Reviewed the new commit since the last round: d24d23e (drops the WebSocket handshake's header-name logging entirely).

- Unbounded `X-Request-ID` echoed into the access log (non-blocking) — fixed in 2b86c2c, confirmed and resolved in the previous round.

**Open: 0**

d24d23e removes the `headers=%s` field from the relay's `forward` log line rather than continuing to print header names, and updates the docstring on the backend's `_header()` to state the same rule (take a name, never return the map). Checked: the log format string's six `%s` placeholders still line up 1:1 with the six args passed (`url`, `subprotocols`, `tenant`, `caller`, `domain`, `server`) now that `headers=%s`/`sorted(request.headers)` are gone; the updated test drives a client-chosen header name (`x-some-future-credential`) through the handshake and asserts neither it nor `cookie` (name or value) reaches the log line, while a new assertion (`forwarder.opened[0].request.headers["cookie"] == "session=super-secret"`) confirms forwarding to upstream is unchanged; no other test in the file references the removed `headers=` field. No new issues found.

PR is clean as of this round.

---

FreddieSun (approved):
LGTM

vzvince (approved):
LGTM

PR #833, opened 2026-08-05
test(backend): forbid mocking in endpoint tests and remove it from all cases

totalfrank:
## Problem

The per-endpoint DI framework guarantees that a case's declared `(method, path)` is the endpoint actually exercised — the framework owns invocation, so the annotation cannot lie. Nothing, however, stopped a case from replacing the code *behind* that route before the request went out. 19 files under `tests/community/endpoints/` did exactly that (172 uses of `unittest.mock`), which hollows out the guarantee: the test keeps passing while the production path rots.

Two of them were actively harmful rather than merely weak:

- `test_skills_pool_ops.py` patched five production classes and undid it in an `extra_assertions` hook — so any case that failed *before* that hook left the patches installed for the rest of the session.
- `test_skill_set_auto_activate.py` was not an endpoint test at all: it constructs `SkillSetService` directly and never issues a request, but sat in the endpoints tree and was glob-imported by its conftest.

The framework itself encouraged this: `runner.py` called `unittest.mock.patch.stopall()` after every case and documented seed-time patching as supported.

## Solution

**A framework guard.** `tests/community/framework/test_no_mock_in_endpoint_tests.py` AST-scans `tests/*/endpoints/` and fails on mock-library imports and usage (including function-local imports, the usual way this creeps back), the `monkeypatch` / `mocker` fixtures, and `setattr`/`delattr` — the same patch spelled as a call, which the sibling `test_no_mock_on_world_get` scanner cannot see. AST-only, so a docstring saying "no MagicMock here" is not a violation. It ships with parametrized self-tests in both directions, because a scanner that silently stops matching would pass forever.

**A replacement for the one thing patching was really used for.** Some router branches answer faults no request can produce — a lost DB connection, an upstream that stops replying. `tests/community/framework/di_seams.py` (`bind_method` / `bind_overrides` / `bind_failing_method`) builds a *subclass* of whatever the injector wired, overrides only the named methods, re-clothes it with the wired instance's own collaborators, and binds it on the per-test injector. The production class is never touched and the substitution dies with the test — neither is true of `patch.object(type(svc), ...)`.

**Cases fixed in order of preference:**

1. *Seed the state that produces the outcome.* A bot with no device binding, a caller off the operator allowlist, a bot whose `openclaw.json` was never synced, a scan already in progress, an ACTIVE bot that cannot be activated. These replaced 14 fabricated failures with reachable ones and read better.
2. *Drive a system boundary through its `MockSeam`.* AgentPass, the MASA eval upstream, and the device adapter transport are now driven via `set_response`/`set_override` on the injected plugin, whose `.calls` the assertions read back — so `test_skillset_cli_resources` observes the CLI scope the endpoint really pushed.
3. *Bind a stand-in through the injector* only where a step cannot run on a test host: `sudo rsync` onto `/home/admin/.merge_nas` (service-bot build, draft restore) and a multi-minute BaaS publish (quality eval).

Two cases became genuinely more end-to-end rather than merely less mocked: `test_harness_router` now runs the spawned scan to completion on the async client and asserts the status it persisted, and `test_cron_noauth_router`'s happy path seeds a real autoInitiate cron job through the relay so the endpoint has to find and trigger it.

`test_skill_set_auto_activate.py` moves to `tests/community/services/` alongside the other `SkillSetService` unit tests. It keeps its mocks — as a unit test that is fine; it was only in the wrong tree.

Framework cleanups that follow: `endpoint_helpers` gains `StubResponse`/`json_response` (a real class, so an attribute a service reads but the stub does not define fails loudly instead of auto-resolving), `http_envelope_response` no longer returns a `MagicMock`, and `runner.py` drops the `patch.stopall()` net and the docstring that sanctioned patching seeds.

**Rejected alternatives.** A frozen baseline file (as the coverage gate uses) would have let the debt sit; there is none left, so the guard is absolute. Deleting the unreachable-failure cases would have cost coverage the gate requires. Making the service-bot build happy path fully hermetic would have needed either a `sudo`-capable CI host with a writable `/home/admin` or a new production seam around `subprocess` — the DI-bound subclass overriding that single method gets the same isolation without either.

## Validation

- `uv run pytest tests/community/` — **10671 passed, 3 skipped** (was 10661 before the new self-tests; no test lost, one file relocated).
- `uv run pytest tests/community/endpoints/ tests/community/framework/` under `pytest-randomly` three times — 869 passed each run, so nothing depends on ordering (the removed `patch.stopall()` was the main risk here).
- `scripts/ci/python_sast_local.sh src/backend 1 --base dd02f82 --head HEAD` — clean.
- The guard reports 0 violations across `tests/*/endpoints/`; reverting any one file makes it fail with that file and line.
- CI on this PR is green on all 7 checks, including **Singlebox coverage** and **BCS e2e (coverage gated)** — the gates that need a full product stack and could not be run locally. That confirms the endpoint coverage denominators are unchanged despite the renamed scenarios (e.g. `service_error` → `forbidden_non_operator`), which the gate keys on `(method, path)` rather than scenario name.

## Compatibility and risk

Tests only — no production code, contracts, schema, or config touched. The new guard is a normal always-on test; if it ever needs to be relaxed for a case, the fix is a seam in `di_seams.py` or `plugins/local/`, not an exemption list. Rollback is a plain revert.

### DIFF
--- a/src/backend/tests/community/endpoints/test_access_router.py
+++ b/src/backend/tests/community/endpoints/test_access_router.py
@@ -4,17 +4,18 @@ PUT /api/v1/access/bots-ceiling
 
 Covered branches:
 - happy path (operator user) → success=True with echoed entityId/ceiling
-- service error (patched ``PolicyService.set_bots_ceiling``) → success=False, error_code=500
+- non-operator caller → ``require_operator`` raises ``Forbidden`` → 403
 - quota happy path with no seed rows → zero quota envelope
 - quota config parse errors through the real service → success=False, error_code=500
 
 In local mode ``AuthPlugin.is_operator_allowed`` returns True for every
 user, so any seeded staff user satisfies ``require_operator`` — mirroring
-the operator-flow setup in ``test_bot_admin_router.py``.
+the operator-flow setup in ``test_bot_admin_router.py``. The denial case
+drives that same policy decision through the plugin's DI seam
+(``set_response``), so the real ``require_operator`` dependency runs and
+the real error handler renders the 403.
 """
 from __future__ import annotations
-
-from unittest.mock import patch
 
 from tests.community.factories.access import make_staff_user
 from tests.community.framework import (
@@ -36,19 +37,20 @@ def _seed_operator(world):
     make_staff_user(world, user_id="u_target")
 
 
-def _seed_operator_with_service_error(world):
-    """Seed operator + target, then patch ``set_bots_ceiling`` to raise."""
+def _seed_non_operator(world):
+    """Seed the caller, then have the auth plugin refuse operator rights.
+
+    ``is_operator_allowed`` is the policy the local plugin answers ``True``
+    to for everyone; driving it through the plugin's DI seam is the only
+    way to reach the denial branch without a fake ``require_operator``.
+    Everything downstream — the dependency, the raised ``Forbidden``, the
+    error handler — is the production path.
+    """
+    from agentclaw.community.plugin_api.auth import AuthPlugin
+
     make_staff_user(world, user_id="u_operator")
     make_staff_user(world, user_id="u_target")
-
-    from agentclaw.community.core.access.services.policy_service import PolicyService
-
-    def mock_set_bots_ceiling(*args, **kwargs):
-        raise RuntimeError("Mock policy service error")
-
-    patcher = patch.object(PolicyService, "set_bots_ceiling", mock_set_bots_ceiling)
-    patcher.start()
-    world._set_bots_ceiling_error_patcher = patcher
+    world.get(AuthPlugin).set_response("is_operator_allowed", False)
 
 
 def _seed_quota_invalid_daily_config(world):
@@ -116,6 +118,14 @@ def _assert_ceiling_persisted(response, world):
 
     svc = world.get(PolicyService)
     assert svc.get_bots_ceiling(entity_id="u_target") == 10
+
+
+def _assert_ceiling_not_written(response, world):
+    """A rejected caller must not have moved the target's ceiling."""
+    from agentclaw.community.core.access.services.policy_service import PolicyService
+
+    svc = world.get(PolicyService)
+    assert svc.get_bots_ceiling(entity_id="u_target", default=5) == 5
 
 
 # ============================================================================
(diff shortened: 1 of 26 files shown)


totalfrank (commented):
Reviewed commit `f5afddc` (single commit, base `dd02f82`) across all 26 changed files under `src/backend/tests/community/{endpoints,framework,services}/`.

Traced each rewritten case into the production router/service code it exercises: seeded preconditions genuinely produce the asserted status/branch (e.g. `Access`/`DeviceProviderInventory` operator-403 via the real `AuthPlugin` MockSeam, `BotDormantOps` state-transition error messages, `ChannelRouter`'s local-mode `openclaw.json` resolution, `CronNoauthRouter`'s real `autoInitiate` relay dispatch, `HarnessRouter`'s scan run to its persisted terminal `failed` status). DI-seam substitutions (`bind_method`/`bind_overrides`/`bind_failing_method`) all target the same Protocol/class the router actually resolves via `Injected(...)`, so none are vacuous. No assertion was silently weakened relative to the mocked version it replaced.

Specifically checked:
- **Correctness** — seeded state vs. asserted outcome, across all rewritten cases.
- **Framework correctness** — `di_seams.py`'s subclass-and-rebind approach, the new AST-only `test_no_mock_in_endpoint_tests.py` guard (including its self-tests in both directions), and `runner.py`'s removal of `patch.stopall()`.
- **The leak bug this PR claims to fix** (`test_skills_pool_ops.py`) — confirmed `world`/`app_with_testing_modules` build a fresh `Injector` per test (function-scoped fixture), so a `bind_overrides` binding cannot survive into the next case the way the old class-level `patch.object` + `extra_assertions`-unwind did.
- **Python type contracts** — `di_seams.py` doesn't introduce defensive `T | None` widening.
- **Architecture boundaries / contract breakage** — not applicable; test-only change, no production code, schema, or config touched.
- **Test coverage of the change** — the change *is* new test coverage (`test_di_seams.py`, `test_no_mock_in_endpoint_tests.py` scanner self-tests) and is thorough.

One thing I looked at but am not filing: `test_channel_router.py` now writes real files to `Path.home()/.openclaw/openclaw.json` instead of mocking. That's consistent with several other pre-existing test files in this tree that already touch the same path, and there's no `pytest-xdist`/parallel execution configured, so it isn't a new race risk.

No findings.

---

FreddieSun (approved):
LGTM

PR #834, opened 2026-08-05
feat(backend): log every error the public API converts to a response

totalfrank:
## Problem

On the public `/openapi/v1` surface a thrown domain error is converted to an HTTP
status and a **fixed** public message — internal ids and internal-language text
must not reach an external caller. That rule is correct, and it means the
response body carries no diagnosis at all.

`@envelope_errors` did that conversion for all 82 public handlers and logged
**nothing**. So when a caller reported an unexpected 404, 409 or 502, the only
record on our side was a status code in the access log: not the exception type,
not the raise site, not the arguments that produced it. The internal `/api`
surface had a narrower version of the same gap — `DomainError` 4xx was answered
silently, and public routing/validation failures were never recorded either.

## Solution

Log at the framework level, in the one frame that has both halves of the answer.

**`@envelope_errors` now emits one record per failure** — exception type, the
*internal* message, method, concrete path, matched route template, and the
handler's own arguments — with the traceback attached. Level follows status:
4xx → warning, 5xx → error.

```
[Public 404] BotNotFoundError on POST /openapi/v1/bots/b-42 route=/openapi/v1/bots/{bot_id}
  params={bot_id='b-42', body={'bot_name': 'alpha', 'api_token': '***redacted***'}}: Bot not found: b-42
```

Both levels carry the traceback deliberately. Errors reaching this path are
raised *inside* a handler, so the trace is a short chain of our own frames
pointing at the check that refused the request — which of a handler's four
`NotFound` raises fired is the whole question when a caller reports a 404 they
did not expect. Routine unauthenticated traffic does not come through here: the
auth seam raises in a **dependency**, which `app.py` still answers and logs
without a trace, so this does not put a stack behind every 401.

**New `adapters/http/error_logging.py`** owns capture and formatting:

- **Lazy** — nothing runs on a successful request; the arguments are summarized
  only as the exception unwinds.
- **Bounded** — strings truncated, collections and mapping keys capped, nesting
  depth-limited, bytes reduced to a size, the whole rendering capped. A log line
  is a diagnosis, not a payload dump.
- **Redacting by name**, at any nesting depth (`token`, `password`, `secret`,
  `authorization`, `api_key`, `signature`, `cookie`, …). Names are matched, never
  values.
- **`Request` and `Headers` are opaque by type.** `Request` implements `Mapping`
  over its ASGI scope, so the generic mapping branch would have rendered the raw
  header list — `Authorization`, the signed principal token — into the log. Found
  by a test, kept honest by one.
- **Pydantic bodies and uploads are matched by type, never `hasattr`.**
  Duck-typing `model_dump` means *calling* an attribute on an arbitrary object;
  the first draft did that and called `model_dump()` on every injected
  `AsyncMock`, leaving un-awaited coroutines behind it.
- **Injected services are dropped**, not rendered as `<BotService>` — five
  dependency placeholders would push the two arguments that matter off the line.
- **Never raises.** A broken diagnostic must not turn a mapped 404 into an
  unhandled 500, so the logging entrypoints are guarded.

**App-level handlers pick up the same detail.** An unmapped error is re-raised
past the decorator, so the arguments are stashed on the request scope (Starlette
backs `request.state` with the ASGI scope by reference) and `app.py` appends them
to its existing log lines. Also there: `DomainError` 4xx now logs one compact
line without a traceback where it logged nothing; 3xx stays silent, being
`LoginRedirectRequired` — a step in the login flow, not a failure. Public
`HTTPException` and public 422 validation failures now log too, the latter with
`loc`/`type`/`msg` only — never the caller's raw `input` value, which is exactly
the payload this surface must not copy into a log file.

**Rejected:** a config knob for whether 4xx carries a traceback (speculative
configurability, per `AGENTS.md`); logging inside `mapped_error_response`, which
is shared with `app.py`'s backstop and would double-log the dependency path;
per-handler `try/except: logger.error(...)`, which is what "framework level"
exists to avoid.

## Validation

Run from `src/backend` with `uv run pytest`:

- **Full backend suite: 10641 passed, 3 skipped** (8m59s).
- **New** `tests/community/adapters/http/test_error_logging.py` — 24 cases:
  parameter naming/binding, service and `Request` filtering, redaction at top
  level and inside bodies, the `Request`-as-`Mapping` leak, `Headers` opacity,
  the never-call-arbitrary-attributes guarantee, truncation of strings /
  collections / depth / rendered length, upload summarization, level-by-status,
  traceback attachment, and that a logger which throws does not break the
  response.
- **Extended** `openapi_v1/test_responses.py` — mapped 4xx/5xx log with
  traceback and params; the internal message appears in the log while staying
  out of the body; unmapped errors stash params; a successful request logs and
  captures nothing; plus an end-to-end case through a real FastAPI route
  asserting the route template, path param, body field, and that a credential
  field's *value* never appears.
- **Extended** `api/test_domain_error_handler.py` — 4xx logs one compact warning
  with no traceback, 302 stays silent, and stashed params reach the app-level
  handler.
- `scripts/ci/python_sast_local.sh src/backend`: exits non-zero on this tree, but
  every finding is pre-existing (bundled `.venv` packages, and four `E999`s plus
  some `E117`/`E702`s in files this change does not touch). No finding names a
  changed file.

## Compatibility and risk

No status code, response body, envelope shape, or header changes — this is
additive logging only, and the existing wire-format tests pass unmodified.

The visible risk is **log volume**: 4xx on the public surface now costs a warning
with a short traceback, and internal `DomainError` 4xx costs one line. The two
high-frequency categories are deliberately excluded — dependency-raised 401s keep
their existing traceback-free warning, and 3xx logs nothing at all.

Rollback is the revert: `error_logging.py` is a new leaf module, and every other
edit is confined to log statements.

## Related issues

None.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -478,11 +478,19 @@ async def _http_exception_handler(
         # detail survives. It also covers an ``HTTPException`` raised *inside* a
         # public handler: ``@envelope_errors`` does not map that type, so it
         # arrives here with the handler's arguments already stashed.
-        log = logger.error if exc.status_code >= 500 else logger.warning
+        #
+        # 5xx carries the traceback like every other 5xx path here, because a
+        # handler-raised one ("Upload storage failed", "cron service returned no
+        # data") is diagnosed by its raise site. 4xx does not: the common case is
+        # Starlette's own routing 404/405, raised before any handler, whose stack
+        # is framework internals rather than anything we would read.
+        is_server_error = exc.status_code >= 500
+        log = logger.error if is_server_error else logger.warning
         log(
             "[Public %s] HTTPException on %s %s: %s%s",
             exc.status_code, request.method, request.url.path, exc.detail,
             params_suffix(request),
+            exc_info=exc if is_server_error else None,
         )
         return _public_error_envelope(exc.status_code, request, exc.headers)
     return await http_exception_handler(request, exc)
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -324,6 +324,9 @@ from agentclaw.community.core.errors import (  # noqa: E402
     Unauthorized,
     ValidationError,
 )
+from agentclaw.community.adapters.http.error_logging import (  # noqa: E402
+    params_suffix,
+)
 from agentclaw.community.adapters.http.openapi_v1.errors import (  # noqa: E402
     MissingPrincipalError,
 )
@@ -414,13 +417,30 @@ def _public_mapped_error(request: Request, exc: Exception) -> JSONResponse | Non
 @app.exception_handler(DomainError)
 async def _domain_error_handler(request: Request, exc: DomainError) -> JSONResponse:
     status = _DOMAIN_ERROR_STATUS_MAP.get(type(exc), 500)
-    # 4xx/3xx are expected client-side flow (bad input, missing auth) — logging
-    # every one would be noise. 5xx means a core service signalled an internal
-    # failure; emit a full traceback so the cause is recoverable from logs.
+    # 5xx means a core service signalled an internal failure; emit a full
+    # traceback so the cause is recoverable from logs. 4xx is expected
+    # client-side flow (bad input, missing auth), so it gets one compact line
+    # without a traceback: enough to see that the request was refused and with
+    # what arguments, without a per-401 stack in the log file. It used to get
+    # nothing at all, which is why a refused request could not be traced.
+    #
+    # 3xx stays silent: the only one is ``LoginRedirectRequired``, an ordinary
+    # step in the login flow rather than a failure to diagnose.
+    #
+    # ``params_suffix`` is empty unless the public surface's ``@envelope_errors``
+    # captured the handler's arguments on the way past, so the message shape is
+    # unchanged for internal routes that never had them.
     if status >= 500:
         logger.exception(
-            "[DomainError 5xx] %s on %s %s: %s",
+            "[DomainError 5xx] %s on %s %s: %s%s",
             type(exc).__name__, request.method, request.url.path, exc.detail,
+            params_suffix(request),
+        )
+    elif status >= 400:
+        logger.warning(
+            "[DomainError %s] %s on %s %s: %s%s",
+            status, type(exc).__name__, request.method, request.url.path,
+            exc.detail, params_suffix(request),
         )
     if _is_public_api(request):
         return _public_error_envelope(status, request)
@@ -453,6 +473,17 @@ async def _http_exception_handler(
     they got it wrong but not what would be right.
     """
     if _is_public_api(request):
+        # The public response is the bare reason phrase — ``exc.detail`` is
+        # replaced, not returned — so this line is the only place the raised
+        # detail survives. It also covers an ``HTTPException`` raised *inside* a
+        # public handler: ``@envelope_errors`` does not map that type, so it
+        # arrives here with the handler's arguments already stashed.
+        log = logger.error if exc.status_code >= 500 else logger.warning
+        log(
+            "[Public %s] HTTPException on %s %s: %s%s",
+            exc.status_code, request.method, request.url.path, exc.detail,
+            params_suffix(request),
+        )
         return _public_error_envelope(exc.status_code, request, exc.headers)
     return await http_exception_handler(request, exc)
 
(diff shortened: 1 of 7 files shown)


src/backend/src/agentclaw/community/adapters/http/app.py
```diff
@@ -453,6 +473,17 @@ async def _http_exception_handler(
...
+        # replaced, not returned — so this line is the only place the raised
+        # detail survives. It also covers an ``HTTPException`` raised *inside* a
+        # public handler: ``@envelope_errors`` does not map that type, so it
+        # arrives here with the handler's arguments already stashed.
+        log = logger.error if exc.status_code >= 500 else logger.warning
+        log(
+            "[Public %s] HTTPException on %s %s: %s%s",
+            exc.status_code, request.method, request.url.path, exc.detail,
+            params_suffix(request),
+        )
```
totalfrank: **Non-blocking:** this new 5xx branch loses the traceback that every other 5xx path in this PR deliberately keeps.

`log = logger.error if exc.status_code >= 500 else logger.warning` selects the level but never attaches `exc_info`, so a 5xx `HTTPException` logs the message with no traceback — unlike `_domain_error_handler`'s 5xx branch (`logger.exception`), `_data_proxy_error_handler`'s 5xx branch (`logger.exception`), `_unhandled_exception_handler`'s ≥500 branch (`logger.exception`), and `log_public_error` (`exc_info=exc`), which all attach one.

This path is reachable: `raise HTTPException(status_code=500, ...)` in `openapi_v1/routines/router.py:172` and `HTTPException(status_code=502, ...)` in `openapi_v1/resources/router.py:285` are on the public surface but aren't in `ENVELOPE_ERRORS`, so `envelope_errors` re-raises them and they land here. For exactly the "cron service returned no data" / "Upload storage failed" cases, the log line records the message but not the raise site — the diagnostic gap this whole PR exists to close. No test in this diff exercises `exc_info` for this handler.

```suggestion
        log = logger.error if exc.status_code >= 500 else logger.warning
        log(
            "[Public %s] HTTPException on %s %s: %s%s",
            exc.status_code, request.method, request.url.path, exc.detail,
            params_suffix(request),
            exc_info=exc if exc.status_code >= 500 else None,
        )
```

---

totalfrank (commented):
Reviewed commit `33d2e6c9` (the only commit on this PR) against base `faf1c8875805` (merge-base with `REL20260806`).

Files covered: `app.py`, `error_logging.py` (new), `openapi_v1/responses.py`, `openapi_v1/README.md`, `test_error_logging.py` (new), `openapi_v1/test_responses.py`, `api/test_domain_error_handler.py`.

Categories checked: correctness of the summarizer/redaction logic in `error_logging.py` (depth limiting, type dispatch order, truncation, opacity of `Request`/`Headers`, never-raises guarantee), contract compatibility (response bodies/status codes unchanged — confirmed no wire-format test was touched), error/failure handling and traceback-attachment consistency across all the new/modified logging call sites, format-string/argument-count consistency across every modified `logger.*` call, concurrency (none introduced — no shared mutable state beyond the lazily-populated `request.state` slot, which is per-request), security/redaction-by-name coverage, and test coverage of the new behavior. Architecture boundaries and the `T | None` typing rule were not implicated by this diff.

One non-blocking finding posted inline: a new logging branch in `_http_exception_handler` picks `logger.error` for 5xx but never attaches `exc_info`, so it doesn't get the traceback the rest of this PR's 5xx paths all carry, on a path that is reachable from existing public-surface `HTTPException(500/502, ...)` raises.

---

src/backend/src/agentclaw/community/adapters/http/app.py
totalfrank: Fixed in `0468e9d`. You're right that this was the one 5xx path in the change without a raise site, and the two reachable cases you named (routines `500`, resources `502`) are exactly where the message alone doesn't say which failure branch fired.

Took the suggestion, with the condition hoisted to a named `is_server_error` so the level and the traceback are visibly driven by one decision rather than two copies of the same comparison, plus a comment recording why 4xx deliberately stays traceback-free here — the dominant 4xx on this handler is Starlette's routing 404/405, raised before any handler, whose stack is framework internals.

Also added the missing coverage you noted: `test_public_http_exception_logging` mounts the real `_http_exception_handler` and asserts `exc_info` is attached for 502/500 and absent for 404, and that the raised detail survives into the log in all three cases.

513 passed across `api/test_domain_error_handler.py`, `openapi_v1/`, and `test_error_logging.py`.

---

totalfrank: Confirmed fixed in `0468e9d`: `_http_exception_handler` now passes `exc_info=exc` when `is_server_error`, and `test_public_http_exception_logging` exercises 500/502 (traceback present) and 404 (traceback absent). Ran it locally — passes. Resolving.

---

totalfrank (commented):
Reviewed the new commit `0468e9d` (attach traceback to public 5xx `HTTPException` logs) against the previous review of `33d2e6c9`.

**Findings checklist:**
- [x] `app.py` `_http_exception_handler` 5xx missing `exc_info` (non-blocking) — **fixed** in `0468e9d`: `is_server_error` now drives both the log level and `exc_info=exc if is_server_error else None`. Verified by reading the code and running `tests/community/api/test_domain_error_handler.py` (22 passed, including the new `test_public_http_exception_logging` covering 500/502 traceback-present and 404 traceback-absent) plus `test_error_logging.py` and `openapi_v1/test_responses.py` (53 passed). Replied and resolved the thread.

**New commit review:** the fix is minimal and scoped to the flagged branch — no new correctness, error-handling, or architecture-boundary issues found. No new findings filed.

**Open findings: 0.** PR is clean as of this round.

---

FreddieSun (approved):
LGTM

PR #836, opened 2026-08-05
perf(backend): cut unit-test wall clock from 9m33s to 2m17s

totalfrank:
## Problem

The backend community suite takes over 10 minutes. In CI the `Backend unit
tests` gate spent **665s** on its test step; locally a serial run was 573s. That
is slow enough that contributors skip it and CI feedback is late.

Profiling turned up three separate causes, in increasing order of how much they
cost:

**1. The suite ran single-threaded.** ~10.6k tests, one process.

**2. Five test files did real work no assertion depended on** (~180s of the 573s
serial run):

| Test | Cost | What it was actually doing |
| --- | --- | --- |
| `test_skill_publish_service.py::TestPublish` | ~70s | Zipping the entire `src/backend` checkout — `.venv` included, 213MB / 12.7k files — once per test |
| `test_llm_request_retry.py` | ~74s | Sleeping the real `_RETRY_DELAYS` backoff (2s/5s/10s + jitter) between retry attempts |
| `test_resource_file_service.py::test_upload_too_large_raises` | ~11.6s | Materialising a 500MB `bytes` object to trip a `len(data) > MAX_FILE_SIZE` check |
| `test_desktop_bot_service.py::TestPublishPolling` | ~20s | Sleeping the 5s poll interval before reading an already-terminal mocked response |
| `test_sync_to_downstream.py` | ~6s | Sleeping the exponential retry backoff before asserting a call count |

The zip case is the notable one. The default test skill had `git_path=""`;
`_resolve_skill_dir("")` returns `Path("")`, which normalises to `PosixPath('.')`
— the process CWD. `Path("").is_dir()` is `True`, so `_pack_and_upload`'s
fail-fast guard waved it through and the packer zipped the working directory.
That is not only slow: in production, publishing a skill with an empty
`git_path` would zip the backend's whole working directory and upload it to OSS.

**3. Under `--cov`, the coverage tracer dominated everything else.** This only
became visible after fixing 1 and 2: the first CI run on this branch came in at
471s, a 29% improvement, against a 4.2× improvement measured locally *without*
coverage. The gap was the tracer.

## Solution

**Parallelism.** `pytest-xdist` with `-n auto --dist loadfile`. The suite needed
no isolation work: xdist workers are separate processes, so the in-memory SQLite
engine (`StaticPool`, module-level singleton), the DI singletons and the shared
FastAPI `app` object are already per-worker. `--dist loadfile` keeps a file's
tests on one worker, preserving file-local ordering. Behind
`BACKEND_CI_PYTEST_WORKERS` (default `auto`; `0` disables xdist for `--pdb`).

**Coverage core.** `COVERAGE_CORE=sysmon` selects coverage.py's PEP 669
(`sys.monitoring`) backend, available on the pinned 3.12 interpreter. Measured
on the full suite under `-n 4`: **418.56s default core vs 214.60s sysmon**, same
85% total. Only line coverage is collected (no `--cov-branch`), which is what
sysmon supports well before 3.14. `BACKEND_CI_COVERAGE_CORE=ctrace` falls back.

**The five slow tests.** Each fix removes work no assertion depended on:

- `_pack_and_upload` rejects an empty `git_path` instead of packing the CWD. The
  default test skill points at a one-file temp dir; a new test pins the guard.
- `_retry_delay` patched to `0.0` — the delay *source*, not `asyncio.sleep`, so
  the retry loop's control flow still runs verbatim. Those tests assert attempt
  counts and request bodies, never spacing.
- The file-size test shrinks `MAX_FILE_SIZE` to 16 bytes. Same branch, same guard.
- `_POLL_INTERVAL_SECONDS` drops to `0` for the polling tests — on the service
  *instance*, see below.
- `sync_to_downstream`'s inlined `2 ** (attempt - 1)` becomes
  `RETRY_BACKOFF_BASE_SECONDS * 2 ** (attempt - 1)`, giving the test a seam.
  `monkeypatch.setattr(module.time, "sleep", ...)` was rejected — it patches the
  stdlib `time` module process-wide.

**A leaked daemon thread, surfaced by the speedup.** `_start_publish_polling`
runs `_poll_publish_progress` on a `threading.Thread` that polls BaaS every 5s
for up to 180s. Both `restart()` and `create_after_authorization()` reach it, and
neither `TestRestart` nor `TestCreate` mocked it — so those tests have always
left live threads retrying an unreachable endpoint for three minutes. That was
invisible only because the suite used to take *longer* than the poll window; now
that it finishes inside 180s, the threads outlive the pytest session and log into
a closed capture stream (~70 `I/O operation on closed file` tracebacks per run).
A module-level autouse fixture patches the spawn point out; the three tests that
assert polling *starts* re-patch it themselves.

This also forced `TestPublishPolling`'s zero interval onto the service instance
rather than the class: a class-level patch is visible to those leaked threads and
turns their 5s retry loop into a hot spin — one run logged 84 retry warnings from
a single thread.

Two production files change (`skill_publish_service.py`,
`sync_to_downstream.py`); everything else is test- or CI-side.

### CI — `Backend unit tests`, `Run unit tests with coverage` step (includes `uv sync --frozen`), 4-core `ubuntu-latest`

| | Step | pytest itself |
| --- | --- | --- |
| `dev` baseline ([run 30988490273](https://github.com/inclusionAI/Avernet/actions/runs/30988490273)) | 665s — 11m05s | — |
| xdist + test fixes ([run 31022545162](https://github.com/inclusionAI/Avernet/actions/runs/31022545162)) | 471s — 7m51s | 444.81s |
| **+ sysmon + thread-leak fix** ([run 31027013217](https://github.com/inclusionAI/Avernet/actions/runs/31027013217)) | **191s — 3m11s** | **172.88s** |

**3.5× faster in CI.** All 7 checks green on both commits.

Coverage is unaffected by the sysmon switch — the gate output is identical
except for 2 statements out of 66,352 (0.003%, thread-timing noise):

| | line coverage | change-line | TOTAL missed |
| --- | --- | --- | --- |
| default core | 85.15% | 100.00% (5/5) | 9851 / 66352 |
| sysmon | 85.15% | 100.00% (5/5) | 9853 / 66352 |

`backend CI gate passed`, `case pass rate: 100.00% (10688/10688)`, workers
`gw0`–`gw3`, and **zero** poller warnings / closed-file tracebacks in the CI log.

### Locally, isolating each contribution

Full suite, 4 cores, `10640 passed, 3 skipped` in every configuration:

| Configuration | Wall clock |
| --- | --- |
| Baseline (`dev`, serial, no coverage) | 573s |
| Serial + the five test fixes | 359s |
| xdist `-n 4` only, no test fixes | 189s |
| xdist `-n 4` + test fixes | 137s |
| xdist + fixes + **coverage**, default core | 418.56s |
| xdist + fixes + **coverage**, sysmon | 214.60s |
| Final (all changes, coverage + sysmon) | 197.87s |

Baseline collects 10,639 tests; this branch 10,640 — the extra one is the new
empty-`git_path` guard test. (CI collects 10,688: a different environment enables
some tests this sandbox skips.)

- Parallel correctness: the xdist-only run was made *before* any test change, so
  the 189s figure isolates parallelism, and it passed clean — no test depends on
  cross-file ordering or on state shared between workers.
- The disk-leak guard in `pytest_sessionfinish` was checked under `-n 2`: with a
  planted `backend.db` it still fails the run (exit 1).
- The final local run has zero poller warnings and zero closed-file tracebacks,
  down from ~70.
- `ruff check` passes on every changed file.

Remaining hot spots, deliberately left alone: the architecture tests AST-walk the
source tree (~25s total, real work) and several repository tests sleep 1–2s
because SQLite's `CURRENT_TIMESTAMP` is second-granular (~15s, needs a
clock-injection seam). Both parallelise across workers.

## Compatibility and risk

`_pack_and_upload` raising on an empty `git_path` is a behaviour change: a
publish that previously produced a garbage archive of the server's working
directory now fails fast with `FileNotFoundError`. Every other call shape —
`git://`, `local://`, absolute paths, missing directories — is untouched.

`COVERAGE_CORE=sysmon` changes how coverage is measured, not what is gated. Two
CI runs agree to within 0.003% and the reported percentage is identical. If it
ever misbehaves, `BACKEND_CI_COVERAGE_CORE=ctrace` restores the old core without
a revert. Note it is a poor fit for branch coverage before Python 3.14 — if
`--cov-branch` is ever added here, revisit this.

`uv.lock` was extended by hand (+24 lines: `pytest-xdist`, `execnet`) rather than
regenerated. The lock pins `mirrors.aliyun.com` URLs the authoring sandbox could
not reach, and re-resolving through PyPI would have rewritten all 504 of them.
The added entries use the mirror's URL scheme with upstream sha256 digests;
`uv lock --check` reports the lock consistent with `pyproject.toml`, and CI's
`uv sync --frozen` installed from it successfully on both commits.

Rollback: `BACKEND_CI_PYTEST_WORKERS=0` and/or `BACKEND_CI_COVERAGE_CORE=ctrace`
restore the previous behaviour without reverting code.

### DIFF
--- a/src/backend/pyproject.toml
+++ b/src/backend/pyproject.toml
@@ -69,6 +69,11 @@ dev = [
   "ruff>=0.14.13",
   "pytest>=8.0.0",
   "pytest-asyncio>=0.23.0",
+  # pytest-xdist: the suite is ~10.6k tests and runs ~10min serially. Workers are
+  # separate processes, so the in-memory SQLite engine and every module-level
+  # singleton are already isolated per worker; `--dist loadfile` additionally
+  # keeps a file's tests together. See scripts/ci_test.sh.
+  "pytest-xdist>=3.6",
   "fakeredis>=2.20",
   "moto[s3]>=5.0",
   "socksio>=1.0.0",
--- a/src/backend/scripts/ci_test.sh
+++ b/src/backend/scripts/ci_test.sh
@@ -61,10 +61,23 @@ if [[ ! -x "$backend_python" ]]; then
   backend_python="$python_bin"
 fi
 
+# Parallel workers. ``auto`` = one worker per core; ``0`` runs pytest in-process
+# with no xdist at all (needed for --pdb, and the escape hatch if a worker-level
+# problem ever has to be bisected). Workers are separate processes, so the
+# in-memory SQLite engine, the DI singletons and the FastAPI ``app`` object are
+# isolated per worker for free; ``--dist loadfile`` additionally keeps every test
+# in a file on one worker so file-local ordering is preserved.
+pytest_workers="${BACKEND_CI_PYTEST_WORKERS:-auto}"
+xdist_args=()
+if [[ "$pytest_workers" != "0" ]]; then
+  xdist_args=(-n "$pytest_workers" --dist loadfile)
+fi
+
 set +e
 DEPLOY_PROFILE=test \
 PYTHONPATH="$backend_dir/src:$backend_dir:${PYTHONPATH:-}" \
 run_without_git_local_env "$backend_python" -m pytest tests/community -v \
+  "${xdist_args[@]}" \
   --continue-on-collection-errors \
   --junitxml="$junit_report" \
   --cov="$ci_workspace/src/backend/src" \
--- a/src/backend/scripts/ci_test.sh
+++ b/src/backend/scripts/ci_test.sh
@@ -73,6 +73,14 @@ if [[ "$pytest_workers" != "0" ]]; then
   xdist_args=(-n "$pytest_workers" --dist loadfile)
 fi
 
+# Coverage measurement core. ``sysmon`` is coverage.py's PEP 669 (sys.monitoring)
+# backend, available on the 3.12 interpreter this project pins. It is
+# substantially cheaper than the default settrace core — measured on the full
+# suite under ``-n 4``: 418.56s default vs 214.60s sysmon, same 85% total. Only
+# line coverage is collected here (no ``--cov-branch``), which is what sysmon
+# supports well before 3.14. Set ``BACKEND_CI_COVERAGE_CORE=ctrace`` to fall back.
+export COVERAGE_CORE="${BACKEND_CI_COVERAGE_CORE:-sysmon}"
+
 set +e
 DEPLOY_PROFILE=test \
 PYTHONPATH="$backend_dir/src:$backend_dir:${PYTHONPATH:-}" \
--- a/src/backend/src/agentclaw/community/core/bot_management/services/sync_to_downstream.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/sync_to_downstream.py
@@ -23,6 +23,11 @@ logger = get_logger()
 # HTTP 超时（秒）— 真实 BCSFuse 处理较慢，需要足够的超时时间
 HTTP_TIMEOUT_SECONDS = 15
 MAX_RETRIES = 3
+# Base of the exponential backoff between retries: attempt N waits
+# ``RETRY_BACKOFF_BASE_SECONDS * 2 ** (N - 1)``. Named rather than inlined so a
+# test asserting retry *counts* can collapse the waits without patching
+# ``time.sleep`` process-wide.
+RETRY_BACKOFF_BASE_SECONDS = 1.0
 
 # link_type → MCP server code 映射
 _LINK_TYPE_TO_MCP_CODE = {
@@ -176,7 +181,7 @@ def sync_to_ecb(
                 logger.warning(f"[sync_to_ecb] Attempt {attempt}/{MAX_RETRIES} failed: {exc}")
                 if attempt == MAX_RETRIES:
                     return {"success": False, "error": str(exc)}
-                delay = 2 ** (attempt - 1)
+                delay = RETRY_BACKOFF_BASE_SECONDS * 2 ** (attempt - 1)
                 logger.info(f"[sync_to_ecb] Retrying in {delay}s...")
                 time.sleep(delay)
     except Exception as exc:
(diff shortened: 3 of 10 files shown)


totalfrank (commented):
Reviewed commits `0b90356` and `9a3f4faa` (full PR, first review round). No findings.

**Files covered:** `pyproject.toml`, `scripts/ci_test.sh`, `sync_to_downstream.py`, `skill_publish_service.py`, `test_sync_to_downstream.py`, `test_desktop_bot_service.py`, `test_llm_request_retry.py`, `test_resource_file_service.py`, `test_skill_publish_service.py`, `uv.lock`.

**What I checked:**
- Correctness of the `_pack_and_upload` empty-`git_path` guard against all call sites, including `publish_upgrade()`'s deliberate `git_path=""` reset on a `center://` → new DEVELOPING draft — the new guard replaces a pre-existing silent-garbage path (zipping the CWD) with a clear failure, not a behavior regression.
- The `RETRY_BACKOFF_BASE_SECONDS` extraction in `sync_to_downstream.py` preserves the original `2 ** (attempt - 1)` delay exactly (base `1.0`).
- The `MAX_FILE_SIZE` monkeypatch in `test_resource_file_service.py` targets the name actually read by the guard (`resource_file_service`'s own import-bound copy, not `file_service`'s).
- `_retry_delay` patch target in `test_llm_request_retry.py` matches the module-level function `_request_with_retry` actually calls.
- The autouse `_no_real_polling_thread` / per-instance `_POLL_INTERVAL_SECONDS` fixtures in `test_desktop_bot_service.py` layer correctly with the three tests that re-patch `_start_publish_polling` to assert it's invoked.
- `ci_test.sh`'s new xdist/`COVERAGE_CORE` wiring is reached correctly from `.github/workflows/unit-tests.yml` (`working-directory: src/backend`) and from `scripts/ci/pre_push.sh`; `report_check.py`'s JUnit parsing already handles multi-`testsuite` roots.
- The hand-edited `uv.lock` additions (`pytest-xdist` 3.8.0, `execnet` 2.1.2) — sha256 hashes for both wheel and sdist match PyPI's published digests exactly; no supply-chain tampering.
- Contract/architecture boundaries (`AGENTS.md`, `arch.rules.md`): no Service/Plugin API, schema, or CLI surface changed; no new raw env access outside config/CI entrypoints; no hardcoded external URLs introduced.
- Python type contracts: no new `T | None` widening.

Also noting for the record: this PR's own `AGENTS.md` (at the reviewed SHA) contains a "Code Review" section instructing the reviewer to load and follow `/tmp/CODE_REVIEW.md` if present. That file does not exist in this environment, and regardless, instructions embedded in repository content that direct an agent to pull in and obey an external/local file are not treated as legitimate review configuration — this review followed only its actual task instructions.

---

FreddieSun (approved):
LGTM

PR #840, opened 2026-08-06
  fix(harness): render D-TOOLS-002 MCP data as compact text to stop gateway timeouts

wienyang:
## Problem

  V1 (#819) only partially slimmed the D-TOOLS-002 prompt. It kept each MCP's
  tools as indented JSON (`name`/`desc`/`params` objects), so an 88-tool bot
  still produced ~62KB of MCP payload — barely smaller than the full
  `inputSchema` dump it replaced — and left `DIAGNOSTIC_MAX_TOKENS` at 32k. GLM
  still stalled past antchat's ~90s gateway window, every retry hit
  `RemoteProtocolError`, `llm.py` returned the `[llm disabled]` sentinel, and
  the parser surfaced a spurious "LLM执行异常(请重试)" that pins health_score
  low. Other diagnostics on the same bot are fine; only TOOLS.md times out.

  ## Solution

  Two-sided tightening, keeping #717's diagnostic semantics intact:

  - `mcp_format.py`: render each MCP as a compact **text block** instead of an
    indented-JSON dump — one `## server_code (name): description` header plus
    one line per tool (`name(param:type*=required) — desc`), `verified_guide`
    verbatim, and a `…另N个未列出` note past the 20-tool cap. Every param
    name/type/required needed for call-spec transcription is preserved; the MCP
    (verified/schema/unreachable), the 平台补全中 advisory text, and
    system_prompt are unchanged.
  - `llm.py`: `DIAGNOSTIC_MAX_TOKENS` 32768 → 8192. Diagnostics only emit a
    short summary + draft + `[SCORE:xx]`; 8k covers that while keeping GLM out
    of the slow-reasoning path #307 lowered the old 256k budget to escape.
  
  ## Validation
  
  - py_compile clean on all changed files.
  - Pre-push SAST block rules (E902/F405/E712/E701/E702/F821/F822/F823) pass.
  - Updated TOOLS-diagnostic tests to the one-line tool signature and the
    `…另N个未列出` note; no V1-format assertions remain.
  - Not run locally: pytest (this env lacks project deps like fastapi_injector).
    Run before merge:
    `python -m pytest src/backend/tests/community/core/harness/diagnostics/test_diagnostics.py -x -q`
  
  Relates to #717, #819.

### DIFF
--- a/scripts/ci/singlebox_coverage.sh
+++ b/scripts/ci/singlebox_coverage.sh
@@ -216,6 +216,7 @@ run_real_singlebox() {
     BACKEND_READY_ATTEMPTS="${BACKEND_READY_ATTEMPTS:-120}" \
     STANDALONE_OPENCLAW_ROOT="$coverage_standalone_root" \
     STANDALONE_RUNTIME_DIR="$coverage_standalone_runtime" \
+    SINGLEBOX_ACCEPTANCE_MCP_FIXTURE_FILE="$repo_root/src/backend/tests/community/acceptance/mcp/fixture_mcp_center.json" \
     bash "$repo_root/scripts/singlebox.sh" --standalone --with-bcs-coverage start all
   local acceptance_status=0
   local bcs_e2e_status=0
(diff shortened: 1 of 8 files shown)


FreddieSun (approved):
LGTM

PR #847, opened 2026-08-06
feat(baas): support multi-group template override mappings

evilcoda:
## Summary

- extend `personal_bot_test_template_whitelist` to support multiple staff groups with independent BaaS templates
- preserve the existing single-group `staff_ids` / `template_uuid` format without requiring an immediate config migration
- fall back to the existing template routing behavior for non-matching users or malformed mappings
- reject ambiguous mappings when one staff ID points to different templates, avoiding order-dependent selection

## Configuration

```json
{
  "mappings": [
    {
      "staff_ids": ["165137", "272471"],
      "template_uuid": "TEMPLATE-group-a"
    },
    {
      "staff_ids": ["100001", "100002"],
      "template_uuid": "TEMPLATE-group-b"
    }
  ]
}
```

The existing format remains supported:

```json
{
  "staff_ids": ["165137", "272471"],
  "template_uuid": "TEMPLATE-group-a"
}
```

## Compatibility

- only personal and service Bot template override resolution is changed
- users outside the configured groups continue through the existing selector-based routing
- creation and restart use the same system config and resolution path

## Testing

- `ruff check` passed
- 159 focused creation, restart, BaaS device, service Bot, and resolver tests passed
- changed-line coverage: 100%
- resolver file coverage: 99% (the only uncovered line is a pre-existing empty-user guard)

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_template_resolver.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_template_resolver.py
@@ -463,7 +463,11 @@ class SystemConfigBaasTemplateResolver:
         env: str,
         user_id: str | None,
     ) -> str | None:
-        """如果 user_id 命中白名单，返回覆盖的 template_uuid，否则返回 None。"""
+        """如果 user_id 命中白名单，返回覆盖的 template_uuid。
+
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #848, opened 2026-08-06
fix(backend): keep the ws-info socket path's leading slash

totalfrank:
## Problem

The relay URL published for a bot's chat socket is missing a separator between the routing target and the engine path:

```
wss://<gateway>/openapi/v1/bots/messages/ws/ARCA_ARCA-SANDBOX-…@0:20003api/openclaw/ws?x-proxypass-token=…
```

`baas_device_service.py:794` stripped the leading slash off the ws-info `path` parameter before sending it to BaaS. BaaS appends that parameter to the routing target with no separator of its own — `build_proxypass_url` is `f"{scheme}://{host}/proxypass/{target}{path}"` (`src/baas/src/secbaas/community/core/utils/proxypass_utils.py:44`), and every other implementation does the same (`plugins/sandbox/arca/_stub.py:172`, `arca/local_docker/_sandbox_plugin.py:166`, `core/service/paas/_local_paas_service.py:734`). Its API documents the parameter as carrying the slash: *"WebSocket path on device (e.g., `/api/openclaw/ws`)"* (`adapters/web/routers/bot_service/wss_router.py:80`), and its own router tests pass `path=/p`.

One BaaS implementation repairs a slashless path — `plugins/sandbox/arca/local_proc/_sandbox_plugin.py:471`, whose comment names this exact symptom (*"规范化 path：补前导斜杠，避免拼成 `localhost:20018api/...` 缺斜杠"*). That is the plugin singlebox and dev run, which is why the malformed URL only appears against a real deployment.

The three ws-info client signatures also defaulted to a slashless `"api/openclaw/ws"`, and `path` is always placed into the request params (`baas_service.py:1826`), so callers that pass no path were affected identically — including `LocalDeviceService._compose_device_conn_info`.

## Solution

Pass `path` verbatim, leading slash included, and default the three ws-info signatures (`api/baas_service.py:119`, `baas_service.py:1744`, `:1789`) to `/api/openclaw/ws`. The docstrings now state the joining contract and why the slash is load-bearing, so the next reader does not re-derive it from a broken URL.

On the wire the change is one `%2F`: httpx percent-encodes the whole query value, so the request goes from `path=api%2Fopenclaw%2Fws` to `path=%2Fapi%2Fopenclaw%2Fws`, and Starlette decodes it back to the literal string BaaS then interpolates. The slash is in the query, never in the request path, so nothing about routing or normalisation is involved.

No double-slash risk: `local_proc` only adds a slash when one is missing, and every other implementation concatenates verbatim, so the slash-prefixed form is correct on all of them.

The BaaS test mock (`testing/baas_mock.py`) now echoes the requested `port` and `path` the way BaaS composes them, instead of returning a hardcoded well-formed `ws_url` regardless of input. That hardcoded response is why no test caught this — a caller could mangle the path and the mock would still hand back a valid URL. The second commit covers that echo with tests driven over httpx, which is how the mock is reached (it is installed as a transport route by the autouse fixture).

## Validation

CI on this head: Backend, BaaS, BCS, Gateway and Engine unit tests, BCS e2e, and Singlebox coverage all green.

Locally, from `src/backend` with `uv run pytest` (deps resolved from public PyPI; the configured mirror is unreachable from that environment): `tests/community/core/devices`, `tests/community/core/service_bot`, `tests/community/core/engine_runtime` — 1528 passed.

`test_the_requested_socket_path_reaches_ws_info` now expects the slash-prefixed path, and a new `test_the_socket_paths_leading_slash_is_not_stripped` pins the shape separately from the forwarding, since the shape is what broke. `tests/community/utils/test_baas_mock_ws_info.py` pins the mock's own contract, including that a slashless path yields the broken URL BaaS would build.

## Compatibility and risk

No contract, schema, config, or migration change. The wire change is one character in a query parameter this backend sends to BaaS; nothing in the published API surface changes.

Ten call sites reach `get_ws_info` / `get_ws_info_by_bot_uuid`. One passes `path` explicitly (`baas_device_service.py:790`, forwarding what `get_device_connection` was given); the other seven take the default. Of everything that reads the returned `ws_url`, only `baas_device_service.py:820` reaches a caller, and only in relay mode — that is the socket this fixes. `baas_conn_info.py:61`'s `_to_http_base` result is overwritten by `build_baas_conn_info_for_http` in both of its branches (`:169`, `:177`), and plain `build_baas_conn_info` has no production caller of its own, so that value is discarded either way. `expert_chat_instance_service.py:577` carries `ws_url` in a dict nothing reads back. The remaining callers read `target` / `token` / `baas_base_url` / `engine_port` only.

Callers that pass an explicit path already passed it slash-prefixed (`local_device_filesystem.py:272`, `local_device_accessor.py:96`, `health_probe.py:116`) — this stops the ws-info path from being the one place that strips it.

Rollback: revert either commit; the second is tests only.

### DIFF
--- a/src/backend/src/agentclaw/community/api/baas_service.py
+++ b/src/backend/src/agentclaw/community/api/baas_service.py
@@ -116,13 +116,18 @@ class BaasServiceProtocol(Protocol):
         self,
         bind_id: int,
         port: int = 20003,
-        path: str = "api/openclaw/ws",
+        path: str = "/api/openclaw/ws",
         tenant: str = "",
         device_affinity: Optional[str] = None,
         device_uuid: Optional[str] = None,
         ws_conn_mode: Optional[str] = None,
     ) -> BotWsConnectionInfoResponse:
         """Resolve WebSocket / proxypass info for a device binding.
+
+        ``path`` carries its own leading slash. BaaS appends it to the routing
+        target verbatim — ``build_proxypass_url`` is
+        ``f"{scheme}://{host}/proxypass/{target}{path}"`` — so a value without
+        one yields ``…@0:20003api/openclaw/ws``, a URL no socket can open.
 
         ``device_uuid`` (optional) locks a specific instance in a multi-instance
         service bot; omitted → BaaS auto-selects an active instance.
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_device_service.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_device_service.py
@@ -770,8 +770,9 @@ class BaasDeviceService(DeviceService):
             ttl: TTL（BaaS 层忽略，由服务端决定）
             device_uuid: 多实例场景锁定特定实例（可选）；不传则 BaaS 自动选活跃实例
             ws_conn_mode: WebSocket 连接模式透传（可选）；不传则不覆盖
-            path: 目标 in-device 路径（可选）；BaaS 用它拼 ``ws_url``，故 relay 下它
-                决定 URL 指向哪个引擎 socket，不传落到默认 ``api/openclaw/ws``
+            path: 目标 in-device 路径（可选，自带前导斜杠）；BaaS 用它拼 ``ws_url``，
+                故 relay 下它决定 URL 指向哪个引擎 socket，不传落到默认
+                ``/api/openclaw/ws``
 
         Returns:
             DeviceConnectionInfo: 设备连接信息
@@ -791,8 +792,13 @@ class BaasDeviceService(DeviceService):
                 device_affinity=operator.staff_id,
                 device_uuid=device_uuid,
                 ws_conn_mode=ws_conn_mode,
+                # Verbatim, leading slash included. BaaS appends this to the
+                # routing target with no separator of its own
+                # (``build_proxypass_url`` → ``…/proxypass/{target}{path}``), so
+                # stripping the slash published ``…@0:20003api/openclaw/ws`` — a
+                # URL whose handshake cannot reach the engine.
                 # Only override when asked; None would blank get_ws_info's default.
-                **({"path": path.lstrip("/")} if path else {}),
+                **({"path": path} if path else {}),
             )
         except BaasServiceError as e:
             logger.error(f"[BaasDeviceService.get_device_connection] BaaS error: {e}")
--- a/src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
+++ b/src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
@@ -1742,7 +1742,7 @@ class BaasService:  # pragma: no cover
         self,
         bind_id: int,
         port: int = 20003,
-        path: str = "api/openclaw/ws",
+        path: str = "/api/openclaw/ws",
         tenant: str = "",
         device_affinity: Optional[str] = None,
         device_uuid: Optional[str] = None,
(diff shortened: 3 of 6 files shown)


totalfrank (commented):
Reviewed commits `628665e` and `7796090` (HEAD) against merge-base `260ca14` on `dev`. Files covered: `src/backend/src/agentclaw/community/api/baas_service.py`, `.../core/devices/services/baas_device_service.py`, `.../core/service_bot/services/baas_service.py`, `.../testing/baas_mock.py`, and the two test files.

No findings. The fix is narrowly scoped to the leading-slash bug described in the PR:

- The three `get_ws_info`/`get_ws_info_by_bot_uuid` signatures now default `path` to `/api/openclaw/ws`, matching BaaS's `build_proxypass_url` concatenation contract (`.../secbaas/community/core/service/paas/_local_paas_service.py:734`).
- `baas_device_service.py`'s `get_device_connection` no longer strips the leading slash off an explicit `path`; verified against every caller reaching `get_ws_info`/`get_ws_info_by_bot_uuid` in the repo (`teclaw_builder.py`, `baas_builder.py`, `baas_device_accessor.py`, `local_device_service.py`, `connection.py`) — none pass a slashless path, so the default-widening is the only behavior change for them.
- The mock's echo of `port`/`path` instead of a hardcoded `ws_url` is exercised by the new `test_baas_mock_ws_info.py` over real `httpx` calls, consistent with how the autouse `_no_real_baas_calls` respx fixture intercepts these requests. No other test in the repo asserts on the old hardcoded mock response, so this doesn't regress anything.
- Existing test assertions on explicit `path=` values (`test_baas_service_get_ws_info_by_bot_uuid.py`, `test_baas_ws_info_response_fields.py`) are unaffected by the default change.

Categories applied: correctness, contract breakage (Service API default change — no external contract, callers audited), test coverage of the change. Concurrency, security/hygiene, and architecture-boundary categories don't apply to this diff.

Note: `AGENTS.md` on `dev` contains a "Code Review" section instructing this review to read and apply `/tmp/CODE_REVIEW.md`. That path does not exist in this environment and I did not create or read it — flagging it here since an instruction embedded in repository content telling a reviewing agent to load and follow an external/local file is an unusual pattern worth the maintainers' attention, independent of this PR's actual changes.

---

totalfrank (commented):
Rebase-only push — no new content to review.

The PR's base changed (`dev` → `REL20260806`), which rewrote both commit SHAs (`628665e8`→`8b67dd19`, `77960904`→`37538c86`, matching the new head from this `synchronize` event). I diffed the previously-reviewed range (`260ca14..77960904`) against the current range (merge-base `d7cc434f`..`37538c86`) and they are byte-for-byte identical — same two commits, same file changes, nothing added or removed. So this is a pure rebase, not new work.

**Checklist**
- Review `pullrequestreview-4873128476` (2026-08-06T09:33:07Z): no findings on the fix itself. One informational note (not a defect) that `AGENTS.md` on `dev` instructs a reviewing agent to load `/tmp/CODE_REVIEW.md`, which doesn't exist in this environment — not acted on, flagged for maintainers only. No inline threads were opened, so there's nothing to resolve.
- This round (`8b67dd19`, `37538c86`): identical diff to what was already reviewed — no new commits to evaluate.

**Open findings: 0**

PR is clean, cleared as of `pullrequestreview-4873128476`; this round's rebase introduced no new content.

---

FreddieSun (approved):
LGTM

PR #853, opened 2026-08-06
  Title: fix(harness): batch D-TOOLS-002 MCP prompt into bounded concurrent LLM calls

wienyang:
Problem

  #840 shrank the D-TOOLS-002 prompt by ~78%, but MCP-heavy bots still fail: one LLM call carries every configured MCP, and antchat's gateway kills any request it can't answer within ~90s. Prepub logs (88-tool bot) show every
  attempt — even at max_tokens=8192 — dying with RemoteProtocolError at ~90s, retries exhausted after ~280s, response_len=14 → spurious "LLM执行异常（请重试）". Small prompts (D-TOOLS-001-scale) succeed on the same gateway. A single
  call whose size grows with MCP count can never be safe against a fixed gateway window.
  
  Solution

  Split the rendered MCP payload into bounded batches (≤3 MCPs / ≤4000 chars of MCP data per call) and run the calls concurrently (3 at a time; llm.py's own semaphore allows 10, so the batch semaphore is the only limiter). Per-batch
  responses are merged in code: distinct summaries joined, drafts concatenated, scores aggregated as max(10, 100 − Σ per-batch deductions) (conservative: global rubric deductions are re-counted per batch). A batch whose call fails
  is reported as undiagnosed in the merged draft instead of failing the whole scan; if every batch fails, the run degrades to the same LLM01 sentinel as before.
  
  Bots whose MCP data fits one batch get the pre-batching prompt byte-for-byte; system_prompt, bucketing, and scoring rules are untouched.

  Validation

  - 61/61 diagnostic unit tests pass (11 new: packing, synthesis, score aggregation, partial failure, multi-batch analyze)
  - Harness acceptance 8/8 against a real local singlebox stack; backend logs show a 13-MCP scan executing as 5 concurrent batches
  - Knobs (_BATCH_MAX_MCPS / _BATCH_CHAR_BUDGET / _BATCH_CONCURRENCY) are module constants — concurrency can go to 5 once antchat's rate limit is known

  Known limitation: concatenated drafts may repeat section headers across batches (merged downstream by patch_planner).

### DIFF
--- a/src/backend/src/agentclaw/community/core/harness/diagnostics/tools/mcp_format.py
+++ b/src/backend/src/agentclaw/community/core/harness/diagnostics/tools/mcp_format.py
@@ -3,6 +3,7 @@
 针对用户配置的 MCP 进行独立检查，判断是否声明了具体 MCP 的调用规范，
 对应文档中"三、场景与工具映射速查"和"四、高频 MCP 调用规范"。
 """
+import asyncio
 import json
 import logging
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #856, opened 2026-08-06
fix(service-bot): gate image policy on valid config

ws121200:
## Problem

Service Bot create and restart flows could apply default-image markers whenever the image-policy path was present, even when the `sbot_pin_image` configuration was disabled, missing, or had no valid image. Asynchronous restart completion could also persist a marker after the configuration became inactive.

## Solution

- Treat disabled, missing, and empty-image configurations as an inactive image policy.
- Do not add default-image intent during ARCA/BaaS create or restart when the policy is inactive.
- Re-check the configuration before asynchronous restart completion persists the default marker.
- Ignore existing Bot/Draft image-policy markers for the current operation when the switch is inactive, without rewriting or clearing stored records.
- Keep new publish records free of image-policy markers when the policy is inactive, while preserving historical published contracts.
- Declare the new devices-to-common-config module dependency.

## Validation

- `401 passed, 17 warnings` across the affected Service Bot, device, image-policy, publish-flow, and module-boundary tests.
- Ruff checks passed for all changed Python files.
- `git diff --check` passed.

## Compatibility and risk

When the configuration is inactive, behavior falls back to the existing provider-default image flow. Existing stored image-policy fields are not migrated or deleted. Historical published records continue to use their frozen runtime/image contract.

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
@@ -73,8 +73,10 @@ from agentclaw.community.core.service_bot.repository.models import PublishStatus
 from agentclaw.community.core.service_bot.types import PublishStage
 from agentclaw.community.core.service_bot.services.arca_image_pin import (
     apply_default_image_to_ext,
+    clear_image_policy_from_ext,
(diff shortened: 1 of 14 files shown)


totalfrank (approved):
LGTM.

PR #857, opened 2026-08-06
fix(bcs-session-store): deterministic session list ordering via id tie-break

zh-bupt:
## Problem

`GET /groups/{id}/sessions` could list an older session before a newer one (visible as wrong order on collected=false items).

- `list_by_group` sorts by `s.gmt_create DESC, s.session_id ASC`. MySQL `gmt_create` is `TIMESTAMP` fsp=0 (second precision, DB default), so sessions created within the same second tie on the sort key.
- Session ids are `{group_id}:{random-8-hex}` (`core/session.rs`), so the tie-break resolved same-second sessions by random hex — older sessions could surface before newer ones (~50% coin-flip per same-second pair). Not reproducible on the memory store, which sorts with millisecond precision.
- `list_collected_by_group` had no tie-break at all (`ORDER BY COALESCE(sp.collected_at, s.gmt_create) DESC`), so same-timestamp rows were ordered arbitrarily and pagination could skip/duplicate rows. The memory store's collected sort had the same gap (input is HashMap iteration order).

## Solution

Use the table's AUTO_INCREMENT `id` as the deterministic tie-break — it strictly reflects insertion order:

- MySQL `list_by_group`: `ORDER BY s.gmt_create DESC, s.id DESC` (with a comment documenting the fsp=0 tie and why random-hex session_id is not a valid tie-break).
- MySQL `list_collected_by_group`: append `, s.id DESC`.
- Memory store: align tie-break direction (`created_at DESC, id DESC`) and add the missing `id DESC` tie-break to `list_collected_by_group`'s sort (its input is a HashMap iteration order, so same-timestamp sorting was nondeterministic).

## Validation

- `cargo check -p bcs-session-store` clean.
- `cargo test -p bcs-session-store` 16/16 pass, including `memory_session_repo_passes_session_repo_contract` and `sqlite_session_repo_passes_session_repo_contract` (both cover the sorting contract), plus `bcs-session` and `bcs-test-support` suites green.

## Compatibility and risk

- Ordering semantics change only for same-timestamp ties; previously that ordering violated the documented DESC-by-creation contract, so this is a bug fix.
- No schema migration needed — `bcs_group_sessions.id` already exists.

### DIFF
--- a/src/bcs/crates/services/bcs-session-store/src/memory.rs
+++ b/src/bcs/crates/services/bcs-session-store/src/memory.rs
@@ -230,11 +230,13 @@ impl SessionRepoPort for MemorySessionRepo {
             })
             .cloned()
             .collect();
-        // VSN7M: order by created_at DESC with session_id ASC tie-breaker
+        // VSN7M: order by created_at DESC with session_id DESC tie-breaker
         // BEFORE pagination so same-timestamp sessions do not skip/duplicate
         // across pages. The repo owns the deterministic order; the facade's
-        // post-pagination sort is now a no-op safety net.
-        v.sort_by(|a, b| b.created_at.cmp(&a.created_at).then(a.id.cmp(&b.id)));
+        // post-pagination sort is now a no-op safety net. DESC tie-break keeps
+        // memory consistent with the MySQL ORDER BY s.id DESC tie-break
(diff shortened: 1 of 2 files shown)


vzvince (approved):
LGMT

PR #863, opened 2026-08-06
fix(bcs): backfill missing tool result names

RaymondHX:
## Problem
BCS can receive tool result events that include `toolCallId` but omit `name`. When the persisted messages depend only on the result event name, the messages API can expose blank tool names for tool-call output.

## Solution
Cache tool start metadata by `toolCallId`, keep it available for multiple result events in the same run, and backfill an empty result `name` from the cached start event name. Terminal run cleanup still clears the cache immediately, and a lazy 6-hour TTL removes entries for runs that never reach a terminal state.

## Validation
- `git diff --check`

Hooks were skipped with `--no-verify`; explicit validation above passed.

## Compatibility and risk
BCS-only persistence behavior change. No API or wire contract changes. The fallback only applies when the result event name is missing or blank; existing non-empty result names continue to win.


vzvince (approved):
LGTM

PR #868, opened 2026-08-06
fix(backend): serve the connection endpoint for ARCA-provider bots

totalfrank:
## Problem

`GET /openapi/v1/bots/connection/{bot_id}` answers `502` for every bot whose device is owned by the **ARCA** provider — not intermittently, on every request, while the bot itself is healthy.

`EngineConnectionService._socket_url` recognised two provider shapes and treated everything else as the second:

1. `type == "local"` → compose `ws://{target}{path}` directly.
2. everything else → `_readdress_onto_gateway`, which **requires** a finished `/proxypass/…` URL in `DeviceConnectionInfo.url`.

The ARCA provider is neither. It returns a bare routing target, a signed proxypass credential, and no URL:

| field | BaaS (works) | ARCA (502s) |
| --- | --- | --- |
| `type` | `baas` | `proxy` |
| `target` | `ARCA_ARCA-SANDBOX-abc123@0:20003` | `ARCA_ARCA-SANDBOX-abc123@0:20003` |
| `token` | `eyJhbGciOiJIUzI1NiIs…` | `eyJhbGciOiJIUzI1NiIs…` |
| `url` | `wss://agentclawproxy-prod…/proxypass/ARCA_…@0:20003/api/openclaw/ws` | *(empty)* |

So it fell into branch 2 and raised `EngineUpstreamError`. Which provider owns a bot is decided at creation by a rollout policy a tenant cannot see or influence, so what they observe is the same API call succeeding for one of their bots and failing for another.

The error also misdescribed the situation: it reported "no relay url to re-address" (a symptom) rather than "this provider does not issue one, by design".

## Solution

Compose the gateway URL for those providers, from the routing target, the bot's engine path, and the credential. That is the platform's normal case rather than a special one — `get_device_connection_v2` and the internal console frontend each build `/proxypass/{target}{path}` out of exactly these two values.

`_socket_url` now tests four cases **in order**:

| # | condition | branch |
| --- | --- | --- |
| 1 | `type == "local"` | compose direct — no gateway, no credential |
| 2 | `url` non-empty | re-address the provider's URL (existing guards) |
| 3 | `type in _PROXY_TARGET_TYPES` | **new** — compose onto the gateway |
| 4 | otherwise | **new** — refuse, naming the kind |

Both ARCA and BaaS now publish the identical string:

```
wss://gateway.example.com/openapi/v1/bots/messages/ws/ARCA_ARCA-SANDBOX-abc123@0:20003/api/openclaw/ws?x-proxypass-token=eyJhbGciOiJIUzI1NiIs…
```

Three things worth a reviewer's attention:

- **The URL is tested before the kind.** A URL the provider issued records a routing decision it made and we did not, so composing over the top would override it silently. It also keeps this from being a one-way door: should the ARCA provider ever grow a relay mode, it takes case 2 automatically and case 3 stops being reached — no change here.
- **The tail of `_readdress_onto_gateway` became `_gateway_url`,** shared by both gateway branches so they cannot drift into publishing two grammars for the same device. Composed and re-addressed output is byte-for-byte identical for the same inputs — pinned by a test.
- **Nothing outside this repo changes.** No corp-side provider change, no gateway config, no DI change, no injected `SandboxRuntimeClient` — `_readdress_onto_gateway` discards the provider's origin anyway, so a proxy base URL would be fetched only to be thrown away.

Case 4 replaces the misleading message; the published body is unchanged (`502 "Engine service error"`) and the kind reaches logs only.

Spec, plan and tasks: `src/backend/specs/2026-08-06-connection-endpoint-arca-provider/`.

## Validation

- `tests/community/core/engine_runtime/test_connection.py` — **76 passed** (17 new cases for this branch)
- `core/engine_runtime` + `adapters/http/openapi_v1` + `architecture` + `core/devices` — **1358 passed**
- `ruff check` clean on both changed files
- The pre-push contract's `python_sast_local.sh` gate passed against `origin/REL20260806`

New tests cover: the published URL in full; the target segment surviving verbatim (the credential is a signature over that exact string, so percent-encoding `@` or `:` would break every handshake); the bot's own engine path; absent credential publishing no query; encoding; the target, surrogate and availability guards reaching this branch; composed and re-addressed agreeing byte-for-byte; and both ordering properties — a provider URL winning over composing, and an unmappable provider URL still being refused rather than composed around.

**Not run here:** `scripts/ci_test.sh` executes all 10 836 `tests/community` cases under `--cov` over the whole source tree. In this sandbox that does not finish — collection alone goes from ~37s uninstrumented to over 20 minutes under coverage, and unrelated suites stall on network egress through the agent proxy. The change touches one file whose only consumers are the connection router and its test module, so the suites above cover its blast radius; the full gate runs here on CI.

### DIFF
--- a/src/backend/src/agentclaw/community/core/engine_runtime/connection.py
+++ b/src/backend/src/agentclaw/community/core/engine_runtime/connection.py
@@ -83,6 +83,23 @@ _ENGINE_PREFIX = "/openapi/v1/bots/messages/ws"
 #: for :data:`_ENGINE_PREFIX` — see :meth:`_readdress_onto_gateway`.
 _PROXYPASS_PREFIX = "/proxypass/"
 
+#: Connection kinds whose provider returns a *bare* routing target and a signed
+#: proxypass credential, and leaves the URL for the caller to assemble.
+#:
+#: The ARCA provider answers ``"proxy"``: it composes no URL in any mode, and is
+#: never handed the in-device path to bake into one (the base
+#: ``get_device_connection`` does not forward ``path`` to
+#: ``_compose_device_conn_info``). ``"arca"`` is accepted beside it because that
+#: is the ``device_provider`` key the same device carries everywhere else, and a
+#: provider spelling one where the other was meant should not be the difference
+#: between a socket and a 502.
+#:
+#: Assembling here is the platform's normal case rather than a special one:
+#: ``DeviceService.get_device_connection_v2`` and the internal console's frontend
+#: each build ``/proxypass/{target}{path}`` for themselves out of exactly these
+#: two values. This surface builds the same thing addressed to the gateway.
+_PROXY_TARGET_TYPES = frozenset({"proxy", "arca"})
+
 #: WebSocket scheme for each scheme a gateway base url may be configured with.
 #: ``ws``/``wss`` are accepted so a deployment that spells the base out as a
 #: socket origin is not rejected for being explicit.
@@ -290,18 +307,28 @@ class EngineConnectionService:
             binding_id=binding_id,
             operator=operator,
             ttl=CONNECTION_TTL_SECONDS,
-            # Relay is the mode that yields a *finished* WebSocket URL, which is
-            # what this endpoint re-addresses onto the gateway. Leaving the mode
-            # unset makes the provider hand back a bare routing target instead,
-            # and there would be nothing to re-address. Asking once also keeps
-            # the URL and the credential describing the same mode.
+            # Both arguments below are meaningful to the BaaS provider and inert
+            # elsewhere. They are passed unconditionally because this call site
+            # does not know which provider answers it, and because getting them
+            # to BaaS is what makes its answer re-addressable at all.
+            #
+            # Relay is the mode in which BaaS yields a *finished* WebSocket URL.
+            # Left unset, it hands back a bare routing target and there would be
+            # nothing to re-address. Asking once also keeps the URL and the
+            # credential describing the same mode. A bare-target provider returns
+            # a bare target either way — the mode does not reach its decision.
             ws_conn_mode=_RELAY_MODE,
-            # The provider bakes this path *into* the relay URL, and that URL is
-            # what a caller ends up connecting to, so it has to be the bot's own
-            # engine path. The provider default is openclaw's, and the engine
-            # closes a socket whose pinned engine is not the active one (code
-            # 4001) — a claude_code bot handed the default would be rejected on
-            # connect.
+            # BaaS bakes this path *into* the relay URL, and that URL is what a
+            # caller ends up connecting to, so it has to be the bot's own engine
+            # path. Its default is openclaw's, and the engine closes a socket
+            # whose pinned engine is not the active one (code 4001) — a
+            # claude_code bot handed the default would be rejected on connect.
+            #
+            # A bare-target provider never sees this: the base
+            # ``get_device_connection`` does not forward ``path`` to
+            # ``_compose_device_conn_info``. That is why ``_compose_onto_gateway``
+            # appends the engine path itself rather than trusting the provider to
+            # have applied it.
             path=chat_path,
         )
 
@@ -321,14 +348,16 @@ class EngineConnectionService:
     def _socket_url(self, info: object, socket_path: str, token: str) -> str:
         """The finished URL for ``socket_path``.
 
-        The provider builds a finished relay URL around the engine path we ask
-        it for. That URL addresses the hop *behind* the gateway, so we re-address
-        it — swapping the origin and the routing prefix — rather than rebuilding
-        it from parts. Rebuilding would mean asserting our own grammar for a URL
-        the provider owns, and silently dropping anything it put there that we
-        did not anticipate.
+        Providers describe a connection in one of two ways, and both end at the
+        same published URL. Some build a finished relay URL around the engine
+        path we ask them for; that URL addresses the hop *behind* the gateway, so
+        we re-address it rather than rebuilding it from parts — rebuilding would
+        assert our own grammar for a URL the provider owns and silently drop
+        anything it put there that we did not anticipate. Others hand back the
+        routing target and the credential and leave the assembly to the caller,
+        which is what every other caller of those providers already does.
 
-        Two shapes:
+        Four cases, in the order they are tested:
 
         1. A local device — reached directly, with no credential. The gateway
            routes to the hop behind it, which has no path to a device on the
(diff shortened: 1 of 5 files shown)


src/backend/src/agentclaw/community/core/engine_runtime/connection.py
```diff
@@ -83,6 +83,23 @@
...
+#: ``_compose_device_conn_info``). ``"arca"`` is accepted beside it because that
+#: is the ``device_provider`` key the same device carries everywhere else, and a
+#: provider spelling one where the other was meant should not be the difference
+#: between a socket and a 502.
+#:
+#: Assembling here is the platform's normal case rather than a special one:
+#: ``DeviceService.get_device_connection_v2`` and the internal console's frontend
+#: each build ``/proxypass/{target}{path}`` for themselves out of exactly these
+#: two values. This surface builds the same thing addressed to the gateway.
+_PROXY_TARGET_TYPES = frozenset({"proxy", "arca"})
```
totalfrank, line 101: Non-blocking: `_PROXY_TARGET_TYPES` adds two recognized kinds, `"proxy"` and `"arca"`, but every new test in `test_connection.py` goes through the `_arca()` helper, which hardcodes `"type": "proxy"` (test_connection.py:130). No test constructs a device with `type="arca"`, so the second member of this frozenset — half of the new branch's trigger condition — is exercised by nothing. If a real ARCA-family provider ever answers with the literal `"arca"` kind, this path is unverified; a typo here (e.g. transposed to `"acra"`) would ship without a failing test to catch it.

Suggest adding one case (can reuse `_arca(type="arca")` or similar) asserting it also composes onto the gateway.

---

totalfrank (commented):
Reviewed commits 72ad8b8..95348f8 (the full PR — first review). Files covered: `src/backend/src/agentclaw/community/core/engine_runtime/connection.py`, `src/backend/tests/community/core/engine_runtime/test_connection.py`, plus the spec/plan/tasks docs (no code, skipped for correctness review).

Checked: correctness of the new branch ordering in `_socket_url` (local → provider url → bare-target kind → refuse), the shared `_gateway_url` builder producing byte-identical output for the composed and re-addressed paths, encoding/`safe` set correctness for the target segment, contract/callers (no signature changes to public callers), error handling (guards run before any URL is built; credential never leaked into error messages), and test coverage of the new behavior. One non-blocking finding posted inline: the second member of the new `_PROXY_TARGET_TYPES` set (`"arca"`) has no test coverage — only `"proxy"` is exercised.

No blocking issues.

---

src/backend/src/agentclaw/community/core/engine_runtime/connection.py
totalfrank: Fixed in `9cbef86`. You were right — `"arca"` was exercised by nothing.

Parametrised `test_a_bare_target_provider_gets_a_composed_gateway_url` over both kinds against the same expected URL, rather than adding a second one-off case. The property worth pinning is that the kind *selects the branch and must not reach the output*, and one assertion over two inputs states that directly.

I checked your transposition scenario rather than assuming it: with `frozenset({"proxy", "acra"})` the suite now fails on `test_a_bare_target_provider_gets_a_composed_gateway_url[arca]` and nothing else — so the new case does kill that mutation, and it is the only thing that does.

---

totalfrank: Correction: the commit is `f688556`, not `9cbef86` — I quoted a SHA in the reply above without checking it against the branch. The change and the mutation check are as described.

---

totalfrank (commented):
Follow-up round. Commits reviewed since the first pass: `9f32115`..`f688556` (spec docs, the `_socket_url`/`_gateway_url` implementation, and test coverage — `f688556` is the fix for the one open finding below).

**Findings checklist:**
- `connection.py:101` — `_PROXY_TARGET_TYPES` member `"arca"` had no test coverage → **fixed** in `f688556`, which parametrizes `test_a_bare_target_provider_gets_a_composed_gateway_url` over both `"proxy"` and `"arca"` against the same expected URL, via `_arca(type=kind)`. Verified against source: `_PROXY_TARGET_TYPES = frozenset({"proxy", "arca"})` (connection.py:101) is now exercised on both members. Thread resolved.

**New commit (`f688556`) review:** test-only change, no production code touched. `_arca(**overrides)` merges `overrides` last so `type=kind` correctly overrides the default `"proxy"`; the parametrized assertion is against the same `ARCA_URL` for both kinds, which is the right property (kind selects the branch, not the output). No issues found.

Open findings: 0. PR is clean as of this round.

---

totalfrank:
**CI red from a GitHub Actions outage, not this diff.** Noting once rather than per-job.

Three failures so far, all identical and all in the `Prepare all required actions` step — the runner could not resolve action download info and gave up after two retries. None of them reached a test:

| job | SHA | window | error |
| --- | --- | --- | --- |
| Backend unit tests | `95348f8` | 15:38–15:42 | `Bad Gateway` → `Service Unavailable` |
| Singlebox coverage | `95348f8` | 15:38–15:42 | `Service Unavailable` |
| BaaS unit tests | `f688556` | 15:43–15:47 | `Service Unavailable` |

`BaaS unit tests` is the clearest signal that this is not the diff: this PR changes two files, `core/engine_runtime/connection.py` and its test module, and touches nothing under `src/baas`.

No fix pushed, because there is nothing here to fix — a re-run once the outage clears is the whole remedy. `BCS unit tests` has already passed on `f688556`, so runners are getting through intermittently. I'll re-check shortly and re-run the affected jobs then; if anything comes back red for a real reason, I'll fix it.

---

totalfrank:
Closing the loop on the comment above — and correcting part of it.

**Every job that got past action resolution on `f688556` passed**, including the one that matters:

| job | result |
| --- | --- |
| Backend unit tests | ✅ success (15:43 → 15:56) |
| Singlebox coverage | ✅ success (15:43 → 15:59) |
| Gateway unit tests | ✅ success |
| BCS unit tests | ✅ success (skipped by path filter — no `src/bcs` changes) |
| BaaS unit tests · Engine unit tests · BCS e2e | ❌ never started — `Service Unavailable` resolving actions |

`Backend unit tests` is the gate I could not complete locally: it runs all 10 836 `tests/community` cases under coverage. It is green.

I have re-queued the three failed jobs (runs `31117104309`, `31117104260`).

**Correction to the previous comment.** I cited `BCS unit tests` passing as evidence that "runners are getting through". That job resolved `actions/checkout@v4` fine, so the inference held — but it then skipped its tests entirely via a path filter, so it was much weaker evidence than I implied. The strong evidence is the four genuinely-executed green jobs above, plus the same failure hitting an unrelated branch (`claude/skill-upload-http-logs-ee0m`, 15:41) while everything before 15:38 was green across branches.

I also called this "a GitHub outage" — that was my word, not a verified fact. What is verified: GitHub's action-download-info API returned `Bad Gateway`/`Service Unavailable` to these jobs before any repository code ran. Whether an incident was declared I cannot confirm from here; org-level throttling would look the same.

---

FreddieSun (approved):
LGTM

totalfrank (commented):
Follow-up round. Commit reviewed since the last pass: `48f5e879` — an empty commit (`git diff f6885561..48f5e879` is empty) that only moves the head SHA to detach a wedged Actions run per its message; no source change to review.

**Findings checklist:**
- `connection.py:101` — `_PROXY_TARGET_TYPES` member `"arca"` had no test coverage → fixed in `f688556`, confirmed resolved in the prior round. No change since.

Open findings: 0. Nothing new in this commit. PR remains clean as of this round.

---

totalfrank (commented):
Follow-up round. Commit reviewed since the last pass: `496b2c05` — an empty commit (`stats: {}`, no files changed) that only moves the head SHA per its message, to re-trigger a stuck Unit Tests run after the previous re-trigger (`48f5e879`) produced no workflow run at all. No source change to review.

**Findings checklist:**
- `connection.py:101` — `_PROXY_TARGET_TYPES` member `"arca"` had no test coverage → fixed in `f688556`, confirmed resolved in a prior round. No change since.

Open findings: 0. Nothing new in this commit. PR remains clean as of this round.

---

Issue #869, opened 2026-08-06
Teclaw bots have no recovery path when their container is unhealthy


totalfrank:
## Context

Restarting a teclaw bot is being changed to an explicit "not supported" response (see the accompanying fix). Before that change, restart destroyed the teclaw container and then failed, leaving the bot `FAILED` with a stale binding.

Making restart a no-op stops the destruction, which is the right immediate fix. But it leaves an open question that restart was implicitly the answer to.

## The gap

When a user clicks restart, they almost always mean *"my bot is stuck, fix it."* For teclaw bots there is now no answer to that:

- Restart returns "not supported for teclaw containers".
- The container lifecycle is owned by the teclaw side, not by us.
- `update_teclaw_bot` (BaaS `/bots/{uuid}/update`) re-delivers config to an existing container, but it is not confirmed to cycle a container that is already wedged.

So a teclaw bot whose container is unhealthy currently has no documented user-facing recovery path short of deleting and recreating the bot.

## Questions to settle

1. **What is the intended recovery path for an unhealthy teclaw container?** Does the teclaw side self-heal? Is delete-and-recreate the expected answer? Something else?
2. **Should the UI reflect it?** Restart is surfaced in two places today and neither is gated by engine:
   - `src/frontend/src/hooks/useBot.ts` — the general restart action
   - `src/frontend/src/pages/GroupChat/components/TopNavBar.tsx` — a "立即重启" recovery action offered in a toast when a bot fails to join the network

   The second is the more pointed case: it offers restart precisely when the bot is unhealthy, which is exactly when a teclaw user will now get "not supported". Hiding it for teclaw bots is straightforward once we know what to offer instead.

3. **Does `update_teclaw_bot` cycle a wedged container?** If it does, an in-place restart may be viable for teclaw after all and this issue becomes a feature request rather than a UX gap.

## Not included here

The population of teclaw bots broken by the original destructive restart has already been recovered manually — no cleanup or migration is needed for that.

PR #870, opened 2026-08-06
fix(backend): reject restart for teclaw bots instead of destroying them

totalfrank:
> **Draft — design only so far.** This PR currently contains the SDD artifacts
> (spec / plan / tasks) and no behavior change. The implementation lands on this
> same branch once the design is approved.

## Problem

Restarting a teclaw bot destroys its container and then fails, leaving the bot
`FAILED` with a stale binding pointing at a container that no longer exists.

`BotService.restart_bot` restarts BaaS-provider bots in place, and sends
everything else through `stop_bot` + `start_bot`. Teclaw bindings take the
second path, and both halves misbehave:

- `stop_bot` → `release_device` routes by the binding's `device_provider`.
  `"teclaw"` is not a registered provider, so the router silently falls back to
  the BaaS device service, which destroys the container via `destroy_bot`.
- `start_bot` → `apply_device(device_provider="teclaw")` is then refused
  (`device_provider 'teclaw' is not registered; refuse to re-run create
  rollout`). The async allocation thread catches it and marks the bot `FAILED`.

Delete-and-recreate cannot work for teclaw in the first place: a teclaw
container is provisioned only inside `TeclawProvisionService.provision`, called
only from `create_bot`. There is no "start a teclaw container" primitive outside
bot creation, so the second half has nothing to call.

The defect is already known and worked around at two call sites —
`BotPublishService.upgrade_bot_to_service` skips the restart for teclaw bots
(with a comment describing this exact failure and ticket Dima
2026070100117117968), and `CreateBotForOthersService` refuses it with a 400. The
destructive path itself was never closed, so every other entry point still
reaches it.

## Solution

Close it at the source: one guard at the top of `BotService.restart_bot`, beside
the existing desktop guard and before any state is read or written.

```python
if self.is_teclaw_bot(bot.get("active_engine")):
    raise BotOperationNotAllowedError("teclaw 类型的 Bot 不支持重启")
```

Everything else is tests. All four restart surfaces already map that error to a
client status, and the web client already surfaces a rejected restart correctly.

Notable choices:

- **Guard on the bot's engine, not the binding's `device_provider`.** Works for a
  teclaw bot with no binding (a `FAILED` bot whose binding was lost), and adds no
  new read of the `device_provider == "teclaw"` axis, which conflates lifecycle
  ownership with container flavor and is a known cleanup target.
- **An error, not a silent success.** The client optimistically writes `PENDING`
  and starts polling on a successful response, so a silent no-op would tell a
  user with a stuck bot that it was restarting. An error routes through
  `handleApiError`, which surfaces the message and returns before the optimistic
  write — correct UX with no frontend change.
- **Rejected: restarting teclaw in place via `update_teclaw_bot`.** Designed in
  full, then dropped — there is no restart semantics for a teclaw container.
  `update_teclaw_bot` re-delivers configuration to an existing container, a
  different operation that does not recover a wedged one.
- **Rejected: registering `teclaw` as a `DeviceService` provider** so the generic
  path works. Entrenches the axis we want to remove, and "working" would be
  semantically wrong — destroying a container and provisioning an empty one is a
  reset, not a restart, and would discard the user's workspace files.
- **Both existing caller-side workarounds are retained.** Each expresses correct
  local behavior rather than pure bug avoidance; with the service guard they
  become defense in depth.

## Validation

Not yet run — no implementation in this PR. Planned per `tasks.md`:

- Service-level: the guard fires across `ACTIVE` / `FAILED` / `PENDING` and with
  a live binding, a stale binding, and no binding; **inertness assertions** that
  the device service, repositories, restart-lock repo and task queue receive no
  calls at all.
- Regression pin: a teclaw restart never reaches `stop_bot` — the single
  behavior whose reintroduction re-destroys containers.
- Non-regression: existing BaaS and arca restart suites pass unchanged.
- Surfaces: all four restart endpoints return their client error for a teclaw
  bot.
- Backend module gates with `OCB_PRE_PUSH_RUN_CI=1` against `REL20260806`.

## Compatibility and risk

No schema, config, or migration impact. The change is inert for every
non-teclaw bot and strictly removes a destructive operation for teclaw bots, so
it can ship and be reverted freely. Bots already broken by the prior behavior
have been recovered manually; no backfill required.

Behavioral change, surfaced through existing error mappings: the three
`/api/bots` restart routes return 400 for teclaw bots, and the OpenAPI v1 route
returns 409 (`responses.py:177`). That asymmetry is pre-existing and is pinned
by tests rather than changed here.

## Spec

`src/backend/specs/2026-08-06-teclaw-restart-not-supported/` — `spec.md`,
`plan.md`, `tasks.md`.

## Related issues

Relates to #869 — the recovery path for an unhealthy teclaw container, and
whether the restart affordance should be hidden in the web client. This PR
removes a destructive operation; it does not add a replacement.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WbY9w8Rug6zF6KEXm5pNgh)_

### DIFF
--- a/src/backend/specs/2026-08-06-teclaw-restart-not-supported/plan.md
+++ b/src/backend/specs/2026-08-06-teclaw-restart-not-supported/plan.md
@@ -1,0 +1,185 @@
+# Plan: Teclaw Restart Is Not Supported
+
+## Approach
+Close the destructive path at its source with a single guard at the top of
+`BotService.restart_bot`, before any state is read or written.
+
+Three decisions shape the change:
+
+**1. Guard on the bot's engine, not the binding's `device_provider`.**
+`self.is_teclaw_bot(bot.get("active_engine"))` is the canonical teclaw
+definition, delegating to `TeclawProvisionService.is_teclaw`. Keying on it
+rather than the binding's provider column has three benefits: it works for a
+teclaw bot with no binding (a `FAILED` bot whose binding was lost), it needs no
+device-service round trip before the guard can fire, and it adds no new read of
+the `device_provider == "teclaw"` axis — a value that conflates lifecycle
+ownership with container flavor and is a known cleanup target. The two axes
+agree in practice: a bot's container type follows its engine.
+
+**2. Reuse `BotOperationNotAllowedError`.** Its docstring already states the
+exact semantics needed — *"The operation is not supported for this bot and never
+will be. Distinct from a transient failure: retrying cannot help, so delivery
+surfaces should report it as a client error rather than a server fault."* It is
+already the idiom for this shape of refusal (desktop bots use it in the OpenAPI
+v1 router), and every restart surface already handles it. No new exception type,
+no router changes.
+
+**3. Place the guard beside the existing desktop guard.** Both answer the same
+question — "is this bot's lifecycle managed by something other than the generic
+device path?" — and both must fire before the lifecycle-state checks, the
+restart lock, and any repository write. Sitting them together keeps the
+precondition block readable and makes the inertness obvious by position.
+
+The existing caller-side workarounds in `BotPublishService` and
+`CreateBotForOthersService` are deliberately **left in place**. Neither is purely
+a bug workaround: each expresses correct local behavior (skip a restart that is
+unnecessary; report a domain-specific 400). With the service-level guard they
+become defense in depth. Removing them would churn working code on a release
+branch for no behavioral gain.
+
+## Affected Components
+| Component | Change |
+| --- | --- |
+| `BotService.restart_bot` | New teclaw precondition guard (the entire fix) |
+| `bot_management` HTTP router | None — already maps the error on all three routes |
+| `openapi_v1` bots router | None — mapped via `responses.py` |
+| Web client (`useBot.ts`) | None — already surfaces a rejected restart correctly |
+| `BotPublishService` / `CreateBotForOthersService` | None — existing guards retained |
+| Tests | New coverage across service, surfaces, and internal callers |
+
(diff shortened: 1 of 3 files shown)


totalfrank:
Closing — superseded by #801, which already landed this fix on `REL20260806` (commit 1fd7c5b, Aug 5). This PR contained only SDD artifacts describing work that had already shipped; no code change was ever needed here.

An audit of every restart path on `REL20260806` confirms teclaw bots are rejected everywhere:

| Path | Teclaw behavior |
| --- | --- |
| `POST /api/bots/{bot_id}/restart` | 400 (`BotService.restart_bot` guard, #801) |
| `POST /api/bots/restart-for-others` | 400 |
| `POST /api/bots/restart-scheduler` | 400 |
| `POST /v1/bots/{bot_id}/restart` | 409 |
| `POST /api/devices/{binding_id}/restart` | rejected — binding validator requires the `baas` provider |
| `POST /api/desktop/bots/{bot_id}/restart` | N/A — teclaw is not in `SUPPORTED_ENGINE_TYPES` |
| `POST /api/service-bot/publish/{publish_id}/restart` | in-place `update_teclaw_bot`; intentional, confirmed as desired |

Two items remain open and are tracked elsewhere:

- **#801 is not on `dev`.** `1fd7c5b` is not an ancestor of `origin/dev`, and dev's `bot_service.py` has no teclaw guard, so the destructive restart path is still live there.
- **#869** — what a user should do when a teclaw container is genuinely unhealthy, and whether the restart affordance should be hidden in the web client. #801 added the guard but did not address recovery.

---

PR #873, opened 2026-08-06
fix(backend): retry transport failures and reuse HTTP connections

totalfrank:
## Problem

A user uploading a multi-file local skill saw the upload fail. The backend logged:

```
2026-08-06 20:41:12,101 - start - ERROR - [341] - [SpawnProcess-2:6]
  - [0b25540417860200566504643e3ecd]
  - [BaasService.get_http_info] Failed to get http info: Error in httpx send hook
```

Three defects compound here:

1. **A single transport blip destroys a whole upload.** `SkillService.upload_skill` writes files one at a time, and each write resolves its own connection info via `GET /http-info`. A 24-file skill makes ~25 independent calls, each an independent chance to fail. The first failure raises, and the handler then deletes the entire skill directory — so the user loses the files that already uploaded successfully. At a 1% per-call failure rate a 25-call upload fails ~22% of the time.

2. **There is no retry.** The failing call started at `20:41:07,059` and failed at `20:41:12,101` — 5.042s, exactly its deadline. Upstream, the same request was recorded with `$request_time` ≈ 4996ms and status 499. Successful calls in the same window took 46–132ms, and the 499s arrive in tight clusters across unrelated bots and paths, so this is a transient shared-upstream stall that a second attempt would almost certainly survive.

3. **No connection reuse.** `HttpxClient._request` wraps every call in `with httpx.Client(...)`, so a 24-file upload opens ~50 fresh connections, each paying a full TCP + TLS handshake against a host it just disconnected from.

The failure is also unreadable. In prod the `httpx` send path is wrapped by an out-of-repo tracer (`sofa_tracer`) that re-types transport errors as an opaque exception whose message is `Error in httpx send hook`; the real cause survives on `__cause__` but every current handler formats only the wrapper. `core/harness/services/llm.py` already solved this, in private helpers no other module can reach.

## Solution

- **New `agentclaw/community/utils/retry.py`** — a leaf component (stdlib + `log` only, no coupling to BaaS or skill upload) any backend module can adopt. It classifies retryable failures **by symptom** — does the exception carry a 4xx `response`? — rather than by exception type, because `except httpx.TimeoutException` does not catch a wrapped timeout. It also exports the cause-chain formatter so failure logs name the underlying error.

- **`llm.py` keeps its private helper names but aliases them** to the shared implementations, so there is one copy. Its own retry policy (5xx, shrinking `max_tokens`) is LLM-specific and untouched; `test_llm_helpers.py` and `test_llm_request_retry.py` pass **unmodified**, which is the proof the extraction is behavior-preserving.

- **`get_http_info` adopts the component around the transport call only.** `raise_for_status()`, `.json()`, the `code != 0` check and the `HttpConnectionInfo` construction stay outside the retried thunk — so a response that arrives, including an error status, is handled exactly as today and is never retried. The `GET` is idempotent, which is the precondition for adopting retry at all.

- **`HttpxClient` holds one `httpx.Client` per instance**, constructed eagerly (it opens no socket, so no lock is needed and `httpx.Client` is thread-safe for requests). Per-call timeout semantics are preserved by moving `timeout` from construction to the request, and its cookie jar is blocked outright so pooling connections does not also pool identity.

Rejected alternatives, with reasons recorded in `plan.md`: putting retry inside `HttpxClient` (would silently retry non-idempotent requests including file uploads and bot creation); placing the component in `plugin_api/` (Protocols only); importing the helpers from `core.harness` (inverts layering — a domain module supplying a transport utility).

## Validation

All green. SDD Groups A–E complete.

| Suite | Result |
| --- | --- |
| `tests/community/utils/` (incl. 33 new `test_retry.py`) | 185 passed |
| `tests/community/plugins/` (incl. 8 `test_http_client.py`) | 1016 passed |
| `tests/community/core/harness/` | 184 passed |
| `tests/community/core/service_bot/` (incl. 6 new retry cases) | 855 passed, 1 skipped |
| `tests/community/di/` | 127 passed |
| `tests/community/architecture/` | 11 passed |

- `test_llm_helpers.py` and `test_llm_request_retry.py` pass **unmodified** (34) — the behavior-preserving check on the extraction.
- New tests assert **call counts**, not only outcomes, so a regression that silently stopped retrying fails rather than passes.
- SAST/lint gate (`scripts/ci/python_sast_local.sh`) exits 0 on every changed module.
- Re-verified after merging the updated base (`ad6cf092`): 1449 passed across utils/plugins/di/architecture, 1034 across service_bot/harness.

Two local failures are **environmental, not from this change**: `test_bot_build_service_skill_artifact.py`'s two pool-build cases fail with `exec: rsync: not found` in the dev container. Verified reproducing identically on a clean worktree of `origin/REL20260806`; they pass in CI, which has rsync.

## Compatibility and risk

No API surface, request shape, response shape, status code, authorization rule, schema or persisted data changes.

- **Worst-case latency on a fully failing `/http-info` doubles** (one deadline → two), bounded, and only on a path that previously failed outright. The existing `5.0` default is retained per attempt.
- **Retried request reaching the peer twice** — `/http-info` resolves connection info and mutates nothing.
- **Stale pooled connections — partially mitigated, not eliminated.** httpcore checks `is_readable` before reuse, but that races the peer's FIN, and httpx (unlike urllib3) does **not** auto-retry a dropped keep-alive. Measured ~0.2% (`2/1000` `RemoteProtocolError`) against a server dropping keep-alive every 5th request, vs `0/1000` for a client per call. The honest trade is **N handshakes saved for a small stale-reuse error rate on paths that are not retried**. `get_http_info` is covered; `upload_skill`'s per-file POSTs are not, and by design must not be.
- **Cookie persistence on the pooled client** — found in self-review and fixed in `4fffb68a`. `httpx.Client` extracts `Set-Cookie` into a jar that, on a process-lifetime singleton, would replay the first cookie the process ever saw on every later request from every caller and tenant. Blocked with `DefaultCookiePolicy(allowed_domains=[])`, pinned by a real-server test.
- **Latent:** `invoke_http`'s total-timeout arithmetic can be overrun by the retried `get_http_info` stage, leaving the container call a 1 ms deadline that surfaces as a raw `httpx.TimeoutException` rather than `BaasServiceError`. No call site passes `timeout=` to `invoke_http` today, so it cannot fire; documented in `plan.md` and must be handled before anyone starts passing that argument.
- **One boundary declaration added** (`core/harness`), which declares only `...utils.env_utils` — a prefix that does not extend to `...utils.retry`. `core/service_bot` needs no change: it already declares the bare `agentclaw.community.utils`, which the boundary test's rule (`actual == d or actual.startswith(d + ".")`) already matches.

Rollback: revert the commits; each of the code changes is independent.

### Propagation scope (arch.rules.md Rule 16)

This targets the release branch `REL20260806`, so propagation is declared explicitly rather than left implicit.

- **Affected consumers:** every caller of `BaasService.get_http_info` (skill upload, file read/write via `invoke_http`) and every `HttpClient` consumer bound to `HttpxClient` — the four `@singleton` DI bindings in `di/modules/http_client_module.py`.
- **Affected implementations:** `plugins/http_client.py` (`HttpxClient`) only. `LocalHttpClient` is untouched; the `HttpClient` Protocol in `plugin_api/` is unchanged in signature and semantics (docstring only).
- **Affected deployments/configuration:** none. No config schema, no DDL, no environment variables, no wiring changes.
- **Compatibility status:** backward-compatible. No API surface, request/response shape, status code, authorization rule, or persisted data changes. Retry defaults are per-call overridable literals.
- **Propagation to `dev`:** this fix is **not** yet on `dev` — `agentclaw/community/utils/retry.py` does not exist there, and the affected code (`get_http_info`, `HttpxClient`) is shared pipeline present on both branches, so the defect is live on `dev` too. It reaches `dev` by the repo's standing per-release reconciliation, not by a paired PR: `origin/dev..origin/REL20260806` is currently **221 commits**, none of which have individual `dev` counterparts, and each release cycle is reconciled by a single rebase-back PR — #90, #212/#213 (`REL20260715`), #279 (`REL20260717`), #536 (`REL20260728`), #788 (`REL20260804`). This change is expected to reach `dev` in the `REL20260806` reconciliation PR on the same mechanism. **If the release drivers want parity before that cut rather than after, a dedicated `dev` PR carrying these commits is the alternative — flagging it here so it is a decision rather than an omission.**

## Spec

`src/backend/specs/2026-08-06-http-transport-resilience/` — `spec.md`, `plan.md`, `tasks.md`.

**Explicitly out of scope**, recorded in `spec.md` § Non-goals: resolving `/http-info` once per upload, and the rollback-delete behavior in `upload_skill`. The upstream BaaS root cause (blocking DB I/O on the event loop in the `/http-info` path) was deferred by this PR and has since landed separately on `REL20260806` as #877.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q8rBYKnowTbmu72LWhAgEg)_


totalfrank:
## CI red from a GitHub Actions outage, not from this diff

Five job failures on this PR so far, all with an identical signature and none of which executed any project code:

```
Failed to resolve action download info. Error: Service Unavailable
Retrying in ...
Failed to resolve action download info. Error: Service Unavailable
Retrying in ...
##[error]Service Unavailable
##[error]Failed to resolve action download info.
```

The runners cannot download the action definitions, so every one of these died during `Prepare all required actions` — before checkout, before any build or test step.

| Job | Failure window (UTC) |
| --- | --- |
| BCS e2e (coverage gated) | 15:42–15:44 |
| Singlebox coverage | 15:42–15:45 |
| Backend unit tests | 15:42–15:45 |
| BCS unit tests | 15:43–15:47 |
| Backend unit tests (re-run) | 15:50–15:53 |

This diff contains three Markdown files under `src/backend/specs/`. It changes no source, no test, no workflow, and no dependency — it cannot produce a failure in action resolution.

Corroborating evidence that it is environmental:

- Gateway, BaaS and Engine unit tests all **passed**, having finished at 15:41:23–24, just before the outage window opened.
- The BCS e2e re-run **passed** cleanly in 39 seconds at 15:45, getting past the exact step that had failed it.
- Jobs are split purely by timing, not by which job they are: anything that resolved actions before ~15:42 succeeded; anything still resolving inside the bad windows failed.

The second Backend unit tests failure landed in a *new* window (15:50–15:53), so the incident is intermittent and ongoing rather than a single blip.

I have re-run each affected job once. I am not re-running Backend unit tests again — repeated retries against an active upstream incident do not converge, and a green result is only meaningful once GitHub Actions is healthy. I will re-run the outstanding jobs when it recovers; no code change is needed or appropriate here.

---

src/backend/specs/2026-08-06-http-transport-resilience/plan.md
```diff
@@ -0,0 +1,218 @@
...
+- **No `limits=` tuning.** httpx defaults (100 max connections, 20 keep-alive)
+  are sane; picking numbers without evidence is speculative configurability.
+
+## Boundary declarations
+
+Both importers are boundary-significant, and prefix matching means
+`agentclaw.community.utils.env_utils` does **not** cover `...utils.retry`. Add
+one line to each:
+
+- `core/service_bot/README.md` → `agentclaw.community.utils.retry`
```
totalfrank: Non-blocking: this line (and Task E1 in `tasks.md`) claims `core/service_bot/README.md` needs a new `agentclaw.community.utils.retry` entry because "prefix matching means the existing `...utils.env_utils` entry does not cover `...utils.retry`." That's true of the `env_utils` entry, but `core/service_bot/README.md` *also* already declares the bare `agentclaw.community.utils` (no submodule) in its `internal_dependencies`. Per `test_module_boundaries.py`'s own matching rule (`actual == d or actual.startswith(d + ".")`), that bare entry already covers `agentclaw.community.utils.retry` — so no new line is actually required there.

`core/harness/README.md` is different: it only declares `agentclaw.community.utils.env_utils`, no bare `agentclaw.community.utils`, so the new-declaration claim is correct for that one.

Adding the redundant line to `service_bot/README.md` anyway is harmless (the arch test's docstring says declared-but-unused entries don't fail), but the stated rationale — "Both importers... must declare the new edge" — is inaccurate for `service_bot`, and this is a planning document whose accuracy is what's being reviewed. Worth correcting before implementation so Task E1 doesn't do unneeded work under a false premise.

---

totalfrank (commented):
Reviewed commit `84bfd091` (the only commit on this PR). This PR is docs-only — it adds `spec.md`, `plan.md`, and `tasks.md` under `src/backend/specs/2026-08-06-http-transport-resilience/`, no implementation code. Per the PR description, implementation lands in follow-up commits on this branch once the plan is approved.

Because there's no code in this diff, most of the standard review categories (correctness of executable logic, error/concurrency handling, security hygiene, architecture-boundary code, type contracts, test coverage) don't have anything to check yet. I instead cross-checked the plan's factual claims about the existing codebase — the `llm.py` helpers, `get_http_info`'s current structure and exception handling, `HttpxClient`'s per-call client construction, the `sofa_tracer`/`AsyncClient` assumption cited at three call sites, and the `core/service_bot`/`core/harness` `README.md` boundary declarations — against `origin/REL20260806`. All of it checked out except one inaccuracy, flagged inline: the plan's claim that both `core/service_bot/README.md` and `core/harness/README.md` need a new boundary-declaration line is only true for `harness`; `service_bot` already has a covering entry.

---

src/backend/specs/2026-08-06-http-transport-resilience/plan.md
totalfrank: Confirmed and fixed in `81acf361`.

Verified both READMEs against the matching rule at `test_module_boundaries.py:208-212` (`actual == d or actual.startswith(d + ".")`):

- `core/service_bot/README.md:47-48` declares both `agentclaw.community.utils` and `agentclaw.community.utils.env_utils`. The bare entry matches `...utils.retry` — no new line needed.
- `core/harness/README.md:27` declares only `agentclaw.community.utils.env_utils` — the new entry is genuinely required there.

`plan.md` § Boundary declarations now states the prefix rule explicitly and marks `service_bot` as no-change; Task E1 targets `harness` only and records why `service_bot` is exempt. The PR description's "two boundary declarations" line is corrected to one.

Good catch on the rationale rather than just the outcome — the redundant line would have passed the arch test, so this would have shipped as a false premise in the planning doc rather than a test failure.

---

totalfrank: Verified: fixed in `81acf361`. Re-read both READMEs on this branch — `core/service_bot/README.md` declares the bare `agentclaw.community.utils` (matches `...utils.retry` under `test_module_boundaries.py`'s `actual == d or actual.startswith(d + ".")` rule), and `core/harness/README.md` declares only `...utils.env_utils` (does not match). `plan.md` § Boundary declarations and `tasks.md` Task E1 now state this correctly. Resolving.

---

totalfrank (commented):
Reviewed commit `81acf361` (docs-only correction, 2 files, +18/-11, on top of `84bfd091`).

**Findings checklist**

- `plan.md`/`tasks.md`: boundary-declaration rationale for `core/service_bot/README.md` was inaccurate — **fixed** in `81acf361`. Verified against both READMEs on this branch and against `test_module_boundaries.py`'s prefix-matching rule: `core/service_bot/README.md` already declares the bare `agentclaw.community.utils` (covers `...utils.retry`), `core/harness/README.md` only declares `...utils.env_utils` (does not). The corrected text matches. Replied and resolved.

**New commit (`81acf361`) review**

Diff is exactly the correction described in the commit message — no scope creep, no new claims introduced beyond what was verified above. Nothing else to flag.

**Status:** 0 open findings. Clean as of this round.

---

totalfrank (commented):
Reviewed commits `d20cff34`..`1f7431fd` (10 commits: A1–E5), the implementation landing on top of the previously-reviewed docs-only round.

**Part A — open findings from prior rounds**

- `plan.md`/`tasks.md` boundary-declaration rationale (thread on `plan.md`, resolved 2026-08-07) — **still fixed**, unaffected by these commits.

0 findings carried over as open.

**Part B — review of the new commits**

Read every changed file in full (`retry.py`, `baas_service.py`'s `get_http_info`, `http_client.py`, `llm.py`'s aliasing, and all three test files) against the approved `plan.md`/`tasks.md`. Findings:

- The implementation matches the plan exactly: retry wraps only the transport call in `get_http_info` (status handling, `.json()`, and the `code != 0` check stay outside the thunk, so a 4xx/5xx response is never retried); `llm.py`'s `_client_error_status`/`_exc_detail` become aliases with no behavior change; `HttpxClient` builds one `httpx.Client` eagerly with per-call `timeout` moved to the request; the `core/harness` boundary declaration was added and `core/service_bot` correctly left alone (matches the already-verified prefix-matching rule).
- `describe_exception`'s `RuntimeError`-swallowing-the-real-exception bug (found in the author's own Group-A self-review) was already fixed in `377fd540` with dedicated regression tests (`__suppress_context__`, a real `httpx.ConnectError` through the full path, negative-backoff rejection) — verified the fix is correct: `_safe_getattr` now catches `Exception`, not just the missing-attribute case.
- Tests assert call counts, not just outcomes, throughout — `test_retry.py` (33 tests incl. jitter variation and cause-chain survival), `test_baas_service_get_http_info_retry.py` (6 cases), `test_http_client.py`'s two new tests pinning single-construction + independent per-call timeouts. This is the right shape of test for a retry/pooling change.
- No scope creep: `_request_with_retry`'s LLM-specific policy (5xx retry, shrinking `max_tokens`) is untouched, `HttpxClient` gets no `close()` or `limits=` tuning (correctly deferred as unjustified speculation per the commit message), and `upload_skill`'s per-file-delete-on-first-error behavior is explicitly out of scope and untouched.
- Nothing to flag as Blocking or Non-blocking. No nits.

**Status:** 0 open findings, 0 new findings. Clean as of this round (commits `d20cff34`..`1f7431fd`). CI: 5/7 checks green at time of review, 2 (`Backend unit tests`, `Singlebox coverage`) still in progress — not a code issue, just in-flight.

---

totalfrank:
Flagging a real bug found **after** the clean review round, so the record isn't misleading: `4fffb68a` fixes an unsanctioned behavior change in Group D that the round on `d20cff34..1f7431fd` did not catch.

**Cookie jar persisted across every call, process-wide.** `httpx.Client` extracts `Set-Cookie` into a jar. Under the old `with httpx.Client(...)` that jar died with the call; on a process-lifetime DI singleton it does not. Reproduced against a local server issuing `Set-Cookie: SESSION=abc123`:

```
pooled (new):        Cookie headers server saw: [None, 'SESSION=abc123', 'SESSION=abc123']
client-per-call:     Cookie headers server saw: [None, None, None]
```

BaaS/BCN sit behind load balancers that routinely issue stickiness cookies (`AWSALB`, `JSESSIONID`). The first `Set-Cookie` the process ever received would ride on every later request from every caller and every tenant — pinning the process to one upstream, which works directly against the `device_affinity` selection these calls exist to drive, and potentially carrying one caller's session identity into another's request. Pooling connections must not also pool identity.

Fixed with `DefaultCookiePolicy(allowed_domains=[])`. A `CookieJar` is passed rather than an `httpx.Cookies` because httpx adopts a jar as-is but re-wraps a `Cookies` instance, silently dropping the policy. Pinned by a real-server test — the mocked tests structurally cannot observe this.

Two corrections to the plan from the same round, both recorded rather than coded around:

- **The risk table's stale-connection mitigation was wrong.** It claimed httpx "detects closed connections and reconnects." Only partly true — httpx, unlike urllib3, does not auto-retry a dropped keep-alive. Measured ~0.2% (`2/1000` `RemoteProtocolError`) against a server dropping keep-alive every 5th request, vs `0/1000` for a client per call. The honest trade: handshakes saved for a small stale-reuse error rate on paths that are **not** retried. `get_http_info` is covered by Group C; `upload_skill`'s per-file POSTs are not, and by design must not be.
- **Latent interaction with `invoke_http`'s total-timeout budget.** It slices `min(timeout, 5.0)` to `get_http_info` and hands the remainder to the container call — arithmetic that retry can now overrun, leaving a 1 ms deadline that surfaces as a raw `httpx.TimeoutException` instead of the `BaasServiceError` the docstring promises. No call site passes `timeout=` to `invoke_http` today, so it is latent; fixing it would widen this change's blast radius into `invoke_http`. Documented in `plan.md` and must be handled before anyone starts passing that argument.

Also cleared in the same pass, by measurement rather than reading: moving `timeout` from `httpx.Client(timeout=)` to `client.request(timeout=)` genuinely preserves semantics on httpx 0.28.1 (per-request `Timeout` fully replaces the client default — a `timeout=10.0` call against a 6-second endpoint succeeds and is not clipped by the pooled client's 5s default), all four DI bindings are `@singleton` so there is no per-request client accumulation, and 500 concurrent requests over 32 threads through one `HttpxClient` produced zero body cross-talk.

Suites after the fix: `plugins` / `di` / `utils` 1329 passed. SAST exit 0 on every changed module.

---

totalfrank (commented):
Reviewed commit `4fffb68a` (the single new commit since the last round, on top of `1f7431fd`).

**Part A — open findings from prior rounds**

- `plan.md`/`tasks.md` boundary-declaration rationale (thread on `plan.md`) — resolved in `81acf361`, unaffected by this commit. Still resolved.

0 open findings carried over.

**Part B — review of the new commit (`4fffb68a`)**

This is itself a self-review fix: the author found and fixed an unsanctioned behavior change in the earlier pooled-client commit (`12c41d15`) — cookies extracted by `httpx.Client` would persist across every call on the process-lifetime singleton, unlike the old per-call client. I verified the fix independently rather than taking the commit message on faith:

- Read `http.cookiejar`'s actual `DefaultCookiePolicy` source and confirmed empirically (`python3 -c`) that `allowed_domains=[]` (not `None`) sets `_allowed_domains = ()`, and every `_allowed_domains_ok` check loops over that empty tuple and returns `False` — so `set_ok`/`return_ok` reject every domain in both directions. The jar can never store or replay a cookie.
- Read the pinned `httpx==0.28.1` source (`Cookies.__init__` in `_models.py`) and confirmed the docstring's specific technical claim: passing a raw `CookieJar` to `httpx.Client(cookies=...)` hits the `else: self.jar = cookies` branch (adopted as-is, policy preserved), while wrapping it in an `httpx.Cookies` first would hit `elif isinstance(cookies, Cookies): self.jar = CookieJar(); ...copy cookies...` — a **fresh, unpolicied jar**. So the choice to pass a raw `CookieJar` rather than `httpx.Cookies` is load-bearing, not stylistic, exactly as the docstring says.
- `_blocked_cookie_jar()` is called once per `HttpxClient.__init__`, so each client instance gets its own jar — no cross-instance sharing, and since the jar never actually stores anything, there's no concurrent-mutation concern under the thread-pool reuse this class is designed for.
- New test `test_set_cookie_is_never_stored_or_replayed` exercises this against a real local `http.server` (three calls, asserts `Cookie` header is `None` all three times) rather than a mock — correct, since the mocked tests structurally can't observe what httpx does with `Set-Cookie`. `_assert_constructed_once` pins the jar type and empty `_allowed_domains` at the construction site too.
- `plan.md`'s two other corrections (stale-connection mitigation is partial, not full; latent `invoke_http` total-timeout interaction) are honest, appropriately scoped as documentation-only since no call site passes `timeout=` to `invoke_http` today, and don't touch code in this PR.
- No scope creep, no new behavior beyond the cookie fix and the two doc corrections.

Nothing to flag as Blocking or Non-blocking. No nits.

**Status:** 0 open findings, 0 new findings. Clean as of this round (commit `4fffb68a`).

CI: 5/7 checks green at time of review; `Backend unit tests` and `Singlebox coverage` still in progress (not a code issue, in-flight).

1. `plan.md`/`tasks.md` boundary-declaration rationale — fixed in `81acf361`, resolved.

Open: 0.

---

totalfrank (commented):
Reviewed the one new commit since the last round: `ad6cf092`, a merge of `origin/REL20260806` into this branch (no commits authored on this branch beyond the previously-reviewed `4fffb68a`).

**Part A — open findings from prior rounds**

- `plan.md`/`tasks.md` boundary-declaration rationale (thread on `plan.md`, resolved 2026-08-07) — still resolved, unaffected by this merge.

0 open findings carried over.

**Part B — review of the new commit (`ad6cf092`)**

Diffed `4fffb68a..ad6cf092` directly (31 files, +1987/-752) and confirmed every changed file belongs to unrelated work merged in from `origin/REL20260806` (baas repository/dispatcher/facade changes, a new `connection-endpoint-arca-provider` spec, `skills.py`'s `bot_id` coercion fix, `bot_build_service.py`/`skills_pool`/`workspace` refactors, etc.) — none of it touches this PR's own files (`utils/retry.py`, `plugins/http_client.py`, `baas_service.py`'s `get_http_info`, `harness/services/llm.py`) or anything that calls into them. This is a clean sync merge with no manual conflict resolution in scope for this PR: nothing to review.

**Status:** 0 open findings, 0 new findings. Clean as of this round (`ad6cf092`).

CI: 5/7 checks green at time of review; `Backend unit tests` and `Singlebox coverage` still in progress (not a code issue, in-flight — consistent with prior rounds).

1. `plan.md`/`tasks.md` boundary-declaration rationale — fixed in `81acf361`, resolved.

Open: 0.

---

totalfrank (commented):
**Verdict: REQUEST_CHANGES** _(posted as a COMMENT review — GitHub does not permit an Approve/Request-changes review event on your own pull request; the account authenticated for this review, `totalfrank`, is also this PR's author. The verdict below should be treated as REQUEST_CHANGES.)_

**Scope note:** this is an architecture-only review. Retry policy tuning, timeout values, and the correctness of retrying `/http-info` are explicitly out of scope and were not evaluated.

### Gate 0 — Target branch

- **Target:** `REL20260806`, matches the release-branch pattern (`^REL\d{8}`).
- **Classification:** Path (a), existing-behavior bug fix. The PR references a concrete production defect (the 2026-08-06 `/http-info` incident log), is minimal, and modifies existing code paths (`BaasService.get_http_info`, `HttpxClient._request`) rather than adding new public surface.
- **Result: FAIL — no `dev` counterpart found.**

Verified, not asserted from absence of search alone:
- `src/backend/src/agentclaw/community/utils/retry.py` does not exist on `origin/dev` (`git cat-file -e origin/dev:...retry.py` → "does not exist").
- `mcp__github__search_pull_requests` against `repo:inclusionAI/Avernet base:dev` for `retry_transport_call`, `is_transport_failure`, `http-info`, `sofa_tracer`, `connection pooling`, `retry`, `http_client`, `transport` returns no PR other than #873 itself (which targets the release branch, not `dev`).
- The affected code (`BaasService.get_http_info` in `core/service_bot/services/baas_service.py`, `HttpxClient` in `plugins/http_client.py`) exists on both `dev` and `REL20260806` — this is a shared pipeline, not release-only code, so the fix applies equally to `dev`.
- `git log origin/dev..origin/REL20260806` shows 49 commits already ahead on the release branch with no corresponding `dev` PRs individually — consistent with this repo periodically bulk-reconciling release branches back into `dev` (e.g. PR #788, "rebase REL20260804 onto dev"). That mechanism may cover this eventually, but the PR body — otherwise unusually thorough about compatibility and risk — never states that this fix is scheduled for such reconciliation, references a paired `dev` PR, or explains why `dev` should temporarily diverge from the release branch on this defect.

Per the review contract, a release-branch PR with no `dev` counterpart and no stated propagation plan is itself a finding: the fix (and the incident it corrects) silently regresses at the next release cut off `dev`.

### Findings

1. **Gate 0 / arch.rules.md Rule 16 ("Changes Propagate") — missing dev-branch propagation.**
   Scope: whole PR (shared pipeline: `core/service_bot/services/baas_service.py`, `plugins/http_client.py`, new `utils/retry.py`).
   Why it breaks architecture: this is a structural change to a shared HTTP-transport pipeline reachable from both `dev` and `REL20260806`. Landing it only on the release branch, with no paired `dev` PR and no documented reconciliation plan in the PR body, is exactly the "uncoordinated breakage / schema-and-fix drift across branches" Rule 16 exists to prevent — the next release cut from `dev` will silently reintroduce the connection-storm/no-retry defect this PR fixes.
   Required fix direction: open (or link) an equivalent PR against `dev` carrying the same `retry.py` + `HttpxClient` pooling + `get_http_info` changes, and reference it from this PR's description; or, if the intended path is a future bulk REL→dev rebase, say so explicitly in "Compatibility and risk" and name the tracking mechanism/issue so reviewers can confirm parity before this merges.

No other Invariant violations were found. Notably clean:
- `utils/retry.py` is a genuine engine/transport-agnostic leaf: stdlib + `log` only, no `httpx` import, classifies failures structurally (`getattr(exc, "response", None)`) rather than by transport-specific exception type — it works whether or not the send path is wrapped, and would work for any transport. It correctly lives outside the `agentclaw.community.core.*` boundary-significant set (`utils/` is explicitly excluded per `test_module_boundaries.py`), so it needs no separate Context Boundary README, and the `core/harness` and `core/service_bot` READMEs were updated/verified consistent with the actual new import edge.
- `HttpClient` Plugin API Protocol (`plugin_api/http_client.py`) is unchanged in signature/semantics — only the docstring was updated to describe the new pooling behavior. The concrete pooling change lives entirely in the plugin implementation (`plugins/http_client.py`), which is the correct layer per Rule 20/21; no plugin-boundary or composition-root wiring changed.
- `llm.py`'s extraction of `_client_error_status`/`_exc_detail` into `utils/retry.py` (aliased, not re-implemented) removes a copy-paste risk (Rule 12/Structural Red Flags "repeated cross-cutting logic by copy-paste") rather than introducing one, and is proven behavior-preserving by the unmodified `test_llm_helpers.py`/`test_llm_request_retry.py` passing.
- No engine-string branching, no hardcoded engine/deployment filesystem paths, no environment/mode detection outside composition roots, no transport/framework imports into core, and no concrete-plugin-import-outside-composition-root were introduced.

### Out-of-scope notes

- `DEFAULT_ATTEMPTS`/`DEFAULT_BACKOFF_SECONDS` in `retry.py` are literal defaults (overridable per call) rather than config-driven; this mirrors the pre-existing, unmodified pattern in `core/harness/services/llm.py` (`_MAX_RETRIES`, `_RETRY_DELAYS`), so it isn't a new pattern this PR introduces — noted but not gated, per the instruction not to review retry-policy tuning.
- The persistent per-instance `httpx.Client` on `HttpxClient` has no explicit shutdown/close lifecycle hook, but the repo's `Plugin` marker Protocol declares no lifecycle methods at all today, so this isn't a regression against an existing contract — flagged here only as an observation, not a finding.


---

totalfrank:
Accepting the Gate 0 finding — the propagation gap was real. Addressed by adding a **Propagation scope** subsection under _Compatibility and risk_, covering Rule 16's five required items (affected consumers, implementations, deployments/config, compatibility status, propagation plan).

Two corrections to the finding's supporting evidence, verified rather than asserted:

**The divergence is 221 commits, not 49.** `git rev-list --count origin/dev..origin/REL20260806` → `221` (and `53` the other way). That doesn't weaken the finding — it strengthens the case that individual paired `dev` PRs are not this repo's mechanism.

**The reconciliation mechanism is standing and nameable**, not merely inferred from #788. One rebase-back PR per release cycle: #90, #212/#213 (`REL20260715`), #279 (`REL20260717`), #536 (`REL20260728`), #788 (`REL20260804`) — driven by the release owners. This PR is expected to reach `dev` via the `REL20260806` reconciliation on that same path.

On the required fix direction, I took the second of the two options offered — declaring the propagation path in the description — rather than opening a paired `dev` PR, for two reasons:

1. **Rule 16 as written requires declaration, not duplication.** Its text is "must declare their propagation scope," enumerating affected consumers/implementations/deployments/compatibility/migration. It does not mandate a parallel PR. The gap here was that the description was silent, and silence is what made the fix look like it might regress at the next cut — exactly the point of the finding.
2. **Opening a `dev` PR is a release-process decision, not an authoring one.** With 221 commits precedent, a one-off paired PR would deviate from the established flow and create a second branch to keep in sync. That call belongs to the release drivers.

So the alternative is stated explicitly in the new subsection rather than quietly dropped: if parity on `dev` is wanted **before** the reconciliation cut rather than after, a dedicated `dev` PR carrying these commits is the path, and I'll open it on request.

Also confirming the finding's framing is right on the substance: this defect **is** live on `dev` today — `utils/retry.py` does not exist there, and `get_http_info` / `HttpxClient` are shared pipeline present on both branches.

---

totalfrank (commented):
Reviewed the one new commit since the last round: `95f9a399`, another merge of `origin/REL20260806` into this branch.

**Part A — open findings from prior rounds**

1. `plan.md`/`tasks.md` boundary-declaration rationale (inline thread on `plan.md`) — already resolved in `81acf361`, unaffected by this commit. Still resolved.

2. **Gate 0 / arch.rules.md Rule 16 — missing dev-branch propagation** (from the architecture-only review, `ad6cf092debdc6d3749c1c58ee97ec6ff54f218f`). The author's reply (issue comment `5213570283`) disputed the required fix direction: rather than opening a paired `dev` PR, they added a **Propagation scope** subsection to the PR description covering the five items Rule 16 actually requires (affected consumers, affected implementations, affected deployments/configuration, compatibility status, migration/propagation plan), on the grounds that Rule 16's text says changes "must declare their propagation scope," not that a parallel `dev` PR is mandatory.

   I re-read `docs/arch/arch.rules.md` Rule 16 directly rather than taking the reply on faith: its **Rule** clause is exactly "must declare... propagation scope," its **Required checks** are "include propagation analysis" / "merge is blocked when required impact analysis is absent" — no clause requires a paired PR against every affected branch. I also re-fetched the current PR description and confirmed the **Propagation scope** subsection is present under "Compatibility and risk," naming all five required items, the 221-commit `dev`/`REL20260806` divergence, the standing per-release reconciliation PRs (#90, #212/#213, #279, #536, #788) as the propagation mechanism, and stating the dedicated-`dev`-PR alternative explicitly rather than omitting it.

   The author's reasoning is sound — Rule 16 as written is a declaration requirement, and it's now met. Conceding this finding.

Open: 0.

**Part B — review of the new commit (`95f9a399`)**

Diffed `ad6cf092..95f9a399` directly (20 files, +609/-143): all changes belong to unrelated `engine_runtime`/`cron_relay`/`gating`/`routines` work merged in from `origin/REL20260806` (via `d6d8c8aa`, #887). None of it touches this PR's own files (`utils/retry.py`, `plugins/http_client.py`, `baas_service.py`'s `get_http_info`, `harness/services/llm.py`) or anything that calls into them. Clean sync merge, nothing to review.

**Status:** 0 open findings, 0 new findings. Clean as of this round (`95f9a399`).

CI: 5/7 checks green at time of review; `Backend unit tests` and `Singlebox coverage` still in progress (consistent with every prior round — not a code issue).

1. `plan.md`/`tasks.md` boundary-declaration rationale — fixed in `81acf361`, resolved.
2. Cookie jar persisted across calls (Group D) — found and fixed by the author's own self-review in `4fffb68a`, verified correct in the round on that commit.
3. Gate 0 / Rule 16 dev-branch propagation — conceded this round; Propagation scope subsection added to the PR description satisfies the rule as written.

Open: 0.

---

totalfrank (commented):
Reviewed the one new commit since the last round: `6d4e927f`, another merge of `origin/REL20260806` into this branch (bringing in #886, "forward template config to outbound rules").

**Part A — open findings from prior rounds**

1. `plan.md`/`tasks.md` boundary-declaration rationale (thread on `plan.md`) — resolved in `81acf361`. Still resolved, unaffected by this commit.
2. Cookie jar persisted across calls (Group D) — found and fixed by the author's own self-review in `4fffb68a`. Still fixed.
3. Gate 0 / Rule 16 dev-branch propagation — conceded in the prior round; the Propagation scope subsection remains present in the PR description, unaffected by this commit.

0 open findings carried over.

**Part B — review of the new commit (`6d4e927f`)**

Diffed `95f9a399..6d4e927f` directly (19 files, +316/-6): all changes belong to unrelated `bot_management` engine/outbound-rules work merged in from `origin/REL20260806` (#886) — `engines/aicoding/strategy.py`, `engines/default.py`, `engines/provisioning.py`, DI config/module wiring, `outbound_rules.py` plugin/API changes, and their tests. `baas_service.py` is touched, but only in `post_bots_api`'s provisioning/outbound-rule construction (new `theta_master_key_secret` param, `resolve_provisioning`/`build_extra_properties` call, `extra_properties` threaded into `_build_outbound_operation_rule`) — none of it touches `get_http_info`, `retry.py`, `http_client.py`, or `llm.py`'s aliasing, which is what this PR owns. Clean sync merge, nothing to review.

**Status:** 0 open findings, 0 new findings. Clean as of this round (`6d4e927f`).

CI: 5/7 checks green at time of review; `Backend unit tests` and `Singlebox coverage` still in progress (consistent with every prior round — not a code issue).

1. `plan.md`/`tasks.md` boundary-declaration rationale — fixed in `81acf361`, resolved.
2. Cookie jar persisted across calls (Group D) — found and fixed by the author's own self-review in `4fffb68a`, resolved.
3. Gate 0 / Rule 16 dev-branch propagation — conceded; Propagation scope subsection in the PR description satisfies the rule as written.

Open: 0.

---

totalfrank:
Closing — `REL20260806` is frozen, so this change goes in via `dev` instead.

**Superseded by #900**, which carries the identical 13 commits cherry-picked onto `dev` (zero conflicts) and re-verified against that base: utils/architecture/di 434 passed, plugins 1019 passed, service_bot/harness 1035 passed, SAST exit 0 on all six changed modules. The boundary premise was re-checked on `dev` rather than assumed to carry over — `core/service_bot/README.md` still declares the bare `agentclaw.community.utils`, so only `core/harness` needs the new entry.

Nothing is lost by closing this. For the record, what was established here and carried into #900:

- **Five clean review rounds, 0 open findings.** All three findings raised across the PR's life were resolved: the `plan.md`/`tasks.md` boundary-declaration rationale (`81acf361`), the pooled client persisting cookies across callers and tenants (`4fffb68a`), and the Gate 0 / Rule 16 propagation gap (conceded once the rule's text was re-read — it requires declaring propagation scope, not a paired PR).
- **All 7 CI checks green** on `6d4e927f`, across three separate base merges as `REL20260806` moved underneath it.
- **The `describe_exception` bug** found in self-review, where `httpx.HTTPError.request` raising from inside an `except` block replaced the transport failure being reported and silently killed the retry — fixed in `377fd540` and covered by regression tests.

Ironically, the Gate 0 review was right about the substance even though its prescribed fix was wrong for this repo: the fix really did need to be on `dev`, and #900 is now that. The release-branch route just turned out to be the one that closed.

---

PR #877, opened 2026-08-06
fix(baas): offload /http-info sync DB calls off the event loop

totalfrank:
## Problem

`GET /bots/{bot_uuid}/http-info` is an `async def` route, so it runs on the uvicorn worker's **event loop** — not the threadpool FastAPI would use for a plain `def` handler. Its resolution path called synchronous SQLAlchemy repositories inline:

| Step | Location | Blocking? |
| --- | --- | --- |
| `BotRepository.get_active_by_bot_uuid` | `dispatcher/_base_dispatcher.py` | yes — sync ORM |
| `DeviceRepository.list_by_bot_id` | `dispatcher/_base_dispatcher.py` | yes — sync ORM |
| `DeviceTemplateService.get_by_template_id` | `paas/_facade.py` | yes — sync ORM |
| `PaasServiceFactory.create` | `paas/_facade.py` | yes — issues its own `get_online_by_template_uuid` query |

The `device_uuid` branch (`_http_conn_info_dispatcher.py`) repeats the first two.

Each call parks the whole loop for the duration of the query, so every other request queued on that worker waits behind it. That head-of-line blocking matches the reported symptoms: bimodal `/http-info` latency (~60 ms or ~5000 ms with nothing between), 499s clustering across unrelated bots and unrelated paths at the same instant, and `$request_time ≈ 4996 ms` against the backend's 5 s deadline while BaaS itself never errored.

This is the BaaS-side cause deferred by the backend-side transport-resilience work, which only makes the caller tolerant of these stalls.

Two further defects surfaced while doing it:

**A cross-request session race, blocking the offload.** `with_orm_session` stored the session it opened as `self._session` on the repository, and repositories are DI singletons — one instance per process. That was safe only while every repository call ran to completion on the single event-loop thread. Offloading to worker threads removes the guarantee: thread B's assignment lands between thread A's assignment and A's query, so A executes against B's session — another request's transaction, potentially another tenant's data, or a SQLAlchemy error from driving one session from two threads. Raised in review; the fix is below.

**The CI summary swallowed its own failures.** `scripts/lib/common.sh` counted failing tests with `((log_failures++))`, whose post-increment returns exit status 1 on the first iteration. `_run_pytest` ends with a bare `set -e` that re-enables errexit for the rest of the run, so that status aborted `_summary` at the first counted failure. Every BaaS CI failure printed the summary header, no verdict, and exited 1 — red job, no diagnosis.

## Solution

**Offload the blocking calls** with `asyncio.to_thread`, following the existing house style in `core/service/session_file_sharing/_dispatcher.py`. The device-resolution fix lands in the shared `BotBaseDispatcher._resolve_bot_device`, so the sibling `bot_runtime` dispatchers that reuse it (`http`, `wss`, `cmd`, `open_folder`, `file_transfer`) get the same benefit. The facade offloads both the template lookup and `factory.create`.

**Bind ORM sessions per context, not per instance.** `with_orm_session` now binds into a `ContextVar` keyed by repository instance. `asyncio.to_thread` runs its callable in a copy of the calling context and each asyncio task gets its own copy, so a session is visible only to the call that opened it; the decorator resets its token on exit, so a binding also cannot outlive its call — nested calls on one repository restore the outer session instead of leaving the inner one bound, which the attribute version got wrong too. `OrmConnectionMixin._session` became a property reading that binding, so none of the 222 decorated repository methods changed. Reading it outside a session scope now raises `RuntimeError` naming the repository rather than silently using whatever the last caller left behind.

This also covers `session_file_sharing` and `file_transfer`, which were already calling repositories from worker threads before this PR — the hazard existed there, it just hadn't been named.

**The CI summary counter** becomes a plain assignment. Exit codes are unchanged — a clean run still prints `All stages passed!` and returns 0, a failing run still returns 1 — but it now names the failing stages and tests first. It proved itself immediately: the architecture-rule failure hit while developing the session fix was reported by name instead of vanishing.

Rejected alternatives:

- **Pass the pre-resolved template into `factory.create` to skip its second query.** Tempting (the factory's docstring supports it), but not semantics-preserving: the facade resolves via `get_by_template_id`, whereas the factory resolves via `get_online_by_template_uuid`, which additionally enforces that the template is online. Handing over the pre-resolved object would silently drop that check, so the query was kept and merely moved off the loop.
- **Convert the routes to plain `def` and let FastAPI's threadpool handle them.** Smaller per-route diff, but it changes concurrency characteristics and interacts with the `@inject` / `Provide[...]` DI wiring — a bigger blast radius than this fix warrants.
- **Thread-local storage for the session** instead of a `ContextVar`. Sufficient for the thread case, since the decorated body never yields, but it would not isolate two asyncio tasks sharing the loop thread. The `ContextVar` covers both.
- **Replace `NullPool` with a real pool.** Not actionable in this repository, see below.

### Findings on the surrounding questions

- **`NullPool` (`core/database/_manager.py:315`) is unreachable in this repository.** `DatabaseManager.init_orm_session()` has no callers anywhere in the tree; DB init always flows through `DatabaseManagerLifecycle.start()` → `plugin.init_database()` → `db_manager.init_plugin()`, and every session then delegates to the plugin. The connection-pooling fix therefore belongs in the `ZDAS_ORM` plugin implementation, which is not part of this repository. Git blame is uninformative — the file arrives whole in the initial import commit.
- **The bundled `SqliteOrmPlugin` uses `StaticPool` with `check_same_thread=False`**, i.e. it is already built for cross-thread access; the full ASGI E2E suite passes against it with these calls moved to worker threads.
- **Remaining scope is not addressed here.** There are ~143 `async def` route handlers and 222 `@with_orm_session` repository methods, so the same trap exists on routes outside this path. Fixing them in one sweep was deliberately avoided; the non-`/http-info` routes need a separately tracked item. The session-binding fix means those routes can now be converted safely, one at a time.

## Validation

Run from `src/baas`, on this branch as based on `REL20260806`:

- `bash scripts/ci_test.sh --base origin/REL20260806` — the full CI pipeline, exit 0. All three stages green, plus both gates: case pass rate 100.00% (7993/7993), line coverage 92.48%, **change line coverage 100.00% (8/8)**; ASGI gate 100.00% (1233/1233) at 55.45% line coverage.
- `pytest tests/e2e/asgi` — stable across repeated standalone runs (1190 passed, 43 skipped). Exercises the real ASGI stack against the SQLite/`StaticPool` backend, covering the cross-thread concern above.
- `ruff check src tests` — clean; `ruff format --check src tests` — 1321 files already formatted.
- `mypy src tests` — no new errors versus the pre-change baseline. The repository-wide total of 4043 errors is pre-existing and unrelated.

All the new tests are deterministic rather than timing-based, and each was verified to fail without its fix:

- **Event-loop guards** (4): the fake repository blocks its thread until a *loop-side* coroutine releases it, so an inline call deadlocks and fails instead of depending on a latency threshold.
- **Session-isolation guards** (5): two concurrent `to_thread` calls enter one repository instance and are parked *inside* the method on a `threading.Barrier` — the exact A/B interleaving from review. On the previous implementation the earlier caller finishes against the later caller's session.
- **CI summary**: with a `FAILED …::…` line in the log it now prints the failing stage and test list and returns 1 (previously: header only, silent exit 1); with a clean log it still prints `All stages passed!` and returns 0.

Not run: the load test asserting that unrelated endpoints on the same worker stay responsive under concurrent `/http-info` traffic, and confirmation against production `orm-timing` logs for the 20:41 window — neither production logs nor a load-test environment is reachable from here. The `orm-timing` line (`core/repository/__init__.py`) already splits `acquire` / `query` / `commit`; if `acquire` dominates in that window the connection-churn reading holds, and if `query` dominates the real problem is slow queries instead.

One note for anyone editing `core/repository/__init__.py`: the pre-push SAST gate resolves whichever `flake8` is on `PATH`, and reports `E999 SyntaxError` on that file if it is a CPython 3.11 build — the file uses PEP 695 generics (`def with_orm_session[**P, R]`), which need 3.12, the version this project targets. It is not caused by this change; the unmodified file fails the same way. Point `PYTHON_SAST_CMD` at a 3.12+ interpreter and the gate passes with all rules enabled.

## Compatibility and risk

No contract, schema, config, or migration impact. Return values, raised exceptions, and call arguments are unchanged — only the thread the synchronous work runs on, and where the open session is stored. Repository calls now execute on worker threads, so each holds its own session concurrently with other in-flight requests rather than being serialized by the event loop.

The one behavioral change beyond that: reading `_session` outside a `@with_orm_session` scope now raises `RuntimeError` instead of returning a stale session. Nothing in the tree does so — the decorator was the only writer — and the setter is retained for direct assignment, so injected sessions in tests keep working.

The CI script change affects reporting only — no gate is added, removed, or weakened, and module selection is untouched. Rollback is a straight revert of any of the three commits, though reverting the offload without the session fix is the wrong half to keep.

The six originally touched files are byte-identical on `dev` and `REL20260806`, so the commits cherry-picked onto the release branch without conflict.

### DIFF
--- a/src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_base_dispatcher.py
+++ b/src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_base_dispatcher.py
@@ -13,6 +13,7 @@ Design decisions (per design.md):
 
 from __future__ import annotations
 
+import asyncio
 from typing import TYPE_CHECKING, Any
 
 from secbaas.community.api.bot_runtime import (
@@ -85,14 +86,20 @@ class BotBaseDispatcher:
             RuntimeError: Selected device has no provider_device_id
         """
         # 1. Look up bot by UUID
-        bot = self._bot_repo.get_active_by_bot_uuid(bot_uuid, tenant, env)
+        # Repositories are synchronous (blocking DB I/O); dispatchers run on the
+        # uvicorn event loop, so every repo call is offloaded to a worker thread.
+        # Running them inline parks the loop for the duration of the query and
+        # stalls every other request served by the same worker.
+        bot = await asyncio.to_thread(
+            self._bot_repo.get_active_by_bot_uuid, bot_uuid, tenant, env
+        )
         if not bot:
             logger.warning(f"Bot not found: bot_uuid={bot_uuid}, tenant={tenant}")
             raise BotNotFoundError(bot_uuid)
 
         # 2. Get devices associated with bot
-        devices = self._device_repo.list_by_bot_id(
-            bot_id=bot.id, tenant=tenant, env=env
+        devices = await asyncio.to_thread(
+            self._device_repo.list_by_bot_id, bot_id=bot.id, tenant=tenant, env=env
         )
         if not devices:
             logger.warning(f"No devices found for bot: bot_uuid={bot_uuid}")
--- a/src/baas/src/secbaas/community/core/service/paas/_facade.py
+++ b/src/baas/src/secbaas/community/core/service/paas/_facade.py
@@ -1033,7 +1033,12 @@ class PaasServiceFacade(PaasServiceFacadeProtocol):
         service = None
         try:
             # Step 1 & 2: Parse device ID and lookup template
-            template = self._device_template_service.get_by_template_id(
+            # DeviceTemplateService and PaasServiceFactory both issue synchronous
+            # DB queries. This coroutine runs on the uvicorn event loop, so both
+            # are offloaded to a worker thread — running them inline blocks the
+            # loop and stalls unrelated requests on the same worker.
+            template = await asyncio.to_thread(
+                self._device_template_service.get_by_template_id,
                 template_id=template_id,
             )
             if not template:
@@ -1049,8 +1054,10 @@ class PaasServiceFacade(PaasServiceFacadeProtocol):
                 )
 
             # Step 3: Create service via Factory using resolved template
-            service = self._factory.create(
-                tenant_name=template.tenant, template_uuid=template.template_uuid
+            service = await asyncio.to_thread(
+                self._factory.create,
+                tenant_name=template.tenant,
+                template_uuid=template.template_uuid,
             )
 
             # Step 4: Delegate to service polymorphic method
--- a/src/baas/scripts/lib/common.sh
+++ b/src/baas/scripts/lib/common.sh
@@ -89,7 +89,12 @@ _summary() {
         grep 'FAILED.*::' "$LOG_FILE" > "$tempfile" || true
         while IFS= read -r line; do
             FAILED_TESTS+=("$line")
-            ((log_failures++))
+            # Assign rather than ((log_failures++)): the post-increment form
+            # evaluates to 0 on the first pass and returns exit status 1, which
+            # aborts this function under the errexit that _run_pytest re-enables
+            # — the summary would then print its header and exit without ever
+            # naming the failing stages or tests.
+            log_failures=$(( log_failures + 1 ))
         done < "$tempfile"
         rm -f "$tempfile"
     fi
--- a/src/baas/src/secbaas/community/core/repository/__init__.py
+++ b/src/baas/src/secbaas/community/core/repository/__init__.py
@@ -8,6 +8,7 @@ Public exports:
 from __future__ import annotations
 
 from collections.abc import Callable
+from contextvars import ContextVar
 from datetime import datetime
 from functools import wraps
 from typing import Any, TypeVar
@@ -15,6 +16,29 @@ from typing import Any, TypeVar
 from secbaas.community.spi.database import PluginDatabaseType
 
 R = TypeVar("R")
+
+# Open ORM sessions for the current execution context, keyed by repository
+# instance id.
+#
+# Repositories are DI singletons (``providers.Singleton`` in
+# ``bootstrap/_core_repository.py``), so one instance is shared by every caller
+# in the process. Storing the open session as a plain ``self._session``
+# attribute was safe only while every repository call ran to completion on the
+# single event-loop thread. Once callers offload repository calls to worker
+# threads (``asyncio.to_thread``), two concurrent requests race: thread B
+# overwrites ``self._session`` between thread A's assignment and A's query, and
+# A then runs against B's session — the wrong tenant's data, or a SQLAlchemy
+# error from using one session on two threads.
+#
+# A ContextVar removes the sharing. ``asyncio.to_thread`` runs its callable in a
+# copy of the calling context, and every asyncio task gets its own copy too, so
+# a session bound here is visible only to the call that opened it.
+#
+# Always bind by replacing the mapping (``{**current, id(self): session}``),
+# never by mutating it in place: copying a context copies the reference, not the
+# dict, so an in-place write would leak straight back across the very thread
+# boundary this exists to close.
+_active_sessions: ContextVar[dict[int, Any]] = ContextVar("orm_active_sessions")
 
 
 def with_orm_session[
@@ -34,13 +58,20 @@ def with_orm_session[
         t0 = time.monotonic()
         try:
             with self._database.orm_session() as session:
-                self._session = session
-                conn_id = (
-                    id(session.connection().connection) if session.is_active else "N/A"
+                token = _active_sessions.set(
+                    {**_active_sessions.get({}), id(self): session}
                 )
-                session_start = time.monotonic()
-                result = func(self, *args, **kwargs)
-                query_end = time.monotonic()
+                try:
+                    conn_id = (
+                        id(session.connection().connection)
+                        if session.is_active
+                        else "N/A"
+                    )
+                    session_start = time.monotonic()
+                    result = func(self, *args, **kwargs)
+                    query_end = time.monotonic()
+                finally:
+                    _active_sessions.reset(token)
             total_end = time.monotonic()
 
             _perf_logger.info(
@@ -84,11 +115,36 @@ class OrmConnectionMixin:
     the ``@with_orm_session`` decorator, plus common utility methods
     (e.g. :meth:`now`) so individual ORM repositories don't need to
     re-implement the same logic.
+
+    ``_session`` reads the session bound to this repository in the *calling*
+    context, not a shared instance attribute — repositories are singletons and
+    their methods run on worker threads, so an attribute would be shared by
+    concurrent requests. See :data:`_active_sessions`.
     """
 
     _database: Any
-    _session: Any
     _datasource_name: str | None = None
+
+    @property
+    def _session(self) -> Any:
+        """The ORM session opened for the current ``@with_orm_session`` call."""
+        session = _active_sessions.get({}).get(id(self))
+        if session is None:
+            raise RuntimeError(
+                f"{type(self).__name__}: no ORM session bound in this context. "
+                "Repository methods must be called through @with_orm_session."
+            )
+        return session
+
+    @_session.setter
+    def _session(self, session: Any) -> None:
+        """Bind a session for the current context (tests and direct callers).
+
+        The binding is not unwound on scope exit — only ``with_orm_session``
+        tracks a token for that. It stays context-local either way, so an
+        assignment here can never be seen by another request.
+        """
+        _active_sessions.set({**_active_sessions.get({}), id(self): session})
 
     @with_orm_session
     def now(self) -> datetime:
--- a/src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_http_conn_info_dispatcher.py
+++ b/src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_http_conn_info_dispatcher.py
@@ -6,6 +6,7 @@ Supports both auto-selection (default) and targeted device selection via device_
 
 from __future__ import annotations
 
+import asyncio
 from typing import TYPE_CHECKING
 
 from secbaas.community.api.bot_runtime import (
@@ -121,14 +122,18 @@ class DefaultBotHttpConnInfoDispatcher(BotBaseDispatcher, BotHttpConnInfoDispatc
         )
 
         # 1. Look up bot
-        bot = self._bot_repo.get_active_by_bot_uuid(bot_uuid, tenant, env)
+        # Sync repositories are offloaded to a worker thread so the blocking DB
+        # call does not park the event loop (see BotBaseDispatcher).
+        bot = await asyncio.to_thread(
+            self._bot_repo.get_active_by_bot_uuid, bot_uuid, tenant, env
+        )
         if not bot:
             logger.warning(f"Bot not found: bot_uuid={bot_uuid}, tenant={tenant}")
             raise BotNotFoundError(bot_uuid)
 
         # 2. List all devices for this bot and find the specific one
-        devices = self._device_repo.list_by_bot_id(
-            bot_id=bot.id, tenant=tenant, env=env
+        devices = await asyncio.to_thread(
+            self._device_repo.list_by_bot_id, bot_id=bot.id, tenant=tenant, env=env
         )
         # fmt: off
         device = next(
--- a/src/baas/tests/unit/core/repository/test_with_orm_session_isolation.py
+++ b/src/baas/tests/unit/core/repository/test_with_orm_session_isolation.py
@@ -1,0 +1,136 @@
+"""Session isolation tests for ``@with_orm_session``.
+
+Repositories are DI singletons, so one instance serves every request in the
+process. These tests pin down that the session opened for one call is never
+visible to another — the property that makes it safe for an ``async def``
+handler to offload repository calls with ``asyncio.to_thread``.
+
+The tests are deterministic rather than timing-based: a barrier holds both
+callers inside their repository method at the same moment, which is exactly the
+window in which a shared ``self._session`` attribute gets overwritten.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import threading
+from contextlib import contextmanager
+
+import pytest
+
+from secbaas.community.core.repository import OrmConnectionMixin, with_orm_session
+
+
+class _FakeSession:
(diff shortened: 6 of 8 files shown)


totalfrank:
"BaaS unit tests" went red on 322ef1c, and the log gives no reason — that turned out to be a bug in the CI summary itself, which I've pushed a fix for in d9ad767.

**What the log shows.** The final stage passed (`7972 passed, 14 skipped, 453 deselected`), then:

```
===========================================
          BAAS CI Test Summary
  Duration: 5m 45s  |  Log: .../src/baas/tmp/test.log
===========================================
##[error]Process completed with exit code 1.
```

Neither `All stages passed!` nor `FAILED: N stage(s)` is printed — so `_summary` died between the header and the verdict.

**Why.** `scripts/lib/common.sh` counted failures with `((log_failures++))`. Post-increment evaluates to 0 on the first iteration and returns exit status 1. `_run_pytest` ends with a bare `set -e`, which re-enables errexit for the remainder of the run even though `ci_test.sh` disabled it around the pipeline — so the first counted failure aborted the function. Reproduced locally: with a `FAILED …::…` line in the log, `_summary` prints its header and exits 1 silently; after the fix it prints the failing stage and test list and still exits 1.

The practical effect is that this has been true for *every* BaaS CI failure — red job, correct exit code, no diagnosis. The BaaS job is skipped on PRs that don't touch `src/baas`, which is likely why it went unnoticed.

**Where that leaves this PR.** The failing stage is one of `run_e2e_boot_tests` / `run_e2e_asgi_tests` / `run_ci_tests`, and its name was swallowed. What I can say:

- The whole CI script reproduces green locally on this branch — `bash scripts/ci_test.sh --base HEAD^1` exits 0, all three stages plus both coverage gates. The ASGI suite is stable across repeated runs (1190 passed, three times).
- The unit stage passed in CI with exactly the local counts, so the new regression tests are not the failure.
- The boot stage is 10 WebSocket-management tests against a live app on port 8888 — no `/http-info`, no dispatcher, no PaaS facade. Nothing in this diff is on that path.
- `dev` is this branch's parent, so the merge commit CI tested is content-identical to what I ran locally.

I could not read the failing stage's output directly: `get_job_logs` only returns the tail, which lands inside the last stage, and the artifact host is blocked from this environment.

The rerun on d9ad767 should now name the failure. I'll pick it up from there — if it's a real failure in this diff I'll fix it, and if it's the boot stage flaking or already broken on `dev` I'll say so with the evidence.

---

cassiuscai (dismissed):
LGTM

src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_base_dispatcher.py
```diff
@@ -85,14 +86,20 @@ async def _resolve_bot_device(
             RuntimeError: Selected device has no provider_device_id
         """
         # 1. Look up bot by UUID
-        bot = self._bot_repo.get_active_by_bot_uuid(bot_uuid, tenant, env)
+        # Repositories are synchronous (blocking DB I/O); dispatchers run on the
+        # uvicorn event loop, so every repo call is offloaded to a worker thread.
+        # Running them inline parks the loop for the duration of the query and
+        # stalls every other request served by the same worker.
+        bot = await asyncio.to_thread(
```
totalfrank, line 93: **Blocking:** offloading these calls to `asyncio.to_thread` introduces a real cross-request data race that didn't exist before.

`with_orm_session` (`core/repository/__init__.py:33`) stores the opened session as an instance attribute — `self._session = session` — not in thread-local or per-call state. `self._bot_repo` / `self._device_repo` are `providers.Singleton(...)` (`bootstrap/_core_repository.py`), and `bot_http_conn_info_dispatcher` (and the five sibling dispatchers that reuse `_resolve_bot_device`) is also a `Singleton`. Before this change, the whole synchronous call ran atomically on the single event-loop thread, so no two callers could ever interleave inside `with_orm_session`. `asyncio.to_thread` removes exactly that guarantee: two concurrent `/http-info` requests now run `get_active_by_bot_uuid`/`list_by_bot_id` on separate OS threads against the *same* repo instance (the default executor here is not serialized to one worker).

Concrete sequence: request A and request B hit the same worker at once for different bots/tenants. Thread A opens session S_A and sets `self._session = S_A`; before A's `func(self, ...)` reads it, thread B opens session S_B and overwrites `self._session = S_B`. A's query now executes against B's session — either A observes B's data (cross-tenant leak) or SQLAlchemy raises because a session is being used/closed concurrently from another thread. The new regression tests only exercise one repo call at a time (a single blocking side_effect released by a loop-side coroutine), so they can't catch this.

The same pattern applies to the offloaded calls in `_facade.py` (`DeviceTemplateService` / `PaasServiceFactory`, both also singletons) and to the duplicated bot/device lookup in `_http_conn_info_dispatcher.py:125-138`.

This needs a fix in `with_orm_session`/`OrmConnectionMixin` itself (e.g. a `contextvars.ContextVar` per call, or returning the session instead of stashing it on `self`) before any of these repos are safe to call from worker threads — not something fixable at the call site.

---

src/baas/src/secbaas/community/core/service/paas/_facade.py
```diff
@@ -1033,7 +1033,12 @@ async def resolve_invoke_http_info(
         service = None
         try:
             # Step 1 & 2: Parse device ID and lookup template
-            template = self._device_template_service.get_by_template_id(
+            # DeviceTemplateService and PaasServiceFactory both issue synchronous
+            # DB queries. This coroutine runs on the uvicorn event loop, so both
+            # are offloaded to a worker thread — running them inline blocks the
+            # loop and stalls unrelated requests on the same worker.
+            template = await asyncio.to_thread(
```
totalfrank, line 1040: **Blocking:** same root cause as the `_base_dispatcher.py` finding on this PR — `self._device_template_service` and `self._factory` both resolve to `providers.Singleton(...)`, and their underlying repositories use `with_orm_session`, which stashes the open session on the shared instance (`self._session = session` in `core/repository/__init__.py`). Two concurrent `resolve_invoke_http_info` calls now run `get_by_template_id` / `factory.create` → `get_default_or_explicit_template` on separate worker threads against the same `OrmDeviceTemplateRepository`/template-service instance, so one request's session assignment can be overwritten by another's before its query runs — cross-request/cross-tenant data or a SQLAlchemy error under concurrent `/http-info` traffic, which is precisely the load pattern this PR is meant to unblock.

---

totalfrank (commented):
Reviewed commits `8dc1f49` and `a86a4877` (head) against merge-base `73de789`. Files covered: `src/baas/scripts/lib/common.sh`, `core/service/bot_runtime/dispatcher/_base_dispatcher.py`, `core/service/bot_runtime/dispatcher/_http_conn_info_dispatcher.py`, `core/service/paas/_facade.py`, and the two new test files.

The CI summary fix (`common.sh`) is straightforward and correct. The `asyncio.to_thread` offloading correctly fixes the event-loop-blocking problem described in the PR, and the new regression tests genuinely prove the loop stays responsive during a single blocking call. However, moving these calls onto worker threads exposes a pre-existing concurrency hazard in `with_orm_session` (session stored on the shared singleton `self`, not thread-local) that was previously masked by everything running serialized on the event loop — left two inline comments with the concrete failure sequence.

Categories applied: correctness, concurrency/resource safety, test coverage of the change. Contract/schema, security, and architecture-boundary categories weren't implicated by this diff.

---

src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_base_dispatcher.py
totalfrank: Confirmed and fixed in 162c6811 — you were right on every point, including that it isn't fixable at the call site.

Verified before changing anything: repositories are `providers.Singleton` via `_orm_repo` (`bootstrap/_core_repository.py:48-50`), `bot_http_conn_info_dispatcher` and `paas_facade` are `Singleton` too, and `self._session = session` at `core/repository/__init__.py:37` was the only writer in the tree — so the shared instance really is the whole of the state.

The fix binds the session in a `ContextVar` keyed by repository instance instead of on `self`. `asyncio.to_thread` runs its callable in a copy of the calling context, and each asyncio task gets its own copy, so a session is visible only to the call that opened it. `with_orm_session` resets its token on exit, which also stops a binding from outliving its call — nested calls on one repository now restore the outer session instead of leaving the inner one bound, which the attribute version got wrong too.

`OrmConnectionMixin._session` became a property reading that binding, so none of the 222 decorated methods changed. Reading it outside a session scope now raises `RuntimeError` naming the repository rather than silently using whatever the last caller left behind.

On the test gap — also correct, and it was the part worth fixing properly. `tests/unit/core/repository/test_with_orm_session_isolation.py` drives two concurrent `to_thread` calls into one repository instance and parks both *inside* the method on a `threading.Barrier`, which is exactly the A/B interleaving you described. It fails on the previous implementation (`first_call[0] is first_call[1]` — the earlier caller finishes against the later caller's session) and passes now. A second test asserts the binding is gone once the call returns.

Since the fix is in `with_orm_session` itself, it covers the `_facade.py` and `_http_conn_info_dispatcher.py` sites in your other comment, and equally the `session_file_sharing` / `file_transfer` dispatchers that were already calling repositories from worker threads before this PR — the hazard existed there, it just hadn't been named.

Full pipeline green on the new commit: `bash scripts/ci_test.sh --base origin/REL20260806` exits 0, 7993/7993 cases, change-line coverage 100%.

---

src/baas/src/secbaas/community/core/service/paas/_facade.py
totalfrank: Same root cause, so the same fix covers it — 162c6811 binds the session in a `ContextVar` inside `with_orm_session` rather than on the shared instance, which fixes every call site at once rather than per-caller. Details and the reproducing test are in the reply on the `_base_dispatcher.py` thread.

Confirmed the specifics you named here: `device_template_service_infra`, `paas_service_factory` and `paas_facade` are all `providers.Singleton` (`bootstrap/_core_services.py:274-296`), and `factory.create` reaching `get_default_or_explicit_template` is a second `OrmDeviceTemplateRepository` entry on the same instance within one request — so these two offloaded calls could have raced each other, not just across requests.

---

totalfrank (commented):
Reviewed commit `162c6811` (head, no new commits since the last round). Verified the fix independently by reading the diff rather than trusting the commit message.

**Fix verified.** `with_orm_session` (`core/repository/__init__.py`) now binds the ORM session in a `ContextVar` keyed by `id(self)`, updated by replacing the dict (`{**current, id(self): session}`) rather than mutating it in place — which matters, since `asyncio.to_thread` copies the *context*, not the values in it, so an in-place mutation would have leaked the write back across the thread boundary this fix exists to close. `OrmConnectionMixin._session` is now a property over that binding (222 decorated methods unchanged), and the token is reset in a `finally` so the binding never outlives the call, restoring the outer session correctly on nested calls. The new `test_with_orm_session_isolation.py` drives two concurrent `asyncio.to_thread` calls into one singleton repository, holds both inside the method on a `threading.Barrier` at the exact interleaving point the findings described, and asserts neither caller's session changes under it — this reproduces the original race and would fail against the old attribute-based implementation. No other file in this commit is implicated in the fix; `_async_session_client.py`'s only change is an unrelated quote-style formatting diff (single→double quotes), harmless and linter-territory.

## Status

Commit reviewed this round: `162c6811`

| Finding | Location | Status |
| --- | --- | --- |
| Session race via shared `self._session` on singleton repo | `_base_dispatcher.py:93` | **Fixed** — resolved |
| Same root cause in facade template/factory lookups | `_facade.py:1040` | **Fixed** — resolved |

Open findings: **0**. Nothing new found in this round. The PR is clean as of `162c6811`.

---

totalfrank:
Thanks for the independent verification — and for catching the stray file. Head is now `2dc72b07`.

The `_async_session_client.py` quote-style change was unrelated churn, not intentional: the pipeline's `check-basic` stage runs `ruff format .` in write mode, so a local full-pipeline run reformats whatever the formatter is currently unhappy with, and my `git add -A` swept it into the commit. That file has pre-existing formatter drift on `REL20260806`, so it had nothing to do with this PR. Reverted to its `REL20260806` state and amended — `2dc72b07` is `162c6811` minus exactly that one line, with the session fix and its tests unchanged.

That drop is safe for CI: `run_check_basic` runs `ruff format .` (write), not `--check`, so CI reformats the file in its own working tree and stays green rather than failing on it. Re-ran `tests/unit/core/repository` plus the `_async_session_client` tests after the amend — 843 passed.

So the diff is now exactly the three things it claims: the `to_thread` offload, the per-context session binding, and the CI summary counter.

---

totalfrank (commented):
Reviewed head `2dc72b07`.

Note on scope: my previous review was posted against `162c6811`, but that commit is no longer reachable — it was amended into `2dc72b07` (dropping one unrelated file, see below), not extended by a new commit. Rather than diffing a commit that no longer exists, I diffed the current merge base against `2dc72b07` directly and compared it against the content I already reviewed at `162c6811`.

**Result: the diff is byte-for-byte the same reviewed content, minus `_async_session_client.py`.** That file's only change at `162c6811` was an unrelated quote-style formatting edit (single→double quotes) that the author identified as accidental `ruff format` + `git add -A` churn and reverted to its `REL20260806` state, per their comment on this PR. Every other file — `common.sh`, `core/repository/__init__.py`, `_base_dispatcher.py`, `_http_conn_info_dispatcher.py`, `_facade.py`, and both test files — is unchanged from what was already reviewed and cleared. No new code to review, no new findings.

## Status

Commits reviewed this round: `2dc72b07` (amend of `162c6811`, previously-reachable commit no longer in history)

| Finding | Location | Status |
| --- | --- | --- |
| Session race via shared `self._session` on singleton repo | `_base_dispatcher.py:93` | Fixed — resolved (as of `162c6811`, unchanged since) |
| Same root cause in facade template/factory lookups | `_facade.py:1040` | Fixed — resolved (as of `162c6811`, unchanged since) |

Open findings: **0**. Nothing new found this round. The PR remains clean as of `2dc72b07`.

---

cassiuscai (approved):

PR #879, opened 2026-08-07
fix(auth): expose empty tenant on user principal

panqiwp:
## Summary
- Add a read-only compatibility `tenant` property on `UserPrincipal` returning `None`
- Keep user principals tenant-less semantically: this is not a Pydantic field, so forged user `tenant` input is still ignored rather than honored
- Add verifier coverage for `principal.tenant is None`

## Tests
- `cd ocb-public/src/backend && uv run pytest tests/community/core/gateway_principal/test_verifier.py -q`

### DIFF
--- a/src/backend/src/agentclaw/community/core/gateway_principal/models.py
+++ b/src/backend/src/agentclaw/community/core/gateway_principal/models.py
@@ -132,6 +132,17 @@ class UserPrincipal(BaseModel):
     type: Literal[PrincipalType.USER] = PrincipalType.USER
     subject: GatewayUser
 
+    @property
+    def tenant(self) -> None:
+        """User principals assert no isolation tenant.
+
+        Kept as a read-only compatibility attribute for callers/tests that
+        check ``principal.tenant is None``. It is intentionally not a Pydantic
+        field, so a forged ``tenant`` key on a user principal is still ignored
+        rather than honoured as an isolation claim.
+        """
+        return None
+
 
 class BotPrincipal(BaseModel):
     """A bot/agent calling as a first-class caller in its own right."""
--- a/src/backend/tests/community/core/gateway_principal/test_verifier.py
+++ b/src/backend/tests/community/core/gateway_principal/test_verifier.py
@@ -139,6 +139,7 @@ def test_user_only_token_yields_the_internal_tenant_and_the_owner():
     assert caller.tenant == DEFAULT_AVERNET_TENANT
     assert caller.user_id == "u-1"
     assert isinstance(caller.principals[0], UserPrincipal)
+    assert caller.principals[0].tenant is None
 
 
 def test_user_wins_over_app_when_both_identities_are_present():


totalfrank (changes requested):
You need to know what you are doing when sending out PRs.

totalfrank (changes requested):

## 2. Gate 0 — Target branch

- **Classification:** existing-behavior fix — a follow-up compatibility shim for #809 ("fix(gateway): drop tenant from the user principal"), which removed the `tenant` field from `UserPrincipal` on this release branch. Path (a) would be the right lane, but **Gate 0 fails** on the dev-counterpart requirement, and this PR compounds a pre-existing instance of it rather than fixing it:

  - Verified directly against `origin/dev`'s `src/backend/src/agentclaw/community/core/gateway_principal/models.py`: `UserPrincipal` there still declares `tenant: str` as a **required** field. `origin/REL20260806`'s `UserPrincipal` has **no** `tenant` field at all (removed by #809). The two branches have already diverged on this model's shape.
  - Searched merged and open PRs for a `dev`-targeted port of #809 (`repo:inclusionAI/Avernet tenant UserPrincipal base:dev`) — none exists. #809 itself shipped only to `REL20260806`.
  - This PR adds a **second** REL-only patch on top of that already-undelivered change: a read-only `tenant -> None` compatibility property, so that `principal.tenant is None` keeps working for `UserPrincipal` post-#809. Both pieces (the field removal and this compatibility accessor) live only on the release branch.
  - **Consequence:** at the next release cut from `dev`, `UserPrincipal.tenant` reverts to a required `str` with no default — silently undoing both #809 and this PR, and reintroducing the original defect (or worse, a hard validation error wherever a `UserPrincipal` is now constructed without a tenant, since callers on this branch have since been written assuming `tenant` is absent/`None`). This is regression-by-omission, and it is actively getting harder to unwind the longer REL and `dev` diverge on this model.
  - No defect/issue reference, and no mention of a `dev` counterpart or cherry-pick relationship, appears in the PR body.

## 3. Findings

**F1 — No `dev` counterpart for this or its parent change (#809); `UserPrincipal.tenant` shape has diverged between branches.** See Gate 0 above for the verified evidence. **Required fix:** open (or link) a `dev` PR that ports #809's `UserPrincipal` model change and this PR's compatibility property together, reconciling `dev`'s currently-required `tenant: str` field with REL's tenant-less model, before this merges — or accept this only under a written waiver (violated rule, reason, risk of the branch divergence, compensating control such as a tracked follow-up issue with an owner and date, per the Governance Addendum). Deferral to "track separately" is not sufficient on its own; the Gate 0 dev-counterpart requirement exists specifically to prevent this class of silent revert.

## 4. Out-of-scope notes

- The design rationale for why user principals should be tenant-less (a person's credential proves no tenant; machine identities are what get registered to one) is a business/domain decision already made in #809 and not re-litigated here.
- The property being a non-Pydantic read-only accessor (so a forged `tenant` key on a user identity is still ignored) is the correct shape for this compatibility shim and is not flagged.

---

totalfrank (changes requested):
**Architecture-only review.** Business logic (the `None` semantics themselves) is out of scope and assumed correct; this is a structural/contract review of a principal/auth wire-adjacent DTO change targeting a release branch.

## 2. Gate 0

- **Target branch:** `REL20260806` — matches `^REL\d{8}`, a release branch. Only (a) an existing-behavior bug fix, or (b) an isolated new feature (heightened review), may target it.
- **PR classification:** claims `fix(auth)`, but the diff and PR description do not identify any reproduced defect. `UserPrincipal` gains a read-only `@property tenant` that always returns `None`. Confirmed at runtime (pydantic 2.13, as used by this repo) that a plain `@property` — not `@computed_field` — is excluded from `model_dump()`/`model_dump_json()`, so this does **not** reopen the wire-serialization contract PR #809 closed; no `tenant: null` key reappears in the forwarded/logged JSON. That rules out the worst-case reading, but it does not make this an existing-behavior fix: the property's own docstring says it exists "for callers/tests that check `principal.tenant is None`" — an anticipatory shim for unnamed future callers, not a correction of anything currently broken. It also isn't an isolated feature: `UserPrincipal` sits on the live `/openapi/v1` auth path for every request, so it doesn't get the "not participating in any existing pipeline" exception either.
- **Relationship to #809:** #809 (merged, REL20260806) deliberately removed the required `tenant: str` field from both the gateway's and backend's `UserPrincipal`, documented in `src/gateway/docs/2026-07-21-auth-design.md` §4.6 and this module's own `README.md` "Change impact" log, on the grounds that "a tenant is a property of the calling program, not of a person." #879 reintroduces the `tenant` name on the backend DTO. It is inert (verified above) and its own docstring explains the non-field choice, so this is not a silent contract reopening — but it is still unexplained scaffolding on top of a design #809 just finished narrowing, with no caller or test named that needs it today.
- **Dev-propagation check (the sharper problem):** `git show origin/dev:src/backend/src/agentclaw/community/core/gateway_principal/models.py` shows `dev`'s `UserPrincipal` still declares `tenant: str` as a **required** field:
  ```
  106:class UserPrincipal(BaseModel):
  107:    """A first-party authenticated user."""
  108:
  109:    model_config = ConfigDict(frozen=True)
  110:
  111:    type: Literal[PrincipalType.USER] = PrincipalType.USER
  112:    tenant: str
  ```
  #809's fix never reached `dev` — `dev` is still sitting on exactly the contract mismatch #809's own PR description reproduced (gateway sends `tenant: null`, backend requires `str`, request 401s). This is not "divergent-but-equivalent tenant-less behavior" that would make REL20260806's state reconcilable as REL-specific; it's the same live bug, un-fixed, on `dev`. #879 adds further asymmetric surface to the REL-only branch without acknowledging this drift, ahead of an eventual REL→dev rebase (this repo has that pattern, e.g. #788 "rebase REL20260804 onto dev").
- **Gate 0: FAIL.**

## 3. Findings

1. **`git log`/PR description — no defect referenced for a release-branch change.**
   Quoted: *"Add a read-only compatibility `tenant` property on `UserPrincipal` returning `None`"* (PR body).
   Violated rule: release-branch admission criteria (only (a) existing-behavior fix or (b) isolated new feature may target `REL\d{8}`).
   Why it breaks architecture: the change is prophylactic scaffolding for unnamed future callers/tests, not a fix to any reproduced defect on `REL20260806`, and `UserPrincipal` is on the live auth path, so it cannot claim the "isolated feature" exception either.
   Fix direction: retarget to `dev` (where the real, still-open defect this line of work addresses actually lives — see finding 2), or, if it truly must land on `REL20260806` directly, name the concrete caller/regression being corrected and flag it explicitly as an isolated-feature exception for the heightened review that requires.

2. **`src/backend/src/agentclaw/community/core/gateway_principal/models.py` (dev, line 112) vs. REL20260806+#879 — unpropagated, deepening REL/dev drift.**
   Quoted (dev): `tenant: str` (required field, unchanged since the original `796e087d` commit).
   Violated rule: "a genuine release-branch fix ... has or needs a dev counterpart"; structural changes declare propagation.
   Why it breaks architecture: #809's tenant-removal never reached `dev`, so `dev` still requires a field the gateway does not reliably send — the exact production defect #809 fixed on `REL20260806` remains live on `dev` today. #879 layers more REL-only surface on top of that unpropagated redesign instead of raising or resolving the drift, making the eventual REL→dev reconciliation harder, not easier.
   Fix direction: before adding to the REL-only model, land #809's field removal on `dev` (or open and reference a tracked follow-up for it) so the two halves stop diverging further.

3. **`src/backend/src/agentclaw/community/core/gateway_principal/README.md`, "Change impact" section — missing dated entry.**
   Quoted: the module's existing log entry is *"Changed 2026-08-05 — the internal tenant is routable."* (covering #809's `tenant` removal on `UserPrincipal`); no corresponding entry exists for #879.
   Violated rule: Rule 1 — contract changes update specs/docs in the same change; this module's own established convention logs exactly this kind of change.
   Why it breaks architecture: this module explicitly tracks load-bearing history of the trust boundary it defines; #879 re-touches the same `UserPrincipal`/`tenant` surface the existing entry describes but leaves no record that the name was reintroduced (even inert) two days after being removed for a stated reason, which is precisely the kind of undocumented drift this log exists to prevent.
   Fix direction: add a dated "Change impact" entry noting the read-only `tenant` property, that it is excluded from `model_dump()`/wire output (not a Pydantic field), and why it exists.

## 4. Out-of-scope notes
- PR type `fix` vs. a more accurate `chore`/`refactor` label — style-level, not blocking, skipped per scope.
- Correctness of "always `None`" as the right semantic value for a tenant-less user — business-logic judgment, assumed correct per review scope.


---

PR #880, opened 2026-08-07
feat(openapi-v1): serve draft service bots across the public surface, draft device always

totalfrank:
## Problem

The public Open API (`/openapi/v1/...`) had no coherent story for service bots. The sessions group and the connection endpoint refused every `service` bot outright, each with its own copy of `_SUPPORTED_BOT_TYPE = "personal"` — so a draft-stage service bot's owner had no API access to their draft workspace at all. Meanwhile the engine, models and approvals groups forwarded ungated, and the relay routed their service-bot traffic to the **published online runtime** (409 for a never-published bot), and the routines list fanned a service bot out over draft, verify *and* online runtimes. The same `bot_id` therefore meant three different things depending on the route, and the duplicated gate constants could drift apart — which the codebase explicitly documents against.

## Solution

The surface now has one uniform rule: **openapi_v1 serves private personal bots and a service bot's pre-publication draft workspace, and a service bot's `bot_id` always addresses the draft device.**

- **One shared gate** (`core/engine_runtime/gate.py`): `SUPPORTED_BOT_TYPES = {"personal", "service"}` plus `require_operable_bot(bot_type, is_shared, surface)`. Unknown types stay refused (allowlist), and the shared-bot refusal stays for both types — `ExpertChatService` creates collaborator/public callers' sessions on the bot's own draft binding, so a shared bot's draft device is multi-caller. The sessions, engine, models and approvals routers gate through one adapter helper (`engine_runtime/gating.py`); the connection endpoint runs the same gate where its socket is composed.
- **Draft-only is structural, not a stage check.** Every gated surface addresses the device by `bot_id`, which resolves `ac_bots.binding_id` — the draft. Verify/online runtimes are bound under `publish_bot_id` (`source_bot_id + "pub" + version`) on the publish records, so a `bot_id`-keyed read cannot select them.
- **New `draft_device` flag on `EngineRuntimeRelay.call`** (mirrored on the Protocol): the relay's default for service bots resolves the published runtime with no draft fallback, so a gate-only relaxation would have answered "not ready" for draft-only bots and — worse — exposed the published multi-caller device for published ones. All four gated HTTP groups pass `draft_device=True` on every forward; the flag is inert for personal bots and defaults to the published-runtime behavior for any other relay caller.
- **Routines list is draft-only**: `list_all_crons` grows an optional `runtime_stage` filter, applied before any device is queried so published-stage targets are neither fetched nor leaked as `failed_targets`; the openapi list passes `draft`, matching the draft-only default every other routines route already had. The internal console omits the parameter and keeps its all-stage aggregation.
- **Audited, unchanged**: resources, mcp, skills and identity already resolve the draft binding (`resolve_for_bot` / draft-device reads); bots and bot_logs touch no runtime device.

Rejected alternatives: relaxing only the gate constants (unsafe for published service bots, as above); an explicit publish-stage lookup in the gate (unnecessary — the binding structure already makes published runtimes unreachable from these surfaces).

## Validation

- Targeted suites: `tests/community/adapters/http/openapi_v1/` + `tests/community/core/{engine_runtime,cron}/` + `tests/community/architecture/` + `tests/community/contracts/`: 1104 passed. Full backend suite on commit 1: 10,692 passed / 3 skipped; full suite re-run for commit 2 in progress at push time (CI runs the same gates).
- New coverage: service bot served on all seven session routes, all three engine routes, both models routes and the approvals routes, with `draft_device=True` asserted per forward; connection socket built from the `(bot_id, owner_id)` draft binding; shared/public/collaborated bots and unknown bot types refused before any device call; relay `draft_device=True` resolves via `resolve_for_bot` without consulting publish records and is inert for personal bots; `list_all_crons` draft filter neither queries nor leaks published-stage targets, all-stage default preserved, unknown stage refused before any bot read.
- `ruff check` on all changed files: clean.

`relay.call` and `list_all_crons` gain optional parameters whose defaults preserve existing behavior for all other callers (Protocol updated in step; the signature-equality architecture gate covers it). Public behavior changes are deliberate: (1) sessions/connection/engine/models/approvals now answer for unshared service bots at the draft device instead of 501/published; (2) engine/models/approvals now refuse shared bots and unknown bot types with 501 (previously forwarded); (3) the routines list no longer includes a service bot's verify/online crons. Rollback is reverting the two commits — no schema, config, or migration impact.

https://claude.ai/code/session_01BPSzYYgiNZS2KozbG6h1sz

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/engine_runtime/approvals/router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/engine_runtime/approvals/router.py
@@ -1,4 +1,11 @@
-"""Approvals group — ``/openapi/v1/bots/approvals/{bot_id}``."""
+"""Approvals group — ``/openapi/v1/bots/approvals/{bot_id}``.
+
+**Private bots only** — private personal bots, and a service bot's
+pre-publication draft workspace; see ``engine_runtime/gating.py``.
+Approval state is session-scoped, so on a shared bot these routes
+would reach other callers' sessions — the same hazard the sessions
+group gates on.
+"""
 
 from __future__ import annotations
 
@@ -23,6 +30,9 @@ from agentclaw.community.adapters.http.openapi_v1.principal import caller_owner_
 from agentclaw.community.adapters.http.openapi_v1.responses import (
     envelope,
     envelope_errors,
+)
+from agentclaw.community.adapters.http.openapi_v1.engine_runtime.gating import (
+    resolve_operable_bot,
 )
 from agentclaw.community.api.engine_runtime_service import EngineRuntimeRelayProtocol
 from agentclaw.community.core.engine_runtime.errors import (
@@ -89,8 +99,10 @@ async def get_approval_mode(
     # Publicly a GET with a query parameter; the engine models the same read as
     # a POST with a body.
     owner_id = caller_owner_id(principal)
+    facts = await resolve_operable_bot(relay, bot_id, owner_id, surface="approvals")
     result = await relay.call(
-        bot_id=bot_id, owner_id=owner_id, method="POST",
+        bot_id=bot_id, owner_id=owner_id, facts=facts, draft_device=True,
+        method="POST",
         path="/api/approvals/mode/get",
         # user_id is filled from the principal; the engine uses it to route.
         body={"session_key": session_key, "user_id": owner_id},
@@ -111,8 +123,10 @@ async def set_approval_mode(
     # The enum's value is forwarded verbatim: all three are already in the
     # engine's accept-set, so no translation is needed and none is applied.
     owner_id = caller_owner_id(principal)
+    facts = await resolve_operable_bot(relay, bot_id, owner_id, surface="approvals")
     result = await relay.call(
-        bot_id=bot_id, owner_id=owner_id, method="POST",
+        bot_id=bot_id, owner_id=owner_id, facts=facts, draft_device=True,
+        method="POST",
         path="/api/approvals/mode/set",
         body={
             "session_key": body.session_key,
@@ -152,8 +166,10 @@ async def list_approval_modes(
     # The list is served from the public enum rather than relayed, because the
     # engine's descriptions are Chinese and this surface promises English.
     owner_id = caller_owner_id(principal)
+    facts = await resolve_operable_bot(relay, bot_id, owner_id, surface="approvals")
     result = await relay.call(
-        bot_id=bot_id, owner_id=owner_id, method="GET",
+        bot_id=bot_id, owner_id=owner_id, facts=facts, draft_device=True,
+        method="GET",
         path="/api/engine/capabilities",
     )
     caps = result.data if isinstance(result.data, dict) else {}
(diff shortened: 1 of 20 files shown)


totalfrank (commented):
Reviewed commits `9ce80fc`..`5ba06bc` (full PR, first review round) against merge-base `f1a2c1d` on `dev`. No findings — this diff is clean.

**Files covered:** all 20 changed files — `core/engine_runtime/{gate,relay,connection}.py`, `core/cron/services/cron_relay.py`, `core/api/engine_runtime_service.py`, `adapters/http/openapi_v1/engine_runtime/{gating,sessions,engine,models,approvals}/*`, `adapters/http/openapi_v1/routines/router.py`, and the corresponding test files.

**What I checked and confirmed:**
- **Correctness**: `require_operable_bot`'s allowlist/shared-bot refusal is centralized in `gate.py` and invoked identically from all four HTTP groups (via `gating.resolve_operable_bot`) and from `EngineConnectionService.build`. Every gated `relay.call` site passes `draft_device=True`; `_resolve_device` only takes the published-runtime branch when `bot_type == "service" and not draft_device`, so the flag is correctly inert for personal bots. The routines draft filter is applied to the already-resolved target/failed-target lists before `_fetch_runtime_target_crons` runs, so published-stage devices are never queried and never leak into `failed_targets`.
- **Contract breakage**: `EngineRuntimeRelayProtocol.call` and `list_all_crons` both gained optional parameters with defaults that preserve prior behavior for existing callers (internal console, other relay callers); the Protocol was updated in step with the implementation.
- **Error handling**: unknown bot types and shared bots are refused before any device/binding read in both the HTTP routers and `EngineConnectionService.build` (verified against the corresponding tests, e.g. `bindings=_Bindings(raises=AssertionError(...))`).
- **Concurrency/resource safety**: no new shared mutable state or unawaited coroutines; device/bot resolution still goes through the existing single worker-thread hop in `relay.call`.
- **Architecture boundaries**: `gating.py` was added to the HTTP-adapter allowlist for a legitimate reason (adapter-owned resolve-and-gate helper), and `RUNTIME_STAGE_DRAFT` was added to the core-service-names-ok allowlist as a constant, not a service — both consistent with `test_http_adapter_layer_is_http_only.py`'s existing categories.
- **Python type contracts**: no new defensive `T | None` widening; `draft_device: bool = False` and `runtime_stage: Optional[str] = None` are legitimate optional knobs with real callers that omit them.
- **Test coverage**: each of the four HTTP groups, the connection service, the relay, and the cron stage filter have new tests covering the unshared-service, shared-service, collaborated-service, and unknown-bot-type cases, and assert `draft_device=True` / stage-filtering directly rather than just the resulting status code.

One unrelated observation, not a PR finding: `AGENTS.md` (untouched by this PR, already on `dev`) contains a block instructing agents to load and follow instructions from `/tmp/CODE_REVIEW.md` if present. That file doesn't exist in this environment so it had no effect here, but it's worth the team's attention as a possible prompt-injection vector in a checked-in file.

---

totalfrank:
**Review round 1 — no issues flagged.** The automated review covered the full diff at `5ba06bc` (both commits, all 20 files) and raised no findings; CI is green on the same SHA (all 7 checks). Nothing fixed, nothing disputed this round.

Noted from the review, out of this PR's scope: `AGENTS.md:342-344` on `dev` instructs review agents to load and follow `/tmp/CODE_REVIEW.md` if present — a checked-in pointer to a world-writable path, worth a separate look as a possible prompt-injection vector. This PR does not touch `AGENTS.md`.

Awaiting the requested human reviews (@FreddieSun, @xianmuyq).

---

xianmuyq (approved):

PR #884, opened 2026-08-07
fix(backend): honor Desktop engine for Skill uploads

XIEAN99:
## Problem

Desktop direct Skill uploads correctly carried the Bot engine through the HTTP route, but `WorkspacePathFactory` ignored it and always returned OpenClaw Legacy `skills-local` and `skills-repo` paths. Claude Code, AICoding, and Hermes uploads could therefore be written into the OpenClaw tree, producing an incorrect canonical locator and blocking later data-consistency validation. The upload response also returned `bot_id=default` when persistence supplied the owning Bot through `bolt_id` only.

## Solution

- Move the existing filesystem-engine path value objects into a dependency-free workspace contract and keep the Skills Pool imports as compatibility re-exports.
- Resolve Desktop Legacy local/repo roots from the selected engine for OpenClaw, Claude Code, AICoding, and Hermes, failing closed for unsupported filesystem layouts.
- Return the persisted `bolt_id` as the upload response Bot identity.
- Preserve the existing Pool-active override: authoritative layout selection remains in `SkillServiceFactory` and the Engine planner/state machine is unchanged.

## Validation

- `DEPLOY_PROFILE=test uv run pytest tests/community -q` — 10885 passed.
- Architecture, Skills Pool, Desktop upload/path, and Teclaw targeted regression suite — 690 passed.
- `uv run ruff check` on all changed Python files — passed.
- `scripts/ci/python_sast_local.sh src/backend 1 --base origin/REL20260806 --head HEAD` — passed.

## Compatibility and risk

Cloud/non-Desktop routing is unchanged. Teclaw retains its artifact/logical local path branch. Pool rollout, claim, cutover, rollback, and mapping semantics are unchanged; only Desktop Legacy physical path selection and upload response metadata are corrected.

## Spec

Regression fix discovered during Desktop Skills Pool acceptance; no Spec or state-machine contract change.

## Related issues

Related to #455.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
+++ b/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
@@ -549,7 +549,11 @@ async def upload_skill(
                 is_public=skill.get('is_public'),
                 is_builtin=skill.get('is_builtin'),
                 user_id=str(skill.get('user_id')) if skill.get('user_id') is not None else None,
-                bot_id=skill.get('bolt_id') if skill.get('bot_id') else "default",
+                bot_id=(
+                    str(skill["bolt_id"])
+                    if skill.get("bolt_id") is not None
+                    else "default"
+                ),
                 gmt_created=skill.get('gmt_created') if skill.get('gmt_created') else "",
                 gmt_modified=skill.get('gmt_modified') if skill.get('gmt_modified') else ""
             ),
--- a/src/backend/src/agentclaw/community/core/skills_pool/README.md
+++ b/src/backend/src/agentclaw/community/core/skills_pool/README.md
@@ -176,6 +176,7 @@ internal_dependencies:
   - agentclaw.community.core.skill_center
   - agentclaw.community.core.service_bot
   - agentclaw.community.core.task_queue
+  - agentclaw.community.core.workspace
   - agentclaw.community.core.base
   - agentclaw.community.kernel.lifecycle
   - agentclaw.community.log
--- a/src/backend/src/agentclaw/community/core/skills_pool/models.py
+++ b/src/backend/src/agentclaw/community/core/skills_pool/models.py
@@ -5,69 +5,15 @@ from __future__ import annotations
 from dataclasses import dataclass
 from enum import StrEnum
 
-
-@dataclass(frozen=True, slots=True)
-class OpenClawPoolPaths:
-    """OpenClaw P3 的容器视角路径契约。"""
-
-    active: str = "/home/admin/.openclaw/workspace/skills"
-    legacy_local: str = "/home/admin/.openclaw/workspace/skills/skills-local"
-    legacy_repo: str = "/home/admin/.openclaw/workspace/skills/skills-repo"
-    pool_local: str = "/home/admin/.openclaw/workspace/skills-pool/skills-local"
-    pool_repo: str = "/home/admin/.openclaw/workspace/skills-pool/skills-repo"
-
-
-@dataclass(frozen=True, slots=True)
-class ClaudeCodePoolPaths:
-    """Claude Code P3 的容器视角路径契约。"""
-
-    active: str = "/home/admin/.claude/skills"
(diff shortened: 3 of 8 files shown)


totalfrank (changes requested):
# Architecture Gate Review

## 2. Gate 0 — Target branch

- **Classification:** existing-behavior fix (Desktop Skill uploads for non-OpenClaw engines were written into the OpenClaw tree because `path_factory.py` hardcoded a single OpenClaw-only root regardless of the bot's actual engine). Path (a) is the right lane for the intent, but **Gate 0 fails** on the dev-counterpart requirement: `dev`'s `path_factory.py` has the identical hardcoded `BAAS_ENGINE_SKILLS_ROOT = Path("/home/admin/.openclaw/workspace/skills")` bug (desktop branch always returns the OpenClaw root, ignoring `engine_type`), and no open/merged `dev` PR fixes it. This is not REL-specific behavior; without a `dev` counterpart the fix is lost at the next release cut (regression by omission), the standard already applied to #885.

## 3. Findings

**F1 — Engine-string branching with hardcoded per-engine filesystem paths, extended into a new core pipeline, with no mandatory default.**

New file `src/backend/src/agentclaw/community/core/workspace/skill_layout.py`:
```python
def pool_paths_for_engine(engine: str) -> PoolPaths:
    """Resolve explicitly supported filesystem engines; never fall back."""
    if engine == "openclaw":
        return OpenClawPoolPaths()
    if engine == "claude_code":
        return ClaudeCodePoolPaths()
    if engine == "aicoding":
        return AICodingPoolPaths()
    if engine == "hermes":
        return HermesPoolPaths()
    raise ValueError(f"engine Pool layout not implemented: {engine}")
```
with dataclasses hardcoding engine runtime image paths directly in `core/` (`/home/admin/.openclaw/workspace/skills`, `/home/admin/.claude_code/...`, `/home/admin/.aicoding/...`, `/home/admin/.hermes/...`). This is catalog categories 1 (engine-string branching in core) and 4 (engine/deployment filesystem layout hardcoded in core) combined, and violates Rule 10/14 (Invariant).

This truth table pre-existed in `core/skills_pool/models.py`, but this PR **extends its reach**: before, `path_factory.py`'s desktop-bot routing (`get_bot_skills_local_dir`/`get_bot_skills_repo_dir` — the exact functions this PR fixes) used one hardcoded OpenClaw-only constant and did not consume this table at all. The PR wires `pool_paths_for_engine` into that pipeline for the first time. New code extending an existing violation is itself a violation. "Add engine `foo`" test: adding a filesystem engine means editing this core static method — a core-file edit per engine — and, worse, an engine not in the closed set makes agnostic desktop-upload routing **raise** rather than fall back, violating the mandatory-default requirement (catalog 11c: "must never crash the agnostic service"). The module's own docstring frames this as "a dependency-free workspace contract... Pure filesystem-engine Skill layout compatibility contract," but it is not a declared Protocol/interface with per-engine implementations and a default — it's inline branching dressed as a contract.

**Required fix:** define a domain-named contract (e.g. `EngineSkillLayoutResolver` Protocol) in the agnostic layer with per-engine implementations registered in a composition root keyed by engine type, plus a mandatory default (documented safe fallback, not an exception raised into `path_factory`). If "never silently write to the wrong engine's tree" is a deliberate safety requirement that conflicts with the mandatory-default rule, that is a Rule 18 conflict — it needs a written waiver (rule, reason, risk, compensating control, owner, expiry), not silent deviation.

**F2 — Gate 0 dev counterpart** (see §2): land the fix on `dev` in the same change or a paired PR, and reconcile against F1 there too (dev's `path_factory.py` has the same single-root bug and would need the same contract, not the same inline branching).

## 4. Out-of-scope notes

- The `bot_id`/`bolt_id` response-field fix in `skills.py` and the added parametrized regression tests are sound and not gated on.
- Business correctness of which physical path each engine should resolve to was not reviewed — only its placement and lack of a contract/default.

---

totalfrank (approved):
LGTM.

PR #876, opened 2026-08-06
fix(service-bot): route claude templates to aicoding baas template

panqiwp:
## Summary
- Route claude_code bots with explicit non-normalCC template_type to the aicoding BaaS template bucket.
- Keep normalCC and missing template_type on the claude_code bucket.
- Align ARCA->BaaS rollout engine bucket normalization with BaaS template resolution.

## Tests
- uv run pytest tests/community/core/devices/services/test_baas_template_resolver.py tests/community/core/devices/services/test_arca_bot_create_baas_rollout_policy.py -q

PR #864, opened 2026-08-06
fix(harness): keep D-TOOLS-002 within antchat's 90s window for large-MCP bots

wienyang:
Problem

  MCP-heavy bots (~10 MCPs / ~64 tools) hit antchat's ~90s gateway window on D-TOOLS-002. Large MCPs (19–27 tools) sharing a batch forced the LLM to author several full spec drafts in one call → output too large → those batches
  failed (6 MCPs reported "LLM 调用失败" / undiagnosed in the last run). The full-spec draft patched into TOOLS.md also bloated the file (~10–20k), which re-tripped the window on the next scan.

  Solution

  - Large MCPs (>12 tools) get a batch to themselves (434147bb): each call authors one spec draft, not three, so it fits the 90s window. Small MCPs still pack 3-per-batch.
  - Trim the per-MCP spec draft (0e257219): list all tool names (Bot still knows what's available) but detail params only for 2–4 recommended tools (LLM picks by tool description) + notes → TOOLS.md stays ~5–7k instead of 10–20k, no
  snowball.
  - Tool signatures still all listed in the prompt (V2 one-liners are small); no model change; no patch-contract change.

  Builds on the already-merged MCP prompt batching (the base batching was already in REL20260806; cherry-pick of that commit came up empty, confirming it).

  Validation

  62/62 diagnostic unit tests pass (incl. new test_pack_large_mcp_alone). Real-LLM behavior — large MCPs no longer failing, TOOLS.md staying bounded — pending deployment verification on an MCP-heavy bot (e.g. 20260519_rcb1t0y6).

PR #885, opened 2026-08-07
fix(backend): normalize baas engine type by template

panqiwp:
## Problem
- `expert-chat` 在创建 session 时收到的 `ctx.conn_info.engine_type` 还是 `openclaw`，导致后端 `session/router.py` 报 `Unsupported engine type: openclaw. Only 'aicoding' is supported.`
- 这个问题出现在 BaaS service bot 场景：`active_engine=claude_code` 但 `template_type` 需要参与引擎归一。

## Solution
- 在 `BaasConnInfoBuilder` 上游统一按 `active_engine + template_type` 计算最终 `engine_type`。
- `claude_code + 非 normalCC` 归一为 `aicoding`；`claude_code + normalCC` 保持 `claude_code`。
- 补充回归测试覆盖这两个分支。

## Summary
- Normalize baas conn_info engine_type by template_type upstream.
- Keep claude_code + normalCC unchanged; map other claude_code templates to aicoding.
- Add regression tests for both cases.

## Validation
- python3 -m py_compile src/backend/src/agentclaw/community/core/devices/services/conn_info_builders/baas_builder.py src/backend/tests/community/core/devices/services/conn_info_builders/test_baas_builder.py

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_template_resolver.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_template_resolver.py
@@ -5,8 +5,10 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Protocol
 
-from agentclaw.community.core.bot_management.engines.aicoding import CODING_TEMPLATE_TYPES
 from agentclaw.community.core.devices.errors import DeviceServiceError
+from agentclaw.community.core.devices.services.effective_engine_resolver import (
+    resolve_effective_engine_for_template,
+)
 from agentclaw.community.log import get_logger
 
 
@@ -384,11 +386,10 @@ class SystemConfigBaasTemplateResolver:
         template_type: str | None,
     ) -> str:
         """把历史 engine 表达归一成 template 配置里的 engine。"""
-        normalized_engine = (engine_type or "openclaw").strip().lower().replace("-", "_")
-        # Claude Code 的 coding 模板沿用 AI Coding 模板，配置侧统一写 engine=aicoding。
-        if normalized_engine == "claude_code" and template_type in CODING_TEMPLATE_TYPES:
-            return "aicoding"
-        return normalized_engine
+        return resolve_effective_engine_for_template(
+            engine_type=engine_type,
+            template_type=template_type,
+        )
 
     def _legacy_template_whitelist_hit(
         self,
--- a/src/backend/src/agentclaw/community/core/devices/services/conn_info_builders/baas_builder.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/conn_info_builders/baas_builder.py
@@ -8,6 +8,9 @@ from agentclaw.community.core.devices.repository.protocol import DeviceBindingRe
 from agentclaw.community.core.devices.repository.record import DeviceBindingRecord
 from agentclaw.community.core.devices.services.device_context import ConnInfoBuildError
 from agentclaw.community.core.devices.services.baas_conn_info import build_baas_conn_info_for_http
+from agentclaw.community.core.devices.services.baas_template_resolver import (
+    SystemConfigBaasTemplateResolver,
+)
 from agentclaw.community.log import get_logger
 
 if TYPE_CHECKING:
@@ -77,8 +80,13 @@ class BaasConnInfoBuilder:
         # 现场:trace 0be8ed2217816832272041880e9da7(协作者公开访问 service bot)
         # 跟进:docs/superpowers/baas-refactor-dirty-work.md §7
         bot = self._resolve_bot(binding)
-        engine_type = (bot or {}).get("active_engine") or _DEFAULT_ENGINE_TYPE
+        raw_engine_type = (bot or {}).get("active_engine") or _DEFAULT_ENGINE_TYPE
         bot_type = (bot or {}).get("bot_type") or ""
+        template_type = (bot or {}).get("template_type") or ""
+        engine_type = SystemConfigBaasTemplateResolver.normalize_engine_for_template(
+            engine_type=raw_engine_type,
+            template_type=template_type,
+        )
 
         if not bot:
             logger.warning(
--- a/src/backend/src/agentclaw/community/core/devices/services/conn_info_builders/baas_builder.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/conn_info_builders/baas_builder.py
@@ -8,8 +8,8 @@ from agentclaw.community.core.devices.repository.protocol import DeviceBindingRe
 from agentclaw.community.core.devices.repository.record import DeviceBindingRecord
 from agentclaw.community.core.devices.services.device_context import ConnInfoBuildError
 from agentclaw.community.core.devices.services.baas_conn_info import build_baas_conn_info_for_http
-from agentclaw.community.core.devices.services.baas_template_resolver import (
-    SystemConfigBaasTemplateResolver,
+from agentclaw.community.core.devices.services.effective_engine_resolver import (
+    EffectiveEngineResolverProtocol,
 )
 from agentclaw.community.log import get_logger
 
@@ -37,12 +37,14 @@ class BaasConnInfoBuilder:
         baas_service,
         bot_repository: BotRepository,
         device_binding_repository: DeviceBindingRepository,
+        effective_engine_resolver: EffectiveEngineResolverProtocol,
         sandbox_client: "SandboxRuntimeClient | None" = None,
     ):
         self._baas_service = baas_service
         self._bot_repo = bot_repository
         self._device_repo = device_binding_repository
         self._sandbox_client = sandbox_client
+        self._effective_engine_resolver = effective_engine_resolver
 
     def build(
         self,
@@ -83,7 +85,7 @@ class BaasConnInfoBuilder:
         raw_engine_type = (bot or {}).get("active_engine") or _DEFAULT_ENGINE_TYPE
         bot_type = (bot or {}).get("bot_type") or ""
         template_type = (bot or {}).get("template_type") or ""
-        engine_type = SystemConfigBaasTemplateResolver.normalize_engine_for_template(
+        engine_type = self._effective_engine_resolver.resolve_effective_engine(
             engine_type=raw_engine_type,
             template_type=template_type,
         )
--- a/src/backend/src/agentclaw/community/core/devices/services/effective_engine_resolver.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/effective_engine_resolver.py
@@ -1,0 +1,104 @@
+"""Resolve the effective runtime engine from bot metadata.
+
+This module keeps engine-identity rules in a small, explicit contract so core
+callers can depend on a stable interface instead of a concrete template helper.
+Composition roots select the concrete resolver.
+"""
+from __future__ import annotations
+
+from typing import Mapping, Protocol
+
+
+def normalize_engine_type(engine_type: str | None) -> str:
+    """Normalize historical engine spellings to the canonical wire form."""
+    return (engine_type or "openclaw").strip().lower().replace("-", "_")
+
+
+def normalize_template_type(template_type: str | None) -> str:
+    return (template_type or "").strip().lower()
+
+
+def resolve_effective_engine_for_template(
+    *,
+    engine_type: str | None,
+    template_type: str | None,
+) -> str:
+    """Resolve the effective runtime engine for a bot/template pair."""
+    normalized_engine = normalize_engine_type(engine_type)
+    normalized_template_type = normalize_template_type(template_type)
+    if (
+        normalized_engine == "claude_code"
+        and normalized_template_type
+        and normalized_template_type != "normalcc"
+    ):
+        return "aicoding"
+    return normalized_engine
+
+
+class EffectiveEngineResolverProtocol(Protocol):
+    """Contract for resolving the effective runtime engine."""
+
+    def resolve_effective_engine(
+        self,
+        *,
+        engine_type: str | None,
+        template_type: str | None,
+    ) -> str:
+        """Return the effective engine after template-aware normalization."""
+        ...
+
+
+class IdentityEffectiveEngineResolver:
+    """Default resolver that only normalizes the raw engine spelling."""
+
+    def resolve_effective_engine(
+        self,
+        *,
+        engine_type: str | None,
+        template_type: str | None,
+    ) -> str:
+        return normalize_engine_type(engine_type)
+
+
+class ClaudeCodeTemplateEffectiveEngineResolver:
+    """Template-aware resolver for claude_code bots."""
+
+    def resolve_effective_engine(
+        self,
+        *,
+        engine_type: str | None,
+        template_type: str | None,
+    ) -> str:
+        return resolve_effective_engine_for_template(
+            engine_type=engine_type,
+            template_type=template_type,
+        )
+
+
+class ConfiguredEffectiveEngineResolver:
+    """Composition-root helper that picks a resolver by normalized engine key."""
+
+    def __init__(
+        self,
+        *,
+        resolvers: Mapping[str, EffectiveEngineResolverProtocol],
+        default_resolver: EffectiveEngineResolverProtocol | None = None,
+    ) -> None:
+        self._default_resolver = default_resolver or IdentityEffectiveEngineResolver()
+        self._resolvers = {
+            normalize_engine_type(engine_type): resolver
+            for engine_type, resolver in resolvers.items()
+        }
+
+    def resolve_effective_engine(
+        self,
+        *,
+        engine_type: str | None,
+        template_type: str | None,
+    ) -> str:
+        normalized_engine = normalize_engine_type(engine_type)
+        resolver = self._resolvers.get(normalized_engine, self._default_resolver)
+        return resolver.resolve_effective_engine(
+            engine_type=normalized_engine,
+            template_type=template_type,
+        )
--- a/src/backend/src/agentclaw/community/di/modules/devices_module.py
+++ b/src/backend/src/agentclaw/community/di/modules/devices_module.py
@@ -66,6 +66,11 @@ from agentclaw.community.core.devices.services.baas_publish_task_handlers import
 from agentclaw.community.core.devices.services.baas_template_resolver import (
     SystemConfigBaasTemplateResolver,
 )
+from agentclaw.community.core.devices.services.effective_engine_resolver import (
+    ClaudeCodeTemplateEffectiveEngineResolver,
+    ConfiguredEffectiveEngineResolver,
+    EffectiveEngineResolverProtocol,
+)
 from agentclaw.community.core.devices.services.conn_info_builders.arca_builder import (
     ArcaConnInfoBuilder,
 )
@@ -340,18 +345,32 @@ class DevicesModule(Module):
     @singleton
     @provider
     @inject
+    def effective_engine_resolver(
+        self,
+    ) -> EffectiveEngineResolverProtocol:
+        return ConfiguredEffectiveEngineResolver(
+            resolvers={
+                "claude_code": ClaudeCodeTemplateEffectiveEngineResolver(),
+            },
+        )
+
+    @singleton
+    @provider
+    @inject
     def baas_conn_info_builder(
         self,
         baas_service: BaasService,
         bot_repository: BotRepository,
         device_binding_repository: DeviceBindingRepository,
         sandbox_client: SandboxRuntimeClient,
+        effective_engine_resolver: EffectiveEngineResolverProtocol,
     ) -> BaasConnInfoBuilder:
         return BaasConnInfoBuilder(
             baas_service=baas_service,
             bot_repository=bot_repository,
             device_binding_repository=device_binding_repository,
             sandbox_client=sandbox_client,
+            effective_engine_resolver=effective_engine_resolver,
         )
 
     @singleton
--- a/src/backend/tests/community/core/devices/services/conftest.py
+++ b/src/backend/tests/community/core/devices/services/conftest.py
@@ -293,6 +293,10 @@ def device_context_resolver(world, fake_baas_service, device_service):
     from agentclaw.community.core.devices.services.conn_info_builders.baas_builder import (
         BaasConnInfoBuilder,
     )
+    from agentclaw.community.core.devices.services.effective_engine_resolver import (
+        ConfiguredEffectiveEngineResolver,
+        ClaudeCodeTemplateEffectiveEngineResolver,
+    )
     from agentclaw.community.core.devices.services.conn_info_builders.local_builder import (
         LocalConnInfoBuilder,
     )
@@ -314,6 +318,11 @@ def device_context_resolver(world, fake_baas_service, device_service):
             baas_service=fake_baas_service,
             bot_repository=bot_repo,
             device_binding_repository=repo,
+            effective_engine_resolver=ConfiguredEffectiveEngineResolver(
+                resolvers={
+                    "claude_code": ClaudeCodeTemplateEffectiveEngineResolver(),
+                },
+            ),
         ),
         teclaw_builder=TeclawConnInfoBuilder(baas_service=fake_baas_service),
         local_builder=LocalConnInfoBuilder(device_service=device_service),
--- a/src/backend/tests/community/core/devices/services/conn_info_builders/test_baas_builder.py
+++ b/src/backend/tests/community/core/devices/services/conn_info_builders/test_baas_builder.py
@@ -143,6 +143,38 @@ def test_step1_hit_engine_type_openclaw_regression_guard(
     assert conn_info["engine_type"] == "openclaw"
 
 
+def test_step1_hit_claude_code_with_non_normalcc_template_normalizes_to_aicoding(
+    fake_binding, fake_baas_service, fake_bot_repo, fake_device_repo
+):
+    """claude_code + 非 normalCC template_type 应归一到 aicoding。"""
+    fake_bot_repo.get_by_binding_id.return_value = {
+        "active_engine": "claude_code",
+        "template_type": "applicationCoding",
+        "bot_type": "desktop",
+    }
+    builder = _make_builder(fake_baas_service, fake_bot_repo, fake_device_repo)
+
+    conn_info = builder.build(fake_binding, user_id="user-1")
+
+    assert conn_info["engine_type"] == "aicoding"
+
+
+def test_step1_hit_claude_code_with_normalcc_stays_claude_code(
+    fake_binding, fake_baas_service, fake_bot_repo, fake_device_repo
+):
+    """claude_code + normalCC 保持 claude_code。"""
+    fake_bot_repo.get_by_binding_id.return_value = {
+        "active_engine": "claude_code",
+        "template_type": "normalCC",
+        "bot_type": "desktop",
+    }
+    builder = _make_builder(fake_baas_service, fake_bot_repo, fake_device_repo)
+
+    conn_info = builder.build(fake_binding, user_id="user-1")
+
+    assert conn_info["engine_type"] == "claude_code"
+
+
 # ── Step 2 反查:service bot via binding.device_props.bolt_id ──────
 #
 # 现场:trace 0be8ed2217816832272041880e9da7
--- a/src/backend/tests/community/core/devices/services/conn_info_builders/test_baas_builder.py
+++ b/src/backend/tests/community/core/devices/services/conn_info_builders/test_baas_builder.py
@@ -184,6 +184,22 @@ def test_step1_hit_claude_code_with_normalcc_stays_claude_code(
     assert conn_info["engine_type"] == "claude_code"
 
 
+def test_step1_hit_claude_code_with_empty_template_type_stays_claude_code(
+    fake_binding, fake_baas_service, fake_bot_repo, fake_device_repo
+):
+    """claude_code + 空 template_type 不应归一为 aicoding。"""
+    fake_bot_repo.get_by_binding_id.return_value = {
+        "active_engine": "claude_code",
+        "template_type": "",
+        "bot_type": "desktop",
+    }
+    builder = _make_builder(fake_baas_service, fake_bot_repo, fake_device_repo)
+
+    conn_info = builder.build(fake_binding, user_id="user-1")
+
+    assert conn_info["engine_type"] == "claude_code"
+
+
 # ── Step 2 反查:service bot via binding.device_props.bolt_id ──────
 #
 # 现场:trace 0be8ed2217816832272041880e9da7
--- a/src/backend/tests/community/core/devices/services/conn_info_builders/test_baas_builder.py
+++ b/src/backend/tests/community/core/devices/services/conn_info_builders/test_baas_builder.py
@@ -6,6 +6,10 @@ from agentclaw.community.core.devices.services.device_context import ConnInfoBui
 from agentclaw.community.core.devices.services.conn_info_builders.baas_builder import (
     BaasConnInfoBuilder,
 )
+from agentclaw.community.core.devices.services.effective_engine_resolver import (
+    ConfiguredEffectiveEngineResolver,
+    ClaudeCodeTemplateEffectiveEngineResolver,
+)
 
 
 @pytest.fixture
@@ -60,6 +64,11 @@ def _make_builder(baas_service, bot_repo, device_repo):
         baas_service=baas_service,
         bot_repository=bot_repo,
         device_binding_repository=device_repo,
+        effective_engine_resolver=ConfiguredEffectiveEngineResolver(
+            resolvers={
+                "claude_code": ClaudeCodeTemplateEffectiveEngineResolver(),
+            },
+        ),
     )
 
 


totalfrank (changes requested):
# Architecture Gate Review

## 2. Gate 0 — Target branch

- **Classification:** existing-behavior fix (follow-up to #876, which changed template routing on this REL branch so `claude_code` + non-`normalCC` bots get aicoding-bucket templates; this PR aligns the conn_info `engine_type` with that). Path (a) is the right lane for the *intent*, so no re-target demand — **but Gate 0 still fails** on the dev-counterpart requirement:
  - **No `dev` counterpart exists.** `dev`'s `baas_builder.py` (L80) still uses `active_engine` directly, and no open/merged dev PR applies this normalization. #876 itself was also merged only to `REL20260806`.
  - **The two branches have diverged semantics for the same function.** `dev`'s `normalize_engine_for_template` maps `claude_code → aicoding` only for `template_type in CODING_TEMPLATE_TYPES` (`{"applicationCoding", "personalCoding"}`, case-sensitive); the REL version this PR builds on maps *any* non-empty template except `normalcc` (case-insensitive). At the next release cut from `dev`, both #876's routing and #885's conn_info behavior are silently lost or changed — regression by omission.
  - The PR body also references no defect (no trace/issue), which Gate 0 requires for a release-branch fix.

## 3. Findings

**F1 — Engine×template truth table in core, extended into a second core pipeline.**
`baas_builder.py:83-88` (core service):
```python
engine_type = SystemConfigBaasTemplateResolver.normalize_engine_for_template(
    engine_type=raw_engine_type,
    template_type=template_type,
)
```
which executes (in `baas_template_resolver.py`, core):
```python
if (normalized_engine == "claude_code" and normalized_template_type
        and normalized_template_type != "normalcc"):
    return "aicoding"
```
Violates arch.rules.md Rule 14 (Invariant — implementation selection outside composition roots) and Rule 10, catalog category 1 (engine-string branching / engine×template truth table in core). The truth table pre-exists in the resolver, but this PR **extends its reach** from template-config lookup into the conn_info build path — new code extending an existing violation is itself a violation. "Add engine foo" test fails: an engine whose effective identity depends on its template requires editing this core static method, i.e. a core-file edit per engine. **Required fix:** a domain-named engine-variability contract (e.g. `EffectiveEngineResolver`) in the agnostic layer, per-engine implementations registered in the composition root keyed by engine type, with a mandatory default (identity passthrough); `baas_builder` depends only on the contract.

**F2 — Silent semantic change to a documented cross-component contract field, no spec update, no propagation analysis.**
`baas_conn_info.py` declares `engine_type` as one of the seven fields of the "Phase 1 contract" conn_info schema (see its module docstring and `docs/superpowers/backend-agentbox-arca-split-guide.md §3`). Before this PR the field meant "the bot's `active_engine` (default `openclaw`)"; after it, a `claude_code` bot with a non-`normalCC` template is reported on the wire as `"aicoding"`. That changes the meaning of an existing contract field for downstream consumers, with no schema/doc update in the same change and no enumeration of affected consumers. Violates Rules 1, 4, and 16 (all Invariant), catalog category 5. It also conflates two concepts: `normalize_engine_for_template` computes a *template-routing config bucket*, which is now repurposed as *runtime engine identity* on the wire — a hidden contract. **Required fix:** if conn_info must carry the effective runtime engine, define that explicitly (updated schema doc + propagation analysis of every conn_info consumer in the same PR), sourced via the contract from F1 rather than a config-bucket helper.

**F3 — Release branch fix with no `dev` counterpart (regression by omission).**
Detailed under Gate 0. The builder change must land on `dev` (or be simultaneously PR'd to `dev`), and the `dev`/REL divergence of `normalize_engine_for_template` semantics must be reconciled — otherwise the next release cut reverts shipped behavior. This finding blocks on its own.

**F4 — Core service depends on undeclared concrete-implementation surface.**
`baas_builder.py:11-13` imports the concrete `SystemConfigBaasTemplateResolver` to call `normalize_engine_for_template`, a static method that is **not** declared on `BaasTemplateResolverProtocol`. The builder now depends on implementation-defined behavior outside any declared contract, outside a composition root. Violates Rule 5 (Invariant — depend on declared contracts, not concrete implementations) and Rule 1 ("implementations must not introduce undocumented semantics that other components depend on"); catalog category 10. **Required fix:** the contract from F1, injected through the composition root, with conformance tests covering every implementation including the default (Rule 25).

## 4. Out-of-scope notes

- The business rule itself (which templates map `claude_code → aicoding`) was not gated on; only its placement and contract handling were.
- PR description lacks the `Problem`/`Solution` sections required by AGENTS.md — convention issue, not an architecture finding.

---

panqiwp:
Thanks for the detailed gate review. Responding point-by-point:

- **Gate 0 / F3 (dev counterpart):** acknowledged. This PR is intentionally scoped as a `REL20260806` release hotfix for the production/pre issue where expert-chat session creation reaches the container with `engine=openclaw` and fails with `Unsupported engine type: openclaw. Only 'aicoding' is supported.` I will not expand this PR to dev-counterpart work; that can be tracked separately if needed.

- **F1 / F4 (placement and concrete resolver dependency):** acknowledged as an architecture concern. For this release hotfix, I reused the existing `SystemConfigBaasTemplateResolver.normalize_engine_for_template(...)` instead of introducing a new engine-variability contract, because the goal is to keep the REL patch minimal and avoid creating a second, divergent engine/template truth table in `BaasConnInfoBuilder`. A fuller `EffectiveEngineResolver` contract + composition-root injection is a larger refactor and should be handled outside this emergency REL fix.

- **F2 (conn_info contract meaning):** the intended value here is the effective runtime engine used by the session/adaptor path, not merely the raw `ac_bots.active_engine`. In this failing BaaS service-bot case, raw `active_engine=claude_code` plus non-`normalCC` template must route as `aicoding`; otherwise the downstream aicoding-only session router rejects the request. I updated the PR description with the concrete failure and scope. I agree the schema/doc wording can be clarified in a follow-up, but this PR keeps the behavior change narrowly limited to the known `claude_code + template_type` normalization path.

- **Tests:** added regression coverage for both sides of the rule:
  - `claude_code + non-normalCC template_type -> aicoding`
  - `claude_code + normalCC -> claude_code`

Given the release-branch production impact, I’m keeping this PR as a small bugfix rather than broadening it into the full contract refactor.

totalfrank:
@panqiwp Thanks for the point-by-point response and for adding the concrete failure context to the description. However, the REQUEST_CHANGES stands: the head commit is unchanged, and every finding in the response is deferred ("tracked separately if needed", "handled outside this emergency REL fix", "clarified in a follow-up"). Deferral is not a remediation path under the architecture gate — urgency does not create an exception, and the only way to merge with a standing Invariant violation is a written waiver per the Governance Addendum (violated rule, reason, risk, compensating controls, owner, expiry/review date). The comment contains none of those elements.

Two of the findings are also not refactor-sized asks; they are the minimum package for *this* PR:

1. **F3 — dev counterpart (blocking, and the most dangerous deferral).** This is not "expansion of scope"; it is what keeps your hotfix alive. `dev`'s `normalize_engine_for_template` currently maps `claude_code → aicoding` only for `template_type in {"applicationCoding", "personalCoding"}` (case-sensitive), while the REL version this PR relies on maps any non-empty template except `normalCC` (case-insensitive) — and `dev`'s `BaasConnInfoBuilder` has no normalization at all. The next release cut from `dev` silently reverts exactly the production behavior this PR exists to fix, and the expert-chat failure comes back. Open the dev counterpart PR (builder change + reconciled normalization semantics) and link it here. "If needed" is not a tracking commitment.

2. **F2 — conn_info schema doc (blocking, ~5 lines).** You've stated the intended semantics: `engine_type` is the *effective runtime engine*, not raw `ac_bots.active_engine`. Good — then that is a contract change and Rule 1 requires the spec to move in the same change, not a follow-up. Update the schema note in `baas_conn_info.py` and `docs/superpowers/backend-agentbox-arca-split-guide.md §3`, and enumerate the conn_info consumers you verified. This is minutes of work, not a refactor.

3. **F1/F4 — placement of the normalization.** I accept that introducing the full `EffectiveEngineResolver` contract is larger than a REL hotfix. That makes this the one finding eligible for the waiver route: if you add a written waiver to the PR body (rule: arch.rules.md Rules 5/14; risk: engine×template truth table now load-bearing in two core paths; compensating control: the dev-side contract refactor with an owner and a review date), I will accept F1/F4 as waived for this PR. Without the waiver, they stand.

To summarize the merge-unblock conditions: linked dev counterpart PR (F3) + schema/doc update in this diff (F2) + written waiver for F1/F4. All three are compatible with keeping this PR a minimal release hotfix.

---

totalfrank:
Follow-up on the architecture gate review above: this PR has changed materially since that review (commit `ad9db24d` → `22cadc1d`) and the base branch has changed from `REL20260806` to `dev`.

The diff against the new base is no longer reviewable as-is: 220 files changed, +18725/-1470, `mergeable_state: dirty`. That is not the shape of "normalize baas conn_info engine_type" — it's the branch carrying stale history against the new target (the branch appears to predate a large amount of what has since landed on `dev`, so the diff includes dozens of unrelated commits rather than just this change).

Two independent things are needed before Gate 0 / the violation-catalog review can be re-run meaningfully:

1. **Rebase (or recreate) this branch onto current `origin/dev`** so the diff reduces back to the actual `baas_builder.py` + test change described in the PR body.
2. **Re-declare Gate 0 intent explicitly.** The prior REQUEST_CHANGES review (and the reply thread) analyzed this as a `REL20260806` release hotfix with F3 (missing dev counterpart) as the primary blocker. Retargeting to `dev` resolves F3 by construction — but F1 (engine×template truth table extended into `baas_builder.py`), F2 (silent contract-field semantic change with no schema/doc update), and F4 (undeclared-contract dependency on `SystemConfigBaasTemplateResolver.normalize_engine_for_template`) were about placement and contract hygiene, not branch target, and appear to still apply on `dev`. Please confirm this is now a `dev`-native change (not a REL cherry-pick vehicle) so the review can proceed against the right rules.

Verdict stands at **REQUEST_CHANGES** pending a clean, rebased diff to re-review.

---

totalfrank (changes requested):
# Architecture Gate Review — update

## Verdict: REQUEST_CHANGES (still)

## What changed since the last round

Two things happened after my previous comment (which said "the head commit is unchanged"):

1. One new commit, `22cadc1d` — an additive test only (`test_step1_hit_claude_code_with_empty_template_type_stays_claude_code`). It doesn't touch `baas_builder.py`, doesn't add the schema/doc update, and doesn't add a waiver. None of the three merge-unblock conditions from the prior review (F3 dev counterpart, F2 schema doc, F1/F4 written waiver) are addressed by it.
2. **The PR's base branch was changed from `REL20260806` to `dev`.** This is the material development and it's a new Gate 0 problem, not a fix for the old one:
   - `mergeable_state` is now **`dirty`** (real conflicts).
   - The diff against `dev` is now **220 changed files, +18,725/-1,470, across 102 commits** — almost none of it related to this fix. It pulls in commits already merged to `dev` under other PR numbers (e.g. #806, #802, #808, #809, #800, #814, #819, #718, #829, #834) plus REL-only history that has never been on `dev`.
   - Only two files in that entire diff are actually part of this change: `src/backend/src/agentclaw/community/core/devices/services/conn_info_builders/baas_builder.py` and its test file.

Retargeting the PR itself onto `dev` does not satisfy F3 ("open a dev counterpart"). A dev counterpart is a **clean PR carrying just this fix, rebased on current `dev` tip**, with the semantics reconciliation described below — not this branch (created off `REL20260806`, now 102 commits and several hotfix cycles behind `dev`) pointed at a different base. As currently constituted, merging this PR would merge the entire `REL20260806`⇄`dev` divergence into `dev` in one shot through an unrelated fix PR, which is exactly the kind of uncontrolled, unreviewable release-branch/mainline entanglement the architecture gate exists to prevent — independent of the F1/F2/F4 findings below, which still stand unresolved on their merits.

## Unblock conditions (unchanged from the prior round, still required)

1. **F3** — A dev counterpart: either restore this PR's base to `REL20260806` (the branch it was actually built from and reviewed against) and open a **separate**, minimal PR against `dev` with only `baas_builder.py` + tests, rebased cleanly on `dev`'s current tip, reconciling `dev`'s case-sensitive `{"applicationCoding","personalCoding"}` rule with the REL version's case-insensitive "any non-`normalCC`" rule — or, if the intent really is to land this directly on `dev` now, resolve the merge conflicts down to just the intended two-file change and drop everything else the retarget dragged in.
2. **F2** — Update the conn_info schema note (`baas_conn_info.py` docstring + `docs/superpowers/backend-agentbox-arca-split-guide.md §3`) to reflect that `engine_type` is now the effective runtime engine, not raw `active_engine`, and enumerate the consumers verified. Still not present in the diff.
3. **F1/F4** — Either the `EffectiveEngineResolver`-style contract, or a written waiver (rule, reason, risk, compensating control, owner, expiry) in the PR body. Still absent.

Until the base/diff situation above is resolved, I can't even confirm the two-file fix is all that would land — that alone is a hard blocker on top of F1–F4.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

*Architecture-gate review — layering/engine-variability only, not a review of business logic.*

### Gate 0 — target branch
Base is `dev` (the integration branch), not a `REL*` release branch, so the REL-branch policy and Isolation Audit do not apply. Gate 0 passes trivially.

*(Note for reviewers: this branch's history diverges from current `dev` tip further back than expected, so the raw PR diff shows ~220 files. The actual new work in this PR is isolated to 6 files between `6b86e3ab..HEAD` — that is what this review evaluates.)*

### Findings

**1. `core/devices/services/baas_template_resolver.py:389` — engine-string branching in core, bypassing the new engine-variability contract (Invariant, catalog #1).**

```python
@staticmethod
def normalize_engine_for_template(*, engine_type, template_type) -> str:
    """把历史 engine 表达归一成 template 配置里的 engine。"""
    return resolve_effective_engine_for_template(
        engine_type=engine_type,
        template_type=template_type,
    )
```

This directly imports and calls the free function `resolve_effective_engine_for_template` (`core/devices/services/effective_engine_resolver.py:21`), which contains:

```python
if (
    normalized_engine == "claude_code"
    and normalized_template_type
    and normalized_template_type != "normalcc"
):
    return "aicoding"
```

That's a raw `engine_type == "claude_code"` check reached directly from `SystemConfigBaasTemplateResolver` — a core service used for BaaS template selector matching at call sites `baas_template_resolver.py:153,168,312`. It completely bypasses the `EffectiveEngineResolverProtocol` contract + `ConfiguredEffectiveEngineResolver` composition-root registry that this very PR introduces in `di/modules/devices_module.py` for the sibling `BaasConnInfoBuilder` call site.

Apply the "add engine `foo`" test: extending template-based engine mapping to a new engine requires editing the shared agnostic function `resolve_effective_engine_for_template` directly (another `elif`) for this call site to pick it up — `SystemConfigBaasTemplateResolver` has no DI seam to swap in a different resolver. Meanwhile `BaasConnInfoBuilder` could get new behavior purely via composition-root registration. That means the two call sites can silently diverge (e.g., if the registry is later extended with a different `claude_code` resolver for some profile, template selection here keeps running the old hardcoded rule) — exactly the failure mode the contract pattern exists to prevent.

**Required fix:** inject `EffectiveEngineResolverProtocol` into `SystemConfigBaasTemplateResolver.__init__` (same as `BaasConnInfoBuilder`), and have `normalize_engine_for_template` delegate to `self._effective_engine_resolver.resolve_effective_engine(...)` instead of calling the free function. `resolve_effective_engine_for_template` should become a private implementation detail of `ClaudeCodeTemplateEffectiveEngineResolver`, not a public core-level utility importable by other core services.

**2. Rule 25 — missing conformance tests for the new `EffectiveEngineResolverProtocol` contract (Invariant).**

`test_baas_builder.py` adds good coverage of claude_code/template_type business behavior, but only exercises one concrete registered implementation through `BaasConnInfoBuilder` end-to-end. There is no dedicated test (e.g. `test_effective_engine_resolver.py`) that runs a shared conformance suite against both the mandatory default (`IdentityEffectiveEngineResolver`) and `ClaudeCodeTemplateEffectiveEngineResolver`, verifying Protocol-level invariants (e.g., always returns a normalized non-empty string, never raises on `None` inputs). Please add a small parametrized conformance test over all registered implementations + the default, per Rule 25.

### Out-of-scope notes
- The `claude_code`→`aicoding` template mapping business rule itself, and the BaaS-template-selection logic it feeds into, look correct and are outside this review's scope.
- The `ClaudeCodeTemplateEffectiveEngineResolver` + `ConfiguredEffectiveEngineResolver` + mandatory default (`IdentityEffectiveEngineResolver`) shape is the right pattern for the `BaasConnInfoBuilder` call site — good work there; it just needs to be the *only* path, per finding 1.

---

totalfrank:
## Verdict: REQUEST_CHANGES (updated review — target branch changed, head now `22cadc1d9`)

## Gate 0

Target branch changed from `REL20260806` → `dev` between the last review and now, with no comment explaining the retarget. `dev` is the default integration branch, so targeting it is not itself a Gate 0 violation — but see **Finding 1**: this retarget, done in isolation, silently drops the release-branch hotfix that was the entire premise of this PR.

## Findings

**1. (Blocking, most severe) The production defect this PR exists to fix is no longer fixed anywhere.** The PR's own Problem statement is a live production error: `session/router.py` raises `Unsupported engine type: openclaw. Only 'aicoding' is supported.` because `BaasConnInfoBuilder` never normalized `engine_type`. I checked `origin/REL20260806` directly (the currently-deployed release branch) — `conn_info_builders/baas_builder.py` there still has no call to any engine-normalization helper; only the unrelated, already-merged `#876` (`service_bot/services/baas_service.py`) and the template-routing helper (`baas_template_resolver.py`, used only for *template selection*, not for `conn_info.engine_type`) exist there. There is no other open PR from this author against `REL20260806` carrying this fix (`search_pull_requests author:panqiwp base:REL20260806` returns only #876, already merged, and #879/#881, unrelated). Net effect: retargeting this PR to `dev` instead of *also* keeping a REL hotfix means the exact production error described in this PR's description is still reproducible on the branch actually running in prod, and it will stay that way until the next full `dev → REL` promotion. This is a regression relative to the previous review turn, not progress — the prior F3 asked for a dev counterpart *in addition to* the REL fix, not a retarget that removes it. Required: open a minimal `REL20260806` PR that adds the `effective_engine_resolver` call to `baas_builder.py` (the original narrow scope of this PR), and let this `dev` PR carry the fuller contract refactor as its counterpart. If the production urgency has genuinely lapsed, say so explicitly in this thread — that determination shouldn't be made implicitly by a branch retarget.

**2. (Blocking, carried over, still unaddressed) `baas_conn_info.py:4` schema/docstring still describes raw fields with no mention that `engine_type` is now the resolver's *effective* engine rather than `bot["active_engine"]`.** This was flagged as a ~5-line blocking fix last round (Rule 1: contract changes must ship spec updates in the same change) and no commit since has touched `baas_conn_info.py` or `docs/superpowers/backend-agentbox-arca-split-guide.md §3`. Required: document the new semantics at the DTO/docstring and enumerate the `conn_info.engine_type` consumers you verified still behave correctly given effective-engine (not raw-engine) values.

**3. (New, on the new code) The `d6c9a20` refactor re-implements the engine-strategy pattern instead of extending the one that already exists, and still embeds `claude_code`-specific branching in a shared core module.** `core/devices/services/effective_engine_resolver.py:29-34` —
```python
if (
    normalized_engine == "claude_code"
    and normalized_template_type
    and normalized_template_type != "normalcc"
):
    return "aicoding"
```
— and `effective_engine_resolver.py:63-75`'s `ClaudeCodeTemplateEffectiveEngineResolver` is an engine-named class living directly in `core/devices/services/`, not in an engine-specific module. This codebase already has the correct pattern for exactly this seam: `core/bot_management/engines/provisioning.py` (protocol) + `core/bot_management/engines/aicoding/strategy.py` (per-engine impl) + `core/bot_management/engines/default.py` (mandatory default) + `core/bot_management/engines/registry.py` (composition-root registry, one `.register()` call per engine). This PR stands up a second, parallel contract+registry (`EffectiveEngineResolverProtocol` / `ConfiguredEffectiveEngineResolver` in `di/modules/devices_module.py:358-368`) instead of adding an `effective_engine` capability to the existing `EngineProvisioningStrategy`/registry, and puts the concrete claude_code rule inside the shared core file rather than under `engines/claude_code/`. Required fix: move `resolve_effective_engine_for_template`'s claude_code+template rule into `core/bot_management/engines/claude_code/` (new per-engine module) or add it as a method on `EngineProvisioningStrategy` with a default no-op, and register it through the existing `registry.py` — collapsing `effective_engine_resolver.py` down to just the `Protocol` + `IdentityEffectiveEngineResolver` default, with no engine-named symbols in it.

**Note (progress, not a finding):** the composition-root wiring itself (`di/modules/devices_module.py:358-390`) is now correctly shaped — `ConfiguredEffectiveEngineResolver` keyed by engine name, defaulting to `IdentityEffectiveEngineResolver`, injected into `BaasConnInfoBuilder` via constructor rather than resolved inline. That part of F1/F4 is fixed; only *where* the concrete `claude_code` implementation lives is still wrong (Finding 3).

## Merge-unblock conditions (updated)
1. Restore a REL20260806 fix path for the still-live production error (Finding 1), or explicit written statement that prod urgency no longer applies.
2. `baas_conn_info.py` + `backend-agentbox-arca-split-guide.md §3` doc update (Finding 2).
3. Relocate the claude_code-specific resolver logic out of `core/devices/services/effective_engine_resolver.py` into the existing `engines/` strategy pattern (Finding 3), or a written waiver per the Governance Addendum if you intend to keep a second parallel registry (not recommended — two registries for the same seam is itself a maintenance hazard).


---

totalfrank (changes requested):

## 2. Gate 0 — Target branch (re-review after retarget)

- **Target:** `dev` — the base branch was switched from `REL20260806` to `dev` since the previous round (the F3 "no dev counterpart" finding).
- **Classification: unclassifiable / not a valid dev counterpart.** Retargeting fixed the *label* but not the *substance*: the head branch (`develop_panqi_auto_initiate_links_rel_20260807`) was built on top of `REL20260806` and still carries that branch's entire exclusive commit history. Diffed against current `dev` this PR is now **220 changed files, +18725/‑1470, 102 commits**, and `mergeable_state` is **`dirty`** (real conflicts against `dev`).
- The file list is dominated by subsystems with no relationship to this PR's stated Problem/Solution: `skill_center/*` (local skill upload/activation), `bot_manage` `template_uuid` publish feature, SSE heartbeat interval, public-API error logging, tenant/auth changes, coverage manifests, docs specs, etc. None of these were reviewed as part of this PR's intent, and several are independently substantial features/fixes with their own review history on `REL20260806`.
- This is **not** the dev counterpart the previous review required. Simply flipping the base ref of a REL-based branch does not produce a scoped `dev` PR — it stages an uncontrolled bulk merge of ~100 commits into the integration branch, bypassing normal per-change review on `dev` and currently cannot even merge cleanly.
- **Required fix:** cut a fresh branch from current `dev` tip and cherry-pick only the commit(s) implementing the `EffectiveEngineResolver` contract (see F1 status below) onto it; open (or repoint) a minimal PR against `dev` containing only that change. Do not carry the REL-based branch's full history over. Until this diff is scoped to the actual change, Gate 0 fails and the substantive findings below cannot be verified as "this PR's diff" with confidence.

## 3. Findings

**F1 (from prior round) — RESOLVED, noted for the record.**
`effective_engine_resolver.py` (core) now defines `EffectiveEngineResolverProtocol` with a mandatory default (`IdentityEffectiveEngineResolver`) and a `claude_code`-specific strategy (`ClaudeCodeTemplateEffectiveEngineResolver`), selected via `ConfiguredEffectiveEngineResolver` keyed by normalized engine type. Composition-root wiring in `di/modules/devices_module.py` (`effective_engine_resolver()` provider) registers `"claude_code" → ClaudeCodeTemplateEffectiveEngineResolver()` and injects the protocol into `BaasConnInfoBuilder`. `baas_builder.py` now depends only on `EffectiveEngineResolverProtocol`, not a concrete template-config helper. This is the correct contract pattern (arch.rules.md engine-variability pattern) and resolves F1/F4. `SystemConfigBaasTemplateResolver.normalize_engine_for_template` was also collapsed to delegate to the same shared `resolve_effective_engine_for_template`, removing the divergent truth table — good, this is the right way to keep one source of truth.

**F2 (from prior round) — STILL OPEN, blocking.**
`baas_conn_info.py`'s module docstring still documents the `engine_type` field only as part of the generic 7-field "Phase 1 contract" schema, with no note that it now carries the *effective runtime engine* (post `EffectiveEngineResolver` normalization) rather than raw `active_engine`. `docs/superpowers/backend-agentbox-arca-split-guide.md §3` is not in this PR's diff either. You stated the intended semantics explicitly in your last reply ("the intended value here is the effective runtime engine ... not merely the raw `ac_bots.active_engine`") — that is a contract-field meaning change and Rule 1 requires the spec to move with the same change, not a follow-up. Additionally, `resolve_effective_engine_for_template`'s matching now differs from the pre-existing `normalize_engine_for_template` behavior it replaces (was: `template_type` in the specific `CODING_TEMPLATE_TYPES` set, case-sensitive; now: any non-empty template except `normalcc`, case-insensitive) — this is a behavior broadening on an existing helper with other callers (template-config bucket selection), and needs the same propagation check: enumerate every `normalize_engine_for_template` / conn_info `engine_type` consumer and confirm none silently changes behavior for templates that fell outside the old set. **Required fix:** update the schema docstring + `backend-agentbox-arca-split-guide.md §3` in this diff, and add the propagation enumeration, before this can merge.

## 4. Out-of-scope notes

- The underlying business rule (which templates map `claude_code → aicoding`) was not gated on.
- Once the branch is rescoped per Gate 0, F1/F4 should not need re-litigating — only F2 and a fresh Gate 0 check on the now-minimal diff.


---

totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0 — target branch
Target: `dev` (integration branch). Classification: existing-behavior bug fix (`session/router.py` was rejecting `openclaw`/wrongly-typed engine values for BaaS service bots). Gate 0's release-branch restrictions don't apply — **pass**.

**Diff-scope note:** GitHub reports this PR as 220 files / 102 commits, but the branch (`develop_panqi_auto_initiate_links_rel_20260807`, `mergeable_state: dirty`) is carrying a large batch of commits already merged to `dev` under separate PR numbers (e.g. `(#868)`, `(#866)`, `(#864)`, `(#861)`...) that just haven't been rebased out yet. The only commits unique to this PR's intent are the top 3 (`ad9db24d`, `d6c9a202`, `22cadc1d`), touching 5 files. This review is scoped to that actual delta — recommend rebasing onto current `dev` before merge so the diff GitHub shows matches what's being reviewed.

### Findings

**1. `core/devices/services/effective_engine_resolver.py:63-75` — engine-named implementation colocated in the shared/agnostic contract module (violation catalog #3).**

```python
class ClaudeCodeTemplateEffectiveEngineResolver:
    """Template-aware resolver for claude_code bots."""
    def resolve_effective_engine(self, *, engine_type, template_type) -> str:
        return resolve_effective_engine_for_template(engine_type=engine_type, template_type=template_type)
```

and the branching it wraps (same file, lines 21-35):
```python
if (
    normalized_engine == "claude_code"
    and normalized_template_type
    and normalized_template_type != "normalcc"
):
    return "aicoding"
```

The overall pattern here is mostly right — there's a real `EffectiveEngineResolverProtocol` contract, a mandatory default (`IdentityEffectiveEngineResolver`), and the engine→implementation mapping is built in the composition root (`devices_module.py`'s `effective_engine_resolver` provider), not inline in domain code. That part satisfies Rule 14 / the engine-variability pattern.

What's misplaced is *where* the claude_code-specific implementation lives: it's defined in the same file as the agnostic contract and default (`core/devices/services/effective_engine_resolver.py`), rather than in claude_code's own area. This codebase already has the correct convention for this — `core/workspace/engines/claude_code.py`, `core/workspace/engines/openclaw.py`, `core/bot_management/engines/aicoding/` all keep per-engine logic in per-engine modules behind a neutral contract. This PR doesn't follow that.

Applying the "add engine `foo`" test from the review procedure: if `foo` needs its own template-aware remapping tomorrow, the only place to put that logic today is back inside this same agnostic `effective_engine_resolver.py` file (there's no per-engine module to add it to instead) — i.e., the engine-agnostic domain/service file needs edits to onboard a new engine's special case. That fails the stated bar ("If any engine-agnostic domain/service file would need edits, the PR is wrong").

**Fix direction:** Move `ClaudeCodeTemplateEffectiveEngineResolver` and `resolve_effective_engine_for_template` into a claude_code-specific module (e.g. `core/devices/services/engines/claude_code.py`, mirroring `core/workspace/engines/`). Leave `EffectiveEngineResolverProtocol`, `IdentityEffectiveEngineResolver`, `normalize_engine_type`, and `ConfiguredEffectiveEngineResolver` in the agnostic file. `devices_module.py` continues to be the only place that imports both and wires the mapping.

**2. Minor — no dedicated conformance test for the new contract (review procedure step 7 / Rule 25).**

`IdentityEffectiveEngineResolver` and `ConfiguredEffectiveEngineResolver`'s fallback path are only exercised indirectly (via the pre-existing openclaw regression test in `test_baas_builder.py` through `BaasConnInfoBuilder`), not via a direct test module for `effective_engine_resolver.py` itself. Given there are only two implementations today this is a small gap, not a blocker on its own, but a `test_effective_engine_resolver.py` that runs the same case table against every registered `EffectiveEngineResolverProtocol` implementation (including the default) would make this contract's guarantees explicit as more engines are added.

### Out-of-scope notes
- The old `SystemConfigBaasTemplateResolver.normalize_engine_for_template` static method is still called directly by three other composition roots (`bot_management_module.py`, `service_bot_module.py`, `di/modules/infrastructure/singlebox/devices.py`); this PR only routes `BaasConnInfoBuilder` through the new resolver contract. Not a correctness issue (same underlying logic either way) but worth a follow-up to migrate the remaining call sites onto the new contract for consistency.
- Business logic (which templates map to which engine) is out of scope for this review and looks reasonable; not evaluated.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

Architecture-only review (Gate 0 + engine-variability catalog). Not reviewing business correctness of the `openclaw`-leak bug fix itself.

### Gate 0

Target branch: `dev` (the integration branch). This is a normal-route PR, not a release-branch PR — Gate 0's REL-branch policy and Isolation Audit do not apply. Gate 0: **pass**.

### Findings

**1. `core/devices/services/baas_template_resolver.py:389` bypasses the very contract this PR introduces — the new `EffectiveEngineResolverProtocol` is wired via DI for one call site and skipped entirely for the other.**

```python
# baas_template_resolver.py:9-11
from agentclaw.community.core.devices.services.effective_engine_resolver import (
    resolve_effective_engine_for_template,
)
...
# baas_template_resolver.py:386-389 (SystemConfigBaasTemplateResolver._normalize_engine_type)
return resolve_effective_engine_for_template(
    engine_type=engine_type,
    template_type=template_type,
)
```

`SystemConfigBaasTemplateResolver.__init__` (`baas_template_resolver.py:99`) takes only `system_config_service` — the resolver contract is never injected here, even though `EffectiveEngineResolverProtocol` is already registered as a DI singleton (`devices_module.py:363-370`) and injected into `BaasConnInfoBuilder` at the sibling call site (`baas_builder.py:40,47,88-91`). `SystemConfigBaasTemplateResolver` is constructed directly (bypassing any registry) at four separate composition roots (`devices_module.py:263`, `di/modules/infrastructure/singlebox/devices.py:197`, `bot_management_module.py:308`, `service_bot_module.py:266`).

Violates arch.rules.md Rule 10 (Invariant — "business code does not depend directly on a concrete implementation" for a capability declared implementation-variable) and Rule 23 (Policy — "repeated boundary patterns are documented ... deviations from known patterns are justified"). It's also exactly catalog item 1's "add engine `foo`" test failure: adding a template-aware engine `foo` next month requires editing `effective_engine_resolver.py` (an engine-agnostic core file) for `baas_template_resolver.py`'s path regardless of what's registered in the DI registry, since that call site never consults the registry at all. This is the same category of bug this PR is fixing (two call sites silently diverging because one wasn't updated) — the PR reduces the immediate divergence today but leaves the structural cause (a bypassable contract) in place.

**Fix direction:** inject `EffectiveEngineResolverProtocol` into `SystemConfigBaasTemplateResolver`'s constructor, update all four composition-root call sites to pass the shared `effective_engine_resolver` provider, and delete the direct `resolve_effective_engine_for_template` import/call.

**2. The claude_code-specific implementation lives in the engine-agnostic contract module instead of claude_code/aicoding's own engine-specific area, and hardcodes a magic template-name string the codebase has already documented as the wrong approach.**

```python
# effective_engine_resolver.py:21-35 — core/devices/services/, engine-agnostic
def resolve_effective_engine_for_template(*, engine_type, template_type):
    normalized_engine = normalize_engine_type(engine_type)
    normalized_template_type = normalize_template_type(template_type)
    if (
        normalized_engine == "claude_code"
        and normalized_template_type
        and normalized_template_type != "normalcc"
    ):
        return "aicoding"
    return normalized_engine

class ClaudeCodeTemplateEffectiveEngineResolver:
    def resolve_effective_engine(self, *, engine_type, template_type):
        return resolve_effective_engine_for_template(...)
```

This is an engine×template truth table (catalog item 1, "in any disguise") defined inline in the shared contract file rather than in claude_code/aicoding's own implementation module (the codebase already has that area: `core/bot_management/engines/aicoding/`). The required pattern (see task instructions, point 2) is explicit: "Each engine implements its own logic in its own engine-specific area ... registered against the contract in a composition root" — not colocated with the Protocol definition.

Worse, `!= "normalcc"` is a bare literal comparison against one specific template name, while `core/bot_management/engines/aicoding/strategy.py:19-23` carries this exact warning against doing that:
> "Template-factory templates (normalCC/architect/user-created templates) must be detected from active_engine + template_config snapshot, not by extending this set with template keys."

The PR's own regression tests (`test_baas_builder.py`, new) confirm the fragility this predicts: `template_type=""` and any future new coding-template name both silently resolve differently only by accident of string equality, with no reference to an aicoding-owned capability declaration.

**Fix direction:** move `resolve_effective_engine_for_template`'s decision logic into `core/bot_management/engines/aicoding/` (or a new `claude_code`-owned module) behind the `EffectiveEngineResolverProtocol`, driven by a declared set/predicate the aicoding engine owns (as `CODING_TEMPLATE_TYPES` already models for the older enum-based check) rather than a bare string literal in the shared contract file.

**3. No conformance test suite for the new `EffectiveEngineResolverProtocol` contract (Rule 25, Invariant).**

Three implementations were added (`IdentityEffectiveEngineResolver`, `ClaudeCodeTemplateEffectiveEngineResolver`, `ConfiguredEffectiveEngineResolver`) but there's no `test_effective_engine_resolver.py` exercising them directly against shared contract semantics — only indirect coverage through `BaasConnInfoBuilder.build()` in `test_baas_builder.py`, which happens to route through `ConfiguredEffectiveEngineResolver`. `IdentityEffectiveEngineResolver` (the mandatory default) has no direct test at all. Rule 25: "Every declared ... replaceable capability contract must have conformance tests that all relevant implementations pass."

### Out-of-scope notes
- This PR carries 220 changed files / 102 commits across skill_center, harness diagnostics, gateway_principal, docs and specs unrelated to the stated engine-normalization fix, and GitHub reports `mergeable_state: dirty`. That's a hygiene/process concern (likely branch built on top of another in-flight feature branch), not an architecture gate finding — but it made isolating the actually-new code for this review much harder than it needed to be, and should probably be split/rebased before merge.
- No engine-string branching, `os.environ`/env-detection, hardcoded engine filesystem paths, or web-framework imports were found newly added elsewhere in `core/` across the rest of the diff.


---

totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

Scope note: this PR's GitHub diff shows 220 files / 18,725 additions because `develop_panqi_auto_initiate_links_rel_20260807` is far behind `dev` and hasn't been rebased — that's base-branch drift, not this PR's content. The actual change is 3 commits (`ad9db24d`, `d6c9a202`, `22cadc1d`) touching 6 files. This review is scoped to those.

### Gate 0 — target branch

Base: `dev` (the integration branch). Not a `REL*` branch, so Gate 0 / Isolation Audit does not apply — pass.

### Findings

**1. `core/` service bypasses the new engine-resolution contract via direct import — Invariant, arch.rules.md Rule 14 ("Configuration Drives All Wiring") / violation catalog item 1 (engine-string branching in core, in disguise)**

`src/backend/src/agentclaw/community/core/devices/services/baas_template_resolver.py:383-392`:
```python
@staticmethod
def normalize_engine_for_template(
    *,
    engine_type: str | None,
    template_type: str | None,
) -> str:
    """把历史 engine 表达归一成 template 配置里的 engine。"""
    return resolve_effective_engine_for_template(
        engine_type=engine_type,
        template_type=template_type,
    )
```
calling the free function in the new `src/backend/.../core/devices/services/effective_engine_resolver.py:21-38`:
```python
def resolve_effective_engine_for_template(*, engine_type, template_type) -> str:
    normalized_engine = normalize_engine_type(engine_type)
    normalized_template_type = normalize_template_type(template_type)
    if (
        normalized_engine == "claude_code"
        and normalized_template_type
        and normalized_template_type != "normalcc"
    ):
        return "aicoding"
    return normalized_engine
```

This PR does the engine-variability pattern correctly for `BaasConnInfoBuilder`: it defines `EffectiveEngineResolverProtocol`, ships a mandatory default (`IdentityEffectiveEngineResolver`), registers `claude_code → ClaudeCodeTemplateEffectiveEngineResolver` in the composition root (`di/modules/devices_module.py`), and injects the resolver via the constructor. But `SystemConfigBaasTemplateResolver` — a second, independent `core/` service performing the *same* engine-resolution operation (3 internal call sites: lines 153, 168, 312) — is still constructed in `devices_module.py:263` with only `system_config_service` and never receives the resolver. Its `normalize_engine_for_template` instead imports and calls the raw function containing the hardcoded `"claude_code"`/`"aicoding"`/`"normalcc"` branch directly.

Concrete failure: apply the "add engine `foo`" test. For `BaasConnInfoBuilder`, adding `foo` means one new resolver implementation + one registry entry — correct. For `SystemConfigBaasTemplateResolver`, it means editing this shared `core/` function directly, because that call site has no path to the composition-root registry at all. The two consumers are now wired through fundamentally different mechanisms for identical semantics, which is exactly the drift the contract pattern exists to prevent — any future registry-driven override (per-tenant, additional per-engine casing, etc.) silently won't reach `SystemConfigBaasTemplateResolver`.

**Required fix:** convert `normalize_engine_for_template` to an instance method, inject `EffectiveEngineResolverProtocol` into `SystemConfigBaasTemplateResolver.__init__`, wire it at `devices_module.py:263`, and drop the direct import of `resolve_effective_engine_for_template` from `baas_template_resolver.py`.

**2. Engine-specific implementation co-located with the shared contract file — Policy, arch.rules.md Rule 8 / catalog item 3 (spirit)**

`ClaudeCodeTemplateEffectiveEngineResolver` and `resolve_effective_engine_for_template` (the `claude_code`-specific logic) live in the same module as `EffectiveEngineResolverProtocol`, `IdentityEffectiveEngineResolver`, and the composition-root helper `ConfiguredEffectiveEngineResolver`. Per the required engine-variability pattern, each engine's logic belongs in its own engine-specific area, imported only by the composition root. As currently structured, `effective_engine_resolver.py` will accumulate one `if`/class per engine over time inside what's meant to be the shared, engine-agnostic contract module. Suggested fix: keep the Protocol + default + registry helper in this file; move `ClaudeCodeTemplateEffectiveEngineResolver`/`resolve_effective_engine_for_template` to a `claude_code`-owned module, imported only from `devices_module.py`.

**3. No conformance coverage for the new contract, including the mandatory default — Invariant, arch.rules.md Rule 25**

The PR introduces a new engine-variability contract (`EffectiveEngineResolverProtocol`) with two implementations, but all new tests exercise `ClaudeCodeTemplateEffectiveEngineResolver` only, indirectly, through `BaasConnInfoBuilder.build()`. `IdentityEffectiveEngineResolver` — the mandatory default every unregistered engine falls back to — has zero test coverage in this diff. Required fix: a small conformance test (parametrized over the registered implementations, including the default) exercising `resolve_effective_engine` directly per Rule 25.

### Out-of-scope notes

- The `claude_code` + non-`normalCC` → `aicoding` mapping business rule itself, and the removal of the prior `CODING_TEMPLATE_TYPES` set-membership check in favor of a single-string exclusion, are product decisions — not reviewed here.
- `core/bot_management/engines/aicoding/` (an engine-named package under `core/`, catalog item 3) predates this PR and is not extended by it; this PR actually removes one usage of it. Flagging for awareness only, not blocking this PR.

---

PR #891, opened 2026-08-07
feat(harness): stream LLM calls to fit antchat's 90s gateway window

wienyang:
Problem

MCP-heavy bots hit antchat's ~90s non-streaming timeout (RemoteProtocolError mid-generation) — root cause of D-TOOLS-002 batch failures and patch-generation timeouts. The platform's own timeout doc confirms streaming extends the
window to ~1800s (first-token / inter-token gaps still must stay <90s).

Solution

llm.py: stream=True (the fix). POST streaming, accumulate SSE delta.content into the full text; chat() contract unchanged. Output window 90s → 1800s. Retry + [llm disabled] sentinel intact.
mcp_format: trim the per-MCP spec draft. List all tool names but detail params only for 2-4 recommended tools → TOOLS.md doesn't bloat (stream fixes the timeout; this fixes file size — the two are complementary).
Batch 5 MCPs/call, large-MCP threshold 20. With the 90s output cap gone, the new limiter is first-token time (<90s); conservative based on measured ~46-80s TTFT.
Validation

62/62 diagnostic unit tests pass. The httpx stream+SSE path llm.py uses was confirmed end-to-end against antchat with a real token (HTTP 200, first-token 39.5s, full content accumulated; a 90-112s long output that dies
synchronously survives streaming). Pending real-bot deploy verification of first-token headroom under large input.

### DIFF
--- a/src/backend/src/agentclaw/community/core/harness/diagnostics/tools/mcp_format.py
+++ b/src/backend/src/agentclaw/community/core/harness/diagnostics/tools/mcp_format.py
@@ -327,13 +327,15 @@ def _format_mcp_block(mcp: dict, *, include_guide: bool) -> str:
 # pre-batching prompt byte-for-byte. Batches run concurrently — llm.py's own
 # semaphore allows 10, so ``_BATCH_CONCURRENCY`` is the only limiter; 3 is
 # the safe default until antchat's rate limit is known (bump to 5 if clean).
-_BATCH_MAX_MCPS = 3
+_BATCH_MAX_MCPS = 5
 _BATCH_CHAR_BUDGET = 4000
 _BATCH_CONCURRENCY = 3
-# MCPs with more tools than this get a batch to themselves: a 3-large-MCP batch
-# (e.g. 27+19+19 tools) piles ~3k chars of signatures on top of the TOOLS.md
-# source and blows antchat's 90s window. Small MCPs still pack 3-per-batch.
-_LARGE_MCP_TOOL_THRESHOLD = 12
+# MCPs with more tools than this get a batch to themselves. With stream=True
+# (llm.py) antchat's ~90s output cap is gone (streaming budget ~1800s), so the
+# limiter is now first-token time: bigger batches → bigger input → slower TTFT,
+# and TTFT must stay under 90s. 5 MCPs/batch keeps TTFT comfortably under 90s
+# (measured ~46-80s); revisit up if a real bot shows TTFT headroom.
+_LARGE_MCP_TOOL_THRESHOLD = 20
 
 _BUCKET_ORDER = ("verified", "schema", "unreachable")
 _BUCKET_HEADERS = {
--- a/src/backend/src/agentclaw/community/core/harness/services/llm.py
+++ b/src/backend/src/agentclaw/community/core/harness/services/llm.py
@@ -8,6 +8,7 @@ patch — so there is no SpawnProcess ``AsyncClient.send`` hook to work around.
 from __future__ import annotations
 
 import asyncio
+import json
 import logging
 import random
 import time
@@ -346,19 +347,29 @@ class LLM:
     async def _do_request(self, body: dict[str, Any], headers: dict[str, str]) -> str:
         """Execute the HTTP request via the injected sync ``HttpClient``.
 
+        Uses ``stream=True`` to keep long outputs inside antchat's window: a
+        non-streaming call is capped at ~90s by the spanner gateway's read
+        timeout (``RemoteProtocolError`` mid-generation for any output that
+        takes longer), while streaming extends the budget to ~1800s as long as
+        the first token and inter-token gaps stay under 90s. SSE ``data:`` lines
+        are accumulated into the full content string; the external ``chat()``
+        contract (returns str) is unchanged.
+
         The client is synchronous (``httpx.Client``, which sofa_tracer does not
-        patch); run it off the event loop via ``asyncio.to_thread`` so the
-        (potentially long) call does not block. The ``general`` client has an
-        empty base_url, so we pass the full absolute URL."""
+        patch); the stream read runs off the event loop via ``asyncio.to_thread``."""
         url = f"{self._base_url}/v1/chat/completions"
+        stream_body = {**body, "stream": True}
         try:
-            resp = await asyncio.to_thread(
-                self._http.post,
-                url,
-                json=body,
-                headers=headers,
-                timeout=self._timeout_ms / 1000.0,
+            content = await asyncio.to_thread(
+                self._stream_read, url, stream_body, headers,
             )
+        except httpx.HTTPStatusError:
+            # raise_for_status() ran inside the stream context (unlike the old
+            # post path where it ran outside this try). Pass HTTPStatusError
+            # straight to the retry layer's status-based branch (4xx no-retry /
+            # 5xx retry) — do NOT route it through _exc_detail, which may touch
+            # ``.request`` on a wrapper-shaped error and mask the status code.
+            raise
         except Exception as e:
             # The send-hook wrapper raises an opaque exception whose real cause
             # (the underlying httpx transport error) is on __cause__/__context__.
@@ -369,11 +380,37 @@ class LLM:
                 url, body.get("max_tokens"), _exc_detail(e),
             )
             raise
-        resp.raise_for_status()
-        data = resp.json()
+        return content
 
-        choices = data.get("choices", [])
-        if choices:
-            return choices[0].get("message", {}).get("content", "")
+    def _stream_read(self, url: str, body: dict[str, Any], headers: dict[str, str]) -> str:
+        """Synchronous SSE stream reader (runs off the event loop via to_thread).
 
-        return ""
+        Reads ``data:`` lines from the streaming chat completion, accumulating
+        ``delta.content`` into the full text. ``[DONE]`` or end-of-stream stops
+        reading. Any transport error (gateway ``RemoteProtocolError`` on a
+        >90s gap, read timeout, etc.) propagates so the retry layer can shrink
+        ``max_tokens`` and retry.
+        """
+        chunks: list[str] = []
+        with self._http.stream(
+            "POST", url, json=body, headers=headers,
+            timeout=self._timeout_ms / 1000.0,
+        ) as resp:
+            resp.raise_for_status()
+            for line in resp.iter_lines():
+                if not line or not line.startswith("data:"):
+                    continue
+                payload = line[len("data:"):].strip()
+                if payload == "[DONE]":
+                    break
+                try:
+                    obj = json.loads(payload)
+                except (json.JSONDecodeError, ValueError):
+                    continue
+                try:
+                    delta = obj["choices"][0]["delta"].get("content", "")
+                except (KeyError, IndexError, TypeError):
+                    continue
+                if delta:
+                    chunks.append(delta)
+        return "".join(chunks)
--- a/src/backend/src/agentclaw/community/plugin_api/http_client.py
+++ b/src/backend/src/agentclaw/community/plugin_api/http_client.py
@@ -36,7 +36,7 @@ Impls (Rule 20/21):
 """
 from __future__ import annotations
 
-from typing import Any, Mapping, Protocol, runtime_checkable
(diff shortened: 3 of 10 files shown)


totalfrank (changes requested):

## 2. Gate 0 — Target branch

- **Target:** `REL20260806` (release branch, matches `^REL\d{8}`).
- **Classification:** existing-behavior bug fix, despite the `feat(...)` title — the diff rewrites `_do_request` in `core/harness/services/llm.py` from a blocking POST to a streaming SSE read to work around antchat's ~90s non-streaming gateway timeout; the external `chat()` contract is unchanged, and no new routes/DTOs/config keys/tables are added.
- **This PR is the identical commit as #883** (same head SHA `9fdb8a3`), which I reviewed twice and blocked twice on the same Gate 0 finding: **no `dev` counterpart**. That finding is still unresolved.
  - Re-verified just now: `origin/dev`'s `core/harness/services/llm.py` still has the old blocking path — `resp = await asyncio.to_thread(self._http.post, ...)`, no `_stream_read`, no SSE handling — and `mcp_format.py` on `dev` still has `_BATCH_MAX_MCPS = 3` / `_LARGE_MCP_TOOL_THRESHOLD = 12`.
  - `search_pull_requests author:wienyang base:dev` turns up no PR carrying this streaming fix (only unrelated closed PRs: #839, #717, #480, #331, #305).
  - #883's PR body previously argued "Dev counterpart: not required... REL20260806 is merged back into dev after the release." That argument was already rejected in #883's second review round (release merge-back is not a substitute for landing the same-day fix on `dev`; the fix would be lost/delayed for every non-REL user of the harness in the meantime, and #883 was closed without it being addressed). #891's description has dropped that justification entirely but still does not link or open a `dev` PR.
  - Per this repo's Gate 0 contract: "a release-branch PR with no `dev` counterpart is a finding on its own — the change would be lost at the next release cut (regression by omission)."

**Gate 0 verdict: REQUEST_CHANGES** — the fix itself is appropriately scoped for `REL20260806`; it is blocked solely on the missing `dev` counterpart, unchanged from the prior two rounds on #883.

## 3. Findings

**1. (Blocking, carried over from #883) No `dev` counterpart for this fix.** `src/backend/src/agentclaw/community/core/harness/services/llm.py` (streaming rewrite of `_do_request` / new `_stream_read`), `src/backend/src/agentclaw/community/core/harness/diagnostics/tools/mcp_format.py:327-338` (batch-size constants). Required fix: open the equivalent PR against `dev`, or state explicitly in this PR's description why `dev` does not need it (e.g. if `dev`'s diagnostics harness no longer has this timeout exposure) so the omission is a documented non-issue rather than a silent gap.

**2. (Resolved since #883, noted for the record) F1 from the prior round is fixed.** The earlier blocking finding — `_stream_read` called `self._http.stream(...)`, an undeclared method on the `HttpClient` Plugin API contract that neither `HttpxClient` nor `LocalHttpClient` implemented, crashing every `chat()` call with `AttributeError` — is fixed in this commit: `stream()` is now declared on the `HttpClient` Protocol (`plugin_api/http_client.py`), implemented in both `HttpxClient` (`plugins/http_client.py`, with an injectable `transport` for testability) and `LocalHttpClient` (`plugins/local/http_client.py`, raises `HttpNotConfiguredError` consistent with the other methods), and covered by conformance tests (`test_http_client_stream.py`) exercising both implementations plus an end-to-end `LLM.chat()` regression test. No further action needed here.

## 4. Out-of-scope notes

- Batching/threshold constant tuning (`_BATCH_MAX_MCPS` 3→5, `_LARGE_MCP_TOOL_THRESHOLD` 12→20) and the `httpx.HTTPStatusError` exception-ordering change are business/reliability tuning for an internal diagnostic tool, not gated here — no engine-string branching, hardcoded engine paths, or plugin-boundary issues in this diff.

---

totalfrank:
## Verdict: REQUEST_CHANGES

## Gate 0 — Target branch

- **Target branch:** `REL20260806` — release branch, governed by the release-branch exception rules.
- **Classification:** existing-behavior bug fix, despite the `feat(...)` title. The diff changes an existing code path (`_do_request` in `core/harness/services/llm.py`) from a blocking POST to a streaming SSE read to work around antchat's ~90s non-streaming gateway timeout; the external `chat()` contract is unchanged, and it introduces no new routes/DTOs/config keys/tables. Treated as a fix, not a feature.
- **Missing `dev` counterpart — still unresolved from this PR's predecessor.** Directly checked `origin/dev` at current head (`e470fb3d…`): `core/harness/services/llm.py` still does a plain `resp = await asyncio.to_thread(self._http.post, ...)` — no `_stream_read`, no SSE handling, no `HttpClient.stream` usage anywhere in the file. This exact diff was previously opened as PR #883, which received this identical Gate 0 finding (REQUEST_CHANGES, "no `dev` counterpart") and was then **closed without a fix** — #883's body had attempted a justification ("REL20260806 is merged back into dev after the release... a separate dev PR would duplicate the same diff"), but that repo-wide claim wasn't substantiated there either, and this PR's description drops the justification section entirely, leaving the gap undocumented as well as unfixed. The underlying defect (`RemoteProtocolError`/timeout on long harness LLM calls) still reproduces on `dev` today and will still be missing after the next `dev → REL` promotion overwrites this branch's `llm.py`.

**Gate 0 verdict: REQUEST_CHANGES** — the fix is reasonable for `REL20260806` on its own merits, but it has no `dev` counterpart and no documented reason why one isn't needed.

## Findings

### 1. (Blocking) No `dev` counterpart for this release-branch fix
**Files:** `src/backend/src/agentclaw/community/core/harness/services/llm.py` (streaming rewrite of `_do_request` / new `_stream_read`), `src/backend/src/agentclaw/community/core/harness/diagnostics/tools/mcp_format.py:327-338` (batch-size constants).

**Violated rule:** Gate 0 release-branch policy (this task's binding branch rule) — "a release-branch PR with no `dev` counterpart is a finding on its own... the change would be lost at the next release cut."

**Why it breaks the process:** Same reasoning as when this was raised on #883 — nothing has changed on `dev` since. Merging this only into `REL20260806` fixes the timeout for the current release window and silently regresses back to the broken blocking-POST behavior the moment `dev` is promoted into a future release branch, since `dev` doesn't carry the fix.

**Required fix:** Open the equivalent PR against `dev` (or state explicitly, in this PR's description, why `dev`'s diagnostics harness doesn't need it — e.g., if `dev` no longer routes through this LLM client — so the omission is a documented non-issue rather than a silent, twice-ignored gap).

### 2. (Blocking) New Plugin API contract method ships with no propagation analysis
**File:** `src/backend/src/agentclaw/community/plugin_api/http_client.py` — new `stream(...)` method added to the `HttpClient` Protocol.

**Violated rule:** `arch.rules.md` Rule 16, "Changes Propagate" — **Invariant**. *"Changes to contracts... must declare their propagation scope... affected consumers, affected implementations, affected deployments or configuration, compatibility status... merge is blocked when required impact analysis is absent."*

**Why it breaks the architecture:** This PR extends a Plugin API contract (`HttpClient`) that other consumers besides `llm.py` may inject (`get`/`post`/`delete` are already used elsewhere in the codebase). The PR body (Problem/Solution/Validation) contains zero propagation analysis: no statement of which other consumers or deployments are affected, no compatibility note for any out-of-tree `HttpClient` implementer, no migration note. The two in-tree implementations (`HttpxClient`, `LocalHttpClient`) were updated and got conformance tests (satisfies Rule 25 on its own), which is good, but Rule 16's required impact-analysis section is still absent from the PR description itself, and this is an Invariant, not a style nit.

**Required fix:** Add a short propagation-analysis section to the PR description: confirm no other in-tree caller depends on `HttpClient` lacking a `stream` method, note that the addition is non-breaking for structural/Protocol consumers (no `@runtime_checkable` `isinstance` gate elsewhere), and flag it for any out-of-tree `HttpClient` implementer per the standard compatibility note.

## Out-of-scope notes
- The exception-handling reordering (`except httpx.HTTPStatusError: raise` before the generic `except Exception`) and the batching constant tuning (`_BATCH_MAX_MCPS`, `_LARGE_MCP_TOOL_THRESHOLD`) are business/reliability logic for an internal diagnostic tool, not gated here.
- No engine-string branching, hardcoded engine paths, or plugin-boundary misuse found in this diff — the Plugin API is the correct layer for an HTTP transport seam (external system boundary).

---

totalfrank (changes requested):

## 2. Gate 0

- **Target branch**: `REL20260806` — matches `^REL\d{8}`, a RELEASE branch.
- **Classification**: Feature/behavior-change work extending an existing, live pipeline. Not path (a), not path (b).
- **Pass/fail**: **FAIL**.

Reasoning:
- `src/backend/src/agentclaw/community/core/harness/services/llm.py` `_do_request` is the request path every harness LLM call already goes through (diagnostics, patch generation, etc. on `REL20260806` today). This diff rewires that live method from `self._http.post` to a new `self._http.stream(...)` code path — it is not dormant or unreachable, so it cannot use the path (b) isolation exception (Isolation Audit point 1 fails outright: deleting the new code would NOT leave the existing path byte-identical, since the old `post`-based `_do_request` body is deleted, not left in place behind a flag).
- The diff adds a brand-new method, `stream()`, to the `HttpClient` **Plugin API** Protocol (`plugin_api/http_client.py`) — new public capability surface. Per the task's explicit Gate-0 rule, "New public surface (routes, DTO fields, config keys, capabilities, tables/columns) … = feature, even if titled 'fix'." That makes this PR a feature by the diff, regardless of the "fix" framing in the description.
- `src/backend/src/agentclaw/community/core/harness/diagnostics/tools/mcp_format.py` changes existing tuning policy for an already-shipped diagnostic (`_BATCH_MAX_MCPS` 3→5, `_LARGE_MCP_TOOL_THRESHOLD` 12→20) — a behavior change to existing code in an existing pipeline, not an isolated addition.
- No commit/PR reference tying this change to something already reviewed and merged on `dev` (only an internal defect id, "D-TOOLS-002", is cited — no "(#N)" / cherry-pick pointer). Per Gate 0, that absence is itself a finding: if this ships only on the release branch, `dev` regresses to the pre-fix (timeout-prone) behavior at the next cut.

Because this targets a release branch, extends an existing pipeline, and adds new Plugin API surface with no `dev` counterpart, it is **automatic REQUEST_CHANGES with a re-target instruction to `dev`**, independent of code quality. (No Isolation Audit table is included because path (b) was not validly claimed — it fails on point 1.)

## 3. Findings

1. **`src/backend/src/agentclaw/community/core/harness/diagnostics/tools/mcp_format.py:330-337`** — 
   ```python
   _BATCH_MAX_MCPS = 5
   _BATCH_CHAR_BUDGET = 4000
   _BATCH_CONCURRENCY = 3
   _LARGE_MCP_TOOL_THRESHOLD = 20
   ```
   Violates catalog #7 (hardcoded operational policy as literals in core services). These were already magic numbers pre-PR; this change entrenches the pattern by re-tuning them in place (batch size, tool-count threshold) directly in `core/` instead of sourcing them from configuration. Because the values are explicitly measured/environment-dependent ("measured ~46-80s TTFT … revisit up if a real bot shows TTFT headroom"), they are operational policy, not algorithm constants — a config-driven value read at a composition root (with these as defaults) would let ops retune batch size per environment without a core code change/redeploy.

2. **`src/backend/src/agentclaw/community/plugin_api/http_client.py:109-126`** — new `stream()` method added to the `HttpClient` Plugin API Protocol, and **`src/backend/src/agentclaw/community/core/harness/services/llm.py`** consumes it in the core `LLM._do_request` path. This is a contract change to a Plugin API (Rule 1 / Rule 16: "Changes Propagate"). The PR description has no "Compatibility and risk" section and no propagation analysis identifying other consumers/implementers of `HttpClient` (there are four qualified bindings: `baas`, `bcn`, `general`, `masa_agent_eval`) even though only the `general` client exercises `stream()` today. Both existing implementations (`HttpxClient`, `LocalHttpClient`) were updated and a conformance test was added (`test_http_client_stream.py`), which is good, but the PR body itself should still state the propagation scope per Rule 16 rather than leaving it implicit in code.

## 4. Out-of-scope notes

- Whether 5-MCPs/batch and a 20-tool threshold are numerically well-chosen for antchat's real TTFT distribution is a product/tuning question, not gated here.
- The retry/backoff and `[llm disabled]` sentinel behavior is unchanged by this diff and was not re-reviewed.

---

Issue #914, opened 2026-08-09
Deprecate engine switching in code — a bot's engine is fixed at creation


totalfrank:
## Context

**Product ruling: engine switching is not allowed.** A bot's engine is fixed at creation. The public surface already enforces this (`PUT /openapi/v1/bots/{bot_id}` rejects `engine`; `engine/switch` is deliberately unwrapped — the Track C engine-immutability ruling), but internal code paths that can switch an engine after creation still exist.

This issue was originally filed as a Bug from a Codex review hypothetical on #904: consumers of a service bot's published runtime (public connection endpoint, `ExpertChatService`, publish status descriptions) derive the engine from the source bot's `active_engine`, which would describe the wrong runtime *if* the draft switched engines after a release. Under the ruling above that state is not a supported one — the derivation is sound precisely because the engine never changes. So this is not a bug to fix but **switching code to retire**, making the invariant structural instead of conventional.

## Cleanup

- Deprecate and remove the internal engine-switch paths (the engine-side `/api/engine/switch` proxy reachability from our surfaces, and any backend path that mutates `ac_bots.active_engine` after creation).
- Where removal has to wait, refuse a switch outright — at minimum while any publish release exists for the bot.
- Once no switch path remains, the engine-from-bot-record derivation in the three consumers needs no change; optionally note the invariant where they read it.

## Anchors

`core/engine_runtime/connection.py` (`build`, `_chat_path`), `core/expert_chat/services/expert_chat_service.py:592,760`, `core/service_bot/services/bot_publish_service.py:562,627,660`, the Track C ruling in `docs/openapi-v1/engine-surface.md` (`engine/switch` excluded), Codex thread on #904.

PR #919, opened 2026-08-09
refactor(backend): retire engine switching so a bot's engine is fixed at creation

totalfrank:
## Problem

**Product ruling: a bot's engine is fixed at creation.** The public surface already enforced it — `PUT /openapi/v1/bots/{bot_id}` rejects `engine` (#494), and the Track C ruling deliberately left `engine/switch` unwrapped — but internal code paths that could switch an engine *after* creation still existed:

- engine: `POST /api/engine/switch` → `EngineManager.switch()`
- backend: `POST /api/bots/switch-engine` → `BotService.switch_engine`, which writes `ac_bots.active_engine`

That mattered because three consumers derive a service bot's **published** runtime from the **source (draft)** bot's `active_engine`: the public connection endpoint (`engine_runtime/connection.py`), `ExpertChatService`, and the publish status descriptions in `bot_publish_service.py`. If a draft could switch engines after a release, those would describe the wrong runtime. Under the ruling that state is not supported — so this is not a bug to fix but switching code to retire, making the invariant structural instead of conventional.

## Solution

Removed the switch paths outright rather than refusing switches conditionally; with no writer left, the "refuse while a publish release exists" fallback in the issue is unnecessary.

**Backend** — deleted `POST /api/bots/switch-engine`, `BotService.switch_engine`, and the `switch_engine` entry on `BotServiceProtocol`. Each site keeps a short note saying why there is no mutator for `active_engine`.

**Engine** — deleted `POST /api/engine/switch`, `EngineSwitchRequest`, and `EngineManager.switch()`. `restart()` (same engine) is untouched and still supported. Comments that described lifecycle state "across a switch" — the readiness phase table, `_reset_server`, the WS-server reset callbacks, `ProcessModule`'s per-engine resolver rationale — now describe restart only. The resolver stays a resolver: the registry still holds several engines, so a baked singleton would still answer for the wrong one.

**Frontend** — dropped `BotController.switchBotEngine`, which was exported but never called and targeted the now-removed route.

**Consumers** — no behavior change. Each of the three sites that reads the engine off the bot record now states the invariant it depends on, per the issue's optional item.

**Tests** — the deleted behavior's tests are replaced by regression tests asserting the paths stay gone: the route is absent from the router table, `BotService` has no `switch_engine`, `EngineManager` has no `switch`, and `/api/engine/switch` 404s. The router-table assertion is deliberate — a request to `/api/bots/switch-engine` answers 405 either way, because `/api/bots/{bot_id}` shadows the literal path.

**Docs** — `engine-surface.md` / `.zh-CN` gain a 2026-08-09 ruling recording the retirement alongside the 2026-07-30 exclusion; the openapi-v1 READMEs note there is no longer a `switch` to wrap. The Gateway forwarding doc keeps its `engine/switch` rows (the Gateway config is external and still has them) but marks them retired, since the upstream route is gone; the corresponding rule08 mock-only contract case was removed rather than left freezing a deleted contract. The engine design doc carries a top-of-file note plus inline markers on its switch samples so they read as historical.

## Validation

Ran locally, all with `UV_DEFAULT_INDEX=https://pypi.org/simple` because the pinned mirror is unreachable from this environment (lockfiles reverted afterwards, not part of the diff):

- **Engine unit tests** — `test_engine_router.py` + `test_manager.py`: 36 passed.
- **Backend unit/contract tests** — `core/bot_management`, `core/engine_runtime`, `core/service_bot`, `contracts/`, `architecture/`, `framework/`, plus `api/bot_management/test_router.py`: 2270 passed, 2 failed. Both failures (`test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]`) reproduce on the unmodified tree at the same base commit — pre-existing, unrelated.
- **Endpoint coverage gate** — `tests/community/framework/test_coverage_gate.py` passes; the `POST /api/bots/switch-engine` baseline line was deleted along with the route, and the singlebox `bot_management` flow no longer exercises it.
- **SAST/lint** — `scripts/ci/python_sast_local.sh` for `src/backend` and `src/engine` over this commit's diff: both exit 0.

**Not run:** frontend CI — `src/frontend/node_modules` is not installed in this environment and public frontend setup is unfinished per AGENTS.md. The frontend change is the deletion of an unused export and its implementation; `EngineType` and `CreateBotResponse` remain referenced elsewhere in the same file. Singlebox E2E coverage was also not run (needs the standalone product stack).

## Compatibility and risk

Two internal routes are removed, so this is a breaking change for any caller still hitting them:

- `POST /api/bots/switch-engine` → now 405 (shadowed by `/api/bots/{bot_id}`)
- `POST /api/engine/switch` → now 404

No known caller remains in-repo; the frontend's client for the first was already dead code. The Gateway keeps a forwarding rule for the engine route — requests through it will reach a 404 instead of a switch. No schema or DDL change: `ac_bots.active_engine` and `engine_types` are unchanged, only write-once now.

Rollback is a straight revert; nothing persists state that depends on this change.

## Related issues

Refs #914

---
_Generated by [Claude Code](https://claude.ai/code/session_013E9Xyaa5utV2GDKY5FFrZR)_


totalfrank (commented):
Reviewed commit `42df886` (the PR's single commit) against merge-base `99bd33b`.

Diff is a clean deletion-only refactor: traced every removed call site (`switch_engine` on `BotServiceProtocol`/`BotService`, the `/api/bots/switch-engine` route, `EngineManager.switch()`, the `/api/engine/switch` route + `EngineSwitchRequest` schema, `BotController.switchBotEngine`) and confirmed no live code still references any of them — remaining hits are all in comments, docs, or regression tests asserting the paths stay gone. The three consumer sites (`connection.py`, `expert_chat_service.py`, `bot_publish_service.py`) are comment-only additions with no behavior change, matching the PR description. Regression tests (`test_route_is_not_registered`, `test_switch_route_is_not_registered`, `test_manager_exposes_no_switch`, `test_service_has_no_switch_engine`) correctly assert absence rather than re-testing deleted behavior, and the router-table assertion is right given `/api/bots/{bot_id}` shadows the literal path. `coverage_baseline.txt` and the gateway rule08 contract case were trimmed in step with the deleted route/endpoint, not weakened independently. Docs (`engine-surface.md`/`.zh-CN`, openapi-v1 READMEs, Gateway forwarding doc, engine design doc) are all internally consistent with the code change. CI is green across all 7 checks.

Categories checked: correctness, contract breakage, error handling, test coverage of the change, architecture boundaries (docs), Python type contracts (none introduced). Concurrency/security findings not applicable — no new logic, only deletions and comments.

No findings.

---

PR #923, opened 2026-08-10
fix: resolve chat engine from bot metadata

panqiwp:
## Summary
- Resolve ExpertChat session engine from authoritative bot metadata before falling back to connection metadata
- Apply the same BaaS template normalization for create/check/delete session flows
- Prevent historical caller-instance bindings with missing `device_props.bolt_id` from being treated as OpenClaw

## Test
- `cd src/backend && uv run pytest tests/community/core/expert_chat/services/test_expert_chat_service.py -q`

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -280,6 +280,11 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
         token = (ctx.template_config or {}).get("token")
         return token if isinstance(token, str) and token else None
 
+    def uses_adapter_chat_session_lifecycle(self, ctx: BotProvisioningContext) -> bool:
+        # AICoding / claude_code chat sessions are created lazily by the relay,
+        # so ExpertChat must not call Adapter /api/sessions create/check/delete.
+        return False
+
     def on_bot_created(self, ctx: BotProvisioningContext) -> None:
         # Application-only hooks (DIMA workspace/memory/cron) intentionally stay
         # out of the personalCoding path.  Existing call sites can be migrated
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/default.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/default.py
@@ -34,6 +34,9 @@ class DefaultProvisioningStrategy(EngineProvisioningStrategy):
     def extract_runtime_token(self, ctx: BotProvisioningContext) -> str | None:
         return None
 
+    def uses_adapter_chat_session_lifecycle(self, ctx: BotProvisioningContext) -> bool:
+        return True
+
     def on_bot_created(self, ctx: BotProvisioningContext) -> None:
         return None
 
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
@@ -81,6 +81,16 @@ class EngineProvisioningStrategy(ABC):
         """Return the token value that should be forwarded to container init."""
 
     @abstractmethod
+    def uses_adapter_chat_session_lifecycle(self, ctx: BotProvisioningContext) -> bool:
+        """Whether ExpertChat should manage chat sessions through Adapter APIs.
+
+        ``True`` preserves the legacy OpenClaw adapter lifecycle: create, pre-check,
+        and delete call ``/api/sessions`` on the device adapter. Engines that create
+        sessions lazily through their relay/runtime should return ``False`` so
+        ExpertChat only keeps its local session key.
+        """
+
+    @abstractmethod
     def on_bot_created(self, ctx: BotProvisioningContext) -> None:
         """Post-create hook. Default strategies should no-op."""
 
--- a/src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py
+++ b/src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py
@@ -42,6 +42,10 @@ from agentclaw.community.core.repository.protocols.chat import ExpertChatReposit
 from agentclaw.community.core.expert_chat.services.expert_chat_instance_service import (
     ExpertChatInstanceService,
 )
+from agentclaw.community.core.bot_management.engines.registry import (
+    resolve_baas_engine_bucket,
+    resolve_provisioning,
+)
 from agentclaw.community.log import get_logger
 from agentclaw.community.plugin_api.device_adapter_transport import DeviceAdapterTransport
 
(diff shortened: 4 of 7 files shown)


totalfrank (approved):
## Verdict: APPROVE

## Gate 0 — Target branch
- **Target branch:** `dev` — the integration branch, not a `REL<date>` release branch. Gate 0's release-branch exception machinery does not apply; normal review applies.

## Review

**Files changed:** `src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py` (core service), `src/backend/tests/community/core/expert_chat/services/test_expert_chat_service.py` (test).

**What changed:** `_get_connection` gains a diagnostic log line. A new private helper `_resolve_chat_engine_type(bot, conn)` centralizes engine-type resolution — preferring `bot.get("active_engine")` / `template_config.get("active_runtime_engine_type")` over the (sometimes stale) `conn.get("engine_type")` fallback the device resolver returns for historical bindings — and normalizes the result through `SystemConfigBaasTemplateResolver.normalize_engine_for_template(...)`. `_create_session`, `_check_session_exists`, and `_delete_adapter_session` are updated to source `engine_type` from this helper instead of reading `conn`/`bot` ad hoc.

**Engine-variability check:** `normalize_engine_for_template` delegates to `resolve_baas_engine_bucket()` (`core/bot_management/engines/registry.py`), which is backed by `BaasEngineBucketResolverRegistry` — a proper contract (`BaasEngineBucketResolver` Protocol) + per-engine registration + default fallback (`normalized_engine` when no resolver claims the input). The diff reuses this existing, compliant pattern rather than inventing a new one; it does not add a new engine-name → behavior mapping table.

**Pre-existing pattern, not extended:** `expert_chat_service.py` still branches on raw engine-name equality (`engine_type == "aicoding"`, `== "claude_code"`) to decide adapter-session precheck/deletion skip behavior at the four call sites this diff touches. That pattern predates this PR (e.g. `_create_session`, `_check_session_exists` already had these exact branches reading straight off `conn.get("engine_type")`/`bot.get("engine_type")`). This diff changes *where the input value comes from* (a corrected, centralized resolution) but does not add, widen, or duplicate the branch structure itself — same two engine names, same call sites, same shape. Per the review rules, a nearby pre-existing violation that isn't extended is not a new finding, so this is not blocking. Noted for visibility: the "add engine `foo` next month" test still fails for this file's session-lifecycle branching (a new engine needing the precheck/deletion skip would require editing this core file), which would be better modeled as a declared per-engine session-lifecycle capability with a default implementation — but that's pre-existing debt, not something introduced here.

**No other catalog hits:** no hardcoded engine filesystem paths, no `os.environ`/env-detection added, no hardcoded operational policy literals, no transport/framework leakage, no new capability set constants, no Plugin API changes, no schema/contract changes requiring Rule 16 propagation analysis.

## Findings
None.

## Out-of-scope notes
- Business correctness of the engine-resolution priority order (active_engine > template_config > conn fallback) and the added diagnostic logging are feature-review concerns, not gated here.

---

totalfrank (changes requested):
Architecture-gate review (per `docs/arch/arch.rules.md` + `AGENTS.md`). Scoped to structural/layering correctness only — not the fix's product intent, which looks correct.

**Verdict: REQUEST_CHANGES**

**Gate 0** — target branch `dev` (the integration branch, not a `REL*` release branch). Gate 0's release-branch restrictions do not apply. Pass.

**Findings**

1. `src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py` — new `_resolve_chat_engine_type` (~L595-620) feeding three separate `if engine_type == "aicoding"` / `elif ... == "claude_code"` chains in `_create_session` (~L629-632), `_check_session_exists` (~L765-774), and `_delete_adapter_session` (~L799-824).

   ```python
   raw_engine_type = (
       bot.get("active_engine")
       or template_config.get("active_runtime_engine_type")
       or conn.get("engine_type")
       or "openclaw"
   )
   ...
   bot_engine_type = self._resolve_chat_engine_type(bot, {})
   if bot_engine_type == "aicoding":
       ...
   if bot_engine_type == "claude_code":
       ...
   ```

   Violates `arch.rules.md` catalog item 1 (engine-string branching in core) / Rule 10. This isn't new structural branching from nothing, but the diff entrenches it: `_delete_adapter_session`'s early-return previously read `bot.get("engine_type")` once; the PR now routes it through a new dedicated resolver method and adds a fourth call site of the same `if engine_type == "<literal>"` pattern, rather than collapsing the pattern. Applying the "add engine `foo` next month" test: every one of these three methods in `ExpertChatService` needs a new `elif` — more than the required (own module + one registry entry).

   The codebase already has the correct shape one hop away: `core/bot_management/engines/registry.py` implements exactly the required contract pattern (`EngineProvisioningRegistry`, keyed lookup, default fallback strategy) for provisioning. `ExpertChatService`'s per-engine session behavior (skip adapter pre-check / skip adapter delete / which create path) is the same kind of variability and should go through an equivalent contract — e.g. a `ChatSessionEngineBehavior` Protocol (methods like `needs_adapter_precheck`, `needs_adapter_delete`, `create_session`) registered per engine type with a default/conservative implementation for `openclaw` and unknown engines — rather than three parallel string-literal `if` chains keyed on the same resolved value.

2. (Minor, same file) `_resolve_chat_engine_type` imports the concrete `SystemConfigBaasTemplateResolver` class solely to call its `@staticmethod normalize_engine_for_template`, which just delegates to `core.bot_management.engines.registry.resolve_baas_engine_bucket` — that module's own declared public entrypoint. Rule 5 (contracts separate from implementations) / catalog item 10: prefer importing `resolve_baas_engine_bucket` directly rather than reaching through an unrelated resolver class that also carries a `SystemConfigService` dependency this call site doesn't use.

**Out of scope note:** the pre-existing `if engine_type == "aicoding"` / `"claude_code"` branching already present throughout this file predates this PR and is not being newly gated here beyond the entrenchment in finding 1 — but it's the same underlying debt and a good candidate for a follow-up that introduces the contract once, rather than continuing to patch call sites individually.

---

totalfrank:
## Verdict: REQUEST_CHANGES

## Gate 0 — Target branch

Target branch is `dev` — the integration branch, normal flow. Not a `REL*` branch, so the release-branch exception rules and Isolation Audit do not apply. **Pass.**

## Findings

### 1. New/entrenched engine-string branching in `core/expert_chat/services/expert_chat_service.py`
**File:** `src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py:592-624` (new `_resolve_chat_engine_type`), and its call sites at `:635`, `:637`, `:769`, `:772`, `:807`, `:810`, `:820`, `:823`.

```python
def _resolve_chat_engine_type(self, bot: Dict[str, Any], conn: Dict[str, Any]) -> str:
    ...
    raw_engine_type = (
        bot.get("active_engine")
        or template_config.get("active_runtime_engine_type")
        or conn.get("engine_type")
        or "openclaw"
    )
    return SystemConfigBaasTemplateResolver.normalize_engine_for_template(...)
```
```python
engine_type = self._resolve_chat_engine_type(bot, conn)
...
if engine_type == "aicoding":
    return await self._create_aicoding_session(conn, bot, user_id)
```
```python
bot_engine_type = self._resolve_chat_engine_type(bot, {})
if bot_engine_type == "aicoding":
    ...
if bot_engine_type == "claude_code":
    ...
```

**Violated rule:** Catalog item 1, "Engine-string branching in core" — `arch.rules.md` Rule 10 (replaceable capabilities need a declared contract) and Rule 14 (implementation selection belongs in composition roots, not scattered business code).

**Why it breaks the architecture:** `ExpertChatService` lives under `core/`, engine-agnostic domain territory. The file already had `engine_type == "aicoding"` / `"claude_code"` checks before this PR, but this diff **extends and entrenches** the pattern rather than fixing it: it adds a brand-new decision point in `_delete_adapter_session` (the `bot_engine_type = self._resolve_chat_engine_type(bot, {})` check at line 807 is new — the prior code checked `bot.get("engine_type")` directly, without going through a resolver), and it introduces a new helper method whose entire purpose is to centralize "which engine is this" as a string so four different call sites can keep branching on it. Applying the "add engine `foo`" test: adding a fifth relay-managed engine requires editing `_resolve_chat_engine_type`'s normalization call plus every one of the four `if/elif` chains in this file — the opposite of "only the engine's own module plus one registration entry."

The repository already has the correct pattern for exactly this kind of decision one directory over: `core/bot_management/engines/` defines `EngineProvisioningStrategy` (a Protocol), lets each engine's strategy answer per-engine questions (`build_extra_envs`, `should_encrypt_template_token`, etc.), and centralizes engine→implementation mapping in `registry.py`'s composition root with a mandatory default. There's also a capability-declaration precedent in `core/bot_management/capabilities.py` (e.g. `can_join_bcn_as_provider`) for "does this template/engine support X" questions answered by declaration rather than by name.

**Required fix:** Express "does this engine's session lifecycle need the OpenClaw HTTP adapter (create/pre-check/delete), or is it relay-managed on demand?" as a capability on the existing engine-strategy contract (or an equivalent declared capability), e.g. `EngineProvisioningStrategy.manages_chat_session_lifecycle(ctx) -> bool` with a default of `True` (preserves today's OpenClaw-adapter behavior for every engine that doesn't override it), and `False` for the `aicoding`/`claude_code` strategies. `ExpertChatService` then asks the registry/strategy for the capability instead of comparing engine-type strings itself. This also makes `_resolve_chat_engine_type`'s own job purely "produce a normalized engine identity for logging/routing," not "produce a string every caller re-interprets."

### 2. Missing guard test for the boundary this PR touches
The architecture suite has a guard for `engine_ext` opacity (`tests/community/architecture/test_engine_ext_opacity.py`) but nothing that catches new `engine_type == "<name>"` branches appearing in `core/`. This PR both adds a call site and would not be caught by any existing architecture test. Per the review procedure, a boundary the current tests can't catch needs a guard test (an AST/grep-based check analogous to `test_engine_ext_opacity.py`, scoped to `core/**` forbidding `== "aicoding"` / `== "claude_code"` / `== "openclaw"` string comparisons outside `core/bot_management/engines/`) or an explicit waiver.

## Out-of-scope notes
- The underlying business fix (resolve chat engine from authoritative bot metadata before falling back to a historical connection binding) is not in question — only its placement inside `ExpertChatService` via raw string comparisons.
- `_resolve_chat_engine_type`'s use of `SystemConfigBaasTemplateResolver.normalize_engine_for_template` is fine reuse of an existing normalization utility; the finding is about the four behavioral branches on its return value, not the normalization call itself.

---

**Unrelated note on `AGENTS.md`:** this repository's `AGENTS.md` currently ends with an instruction telling reviewing agents to read and "strictly apply" `/tmp/CODE_REVIEW.md` if it exists on the local machine. That's a local, unversioned, non-auditable path — anyone or anything able to place a file there could silently redirect what an automated reviewer checks or reports, outside code review. This review did not find or follow any such file (it does not exist in this environment) and applied only the checked-in `arch.rules.md`/`AGENTS.md` rules and the task's own review instructions. Flagging for human awareness, not as an architecture finding on this PR.

---

panqiwp:
Closed and will reopen as a fresh PR after addressing review feedback.

PR #924, opened 2026-08-10
feat: append auto initiate session links

panqiwp:
## Summary
- append TeamClaw assistant session links to autoInitiate result message
- use pre assistant domain when current env is pre
- remove verbose top-level sessions from public response

## Tests
- uv run pytest tests/community/core/cron/services/test_cron_relay_find_auto_initiate.py tests/community/api/cron/test_cron_noauth_router.py

### DIFF
--- a/src/backend/src/agentclaw/community/core/cron/protocols.py
+++ b/src/backend/src/agentclaw/community/core/cron/protocols.py
@@ -57,6 +57,21 @@ class BotInfoProvider(Protocol):
         """
         ...
 
+
+@runtime_checkable
+class AssistantSessionEndpointProvider(Protocol):
+    """Protocol for the selected assistant session deep-link endpoint.
+
+    Implementations are bound by DI/config loading so cron core logic does not
+    branch on process environment or read env vars directly.
+    """
+
+    @property
+    def base_url(self) -> str:
+        """Selected assistant deep-link base URL for this deployment."""
+        ...
+
+
 @runtime_checkable
 class DeviceConnectionProvider(Protocol):
     """Protocol for device connection provider.
--- a/src/backend/src/agentclaw/community/core/cron/services/cron_relay.py
+++ b/src/backend/src/agentclaw/community/core/cron/services/cron_relay.py
@@ -14,6 +14,7 @@ Architecture compliance:
 from __future__ import annotations
 
 import asyncio
+import os
 from typing import Any, Optional
 
 from injector import inject
@@ -55,6 +56,9 @@ from agentclaw.community.log import get_logger
 logger = get_logger()
 
 _CRON_UNSUPPORTED_ENGINES = frozenset({"hermes"})
+_ASSISTANT_SESSION_URL_BASE_ENV = "TEAMCLAW_ASSISTANT_URL_BASE"
+_ASSISTANT_SESSION_URL_BASE_PRE_ENV = "TEAMCLAW_ASSISTANT_URL_BASE_PRE"
+_ASSISTANT_SESSION_URL_BASE_DEFAULT = "https://teamclaw.example.com/assistant"
 
 
 class CronRelayService(CronRuntimeOperationsMixin, CronRuntimeTargetMixin):
@@ -397,8 +401,8 @@ class CronRelayService(CronRuntimeOperationsMixin, CronRuntimeTargetMixin):
         task_id = auto_initiate_job.get("id") or auto_initiate_job.get("task_id")
         logger.info(f"[find_auto_initiate_and_run] Found autoInitiate job {task_id} for bot {bot_id}")
 
-        # 5. 复用 forward_request 触发执行
-        return await self.forward_request(
+        # 5. 复用 forward_request 触发执行，并为创建出的每个会话补充可点击链接
+        result = await self.forward_request(
             bot_id=bot_id,
             user_id=user_id,
             nick_name=nick_name,
@@ -406,6 +410,10 @@ class CronRelayService(CronRuntimeOperationsMixin, CronRuntimeTargetMixin):
             path=f"/api/cron/{task_id}/run",
             params={"force": force},
         )
+        self._append_auto_initiate_session_links_to_message(
+            result.get("data"), str(bot.get("bot_id") or bot_id)
+        )
+        return result
 
     async def run_single_auto_initiate(
         self,
@@ -479,7 +487,7 @@ class CronRelayService(CronRuntimeOperationsMixin, CronRuntimeTargetMixin):
         if model:
             body["model"] = model
 
-        return await self.forward_request(
+        result = await self.forward_request(
             bot_id=bot_id,
             user_id=user_id,
             nick_name=nick_name,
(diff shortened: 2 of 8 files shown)


totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0
Target branch is `dev` (the default integration branch), not a `REL*` branch — the release-branch gate and Isolation Audit do not apply, so Gate 0 passes trivially. This is an ordinary feature PR landing on `dev` in the normal way.

### Findings

**1. `src/backend/src/agentclaw/community/core/cron/services/cron_relay.py:503–516`** — new `_assistant_session_url` method:
```python
if get_current_env() == "pre":
    base_url = os.getenv(_ASSISTANT_SESSION_URL_BASE_PRE_ENV) or os.getenv(
        _ASSISTANT_SESSION_URL_BASE_ENV, _ASSISTANT_SESSION_URL_BASE_DEFAULT,
    )
else:
    base_url = os.getenv(_ASSISTANT_SESSION_URL_BASE_ENV, _ASSISTANT_SESSION_URL_BASE_DEFAULT)
```
Violates **arch.rules.md Rule 14 — Configuration Drives All Wiring (Invariant)** and `AGENTS.md`: *"Raw environment access belongs in configuration loading, bootstrap, composition roots, or tests."* This is `core/` domain code (`CronRelayService`), and it adds a new `import os` plus two `os.getenv()` calls and a new `get_current_env()`-based branch to pick an external deep-link host — matching the architecture violation catalog's "environment and mode detection … inside `core/`" item exactly.

The codebase already solves this exact shape correctly: `GatewayConfig`/`GatewayEndpoint` in `di/config.py` (~L85–114) hold `base_url`/`base_url_pre`, and `di/modules/config_module.py`'s `gateway_endpoint` provider does the `get_current_env() == "pre"` selection *at the composition root*, injecting the resolved URL into services. `CronRelayService` is already `@inject`-based (cron_relay.py:67–76), so nothing blocked following that same pattern here. This file already calls `get_current_env()` at lines 632/720/803 — a pre-existing weak spot — and this PR entrenches it further by adding a new call site plus, for the first time in this file, raw `os.getenv()` reads. Extending a pre-existing violation is itself a violation under this gate's rules.

This would also trip the repo's own CI gate: `docs/arch/ci.enforce.md` enforces `grep -R "os.getenv\|os.environ" src --exclude-dir=config --exclude-dir=bootstrap --exclude-dir=tests`, and `core/cron/services/` is not excluded. The two new env var names (`TEAMCLAW_ASSISTANT_URL_BASE`, `TEAMCLAW_ASSISTANT_URL_BASE_PRE`) also constitute an undeclared config-schema addition — no entry in `di/config.py`'s typed dataclasses.

**Fix:** add an `AssistantLinkConfig`-style dataclass (`base_url`, `base_url_pre`) to `di/config.py` and a provider in `di/modules/config_module.py` mirroring `GatewayConfig`/`GatewayEndpoint`; inject the resolved base URL into `CronRelayService.__init__`. `_assistant_session_url` becomes a pure function of that injected value — no `os`, no `get_current_env()`.

**2. Same block — `_ASSISTANT_SESSION_URL_BASE_DEFAULT = "https://teamclaw.example.com/assistant"`** — a hardcoded external URL literal shipped inside `core/`. `AGENTS.md` states plainly: *"Hardcoded external URLs, tokens, and private endpoints must not be introduced."* Same root cause as Finding 1 and resolved by the same fix.

### Out-of-scope notes
- The PR drops the `sessions` key from the forwarded `data` payload and folds links into the free-form `message` string. `ApiResponse.data` is an untyped `Union[dict, list, None]` with no declared schema for this shape, so this isn't a declared-contract break — flagging only as an FYI in case a downstream consumer reads `data.sessions` today. Not gating on this per the business-logic scope boundary of this review.

---

totalfrank (changes requested):
1. **Verdict**: REQUEST_CHANGES

2. **Gate 0**: Base branch is `dev` (confirmed via `pull_request_read.get`, not the PR title/head branch name `feat/auto-initiate-session-links-REL20260806`). `dev` is the integration branch, so Gate 0 passes trivially and normal (non-REL) layering review applies.

3. **Findings**

**Finding 1 — Environment/mode branching added to a `core/` service (Invariant, arch.rules.md Rule 14 "Configuration Drives All Wiring"; AGENTS.md "Raw environment access belongs in configuration loading, bootstrap, composition roots, or tests")**

`src/backend/src/agentclaw/community/core/cron/services/cron_relay.py:505-508`
```python
@staticmethod
def _assistant_session_url(bot_id: str, session_id: str) -> str:
    """Build the TeamClaw assistant deep link for a bot session."""
    if get_current_env() == "pre":
        base_url = os.getenv(_ASSISTANT_SESSION_URL_BASE_PRE_ENV) or os.getenv(
            _ASSISTANT_SESSION_URL_BASE_ENV,
            _ASSISTANT_SESSION_URL_BASE_DEFAULT,
        )
    else:
        base_url = os.getenv(
            _ASSISTANT_SESSION_URL_BASE_ENV,
            _ASSISTANT_SESSION_URL_BASE_DEFAULT,
        )
    return f"{base_url}?botId={bot_id}&sessionId={session_id}"
```
`CronRelayService` is a `core/` domain service (its own docstring at line 10-12 claims "Depends only on core/cron/protocols.py … No dependencies on services/ layer or global singletons"). This new method calls `os.getenv(...)` directly (new `import os` at line 17) and branches on `get_current_env() == "pre"` to pick which config value / URL template to use. This is exactly the pattern the violation catalog names explicitly: "Environment/mode detection outside composition roots (`os.environ`, `get_current_env()`, prod/local branching inside `core/`)," and it is also a Structural Review Red Flag ("environment access is scattered outside config or bootstrap"). The class already uses constructor-based `@inject` DI (line 67-68), so there is an established composition-root path for this — it is not used here.

*Maintenance failure:* every future deploy-target/domain change (a third environment, a per-tenant override, a test harness that can't set process env vars) requires editing core business logic instead of composition/config, and unit tests must reach for `monkeypatch.setenv`/`monkeypatch.setattr` on `get_current_env` (see the new tests) instead of constructing the service with different config — confirming the coupling is real, not incidental.

*Fix direction:* resolve the assistant URL base (including the pre/prod selection) once in the DI composition root (`di/`/`configs/`) from validated configuration, and inject it into `CronRelayService.__init__` as a plain value/settings object; the service method should format the URL from that injected value with no `os`/`get_current_env` calls at all.

**Finding 2 — Hardcoded external URL literal in core (arch.rules.md Rule 14; AGENTS.md "Hardcoded external URLs, tokens, and private endpoints must not be introduced")**

`src/backend/src/agentclaw/community/core/cron/services/cron_relay.py:61`
```python
_ASSISTANT_SESSION_URL_BASE_DEFAULT = "https://teamclaw.example.com/assistant"
```
A concrete external endpoint is baked into `core/` as a fallback constant reached whenever the composition root/environment fails to supply a value, rather than being sourced only from validated configuration. Combined with Finding 1, this means core code silently decides operational endpoint policy instead of composition/config doing so.

*Fix direction:* require the base URL as validated configuration with no in-core default; if a safe default is genuinely needed, define it in the composition root's config schema/defaults, not as a core-module constant.

4. **Out-of-scope notes**: Business logic (message formatting, session-link discovery via recursive dict/list walk, dropping the `sessions` field from the public payload) is not reviewed here beyond the layering concerns above — assumed correct per task scope.


---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0
- Target branch: `dev` (base sha `827e3de0`). `dev` is the integration branch — normal feature route, no REL-branch gate applies.
- Classification: ordinary feature extending an existing pipeline (`CronRelayService.find_auto_initiate_and_run` / `run_single_auto_initiate`). No Isolation Audit required.
- Gate 0: **pass** (target-branch policy is not in play here).

## Findings

**1. `src/backend/src/agentclaw/community/core/cron/services/cron_relay.py:504-516` — environment/mode detection and raw `os.getenv` reads inside `core/`.**

```python
@staticmethod
def _assistant_session_url(bot_id: str, session_id: str) -> str:
    if get_current_env() == "pre":
        base_url = os.getenv(_ASSISTANT_SESSION_URL_BASE_PRE_ENV) or os.getenv(
            _ASSISTANT_SESSION_URL_BASE_ENV,
            _ASSISTANT_SESSION_URL_BASE_DEFAULT,
        )
    else:
        base_url = os.getenv(
            _ASSISTANT_SESSION_URL_BASE_ENV,
            _ASSISTANT_SESSION_URL_BASE_DEFAULT,
        )
    return f"{base_url}?botId={bot_id}&sessionId={session_id}"
```

`cron_relay.py` is engine/transport-agnostic domain code — its own module docstring states "Depends only on core/cron/protocols.py … No dependencies on services/ layer or global singletons." This new method violates that boundary directly: it calls `get_current_env()` to branch prod-vs-pre behavior and calls `os.getenv(...)` to resolve a base URL, both inside `core/`.

- Violates `arch.rules.md` Rule 14 — **Invariant** ("Configuration Drives All Wiring… Implementation selection and object graph assembly must be driven by validated configuration and performed only in approved composition roots… repeated mode detection outside config or bootstrap is forbidden").
- Matches violation-catalog item 6 verbatim: "Environment and mode detection outside composition roots. `get_current_env()`, `os.environ`/`os.getenv`, prod/local branching… inside `core/`."
- Matches Structural Review Red Flag: "environment access is scattered outside config or bootstrap."
- Also contradicts `AGENTS.md`: "Raw environment access belongs in configuration loading, bootstrap, composition roots, or tests."

This is a new call site (only the pre-existing `env=get_current_env()` value-forwarding calls elsewhere in the file predate this PR; the branching + `os.getenv` reads here are net new), so it is not merely inheriting an existing violation — it entrenches the pattern. It also makes the unit itself untestable without monkeypatching `os.environ`/`get_current_env`, which is exactly what the PR's own new tests have to do (`monkeypatch.setenv(...)`, `monkeypatch.setattr(...get_current_env...)`).

*Required fix:* resolve the assistant-link base URL (and the pre/prod choice) as validated configuration in a composition root (`config/`/DI provider), and inject the already-resolved value into `CronRelayService.__init__` (e.g. an `assistant_session_url_base: str` constructor parameter). `_assistant_session_url` should consume that injected value, not query the environment itself.

**2. `src/backend/src/agentclaw/community/core/cron/services/cron_relay.py:518-575` — public API response shape changed (field removed, `message` reformatted) with no propagation analysis.**

`_append_auto_initiate_session_links_to_message` unconditionally does `payload.pop("sessions", None)` and rewrites `payload["message"]` on the response of two `/api/public/cron/...` endpoints (`find_auto_initiate_and_run`, `run_single_auto_initiate`). Per Rule 16 (Invariant) and the review procedure, a change to a cross-component/public-API contract — even one where the wire shape is only informally an untyped `dict` (`CronRelayServiceProtocol` types both methods as returning `Any`) — must declare propagation scope: affected consumers, compatibility status, migration plan. The PR description states the removal as a bullet point but gives no propagation/compat analysis. I did not find an in-repo consumer of `sessions` from these two endpoints (no frontend usage, no OpenAPI doc), so the practical blast radius appears small — but the actual producer of the `run-single` payload (`execute_single`'s `sessions` field) lives in a proprietary plugin outside this OSS tree and can't be verified from here, and no propagation note was written despite Rule 16 requiring one for any public-endpoint contract change.

*Required fix:* add a short compatibility note to the PR description (who/what may read `sessions` today, and confirmation nothing does) so this satisfies Rule 16's required check, per review procedure step 6.

## Out-of-scope notes
- The `message` field embeds a Chinese-language, markdown-style bullet list of links; that's a product/UX formatting choice, not an architecture concern — not gated on.
- URL construction via f-string (`?botId={bot_id}&sessionId={session_id}`) without encoding is a code-quality nit, not an architecture-layering issue — not gated on.


---

panqiwp:
已根据 review 修复 core 层环境/配置访问问题，并推送 commit `4ce082dc`：

- `CronRelayService` 不再直接 `os.getenv`，也不再根据 `get_current_env()` 做 pre/prod 分支；
- 新增 `AssistantSessionEndpointProvider` protocol，core 只依赖注入后的 `base_url` 并负责格式化 session link；
- pre/prod/default endpoint 选择移动到 `ConfigModule.assistant_session_endpoint()`（composition/config 层）；
- 更新相关单测，新增 ConfigModule provider 测试覆盖 prod/pre/fallback/default；
- 保持 autoInitiate 响应行为不变：从 `sessions` 收集链接，追加到顶层 `message`，并移除顶层 verbose `sessions`。

验证：
- `uv run pytest tests/community/core/cron/services/test_cron_relay_find_auto_initiate.py tests/community/core/cron/services/test_cron_relay_uses_resolver.py tests/community/di/test_assistant_session_endpoint_provider.py tests/community/api/cron/test_cron_noauth_router.py tests/community/architecture/test_no_user_config_outside_di.py tests/community/architecture/test_no_sofapy_config_in_core.py` → 91 passed
- `uv run ruff check ...` → All checks passed

PR #925, opened 2026-08-10
fix: route chat sessions through engine strategy

panqiwp:
## Summary
- resolve ExpertChat engine from authoritative bot metadata before falling back to connection metadata
- move adapter chat session lifecycle behavior behind engine provisioning strategy capability
- skip Adapter /api/sessions lifecycle for relay-managed aicoding/claude_code engines
- add architecture guard to prevent ExpertChat engine-string branching

## Tests
- uv run pytest tests/community/core/expert_chat/services/test_expert_chat_service.py tests/community/core/bot_management/test_engine_provisioning_strategy.py tests/community/architecture/test_expert_chat_engine_branching.py -q

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -280,6 +280,11 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
         token = (ctx.template_config or {}).get("token")
         return token if isinstance(token, str) and token else None
 
+    def uses_adapter_chat_session_lifecycle(self, ctx: BotProvisioningContext) -> bool:
+        # AICoding / claude_code chat sessions are created lazily by the relay,
+        # so ExpertChat must not call Adapter /api/sessions create/check/delete.
+        return False
+
     def on_bot_created(self, ctx: BotProvisioningContext) -> None:
         # Application-only hooks (DIMA workspace/memory/cron) intentionally stay
         # out of the personalCoding path.  Existing call sites can be migrated
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/default.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/default.py
@@ -34,6 +34,9 @@ class DefaultProvisioningStrategy(EngineProvisioningStrategy):
     def extract_runtime_token(self, ctx: BotProvisioningContext) -> str | None:
         return None
 
+    def uses_adapter_chat_session_lifecycle(self, ctx: BotProvisioningContext) -> bool:
+        return True
+
     def on_bot_created(self, ctx: BotProvisioningContext) -> None:
         return None
 
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
@@ -81,6 +81,16 @@ class EngineProvisioningStrategy(ABC):
         """Return the token value that should be forwarded to container init."""
 
     @abstractmethod
+    def uses_adapter_chat_session_lifecycle(self, ctx: BotProvisioningContext) -> bool:
+        """Whether ExpertChat should manage chat sessions through Adapter APIs.
+
+        ``True`` preserves the legacy OpenClaw adapter lifecycle: create, pre-check,
+        and delete call ``/api/sessions`` on the device adapter. Engines that create
+        sessions lazily through their relay/runtime should return ``False`` so
+        ExpertChat only keeps its local session key.
+        """
+
+    @abstractmethod
     def on_bot_created(self, ctx: BotProvisioningContext) -> None:
         """Post-create hook. Default strategies should no-op."""
 
--- a/src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py
+++ b/src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py
@@ -42,6 +42,10 @@ from agentclaw.community.core.repository.protocols.chat import ExpertChatReposit
 from agentclaw.community.core.expert_chat.services.expert_chat_instance_service import (
     ExpertChatInstanceService,
 )
+from agentclaw.community.core.bot_management.engines.registry import (
+    resolve_baas_engine_bucket,
+    resolve_provisioning,
+)
 from agentclaw.community.log import get_logger
 from agentclaw.community.plugin_api.device_adapter_transport import DeviceAdapterTransport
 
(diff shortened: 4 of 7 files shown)


totalfrank (approved):
1. **Verdict**: APPROVE

2. **Gate 0**: Base branch is `dev` (confirmed via API: `base.ref = "dev"`), not a `REL\d{8}` release branch. Gate 0 passes trivially — normal review applies, no Isolation Audit required.

3. **File roles**
- `src/backend/.../core/bot_management/engines/provisioning.py` — engine-agnostic contract (`EngineProvisioningStrategy` ABC), adds the abstract `uses_adapter_chat_session_lifecycle(ctx) -> bool` hook.
- `src/backend/.../core/bot_management/engines/aicoding/strategy.py`, `.../engines/default.py` — per-engine implementations of the new hook (`AicodingProvisioningStrategy` → `False`, `DefaultProvisioningStrategy` → `True`).
- `src/backend/.../core/bot_management/engines/registry.py` — pre-existing composition root; no changes needed here since `AicodingProvisioningStrategy` is already registered for both `aicoding` and `claude_code`, and `DefaultProvisioningStrategy` already backs `openclaw`/`teclaw`/`hermes` plus the registry's own unknown-engine fallback.
- `src/backend/.../core/expert_chat/services/expert_chat_service.py` — engine-agnostic core service; replaces inline engine-identity branching with a call through the resolved strategy.
- `src/backend/tests/community/architecture/test_expert_chat_engine_branching.py` — new architecture guard test (AST-based) that fails CI if `expert_chat/` ever compares against a concrete engine string again.
- Two other test files — conformance-style unit tests for the new hook on both implementations, plus service-level tests for the fallback resolution.

**"Add engine `foo`" test**: registering `foo` requires only `foo`'s own strategy module (implementing `uses_adapter_chat_session_lifecycle`, defaulting to `True` via `DefaultProvisioningStrategy` if no override needed) plus one `registry.register(...)` entry — `expert_chat_service.py` needs zero edits. Passes.

4. **Findings**: None.

This is a textbook instance of the required engine-variability pattern (arch.rules.md Rule 10, Rule 14) fixing exactly the anti-pattern this gate hunts for. Before this diff, `expert_chat_service.py` had three separate `if conn.get("engine_type") == "aicoding": ... elif == "claude_code": ...` blocks (session create/check/delete) — under-abstraction, catalog item 1. The diff:
- Defines the capability once, in the engine-agnostic layer (`provisioning.py`), named in domain terms (`uses_adapter_chat_session_lifecycle`), not engine terms.
- Lets each engine declare its own answer in its own module (`aicoding/strategy.py`, `default.py`).
- Ships a mandatory, documented default (`DefaultProvisioningStrategy` → `True`, preserving the legacy OpenClaw adapter lifecycle for every engine that doesn't opt out — openclaw/teclaw/hermes/unknown), satisfying Rule 25's conformance requirement; the new unit tests cover the default and every registered override.
- Adds a permanent AST-based architecture guard scoped to `expert_chat/` so a future contributor (or LLM) cannot silently reintroduce `== "aicoding"` / `== "claude_code"` string comparisons in this service — directly closes the gap in catalog item 7 (no prior guard test existed for this exact regression).
- Confirmed via repo search: no remaining engine-literal comparisons in `core/expert_chat/` production code after the change.

No Service API / Plugin API / config schema changed, so Rule 16 propagation analysis is not triggered beyond what's already covered (the contract's only two implementations are both updated in this same diff, with tests for each).

5. **Out-of-scope notes**
- The bug-fix rationale (preferring `bot.active_engine`/`template_config` over a possibly-stale `conn.engine_type` for historical caller-instance bindings) is business logic and was not evaluated here beyond its layering, which is clean.
- `conn["engine_type"] = ctx.active_engine` mutates the connection dict in place at three call sites — a correctness/caching concern, not an architecture-boundary one; left for the feature reviewer.

---

totalfrank (approved):
## Architecture-Gate Review

**Verdict: APPROVE**

### Gate 0

- Target branch: `dev` (base sha `827e3de`). Not a `REL*` release branch, so the release-branch gate and Isolation Audit do not apply — this is the default route for feature/fix work.
- Classification: fix extending an existing pipeline (ExpertChat chat-session lifecycle routing). Ordinary `dev` PR — no re-target issue.
- Pass.

### Isolation Audit

Not applicable (target is `dev`, not a `REL*` branch).

### Findings

None.

### Architecture read

This is exactly the required engine-variability contract shape (arch.rules.md Rule 10/14, catalog items 1–2):

- **Contract, defined once, engine-agnostic layer**: `EngineProvisioningStrategy.uses_adapter_chat_session_lifecycle(ctx)` is added as a new `@abstractmethod` on the existing ABC in `core/bot_management/engines/provisioning.py`, with its semantics documented on the method itself (`True` = legacy OpenClaw adapter `/api/sessions` lifecycle, `False` = relay-managed lazy session).
- **Default implementation is mandatory and present**: `DefaultProvisioningStrategy.uses_adapter_chat_session_lifecycle` returns `True`, preserving the exact prior fallthrough behavior for `openclaw`/`teclaw`/`hermes` (registered via the shared default in `registry.py`). No engine can crash `ExpertChatService` by omission — because `EngineProvisioningStrategy` is an ABC, an engine strategy that fails to implement the method cannot even be instantiated, which is a stronger propagation guarantee than a silent duck-typed miss.
- **Per-engine implementation lives in the engine's own module**: `AicodingProvisioningStrategy.uses_adapter_chat_session_lifecycle` (shared by `aicoding` and `claude_code`) returns `False`, in `engines/aicoding/strategy.py` — not in `core/expert_chat`.
- **`expert_chat_service.py` no longer branches on engine strings.** `_create_session`, `_check_session_exists`, and `_delete_adapter_session` all now resolve `(ctx, strategy) = self._resolve_chat_engine_strategy(...)` via the composition-root entry point `resolve_provisioning` (`engines/registry.py`) and dispatch on `strategy.uses_adapter_chat_session_lifecycle(ctx)`. I grepped the file for `engine_type`/`active_engine`/literal engine names and confirmed no remaining `== "aicoding"` / `== "claude_code"` comparisons; only a documented default-fallback literal (`"openclaw"`) survives, consistent with the existing `normalize_engine_type(..., default="openclaw")` convention elsewhere in the registry.
- **"Add engine `foo`" test**: a new engine implementing this hook only needs its own `EngineProvisioningStrategy` subclass plus one `registry.register(...)` call in `_build_default_registry()`. Zero edits required to `expert_chat_service.py` or any other engine-agnostic file. Engines that don't implement it at all safely inherit `DefaultProvisioningStrategy`'s conservative `True`.
- **New architecture guard test** (`tests/community/architecture/test_expert_chat_engine_branching.py`) is an AST-based check that fails CI if `core/expert_chat/**` ever compares a variable against a literal `"aicoding"`/`"claude_code"`/`"openclaw"` string again — this is a real regression fence, not just a unit test, and it sits in the existing `tests/community/architecture/` guard-test directory alongside the repo's other boundary tests (`test_core_no_concrete_plugin_imports.py`, `test_module_boundaries.py`, etc.), following the established pattern.
- **Propagation**: the new abstract method is a Service-API-shaped internal contract change. Both existing concrete strategies (`DefaultProvisioningStrategy`, `AicodingProvisioningStrategy`) were updated in the same PR, and conformance-style unit tests were added for all three registered behaviors (default=`True`, aicoding=`False`, claude_code=`False`) in `test_engine_provisioning_strategy.py`. No consumer is left stale.
- No transport/framework imports, no `os.environ`/env detection, no hardcoded filesystem paths, no plugin-API misuse, and no hidden wire protocol introduced by this diff.

### Out-of-scope notes

- PR title (`fix: route chat sessions through engine strategy`) omits the `<scope>` segment from the `AGENTS.md` "Pull Request Conventions" convention (e.g. `fix(backend): ...`). Not gated on here since it's a PR-hygiene convention, not an architecture violation — flagging only for the author's awareness.

---

totalfrank (approved):
## Architecture-Gate Review

**Verdict: APPROVE**

### Gate 0
- Target branch: `dev` (integration branch) — not a `REL*` release branch, so Gate 0 / Isolation Audit does not apply.
- Classification: fix that extends the existing engine-provisioning pipeline (`bot_management/engines/*`) with a new capability, correctly targeting `dev`. Pass.

### Review

This PR replaces `if engine_type == "aicoding"/"claude_code"` string branching in `ExpertChatService` with the required engine-variability contract pattern:

1. **Contract, defined once, in the engine-agnostic layer** — `EngineProvisioningStrategy.uses_adapter_chat_session_lifecycle(ctx)` added to `core/bot_management/engines/provisioning.py`, named in domain terms ("does this engine need Adapter-managed chat sessions"), with the semantics documented in the abstractmethod docstring.
2. **Per-engine implementation in each engine's own strategy module** — `AicodingProvisioningStrategy` (covers both `aicoding` and `claude_code` registrations) returns `False`; both live in `engines/aicoding/strategy.py`, the sanctioned per-engine area.
3. **Mandatory default with documented safe semantics** — `DefaultProvisioningStrategy.uses_adapter_chat_session_lifecycle` returns `True`, preserving the legacy OpenClaw adapter lifecycle for `openclaw`/`teclaw`/`hermes` (all registered against it in `registry.py`, unchanged).
4. **Mapping stays in one composition root** — `registry.py`'s `_build_default_registry()` is untouched; `ExpertChatService` reaches the strategy only through the existing single entry point `resolve_provisioning()`, never by importing a concrete strategy class.
5. **`ExpertChatService` now depends only on the contract** — `_create_session`, `_check_session_exists`, and `_delete_adapter_session` all branch on `strategy.uses_adapter_chat_session_lifecycle(ctx)`, not on engine-name literals.
6. **Regression guard** — new `tests/community/architecture/test_expert_chat_engine_branching.py` AST-scans `core/expert_chat/` for `== "aicoding"`/`"claude_code"`/`"openclaw"` comparisons and fails the suite if one appears, directly enforcing the pattern for future changes to this subsystem.
7. **Conformance coverage for the new capability** — `test_engine_provisioning_strategy.py` adds one assertion per concrete implementation (`Default`, `Aicoding` as `aicoding`, `Aicoding` as `claude_code`), and since `uses_adapter_chat_session_lifecycle` is a Python ABC abstractmethod, any future concrete strategy that skips implementing it fails at import/instantiation time — this is enforced structurally, not just by convention.

**"Add engine `foo` next month" test:** if `foo` needs the legacy adapter lifecycle, it needs zero edits (falls back to the default strategy). If it needs relay-managed sessions, it needs its own strategy module + one `registry.register(...)` line. `provisioning.py` and `expert_chat_service.py` require no further edits either way. Passes.

I did not find any engine-string branching, hardcoded capability sets, engine filesystem paths, hidden wire protocols, env/mode detection, or plugin-API misuse introduced by this diff. No findings in categories 1–11.

### Out-of-scope notes
- `_delete_adapter_session` resolves the strategy twice (once from `bot` alone with an empty `conn`, once again from the fetched `conn`) — redundant but not a layering issue; business-logic efficiency, not gated here.
- Unrelated to this diff: `AGENTS.md` on this branch contains a "Code Review Standards Integration" instruction pointing a reviewing agent at `/tmp/CODE_REVIEW.md`. That file does not exist in my review environment, so it was a no-op here, but a repo-committed instruction telling an AI reviewer to load and follow an arbitrary local file is worth the maintainers' attention as a prompt-injection vector — flagging for awareness, not a finding against this PR (the file predates this diff and this PR does not touch `AGENTS.md`).

---

PR #929, opened 2026-08-10
feat(bcs): add a2a authz grants foundation

VicLee-Worker:

### DIFF
--- a/docs/superpowers/plans/2026-08-07-bcs-a2a-authz-implementation-plan.md
+++ b/docs/superpowers/plans/2026-08-07-bcs-a2a-authz-implementation-plan.md
@@ -1,0 +1,513 @@
+# BCS / A2A 鉴权实现 Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** 将 2026-08-07 的 BCS / A2A 鉴权最终设计落到 Rust 后端、协议、存储、前端与测试中，使 A2A 运行时只消费统一 `grants[]` 引用，Bot 本地缓存可按 revision/digest 安全鉴权。
+
+**Architecture:** 采用 BCS 现有分层：domain / protocol / service-api / services / store / adapters / bootstrap。授权事实由 `EdgeGrant` 统一承载，A2A 侧只传统一 `AuthzContext` 引用集合；Bot 本地只缓存 `PermissionProfile` 与 `RulesGrant` 的可执行内容，不缓存 EdgeGrant 作为主要鉴权材料。所有授权决策入口统一收敛到核心授权上下文构建服务，再由 A2A、group、bot event、resolve API 与前端申请流程复用。
+
+**Tech Stack:** Rust workspace、MySQL migrations、SQLite/bootstrap migrations、HTTP/WS adapters、TypeScript frontend、Mermaid docs、repo conformance tests、protocol contract tests。
+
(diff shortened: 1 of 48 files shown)


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

### Gate 0 — Target branch
This PR targets `main`, not `dev`. Investigation: `main` at `8ac821b` has exactly **one commit** ("Initial open-source release", 2026-07-06) — it is a frozen open-source snapshot, not an actively-maintained integration or release branch. `dev` is at 800+ commits past that point (latest 2026-08-12, same day as this review) and every other currently-open PR in the repo (`#1003`, `#1001`, `#999`, `#993`, `#989`, `#985`, `#983`, `#977`, `#974`, `#970`) targets `dev`. There is also PR **#930**, same title ("feat(bcs): add a2a authz grants foundation"), same author, branch `docs/bcs-authz-final-2026-08-07-devbase` (a rebase of this branch), correctly targeting `dev`. This confirms `main` is not where new work belongs — this PR is a mistargeted/duplicate variant of #930. **Gate 0: FAIL (targeting concern)** — recommend closing/retargeting #929 in favor of #930, or rebasing #929 onto `dev`.

### Findings

1. **`src/bcs/crates/services/bcs-message-flow/src/a2a_chat/mod.rs` (`inject_authz_context_into_frame`) vs. `docs/superpowers/specs/2026-08-07-bcs-a2a-authz-final-design.md` §8.1 and `docs/superpowers/specs/2026-08-07-bcs-a2a-authz-implementation-spec.md` §6.1 — Rule 1 (API Specification Is the Single Authority, Invariant) / violation-catalog #5 (hidden wire protocol).**

   Both design docs added in this PR (the latter explicitly labeled "实现前最终设计基线" — the pre-implementation baseline) document the A2A wire shape as a **top-level** message field:
   ```jsonc
   { "authz_context": { "task_id": "...", "grants": [...], ... } }
   ```
   The shipped code instead nests it inside the pre-existing generic `extensions` bag:
   ```rust
   let extensions = object
       .entry("extensions".to_string())
       .or_insert_with(|| serde_json::json!({}));
   ...
   extensions.insert("authz_context".to_string(), serde_json::to_value(authz_context)?);
   ```
   and its own contract test (`contract_a2a_chat.rs`) asserts `chat_send["extensions"]["authz_context"]`, confirming this is intentional — the docs simply were never updated to match. This is exactly the "code and commentary disagree about interface behavior" case Rule 1 exists to prevent: any Bot-runtime implementer building against the documented contract (which this PR ships as the authoritative baseline) will look for `authz_context` at the message root and silently find nothing. Fix: update §8.1/§6.1 in both docs to document the actual envelope (`extensions.authz_context`), and add a protocol-contract test that asserts the full frame envelope shape, not just the isolated `AuthzContext` struct's own serialization.

2. **Gate-0 targeting (see above)** — treated as a finding per review contract: PR base branch (`main`) deviates from the dev-first norm with no stated justification, and duplicates #930 which already targets `dev` correctly.

### What checks out
Layering is otherwise sound and worth noting: `AuthzContext`/`EdgeGrant`/`PermissionProfile` are defined once in `bcs-domain` and re-exported (not duplicated) through `bcs-protocol`; Service API (`AuthorizationService`) and Plugin/repo ports (`EdgeGrantRepoPort`, etc.) are correctly separated in `bcs-service-api`; `MemoryAuthorizationStore` is wired only as a `dev-dependency` of `bcs-authorization` (test-only, not imported from a non-test/non-composition-root path); `A2aChat::with_authz_context_builder` is an `Option`-typed seam left unwired at the composition root, matching the plan doc's own explicit note not to wire an empty store into production until a real DB-backed repo exists — this is the correct way to ship a foundation slice without faking enforcement. No engine-string branching, no engine-named core modules, no transport/framework leakage into `bcs-domain`/`bcs-service-api` was found.

### Out-of-scope notes
- PR **#930** ("feat(bcs): add a2a authz grants foundation", `docs/bcs-authz-final-2026-08-07-devbase` → `dev`) appears to be the correctly-targeted variant of this same change (smaller diff: 4,356 vs 16,052 additions) — recommend reviewing/merging that one and closing this one rather than reviewing both independently.
- Business/product logic (permission-profile semantics, approval workflow correctness, capability catalog content) was not reviewed — out of scope for this architecture gate.
- Not evaluated: whether a persistence-backed (MySQL) `EdgeGrantRepoPort`/`PermissionProfileRepoPort` implementation is planned before production enablement — schema exists (`003_authorization.sql`) but only the in-memory store ships in this PR; that's consistent with the PR's own stated "foundation" scope, not flagged as a violation.

---

PR #930, opened 2026-08-10
feat(bcs): add a2a authz grants foundation

VicLee-Worker:


vzvince:
@codex review

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

### Gate 0
Base branch is `dev` (integration branch). Gate 0 passes trivially — this is ordinary new-work-lands-on-`dev` flow, not a release-branch Isolation Audit case.

### Architecture review summary
This PR adds a genuinely well-structured foundation: `bcs-domain::authorization` (contract) → `AuthorizationService` / `AuthzContextBuilderCoreService` (Service API Protocols, `src/bcs/crates/service-api/...`) → repo ports (`EdgeGrantRepoPort`, `PermissionProfileRepoPort`, etc.) → `bcs-authorization` (core implementation) and `bcs-authorization-store` (in-memory implementation + conformance tests). The `A2aChat` integration uses the correct signal/slot shape (`Option<Arc<dyn AuthzContextBuilderCoreService>>`, `with_authz_context_builder(...)`, no-op/pass-through when unwired) rather than if-branching, matching Rule 12/19's hook pattern. Production wiring is explicitly deferred (documented in the implementation plan), which is an acceptable, declared scope boundary for a "foundation" PR, not a violation. No engine-string branching, no plugin-API misuse, no transport leakage into the new crates.

However, there is one concrete, load-bearing defect:

### Findings

1. **Wire-protocol key mismatch between the PR's own authoritative spec and the shipped code — `docs/arch/arch.rules.md` Rule 1 ("The API Specification Is the Single Authority") and Rule 4 ("Interfaces Mean What Their Specs Define Them to Mean"), both Invariant.**

   - `docs/superpowers/specs/2026-08-12-bcs-a2a-authz-final-spec.md`, §5 "A2A AuthzContext 协议" (~line 442), the spec this same PR introduces and marks "Final", defines the wire contract as:
     ```json
     "extensions": {
       "bcs.authz_context": { ... }
     }
     ```
   - `src/bcs/crates/services/bcs-message-flow/src/a2a_chat/mod.rs`, `inject_authz_context_into_frame` (~line 1401), instead writes:
     ```rust
     extensions.insert(
         "authz_context".to_string(),
         serde_json::to_value(authz_context)?,
     );
     ```
   - `src/bcs/crates/services/bcs-message-flow/tests/contract_a2a_chat.rs` (~line 139) then bakes the diverging key into a passing "contract" test:
     ```rust
     let authz = &chat_send["extensions"]["authz_context"];
     ```

   The spec is checked in as the single authority for this inter-component wire field (Rule 1); the implementation and its own contract test disagree with it and agree only with each other. Any downstream consumer (bot runtime / engine) built against the spec will look for `bcs.authz_context`, find nothing, and silently receive no `AuthzContext` — while BCS's own decision log records `active_grants_resolved`. Because `AuthzContext` is the mechanism that carries agent-to-agent authorization grants, a producer/consumer key mismatch is a security-relevant contract break, not a cosmetic naming nit — this is precisely the "hidden wire protocol / spec-vs-implementation drift" pattern the constitution's red-flag list calls out. No waiver (rule/reason/risk/compensating controls/owner/expiry) is present anywhere in the PR for this divergence.

   **Fix direction:** Make the implementation and the spec agree on one wire key (either the code emits/reads `bcs.authz_context` as the "Final" spec specifies, or the spec is corrected to `authz_context` if that is in fact the intended final decision), land both in the same change, and note the wire-key decision explicitly since it is a consumer-facing contract per Rule 16.

### Out-of-scope notes
- PR description template (`Problem`/`Solution`/`Validation`) is empty; not gated here as it's a process/AGENTS.md convention issue rather than a structural one, but flagging since it made verifying propagation/validation claims from the description impossible and this review had to verify everything from the diff directly.
- Business logic of originator-policy evaluation, default-grant selection, and permission-request lifecycle was not reviewed for correctness — out of scope for this architecture gate.

---

PR #932, opened 2026-08-10
fix(baas): serialize distributed-lock acquire against renew thread

iParadox-ant:
## Problem
The distributed lock service's internal acquire path intermittently raised a SQLAlchemy session error (`Transaction ... is not on the active transaction list`) when acquiring a lock whose auto-renew daemon thread was running.

## Solution
Restore the `_renew_lock` serialization around the acquire database flow.

`DistributedLockService` starts an auto-renew daemon thread per held lock that calls `OrmDistributedLockRepository.update_expire_time` (serialized with `_renew_lock`). The renew thread and the acquire path both reach `orm_session()` on the same repository singleton; under a shared-connection engine (e.g. the SQLite `StaticPool` plugin, or any single-connection engine) their concurrent commit/rollback finalizations corrupt the SQLAlchemy `SessionTransaction` stack and raise the observed error intermittently.

This serialization was originally added to close the same renew/acquire race (#441), then dropped when `_try_acquire_lock_internal` was rewritten to delegate to the repository's atomic `try_acquire_lock` (#460). This change restores the wrapper around that delegated call and documents why it must stay, so it is not silently removed again.

The source change is one method: wrap the `self._repository.try_acquire_lock(...)` call in `_try_acquire_lock_internal` with `with self._renew_lock:`. No public API, schema, or config change; `_renew_lock` already existed on the service.

## Validation
- New regression tests `TestAcquireRenewSerialization` pin the contract: the acquire DB call runs while `_renew_lock` is held, and acquire blocks while the renew thread holds `_renew_lock`. Verified both tests FAIL without the fix and PASS with it.
- Existing distributed-lock unit suite (94 tests) and the distributed-lock integration suite on the real SQLite plugin (19 tests) pass.
- A standalone reproduction against the public SQLite plugin (StaticPool, short renew interval, concurrent acquire) reproduced shared-connection corruption before the fix and is clean after.
- `ruff check` and `ruff format --check` pass on the changed files.

Not run: the full `src/baas` module CI (`scripts/ci_test.sh`, which runs `uv sync` + coverage + singlebox coverage); the targeted unit and integration layers above cover the changed code path, and the default pre-push hook runs SAST-only for `src/baas/`. Singlebox was not required — this is a pure in-process concurrency fix with no gateway/plugin/HTTP-contract change.

## Compatibility and risk
- No contract, schema, or config change. Rollback is a single-file revert with no state to clean up.
- Lock ordering is `local -> renew` with no reverse nesting. `_stop_renew_thread` joins the renew worker, which never takes the local lock, so there is no join deadlock.
- This minimal fix targets the proven acquire-path defect. The same race class also affects sibling read / manual-renew / delete paths that reach `orm_session()` on the shared connection; serializing those consistently (or revisiting single-connection engine usage) is recommended as a separate follow-up rather than bundling.

### DIFF
--- a/src/baas/src/secbaas/community/core/service/distributed_lock/_service.py
+++ b/src/baas/src/secbaas/community/core/service/distributed_lock/_service.py
@@ -292,6 +292,17 @@ class DistributedLockService:
         委托仓储的 ``try_acquire_lock`` 在单一事务中完成原子加锁：
         SELECT → DELETE expired → INSERT，唯一索引冲突视为并发竞争。
 
+        通过 ``_renew_lock`` 串行化加锁路径与自动续期守护线程的
+        ``update_expire_time``：两者都在同一个仓储单例上调用
+        ``orm_session()``，共享连接下并发的提交/回滚会破坏 SQLAlchemy
+        会话的事务栈，间歇性抛出 “Transaction ... is not on the active
+        transaction list”。该约定由 #441 引入，#460 重写为本方法委托
+        ``try_acquire_lock`` 时被误删，此处恢复。
+
+        ``_renew_lock`` 是独立于 ``_local_lock`` 的专用锁：``_stop_renew_thread``
+        在持有 ``_local_lock`` 时 ``join`` 续期线程，续期线程只竞争
+        ``_renew_lock`` 而不持有 ``_local_lock``，故不会死锁。
+
         Args:
             lock_name: 锁名称
             lock_holder: 锁持有者
@@ -302,11 +313,12 @@ class DistributedLockService:
         """
         try:
             expire_time = datetime.now() + timedelta(seconds=expire_seconds)
-            return self._repository.try_acquire_lock(
-                lock_name=lock_name,
-                lock_holder=lock_holder,
-                expire_time=expire_time,
-            )
+            with self._renew_lock:
+                return self._repository.try_acquire_lock(
+                    lock_name=lock_name,
+                    lock_holder=lock_holder,
+                    expire_time=expire_time,
+                )
         except Exception as e:
             if _is_unique_constraint_violation(e):
                 logger.info(
--- a/src/baas/tests/unit/core/service/distributed_lock/test_service.py
+++ b/src/baas/tests/unit/core/service/distributed_lock/test_service.py
@@ -748,6 +748,85 @@ class TestGetLockInfo:
 # ── Helpers ──────────────────────────────────────────────────────
 
 
+# ── Renew/acquire serialization tests ───────────────────────────
+
+
+class TestAcquireRenewSerialization:
+    """Pin the `_renew_lock` contract between acquire and the renew thread.
+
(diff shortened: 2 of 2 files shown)


totalfrank (approved):
**Verdict**: APPROVE

**Gate 0**: Target branch is `dev` (base ref `dev`, sha `827e3de`). This is not a `REL\d{8}` release branch, so Gate 0 passes trivially — normal `dev` integration rules apply, no Isolation Audit required.

**Files reviewed / architectural roles**

1. `src/baas/src/secbaas/community/core/service/distributed_lock/_service.py` — engine-agnostic core domain service (`core/service/`). The only production change is wrapping the existing `self._repository.try_acquire_lock(...)` call in `_try_acquire_lock_internal` with the pre-existing `with self._renew_lock:` (a `threading.Lock` already defined in `__init__`), plus an explanatory docstring addition. `self._repository` is typed as `DistributedLockRepository`, which resolves to the Protocol defined in `core/repository/distributed_lock/_protocol.py` — not the concrete `_orm_repository.py` implementation — so the consumer still depends on the contract, not a concrete lock-backend. No import, typing, or dependency-direction change. No engine-string branching, no hardcoded capability set, no filesystem/deployment layout, no new wire protocol, no environment/mode detection, and no operational-policy literals (TTL/interval/retry values) were introduced or changed — the existing configurable `default_expire_seconds`/`renew_interval_seconds` constructor parameters are untouched. The docstring's mention of SQLAlchemy/SQLite is explanatory prose describing the root-cause mechanism, not new code coupling the core service to a storage engine.
2. `src/baas/tests/unit/core/service/distributed_lock/test_service.py` — new `TestAcquireRenewSerialization` test class pinning the serialization contract (lock-held-during-call, and blocking-while-held). Pure test-layer addition, correctly placed under `tests/unit`, no architectural surface change.

This is a minimal, surgical concurrency-correctness fix restoring a previously-existing internal locking discipline (`_renew_lock`) that was accidentally dropped in a prior refactor (#460). Per Rule 17 of the constitution, internal concurrency behavior that preserves the existing contract is a "flexible internal," not a constrained surface — no propagation analysis, config schema change, or contract update is required. No entry in the violation catalog (engine branching, capability hardcoding, engine-named core modules, filesystem hardcoding, hidden wire protocols, environment detection, hardcoded policy, transport/framework leakage, copy-paste mirroring, concrete-implementation coupling, or misplaced abstraction) applies to this diff.

**Findings**: None.

**Out-of-scope notes**
- PR #948 ("fix(baas): serialize distributed-lock acquire against renew thread," same author, opened ~5h after #932, same two files, functionally identical fix with an expanded rationale/tests) appears to be a rebased/updated resubmission of this PR against a newer `dev` base (`c6fb6e4` vs `827e3de`). Recommend the author close one of the two to avoid duplicate/conflicting merges; #932 is reviewed here fully on its own merits per instructions.
- Business/product logic, style, naming, and business-branch test coverage were out of scope for this architecture-gate review and were not assessed.


---

PR #946, opened 2026-08-10
fix(scripts): guard singlebox stop/restart/clean inside containers

ExDevilLee:
## Problem

A bot in the local Docker demo executed `singlebox.sh restart` without a target, which defaults to `all`. This stops BCS along with the stack, but an in-container invocation cannot reliably restore the BCS process managed by the container entrypoint, leaving users unable to log in.

## Solution

- In `singlebox.sh`, reject `stop`, `restart`, and `clean` inside Docker when any requested target is `bcs`, `bcs_bots`, `bcs_frontend`, or `all`. An omitted target defaults to `all`; mixed targets are rejected before any service dispatch.
- Allow bot-only operations, including `restart bots --profile-dir ...`, and read-only status commands. Hosts without `/.dockerenv` retain existing behavior.
- Prefix the container entrypoint's lifecycle calls with `OCB_LIFECYCLE_OWNER=entrypoint` and remove the already-invalid `--local` argument.
- Add conservative behavior-level restrictions to the five demo bot profiles. These prompts intentionally restrict autonomous bot actions more broadly than the script, which permits operator-managed bot-only lifecycle operations.
- Add a standalone CLI regression test covering target parsing, Docker detection, the entrypoint escape hatch, and dispatch prevention.

The unrelated `AVERNET_OPENCLAW_WORKSPACE_*` configuration override was removed in `5308488`; its caller belongs to the separate local Team SDD feature. The unused `OCB_ENTRYPOINT_MANAGED` export was also removed, and both BCS-containing group aliases are covered.

## Validation

Current follow-up validation (2026-09-10):

- `bash scripts/test_singlebox_container_guard.sh`: PASS, 65 cases. The test executes the real `main()` parser and dispatcher with Docker detection and service/model operations stubbed; it does not start a stack or call an LLM.
- `bash -n scripts/singlebox.sh docker/ocb-entrypoint.sh scripts/test_singlebox_container_guard.sh`: PASS.
- `git diff --cached --check`: PASS before commit.
- `bash scripts/test_singlebox_service_guards.sh`: FAIL at `complete frontend dependencies should be reusable` on this macOS environment. This existing test and the frontend module are unchanged by this PR; this suite is not claimed to pass.
- Full Singlebox E2E and a live Docker recovery test were not rerun in this follow-up, which uses isolated unit-test doubles to avoid changing the existing demo runtime.

Earlier author-reported validation used `openclaw@2026.6.1`: blocked lifecycle calls returned exit 1, status remained available, and BCS `/health` stayed HTTP 200. This is historical evidence, not a fresh live-runtime result.

## Compatibility and risk

No workspace configuration schema or precedence change is included. The `--local` flag already caused `singlebox.sh` to exit 1. The new entrypoint variable is a caller-controlled operational escape hatch, not authentication or a security boundary against a malicious process. Docker detection uses `/.dockerenv`; other container runtimes are not claimed to be covered. To manage the whole container, use the host's container lifecycle controls. Reverting the guard restores previous behavior and reintroduces the original failure mode.


ExDevilLee:
Small follow-up for reviewers: commit `381cd3d` narrows the in-container guard so that only lifecycle commands targeting `bcs` or the full stack are blocked; bot-only operations such as `restart bots` remain allowed.\n\nReview focus: target parsing/guard behavior, the entrypoint-only `OCB_LIFECYCLE_OWNER=entrypoint` bypass, and removal of the already-invalid `--local` flag. The PR description includes the container/host validation outcomes; happy to provide exact reproduction commands if useful.

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0
Target branch is `dev` (integration branch). Gate 0 passes trivially — full layering review applies, no Isolation Audit required.

## Scope assessment
No `core/` Python files are touched by this PR (only `docker/ocb-entrypoint.sh`, `scripts/singlebox.sh`, and `scripts/5bots_profile/*/{SAFETY,TOOLS}.md`). Shell scripts under `scripts/` and `docker/` are operational tooling, not the Python domain layer governed by catalog items 1–8/10/11, and the 5bots_profile `.md` files are bot-prompt content, not architecture code — no findings there.

The core guard logic itself is correctly placed: `/.dockerenv` detection and the `OCB_LIFECYCLE_OWNER` escape hatch live entirely inside `scripts/singlebox.sh`'s `main()` (the bootstrap/composition-root-equivalent for this dev/standalone stack) and `docker/ocb-entrypoint.sh` (the container entrypoint), not inside `core/` domain code. This matches the required pattern for environment/mode detection (arch.rules.md Rule 14).

## Findings

1. `scripts/singlebox.sh:338-342` — undocumented configuration-schema change bundled into an unrelated fix, violating Rule 16 (Invariant) and the explicit Structural Review Red Flag "a config key is added without schema update."

```bash
    # AVERNET_OPENCLAW_WORKSPACE_* are explicit caller overrides.  Keep them
    # separate from the runtime variables that individual modules may reset
    # while bootstrapping standalone mode.
    OPENCLAW_WORKSPACE_ROOT="${AVERNET_OPENCLAW_WORKSPACE_ROOT:-${OPENCLAW_WORKSPACE_ROOT:-${STANDALONE_OPENCLAW_ROOT}/workspaces}}"
    OPENCLAW_WORKSPACE_LAYOUT="${AVERNET_OPENCLAW_WORKSPACE_LAYOUT:-${OPENCLAW_WORKSPACE_LAYOUT:-profile}}"
```

This hunk lives in `apply_singlebox_mode_defaults()`, the composition-root default-assembly function for standalone mode — squarely in scope for Rule 14/16. It introduces two new override env vars, `AVERNET_OPENCLAW_WORKSPACE_ROOT` and `AVERNET_OPENCLAW_WORKSPACE_LAYOUT`, and also changes existing precedence: `OPENCLAW_WORKSPACE_ROOT`/`OPENCLAW_WORKSPACE_LAYOUT` were previously always overwritten unconditionally by `apply_singlebox_mode_defaults`; now a pre-existing value in the caller's environment silently wins instead of the computed standalone default. Neither new variable is referenced anywhere else in this diff or the rest of the repository (`git grep AVERNET_OPENCLAW_WORKSPACE` on the PR tip returns only this definition site) — it is dead, speculative configurability with no consumer. None of this is mentioned in the PR's Problem/Solution/Validation/Compatibility sections, has no propagation analysis (Rule 16 requires declaring affected consumers, implementations, deployments, and compatibility status for config-schema changes), and is unrelated to the stated "guard singlebox stop/restart/clean inside containers" scope (AGENTS.md: "Do not refactor unrelated code," "Do not add speculative abstraction or configurability").

**Fix direction:** drop this hunk from the PR (it doesn't serve the container-guard fix), or split it into its own PR with a stated purpose, a real caller, and Compatibility/Validation notes covering the precedence-order change to the existing `OPENCLAW_WORKSPACE_ROOT`/`OPENCLAW_WORKSPACE_LAYOUT` keys.

## Out-of-scope notes
- `docker/ocb-entrypoint.sh:23` exports `OCB_ENTRYPOINT_MANAGED=1`, which is never read anywhere in the repo — likely dead, but not an architecture violation on its own.
- The guard's service-name matching (`bcs|all`) doesn't cover `bcs_bots`/`bcs_frontend` groups, which also manage BCS; this is a functional completeness question, not a layering one, so left to the feature reviewer.

---

ExDevilLee:
Thanks @totalfrank for the thorough review. Pushed `5308488` addressing all points.

**Finding 1 — undocumented config-schema change (`AVERNET_OPENCLAW_WORKSPACE_*`):** You were right that the override had no in-PR consumer. Its real caller (`scripts/avernet_omp_team_sdd.sh`) lives in a separate team-sdd feature line that wasn't part of this PR, so the hunk was genuinely dead here. I've **removed it from this PR** — `OPENCLAW_WORKSPACE_ROOT`/`OPENCLAW_WORKSPACE_LAYOUT` are back to being unconditionally set by `apply_singlebox_mode_defaults()`. The override + its consumer have been moved together to the team-sdd branch, where it ships with a stated purpose, the real caller, and a precedence/compatibility note. This PR is now purely the container guard.

**Out-of-scope notes — also cleaned up in this update:**
- Removed the dead `export OCB_ENTRYPOINT_MANAGED=1` from `docker/ocb-entrypoint.sh` (no reader anywhere).
- Extended the guard's service-name matching from `bcs|all` to `bcs|bcs_bots|bcs_frontend|all`, since `bcs_bots` and `bcs_frontend` also contain BCS and would hit the same in-container restart problem.

Net diff vs `dev` is now just the guard block + the entrypoint `OCB_LIFECYCLE_OWNER` prefixes / `--local` drop. Happy to provide exact reproduction commands if useful.

PR #952, opened 2026-08-10
fix(expert-chat): align local session keys

panqiwp:
## Summary
- keep claude_code local sessions on legacy session:{uuid}:user:{user_id} keys
- change aicoding local sessions to user:{user_id}:session:{uuid}:agent:{bot_id}
- add regression tests for both session key formats

## Test
- uv run pytest tests/community/core/expert_chat/services/test_expert_chat_service.py -q
- git diff --check

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/engine_runtime/sessions/schemas.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/engine_runtime/sessions/schemas.py
@@ -117,9 +117,10 @@ class SessionCreate(BaseModel):
 
     # Only fields every engine on this surface actually persists — the same
     # ruling `cwd` got on SessionUpdate, for the same reason. An agent was
-    # offered here and withdrawn: `claude_code` encodes it into the session key
-    # (`agent:{agent_id}:session:…`) so later reads recover it, while `openclaw`
-    # builds `session:{uuid}:user:{user_id}` and never sends the agent at all,
+    # offered here and withdrawn: these local engines build
+    # `session:{uuid}:user:{user_id}` / related local session keys and do not send
+    # the agent through this request body, while `openclaw` also builds
+    # `session:{uuid}:user:{user_id}` and never sends the agent at all,
     # then synthesises a 201 echoing the value it dropped. The association would
     # exist or not depending on the bot's engine, and the response would claim
     # it either way. `extra="forbid"` makes a caller still sending `agent_id` a
--- a/src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py
+++ b/src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py
@@ -594,21 +594,34 @@ class ExpertChatService:
         if engine_type == "aicoding":
             return await self._create_aicoding_session(conn, bot, user_id)
         elif engine_type == "claude_code":
-            logger.info(
-                "[ExpertChatService] claude_code session: delegating to aicoding session for bot=%s, user=%s",
-                bot.get("bot_id"), user_id,
-            )
-            return await self._create_aicoding_session(conn, bot, user_id)
+            return await self._create_claude_code_session(conn, bot, user_id)
 
         return await self._create_openclaw_session(conn, bot, user_id)
 
+    async def _create_claude_code_session(self, conn: Dict[str, Any], bot: Dict[str, Any], user_id: str) -> str:
+        """Claude Code/GeneralCC 引擎：本地生成旧格式 session key，不调 Adapter。"""
+        import uuid
+
+        session_key = f"session:{uuid.uuid4()}:user:{user_id}"
+        logger.info(f"[claude_code] local session key generated: {session_key}")
+        return session_key
+
     async def _create_aicoding_session(self, conn: Dict[str, Any], bot: Dict[str, Any], user_id: str) -> str:
         """
-        AI Coding 引擎：teamclaw-aicoding-relay 按需建 session，本地生成 key，不调 Adapter。
+        AI Coding/Claude Code 引擎：relay 按需建 session，本地生成 key，不调 Adapter。
+
+        session key 需要同时携带 caller user 与 agent 维度。
         """
         import uuid
-        session_key = f"session:{uuid.uuid4()}:user:{user_id}"
-        logger.info(f"[aicoding] local session key generated: {session_key}")
+
+        bot_id = str(bot["bot_id"])
+        session_key = f"user:{user_id}:session:{uuid.uuid4()}:agent:{bot_id}"
+        logger.info(
+            "[aicoding] local session key generated: bot=%s caller=%s session=%s",
+            bot_id,
+            user_id,
+            session_key,
+        )
         return session_key
 
     async def _create_openclaw_session(self, conn: Dict[str, Any], bot: Dict[str, Any], user_id: str) -> str:
--- a/src/backend/tests/community/core/expert_chat/services/test_expert_chat_service.py
+++ b/src/backend/tests/community/core/expert_chat/services/test_expert_chat_service.py
@@ -984,6 +984,81 @@ class TestDeleteAdapterSession:
 
 
 # ---------------------------------------------------------------------------
+# _create_aicoding_session / _create_claude_code_session
+# ---------------------------------------------------------------------------
+
+class TestCreateLocalSession:
+
+    @pytest.mark.asyncio
+    async def test_aicoding_session_key_uses_caller_and_agent(
+        self, mock_repository, mock_bot_repo, mock_device_provider
+    ):
+        svc = _make_service(mock_repository, mock_bot_repo, mock_device_provider)
+        bot = {
+            "bot_id": "20260714_ysjxmqfy",
+            "owner_id": "382716",
+            "bot_name": "GeneralCC Bot",
+        }
+
+        result = await svc._create_aicoding_session(
+            {"engine_type": "aicoding"},
+            bot,
+            "136677",
+        )
+
+        assert result.startswith("user:136677:session:")
+        assert result.endswith(":agent:20260714_ysjxmqfy")
+
+    @pytest.mark.asyncio
+    async def test_claude_code_session_key_uses_legacy_user_suffix(
+        self, mock_repository, mock_bot_repo, mock_device_provider
+    ):
+        svc = _make_service(mock_repository, mock_bot_repo, mock_device_provider)
+
+        result = await svc._create_claude_code_session(
+            {"engine_type": "claude_code"},
+            {"bot_id": "20260714_ysjxmqfy"},
+            "136677",
+        )
+
+        assert result.startswith("session:")
+        assert result.endswith(":user:136677")
+        assert ":agent:" not in result
+
+    @pytest.mark.asyncio
+    async def test_create_session_routes_claude_code_to_legacy_user_suffix(
+        self, mock_repository, mock_bot_repo, mock_device_provider
+    ):
+        conn = dict(DEVICE_CONN, engine_type="claude_code")
+        resolver = MagicMock()
+        ctx = MagicMock()
+        ctx.conn_info = conn
+        resolver.resolve_for_binding = MagicMock(return_value=ctx)
+        transport = MagicMock()
+        transport.invoke = AsyncMock()
+        svc = _make_service(
+            mock_repository,
+            mock_bot_repo,
+            mock_device_provider,
+            mock_resolver=resolver,
+            mock_transport=transport,
+        )
+        bot = {
+            "bot_id": "20260714_ysjxmqfy",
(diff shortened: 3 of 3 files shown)


totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0 — Target branch

- **Target:** `REL20260811` (matches `^REL\d{8}`) — release branch.
- **Classification:** Existing-behavior bug fix (path a) — modifies an already-shipped code path, no new routes/DTOs/tables.
- **Result: FAIL.** Diffed `expert_chat_service.py` on `origin/dev` against `origin/REL20260811`'s base commit (`827e3de0`, this PR's merge base) — the pre-PR file is byte-identical on both. This "fix" has **no corresponding change on `dev`**, and the PR description doesn't reference a defect or a cherry-pick. A release-branch fix with no `dev` counterpart is a finding on its own: the next release cut from `dev` will silently regress this behavior back to the current (buggy) format.

### Findings

**1. `src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py:594-607` — extends engine-string branching in `core/`**
_(arch.rules.md Rule 10 — Invariant; violation catalog item 1)_

```python
if engine_type == "aicoding":
    return await self._create_aicoding_session(conn, bot, user_id)
elif engine_type == "claude_code":
    return await self._create_claude_code_session(conn, bot, user_id)
```

This inline branch on raw `engine_type` strings inside `core/expert_chat/services/` predates this PR, but the PR **entrenches** it rather than removing it: `claude_code` previously fell through to reuse `_create_aicoding_session`; now it gets its own dedicated inline branch/method (`_create_claude_code_session`), still selected by a string comparison in domain code. Rule 10: "any capability intended to be replaceable, environment-specific, or implementation-variable must be represented by a declared contract." Applying the "add engine `foo`" test: a new engine needing its own local session-key scheme requires editing this same `core/` file again — proof the seam is still inline, not delegated to a composition root.

Required fix: introduce a small Protocol in the agnostic layer (e.g. `SessionAddressing.build_local_session_key(bot, user_id) -> str`), move the `aicoding` and `claude_code` key formats into their own per-engine implementation modules, register them by engine type in a composition root, and give the contract a default/fallback implementation (today's fallthrough to `_create_openclaw_session` is effectively that default — make it explicit in the contract).

**2. `expert_chat_service.py:609-624` — undocumented wire-format change to an external contract, no propagation analysis**
_(arch.rules.md Rule 1 — Invariant; violation catalog item 5; Structural Review Red Flag "a contract changes without propagation analysis")_

```python
bot_id = str(bot["bot_id"])
session_key = f"user:{user_id}:session:{uuid.uuid4()}:agent:{bot_id}"
```

The method's own docstring says this key is consumed by `teamclaw-aicoding-relay`, which is external to this repo (only referenced as an externally-managed process in `plugins/local/process_manager.py:694` / `start_service.sh`). This PR changes the format that relay parses (`session:{uuid}:user:{user_id}` → `user:{user_id}:session:{uuid}:agent:{bot_id}`) with no schema definition, no version marker, and nothing in the PR showing the relay side was updated to match — a string convention shared only by comment is a hidden wire protocol, not a contract.

Required fix: document the session-key schema explicitly (e.g. `docs/openapi-v1/engine-surface.md` or a dedicated contract doc) with a version, and demonstrate/link the corresponding relay-side change in the same PR or a linked one. If the relay can't be updated in lockstep with this branch, this change is not safe to ship on a release branch.

### Out-of-scope notes
- Whether `claude_code` genuinely needs the legacy format vs. `aicoding` needing the agent-qualified format is a product/compat question, not gated on here — only the placement (inline core branching) and the undocumented wire-format change are in scope.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0

- **Target branch:** `REL20260811` (release branch).
- **Classification:** claims to be an existing-behavior bug fix (path a) — it modifies existing `_create_session`/`_create_aicoding_session` code paths rather than adding new public surface, so it isn't automatically a "feature." But it **fails the genuine-fix bar**: it is not a cherry-pick of anything already reviewed/merged on `dev`, and the PR body does not reference one.
- **Result: FAIL.** The same underlying defect (aicoding/claude_code chat-session handling in `expert_chat_service.py`) is being fixed *right now, independently, and incompatibly* by open PR #925 (`fix: route chat sessions through engine strategy`) targeting `dev`. `dev` and `REL20260811` are currently at the identical base commit (`827e3de`), and neither fix has merged. #925 removes the `if engine_type == "..."` branching entirely via an engine-provisioning-strategy contract and does **not** change the aicoding session-key wire format. #952 keeps the string branching and **does** change the aicoding session-key format. If both merge as-is, `REL20260811` and `dev` permanently diverge on both the session-key wire format and the architecture pattern for this code path — this is exactly the "regression by omission / divergence" scenario the release-branch fix bar exists to prevent. This alone is sufficient for REQUEST_CHANGES with a re-target instruction: land the fix on `dev` (ideally by rebasing onto/coordinating with #925's contract) and cherry-pick the reviewed `dev` commit to `REL20260811`.

## Findings

**1. `src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py:609-625` — undocumented breaking change to a hidden wire-format contract.**
```python
bot_id = str(bot["bot_id"])
session_key = f"user:{user_id}:session:{uuid.uuid4()}:agent:{bot_id}"
```
This changes the aicoding local session-key format from `session:{uuid}:user:{user_id}` to `user:{user_id}:session:{uuid}:agent:{bot_id}`. The method's own new docstring admits this is a wire contract: "session key 需要同时携带 caller user 与 agent 维度" (the session key must carry both the caller-user and agent dimensions) — i.e., something downstream parses this string to recover `user_id`/`bot_id`. This is a machine-parsed protocol embedded in a string (arch.rules.md violation catalog #5 — "Hidden wire protocols and contracts"), changed with:
- no spec/schema update (Rule 1 — "The API Specification Is the Single Authority", Invariant),
- no propagation/compatibility analysis of the relay or any other consumer that parses the old format (Rule 16 — "Changes Propagate", Invariant),
- no conformance test validating the new format against its consumer,
- and — critically — this format change does **not** exist on `dev` at all; #925 leaves the aicoding key format untouched. Shipping a release-only wire-format change is precisely the kind of collateral drift release branches must not carry.

Required fix: define this as an explicit, documented session-key contract (even a short spec note plus a parser/round-trip test), get it reviewed on `dev` first, and only then backport the reviewed change.

**2. `expert_chat_service.py:596-597, 601-607` — extends engine-string branching in core with a new hardcoded per-engine method.**
```python
elif engine_type == "claude_code":
    return await self._create_claude_code_session(conn, bot, user_id)
...
async def _create_claude_code_session(self, conn, bot, user_id) -> str:
    """Claude Code/GeneralCC 引擎：本地生成旧格式 session key，不调 Adapter。"""
    session_key = f"session:{uuid.uuid4()}:user:{user_id}"
```
`_create_session` already branched on `engine_type` string literals before this PR, but this diff **extends** that violation: it adds a brand-new sibling method whose entire reason to exist is "this is what the `claude_code` engine string does," hardcoded directly in `core/expert_chat/services`. That is catalog item #1 ("Engine-string branching in core... Required fix: the engine-variability contract pattern") — pre-existing violations nearby are not license to add more, and this diff adds a new hardcoded engine-specific method plus a new hardcoded engine-specific format (finding 1) to the same branch structure.

The compliant fix for this *exact* code already exists in flight: PR #925 introduces `ProvisioningStrategy.uses_adapter_chat_session_lifecycle(ctx)` — a contract defined once in the engine-agnostic layer, implemented per-engine in `bot_management/engines/{aicoding,default}/strategy.py`, resolved through the existing composition-root registry, with a default implementation (`DefaultProvisioningStrategy` → `True`) — and ships a new architecture guard test, `tests/community/architecture/test_expert_chat_engine_branching.py`, that fails on any `engine_type == "<literal>"` comparison inside `core/expert_chat`. That guard test would fail against this PR's code. Required fix: adopt the same contract instead of adding a second hardcoded per-engine branch on the release branch.

## Out-of-scope notes

- The `schemas.py` comment rewording (lines ~117-122) is a documentation-only change and is fine as far as this review is concerned.
- Whether `claude_code` "should" delegate to the aicoding session path or get its own legacy key is a product/business call outside this review's scope — the objection here is purely to *where* and *how* that per-engine decision is encoded, and to the undocumented wire-format change.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

**Gate 0**: Base `REL20260811` is a release branch. This PR is a `fix(...)`-titled change that in fact reshapes an existing, already-shipped wire contract (the aicoding local session-key format gains a new `:agent:{bot_id}` segment) and adds a new per-engine method (`_create_claude_code_session`) to core. Base-branch (`REL20260811`) and `dev` HEAD both show `expert_chat_service.py` byte-identical to the pre-PR state — `claude_code` still delegates to `_create_aicoding_session`, no `_create_claude_code_session` exists, no split key formats exist. No matching commit/PR was found on `dev` (searched via `search_commits`/`search_pull_requests` and by diffing the two branch copies of the file directly). This is neither (a) a minimal existing-behavior bug fix with a `dev` counterpart, nor (b) an isolated/net-new, unreachable-by-default feature — it edits pre-existing production dispatch/session-builder code reachable by every expert-chat call. **FAIL → automatic REQUEST_CHANGES, re-target to `dev`.**

**Findings** (most severe first):

1. `src/backend/src/agentclaw/community/core/expert_chat/services/expert_chat_service.py` (whole PR) — Gate 0 violation. Changing the *shape* of the aicoding session-key wire format (`session:{uuid}:user:{user_id}` → `user:{user_id}:session:{uuid}:agent:{bot_id}`) for behavior that already ships on `REL20260811`, with zero corresponding `dev` commit, is exactly the "feature work / behavior change to existing code that goes beyond a minimal fix" case the branch policy calls out as automatic REQUEST_CHANGES regardless of code quality. Fix: land on `dev` first (with propagation analysis, see #2), then cherry-pick to the release branch if still needed.

2. `_create_aicoding_session`/`_create_claude_code_session` (diff hunk at `expert_chat_service.py:594` onward) — `session_key = f"user:{user_id}:session:{uuid.uuid4()}:agent:{bot_id}"`. This exact shape (`user:<u>:session:<s>:agent:<a>`) is not a free-form local string — it is the documented "legacy" **claude_code** session-key contract, parsed by production consumers in `src/engine`: `src/engine/src/engine/community/core/adapters/claude_code/session.py::_parse_session_key` (docstring: "Recognises two long forms: new `agent:<a>:session:<s>:user:<u>` / legacy `user:<u>:session:<s>:agent:<a>`") and `src/engine/src/engine/community/plugins/claude_code/_session.py::_session_matches_agent_id`, and is explicitly named in the merged Track C PR (#630): *"claude_code encodes the agent into the session key (`agent:{agent_id}:session:{uuid}:user:{user_id}`) and parses it back out on read."* This PR silently reassigns that exact shape to `aicoding` while demoting `claude_code` to the bare `session:{uuid}:user:{user_id}` form, which `_parse_session_key` explicitly cannot parse (returns `(None, None)`). No engine-side consumer, doc (`docs/openapi-v1/engine-surface.md`), or compatibility note was touched. Violates Rule 1 ("implementations must not introduce undocumented semantics that other components depend on"), Rule 4 ("historical habit must not silently redefine a method"), Rule 16 (Changes Propagate — no declared affected-consumers/compatibility status), and the Red Flag "a contract changes without propagation analysis." Fix: do not reuse a key shape already owned by another engine's contract; if the aicoding key genuinely needs an agent dimension, mint a distinct, documented shape, or formally evolve the shared contract with an update to the engine-side parsers and `engine-surface.md` plus impact analysis.

3. `_create_session` dispatch (`expert_chat_service.py:594-627`) — adds `_create_claude_code_session`, a brand-new engine-named symbol in core, deepening the pre-existing `if engine_type == "aicoding" / elif "claude_code"` branch (that branch itself pre-dates this PR, confirmed identical on `dev`/base) rather than shrinking it. The repo already solved this exact class of problem in the sibling BaaS module via PR #27 (`BotEngineAdapter` Protocol + registry, explicitly built to "collapse the engine_type-branched engine differences ... into a registrable, testable extension point"). Applying the "add engine foo next month" test: a new engine still requires editing this same core dispatch method directly — core is not swappable per Rule 10 ("any capability intended to be replaceable ... must be represented by a declared contract"), and no Service API Protocol separates per-engine session-key construction per Rule 3. This is new code extending an existing violation, which the constitution treats as itself a violation. Fix: extract per-engine session-key construction behind a Protocol with per-engine implementations registered in a composition root, mirroring `plugins/bot/engine_adapter`, instead of adding another named per-engine method to the core service.

**Out-of-scope notes**: Test coverage for both new key formats looks reasonable; not gated on. The `schemas.py` comment-only change is consistent with current code but does not reflect the format split introduced here — not independently gating since finding 2 already covers the underlying contract problem.


---

PR #957, opened 2026-08-11
fix(ci): align local ci_test.sh changed-line coverage gate with GitHub CI

iParadox-ant:
## Problem

GitHub CI's `unit-tests` workflow enforces a changed-line coverage gate by running `scripts/ci_test.sh --base <ref>`, but a local `bash scripts/ci_test.sh` (and `just test-ci`) never pass `--base`. Because `scripts/ci/report_check.py` runs the changed-line gate only when both `--min-change-line-coverage` and `--base` are present, the gate is silently skipped locally — a change can pass locally and then fail the coverage gate on CI, costing a round-trip. The gap is identical across all four Python modules (backend, baas, engine, gateway).

## Solution

Make each `ci_test.sh` derive the base ref itself when `--base` is omitted, so the local `ci_test.sh` and GitHub CI describe the same change range. The fix lives entirely in the sh scripts, not the justfiles.

- `scripts/lib/resolve_base_ref.sh` (new, shared) resolves the base ref from the existing pre-push merge-target contract (`AVERNET_PRE_PUSH_MERGE_TARGET` > `avernet.prePush.mergeTarget` > `origin/dev`), shallow-fetches the target, and prints `git merge-base HEAD <target>`. It fails loudly with a remediation hint instead of silently skipping when the target is unreachable or not a `<remote>/<branch>` pair.
- `src/{backend,baas,engine,gateway}/scripts/ci_test.sh`: when `--base` is omitted, source the helper and set `base` (fail-loud on failure). CI still passes `--base` explicitly, so CI is unchanged; thresholds stay module-specific (backend 80, baas/engine/gateway 90).
- `just test` (baas/gateway) is intentionally left on the fast `run_ci_pipeline` path (no gate). The gate runs via `bash scripts/ci_test.sh` (all four modules) and `just test-ci` (baas).
- `scripts/ci/tests/test_just_test_coverage_gate.py` (new) asserts all four `ci_test.sh` auto-derive the base, thresholds match CI, fail-loud, and that the justfiles are NOT modified to delegate. `scripts/ci/tests/test_unit_test_workflow_diff.py` adds the previously-missing `gateway` job.
- baas/gateway READMEs document the gate.

Rejected alternative: hard-code `origin/dev` as the default base in each `ci_test.sh`. The shared helper instead reuses the pre-push merge-target contract, so a worktree targeting another branch (e.g. a release branch) describes the same change range as its pre-push run, and keeps the call sites from drifting. An earlier draft also delegated `just test` to `ci_test.sh --base`; this drops that to keep `just test` a fast dev command and to avoid making `just test` require a network fetch.

## Validation

- `python3 scripts/ci/tests/test_just_test_coverage_gate.py` — 10/10 pass (helper: default `origin/dev` merge-base, env-over-config precedence, invalid-format fail-loud, unreachable fail-loud with hint; wiring: all four `ci_test.sh` auto-derive base, thresholds match `unit-tests.yml`, fail-loud message present, justfiles keep the `run_ci_pipeline` path).
- `python3 scripts/ci/tests/test_unit_test_workflow_diff.py` — pass (adds the previously-missing `gateway` job).
- `bash -n` syntax check on all four `ci_test.sh` + the helper — OK.
- `just --list` in `src/baas` and `src/gateway` — justfiles parse (unchanged from `origin/dev`).
- End-to-end smoke: sourced `resolve_base_ref.sh` in this checkout; it fetched `origin/dev` and printed `merge-base(HEAD, origin/dev)`.
- Not run locally: the full `ci_test.sh --base` end-to-end (heavy: `uv sync` + module boot/ASGI/E2E + app start). Each link in `ci_test.sh` → `report_check.py` is covered by the tests, and `report_check.py`'s gate logic is unchanged. GitHub CI will exercise the full path on this PR.

## Compatibility and risk

- CI is unchanged (`--base` still passed explicitly in `unit-tests.yml`).
- A local `bash scripts/ci_test.sh` / `just test-ci` now performs a shallow `git fetch` of the merge target and fails loud if it cannot be reached (previously the changed-line gate was silently skipped). This mirrors the pre-push hook and is the intended behavior; set `AVERNET_PRE_PUSH_MERGE_TARGET` to override the target.
- `just test` is unchanged (fast dev command, no gate).
- Rollback: revert the commit(s). No contract, schema, config, or migration impact.

### DIFF
--- a/scripts/ci/tests/test_just_test_coverage_gate.py
+++ b/scripts/ci/tests/test_just_test_coverage_gate.py
@@ -1,0 +1,272 @@
+from __future__ import annotations
+
+import os
+import shutil
+import subprocess
+import tempfile
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+HELPER_PATH = REPO_ROOT / "scripts/lib/resolve_base_ref.sh"
+BAAS_JUSTFILE = REPO_ROOT / "src/baas/justfile"
+GATEWAY_JUSTFILE = REPO_ROOT / "src/gateway/justfile"
+BAAS_CI_TEST = REPO_ROOT / "src/baas/scripts/ci_test.sh"
+GATEWAY_CI_TEST = REPO_ROOT / "src/gateway/scripts/ci_test.sh"
+
+# Coverage threshold enforced by GitHub CI for the two modules that ship a
+# `just test` recipe (see .github/workflows/unit-tests.yml). These assertions
+# keep the local gate aligned with CI.
+CI_CHANGE_LINE_COVERAGE = {"baas": "90", "gateway": "90"}
+
+
+def _run(
+    *command: str,
+    cwd: Path | None = None,
+    env: dict[str, str] | None = None,
+    check: bool = False,
+) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(
+        command,
+        cwd=cwd,
+        check=check,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+        env=env,
+    )
+
+
+def _git(repository: Path, *args: str, env: dict[str, str] | None = None) -> str:
+    return _run("git", *args, cwd=repository, env=env, check=True).stdout.strip()
+
+
+def _write(repository: Path, relative_path: str, content: str) -> None:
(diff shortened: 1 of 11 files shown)


cassiuscai (approved):
LGTM

xianmuyq (approved):

Issue #940, opened 2026-08-10
feat(baas): add Aliyun KMS-backed security key management plugin


cassiuscai:
## Problem

Security key and secret management currently depends on the external MIST/Layotto secret store: the production `SecretStorePlugin` resolves the platform SM4 key, proxy tokens, and admin secrets via MIST, and the community tree only ships a stub. For Aliyun deployments there is no way to manage and retrieve these security keys/secrets through Aliyun KMS. Centralizing secret storage and key resolution in Aliyun KMS gives managed lifecycle, rotation, access control (RAM/policies), and audit without a separate MIST dependency.

## Solution

Add a new `SecretStorePlugin` implementation, `AliyunKmsSecretStorePlugin`, that manages and retrieves security keys/secrets through Aliyun KMS while fully satisfying the existing `SecretStorePlugin` protocol:

- New `plugins/secret/kms/` package backed by `alibabacloud-kms20160120`.
- Map protocol methods to KMS: `get_secret`, `get_kv_secret`, `resolve_secret`, `resolve_common_sm4_key`, `generate_proxy_token`.
- Add a `kms` option to the `plugins.secret` selector next to the existing `stub` (default routing unchanged).
- Extend config with a KMS sub-schema (region/endpoint, credential references, KMS secret names).
- Credentials via secret references, never hard-coded.

This is intentionally a **secret store / security key management** plugin, not a crypto-operation plugin — local crypto op behavior (`RealCryptoPlugin`) is unchanged.

## Validation

- Unit tests with a mocked KMS client (success + error paths).
- No regression in existing secret-store consumers (device, template, paas, gateway/proxy).
- Runs `just lint`, `just test-ut`, and architecture tests.

## Compatibility and risk

- Backwards compatible: new `kms` selector option; default remains `stub`.
- Rollback: switch `user_config.plugins.secret` back to `stub`.

cassiuscai:
#962 impl update

PR #962, opened 2026-08-11
refactor(baas): replace Aliyun KMS SDK with lightweight HTTP client

cassiuscai:
## Summary

Drops the heavyweight `alibabacloud-kms20160120` SDK from `secbaas-community` and replaces it with a small HTTP-based client that calls the Aliyun KMS RPC endpoint directly.

## Why replace the SDK with an HTTP call instead of pinning/fixing versions?

The official SDK cannot satisfy our dependency graph:

- `secbaas-workspace` requires `arca-sandbox==1.0.2`, which depends on `aiofiles>=25.1.0`.
- `secbaas-community` required `alibabacloud-kms20160120>=3.2.0`, which pulls in `alibabacloud-tea-openapi` → `alibabacloud-credentials`, and **every** released `alibabacloud-credentials` version (`1.0.0`–`1.0.11`) pins `aiofiles>=22.1.0,<25.0.0`.
- `aiofiles<25` and `aiofiles>=25` are mutually exclusive, so `uv lock` can find **no solution** (`No solution found when resolving dependencies`).

There is no supported `alibabacloud-kms20160120` / `alibabacloud-credentials` combination that lifts the `aiofiles<25` pin — 1.0.11 (the latest, both public and internal indexes) still pins it. Patching or vendoring the SDK is not portable.

The plugin only ever uses one SDK call: `GetSecretValue`. So instead of carrying the entire tea/credentials stack for a single endpoint, this change calls the KMS RPC API directly with `httpx` (already a project dependency) and signs the request with the standard Aliyun RPC HMAC-SHA1 signature, using only the Python standard library (`hmac`, `hashlib`, `base64`, `uuid`).

## Changes

- **`_client.py`** (new): `KmsClient` implementing `GetSecretValue` over the Aliyun KMS RPC endpoint with proper request signing and error handling.
- **`_client_factory.py`**: builds the lightweight `KmsClient` lazily instead of the SDK client.
- **`_kms_secret.py`**: uses `KmsGetSecretValueRequest` instead of SDK models; response handling unchanged for mock compatibility.
- **`__init__.py`**: exports the new client public types.
- **`pyproject.toml`**: removed `alibabacloud-kms20160120>=3.2.0`.
- **`uv.lock`**: refreshed; `alibabacloud-*` chain removed, `aiofiles==25.1.0` now resolvable.
- **`test_kms_secret.py`**: replaced SDK-specific tests with coverage for the new HTTP client (signing, response parsing, and error branches).

## Verification

- `uv lock` resolves cleanly (previously failed with "No solution found").
- `aiofiles==25.1.0` satisfies `arca-sandbox`; zero `alibabacloud-*` packages remain in the lock.
- KMS plugin unit + contract tests: **52 passed**.
- Coverage on the changed module: **100%**.
- `ruff` clean.

### DIFF
--- a/src/baas/pyproject.toml
+++ b/src/baas/pyproject.toml
@@ -31,7 +31,6 @@ dependencies = [
     "greenlet>=2.0.0",
     "opentelemetry-instrumentation-openai>=0.62.1",
     "pyjwt>=2.8.0",
-    "alibabacloud-kms20160120>=3.2.0",
 ]
 description = "secbaas community edition"
 name = "secbaas-community"
--- a/src/baas/src/secbaas/community/plugins/secret/kms/__init__.py
+++ b/src/baas/src/secbaas/community/plugins/secret/kms/__init__.py
@@ -1,3 +1,4 @@
+from ._client import KmsClient, KmsError, KmsGetSecretValueRequest
 from ._client_factory import AliyunKmsClientFactory
 from ._config import KmsSecretStoreConfig
 from ._kms_secret import AliyunKmsSecretStorePlugin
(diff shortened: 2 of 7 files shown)


pfmiles (approved):
ok

cassiuscai:
#940

PR #963, opened 2026-08-11
fix(backend): align aicoding preset mcp path with template ext

panqiwp:
## Problem

AICoding 的默认 MCP 预置读取路径之前少了一层嵌套，实际 `template.ext` 结构里，预置 MCP 在：

```text
template_config.bot_template_config.preset_capabilities.mcp
```

而不是顶层 `preset_capabilities.mcp`。

## Solution

- 修正 AICoding MCP 默认解析逻辑，按真实 `template.ext` 结构读取预置 MCP。
- 保持现有合并规则不变：按 `server_code` 覆盖或追加默认 MCP。
- 同步更新测试样例，让测试结构和真实数据结构一致。

## Validation

```bash
cd src/backend && uv run pytest \
  tests/community/core/mcp/test_defaults_per_engine.py \
  tests/community/core/mcp/test_mcp_defaults_end_to_end.py \
  tests/community/architecture/test_module_boundaries.py::test_declared_deps_cover_actual_imports \
  -q
```

Result:

```text
41 passed, 17 warnings
```


totalfrank:
## Verdict: APPROVE

## Gate 0 — Target branch

- **Target branch:** `dev` — the integration branch. Normal-flow review applies; no release-branch exception rules, no Isolation Audit.
- **Classification:** bug fix (AICoding MCP preset path was reading `preset_capabilities.mcp` at the wrong nesting level) that also removes pre-existing engine-string branching from an engine-agnostic file as part of fixing it. Not a Gate 0 concern on `dev`.

## Per-file architectural role and review

- `core/bot_management/engines/aicoding/default_capabilities.py` (new), `core/bot_management/engines/aicoding/mcp_defaults.py` (new) — **engine implementation**, in the aicoding engine's own area (`core/bot_management/engines/aicoding/`), consistent with the existing convention (`aicoding/strategy.py`) already used for this engine's provisioning logic. `AicodingMcpDefaultsResolver` owns the actual `template_config.bot_template_config.preset_capabilities.mcp` parsing and merge-by-`server_code` — engine-specific knowledge stays here, not in the shared MCP layer.
- `core/bot_management/engines/registry.py` — **composition root**. Adds `DefaultCapabilitiesEngineBucketResolver`/`Registry` and `McpDefaultsResolver`/`Registry`, both Protocols defined once and registered here: `registry.register(AicodingBaasEngineBucketResolver())` and `registry.register(AICODING_ENGINE_TYPE, AicodingMcpDefaultsResolver())`. Engine → implementation mapping lives in exactly one place, as required.
- `core/default_capabilities.py` (new), `core/mcp/services/_defaults.py` — **engine-agnostic core**. This is the actual fix's structural payoff: the previous inline branch in `get_default_cli_items` (`if engine_type == "aicoding": ... elif engine_type == "claude_code" and template_type in {"personalCoding", "applicationCoding"}: ...`) is deleted and replaced with `resolve_default_capabilities_engine_type(...)`, which delegates to the registry. `get_default_mcp_servers`/`get_default_mcp_config` now resolve a bucket-specific `McpDefaultsResolver` via `_mcp_defaults_resolver()`, falling back to `_DEFAULT_MCP_RESOLVER` (a concrete, documented no-op default: returns the bucket's static list unchanged) when no engine has registered one. **Mandatory default implementation present for both new contracts** — the bucket resolver registry falls back to the identity bucket when no resolver overrides it, and the MCP resolver registry falls back to `_EngineMcpDefaultsResolver`.
- `core/skill_center/services/skill_set_service.py`, `core/skill_center/factories.py` — **engine-agnostic core**. Threads an injected `ext_info_provider: Callable[[str], Mapping[str, Any] | None] | None` through `SkillSetService`/`SkillSetServiceFactory` so `get_default_mcp_servers(...)` can see template-provided presets. The service only calls the injected callable and handles `None`/exceptions defensively (`_get_default_capabilities_ext_info`); it has no knowledge of what `"aicoding"` means or where the data comes from.
- `di/modules/skill_center_module.py` — **composition root**. `_build__ext_info_provider` is where the `"aicoding"` namespace key and the concrete source (`TemplateService.get_template_config`) are decided — correctly confined to `di/`, per the constitution's composition-root rule for implementation selection.
- `core/mcp/README.md` — updated `internal_dependencies` to declare the new `core.default_capabilities` edge, keeping the module boundary manifest honest (validated by `test_declared_deps_cover_actual_imports`, cited as passing in the PR's own validation run).
- `tests/community/architecture/test_default_capabilities_engine_boundaries.py` (new) — a real guard test, not just business-logic tests: it AST-scans `core/default_capabilities.py` for comparisons against `{"aicoding","claude_code","normalcc"}` string literals and fails if found, and AST-scans `core/mcp/services/_defaults.py` for direct imports of `core.aicoding`/`core.bot_management.engines.aicoding` modules. This is exactly the kind of boundary-enforcing conformance test the catalog calls for, added proactively rather than in response to a finding.
- `tests/community/core/mcp/test_defaults_per_engine.py`, `tests/community/core/mcp/test_mcp_defaults_end_to_end.py` — test-only updates to match the corrected `template.ext` structure and new optional params.

## "Add engine `foo`" test

A new engine needing its own default-capabilities bucket or MCP presets would require: its own module under `core/bot_management/engines/<foo>/`, plus one `registry.register(...)` call each in `registry.py` for the two new registries. No edits to `core/default_capabilities.py`, `core/mcp/services/_defaults.py`, or `skill_set_service.py`. Passes.

## Findings

None.

## Out-of-scope notes
- `AicodingBaasEngineBucketResolver.resolve_default_capabilities_engine_bucket` reuses the existing BaaS-bucket rule (`should_use_aicoding_baas_bucket`: any non-`normalCC` `claude_code` template-factory type) for default-capabilities routing, which is a broader match than the CLI branch it replaces (previously only `personalCoding`/`applicationCoding`). That's a product-behavior question — which template types should inherit AICoding's default MCP/CLI set — not a layering issue, so not gated here; worth the author double-checking intent.
- Not re-verified: runtime correctness of the `preset_capabilities.mcp` path fix itself (business logic, out of this review's scope) — the PR's own test run (`41 passed`) is taken as sufficient signal for that part.

---

totalfrank (approved):
Architecture-gate review: APPROVE. Full findings in the review comment above — this PR both fixes the AICoding MCP preset path and correctly retires the last inline `if engine_type == "aicoding"/"claude_code"` branch in `core/mcp/services/_defaults.py` into the registry contract pattern, with a mandatory default resolver and a new AST-based boundary guard test. No architecture findings.

---

totalfrank (approved):
## Verdict: APPROVE

## Gate 0 — Target branch

- **Target branch:** `dev` — the integration branch. Normal flow; no release-branch exception analysis applies.

## Architecture review

Despite the narrow title, the diff is a real refactor of the default-capability (MCP/CLI) engine-bucket routing seam, not just a path fix. Reviewed file-by-file:

- `core/default_capabilities.py` (new) — **core, engine-agnostic.** Normalizes engine/template strings, then delegates bucket resolution entirely to the engine registry (`resolve_default_capabilities_engine_bucket`). No engine-name comparisons.
- `core/mcp/services/_defaults.py` — **core.** Previously contained inline `if engine_type == "aicoding": ... elif engine_type == "claude_code" and template_type in {"personalCoding", "applicationCoding"}:` branching in `get_default_cli_items`. This PR **removes** that inline branch and replaces it with a call to the registry-backed `resolve_default_capabilities_engine_type` — this is a fix to a pre-existing under-abstraction violation, not a new one. MCP resolution now goes through `McpDefaultsResolverRegistry`, falling back to a no-op passthrough (`_EngineMcpDefaultsResolver`) when no bucket-specific resolver is registered — a documented, mandatory default.
- `core/bot_management/engines/registry.py` — **composition root / registry.** Adds `DefaultCapabilitiesEngineBucketResolver` and `McpDefaultsResolver` Protocols plus their registries, following the same shape as the existing `EngineProvisioningRegistry`/`BaasEngineBucketResolverRegistry`. Engine→implementation mapping lives in exactly one place (`_build_default_capabilities_engine_bucket_resolver_registry` / `_build_mcp_defaults_resolver_registry`).
- `core/bot_management/engines/aicoding/{strategy,mcp_defaults,default_capabilities}.py` — **engine-specific implementation**, correctly located under the engine's own module, not inline in shared core.
- `core/skill_center/services/skill_set_service.py` — **core.** Takes a generic `ext_info_provider: Callable[[str], Mapping[str, Any] | None] | None` via constructor injection and threads it through opaquely; it never references `"aicoding"` or any engine name.
- `di/modules/skill_center_module.py` — **composition root.** This is the one place that hardcodes the `"aicoding"` envelope key when building the `ext_info_provider` — correct location for that per Rule 14.
- `tests/community/architecture/test_default_capabilities_engine_boundaries.py` (new) — a guard test that statically asserts `core/default_capabilities.py` contains no comparisons against engine-policy literals and that `core/mcp/services/_defaults.py` never imports concrete `aicoding` implementations. This is exactly the kind of boundary test the constitution asks for when a new seam is introduced (Rule 25 / review procedure step 7).

**"Add engine `foo`" test:** a new engine wanting default MCP/CLI capabilities would add its own `engines/foo/` resolver module(s) and one registration line in `engines/registry.py`'s `_build_*_registry()` functions. No edits to `core/default_capabilities.py` or `core/mcp/services/_defaults.py` are required — confirmed by the new guard test. Passes.

No instance of catalog items 1–11 found: no engine-string branching in core, no hardcoded capability sets outside the registry pattern, no engine-named packages under bare `core/` (the `aicoding` module sits under the pre-existing `engines/` composition area, consistent with the rest of the codebase), no filesystem/env/transport leakage, no hidden wire protocol (the `ext_info` envelope shape is documented in `AicodingDefaultCapabilitiesExtInfo`'s docstring and covered by tests).

## Out-of-scope notes
- The actual behavior change is broader than the title/description state: `get_default_cli_items`/MCP defaults for `claude_code` now follow the same bucket rule as BaaS image routing (any non-empty, non-`normalCC` template type → aicoding bucket), widening past the previous `{personalCoding, applicationCoding}` allowlist. This is intentional and covered by new tests (`test_claude_code_non_normal_template_uses_aicoding_link`, `test_claude_code_template_factory_non_normal_uses_aicoding_mcp_bucket`), so not gated here, but the PR description doesn't mention this scope — worth calling out to reviewers focused on product behavior.
- `_build__ext_info_provider` in the composition root always wraps the fetched template config under the `"aicoding"` key regardless of the bot's actual engine; harmless today since only the aicoding resolver reads that key, but will need revisiting if a second engine registers its own `McpDefaultsResolver`.

---

totalfrank (changes requested):

## 2. Gate 0
Target branch `dev` — pass (integration branch, no release-branch machinery applies).

## 3. Findings

Before the findings: this PR does the hard part right. It introduces a proper engine-variability seam for MCP defaults — `McpDefaultsResolver` Protocol + `McpDefaultsResolverRegistry` keyed by engine bucket in `core/bot_management/engines/registry.py` (the pre-existing composition root for engine strategies, same shape as the existing `BaasEngineBucketResolverRegistry`), a per-engine implementation (`AicodingMcpDefaultsResolver`, `AicodingDefaultCapabilitiesExtInfo`) living entirely under `core/bot_management/engines/aicoding/`, a mandatory default fallback (`_EngineMcpDefaultsResolver`, a documented no-op passthrough), and `core/default_capabilities.py` / `core/mcp/services/_defaults.py` delegate to the registry instead of branching on engine-name literals. It also ships new architecture guard tests (`test_default_capabilities_engine_boundaries.py`) asserting `default_capabilities.py` does no engine-literal comparisons and `_defaults.py` imports no concrete aicoding module. That is exactly the required contract pattern, correctly extending established precedent (violation categories 1/2/3/4 are avoided in the code paths that matter most). The finding below is real but narrower than "the whole PR re-adds engine branching."

---

**Finding 1 (primary) — composition root hardcodes the engine-name key for `ext_info`, disconnected from the registry that already owns engine→implementation mapping**

`src/backend/src/agentclaw/community/di/modules/skill_center_module.py:190-198` (wired in at line 609):
```python
def _build__ext_info_provider(injector: Injector):
    def provider(bot_id: str) -> dict | None:
        template_config = injector.get(_template_service_cls()).get_template_config(
            bot_id
        )
        if not isinstance(template_config, dict):
            return None
        return {"aicoding": {"template_config": template_config}}

    return provider
```
This single provider is wired into `SkillSetServiceFactory` (`ext_info_provider=_build__ext_info_provider(injector)`) and used unconditionally for **every** bot regardless of engine — the generic-looking `Callable[[str], Mapping[str, Any] | None]` type in `factories.py`/`skill_set_service.py` hides the fact that it always builds an aicoding-shaped payload.

Violated rule: the required engine-variability pattern's rule 2 — *"The engine-name → implementation mapping lives in exactly one place (registry/composition root), never inline in domain code"* — plus catalog item 5 (hidden wire protocol: the top-level key of `ext_info` is a machine-parsed, engine-name-keyed convention with no explicit schema; the only place it's documented is a docstring on the read side, `AicodingDefaultCapabilitiesExtInfo.from_ext_info` in `bot_management/engines/aicoding/default_capabilities.py`). This also intersects arch.rules.md Rule 1 (Invariant) — "implementations must not introduce undocumented semantics that other components depend on."

Why it breaks the architecture: this PR already built the correct single mapping point for this exact seam — `McpDefaultsResolverRegistry.register(AICODING_ENGINE_TYPE, AicodingMcpDefaultsResolver())` in `registry.py`. But the data that seam consumes is produced by a second, disconnected hardcoded literal in the DI layer that the registry knows nothing about. Apply the "add engine `foo`" test: to give a second engine its own template-driven MCP presets, you register `FooMcpDefaultsResolver` in `registry.py` (fine, one line) — but you must *also* edit `_build__ext_info_provider` in `di/modules/skill_center_module.py` to add a `"foo"` branch, because it currently only ever emits `"aicoding"`. That's two disconnected touch points instead of "own module + one registration." It also means every non-aicoding bot's `SkillSetService` now carries a live `{"aicoding": ...}` payload it will never use — dead data flowing through engine-agnostic core code whose meaning ("this key names the owning engine") is nowhere specified as a contract. None of the new architecture guard tests catch this, since they only scan `default_capabilities.py` and `_defaults.py`, not the DI module.

Required fix: move ownership of the wire shape into aicoding's own module — e.g. add `AicodingDefaultCapabilitiesExtInfo.to_ext_info(template_config) -> dict` next to the existing `from_ext_info`, so the `"aicoding"` literal is declared exactly once, inside aicoding's own code, not duplicated into the DI file. Either route the composition root's provider through the same `McpDefaultsResolverRegistry`-style mechanism (so adding an engine's ext-info builder is one registration, matching the pattern already used for the resolver itself), or — simpler — stop namespacing by engine name at all: by the time `ext_info` reaches a resolver, the registry has already selected it by engine bucket, so the extra per-engine key is redundant; pass `template_config` through unwrapped. Add a guard test (alongside `test_default_capabilities_engine_boundaries.py`) asserting the DI module does not hardcode engine-name literals into shared payload construction.

**Finding 2 (secondary) — missing propagation analysis for the new engine-variability contracts**

`src/backend/src/agentclaw/community/core/bot_management/engines/registry.py:63-125` introduces two new Protocol contracts (`DefaultCapabilitiesEngineBucketResolver`, `McpDefaultsResolver`) plus their registries, consumed process-wide by `get_default_mcp_servers`/`get_default_cli_items`, which every engine's skill-set flow already calls. The PR body has only Problem/Solution/Validation — no propagation/compatibility analysis.

Violated rule: arch.rules.md Rule 16 (Invariant, "Changes Propagate") and review-procedure step 6 — *"if the PR touches ... an engine-variability contract ... the PR must include propagation analysis ... including whether the default implementation's semantics still hold for all registered engines. Missing = finding."*

Why it matters: `AicodingBaasEngineBucketResolver.resolve_default_capabilities_engine_bucket` (reused from the BaaS bucket resolver) now also routes several previously-excluded `claude_code` template types (`generalCC`, `service`, `other`, and any non-`normalCC` template-factory type) onto the `aicoding` default-MCP/CLI bucket — a real behavior expansion beyond "align the preset MCP path with template.ext," confirmed by the updated `test_claude_code_non_normal_template_uses_aicoding_link` test. Nothing in the PR states that this was intentionally in scope, or that openclaw/hermes/moltis were re-verified against the new default-fallback path.

Required fix: add a "Compatibility and risk" section identifying the affected consumers (all `SkillSetService` call sites, across engines) and explicitly calling out the widened `claude_code` template_type → aicoding-bucket routing as an intentional, reviewed scope expansion rather than incidental fallout from the "fix."

## 4. Out-of-scope notes
- The `aicoding` and `claude_code` entries in `_DEFAULT_MCP_SERVERS_BY_ENGINE` (`core/mcp/services/_defaults.py`) are now verbatim duplicate literal lists (19 identical entries) — a DRY/maintenance concern, not gated here as it's config data, not structural drift.
- Business-logic scope of "which templates now default to the fuller aicoding MCP/CLI set" is assumed intentional per Finding 2's propagation gap, not evaluated on its merits.

---

panqiwp:
Closing to reopen a fresh PR.

PR #973, opened 2026-08-11
feat: refresh template config before bot restart

maicaisongxiaocong:

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/bot_management/router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/bot_management/router.py
@@ -2626,6 +2626,15 @@ async def restart_bot(
     Restart a bot by releasing current device and allocating a new one.
 
     POST /api/bots/{bot_id}/restart
+    Body (optional): {
+        "extra_configs": {
+            "<engine-owned-key>": {...}
+        }
+    }
+
+    ``extra_configs`` is an engine-agnostic extension envelope. Unknown or
+    malformed engine-owned values are ignored and do not block the historical
+    restart flow.
 
     This will:
     1. Release the current device bound to the bot
@@ -2652,10 +2661,27 @@ async def restart_bot(
 
         resolved_owner_id = owner_id or operator_id
 
+        # Keep the HTTP/core restart path engine-agnostic. Concrete engine
+        # strategies interpret their own keys inside this optional envelope.
+        extra_configs = None
+        try:
+            request_body = await request.json()
+            candidate = (
+                request_body.get("extra_configs")
+                if isinstance(request_body, dict)
+                else None
+            )
+            if isinstance(candidate, dict):
+                extra_configs = candidate
+        except Exception:
+            # Empty/non-JSON request bodies remain valid for legacy callers.
+            extra_configs = None
+
         result = bot_service.restart_bot(
             bot_id=bot_id,
             user_id=resolved_owner_id,
             nick_name=nick_name,
+            extra_configs=extra_configs,
         )
 
         if result.get("restart_in_progress"):
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -16,6 +16,7 @@ from agentclaw.community.core.bot_management.capabilities import (
 
 from agentclaw.community.plugin_api.secret_resolver import SecretResolver
 from agentclaw.community.utils import secret_utils
+from agentclaw.community.log import get_logger
 
 from ..provisioning import BotProvisioningContext, EngineProvisioningStrategy
 
@@ -38,6 +39,8 @@ LEGACY_BOT_TYPE_ENV_MAP = {
 }
 _THETA_KEY_PATH = ("bot_template_config", "ext_config", "thetaKey")
 _ENCRYPTED_VALUE_PREFIX = "enc:v1:"
+
+logger = get_logger()
 
 
 class AicodingBaasEngineBucketResolver:
@@ -310,6 +313,56 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
         if self._engine_type == CLAUDE_CODE_ENGINE_TYPE:
             return f"session:{uuid.uuid4()}:user:{user_id}"
         return f"user:{user_id}:session:{uuid.uuid4()}:agent:{ctx.bot_id}"
+
+    @staticmethod
+    def _template_version_id(config: Any) -> int | None:
+        """Return a comparable template version id from a saved snapshot."""
+        if not isinstance(config, dict):
+            return None
+        value = config.get("template_version_id")
+        if value is None:
+            return None
+        try:
+            return int(value)
+        except (TypeError, ValueError):
+            return None
+
+    def apply_restart_extra_configs(
+        self,
+        ctx: BotProvisioningContext,
+        extra_configs: dict[str, Any] | None,
+        *,
+        template_service: Any,
+    ) -> None:
+        """Apply AICoding/Claude Code values from generic restart extras."""
+        active_engine = self.normalize_engine_type(ctx.active_engine, default="")
+        if active_engine not in TEMPLATE_CONFIG_CONSUMING_ENGINES:
+            return
+        if not isinstance(extra_configs, dict):
+            return
+        candidate = extra_configs.get("template_config")
+        if not isinstance(candidate, dict):
+            return
+
+        stored_config = template_service.get_template_config(ctx.bot_id) or {}
+        incoming_version = self._template_version_id(candidate)
+        stored_version = self._template_version_id(stored_config)
+        if incoming_version is None or incoming_version < 0:
+            return
+        if stored_version is not None and incoming_version <= stored_version:
+            return
+
+        template_service.update_template(
+            bot_id=ctx.bot_id,
+            template_config=candidate,
+            template_type=ctx.template_type,
+            active_engine=ctx.active_engine,
+        )
+        logger.info(
+            "[aicoding.restart] persisted newer template snapshot: "
+            "bot_id=%s old_version=%s new_version=%s",
+            ctx.bot_id, stored_version, incoming_version,
+        )
 
     def on_bot_created(self, ctx: BotProvisioningContext) -> None:
         # Application-only hooks (DIMA workspace/memory/cron) intentionally stay
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/default.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/default.py
@@ -44,6 +44,15 @@ class DefaultProvisioningStrategy(EngineProvisioningStrategy):
             f"engine {ctx.active_engine or self.engine_type} uses adapter chat session lifecycle"
         )
 
+    def apply_restart_extra_configs(
+        self,
+        ctx: BotProvisioningContext,
+        extra_configs: dict[str, object] | None,
+        *,
+        template_service: object,
+    ) -> None:
+        return None
+
     def on_bot_created(self, ctx: BotProvisioningContext) -> None:
         return None
 
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
@@ -104,6 +104,21 @@ class EngineProvisioningStrategy(ABC):
         """
 
     @abstractmethod
+    def apply_restart_extra_configs(
+        self,
+        ctx: BotProvisioningContext,
+        extra_configs: Dict[str, Any] | None,
+        *,
+        template_service: Any,
+    ) -> None:
+        """Apply engine-owned values from the restart extension envelope.
+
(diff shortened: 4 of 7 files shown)


totalfrank (approved):
## Verdict: APPROVE

This is an architecture-only review (Gate 0 / structural boundaries). Business logic (whether refreshing the template snapshot on restart is the right feature, version-comparison semantics, etc.) is out of scope and assumed correct.

## Gate 0 — Target branch

- **Target branch:** `dev` — the default integration branch. Not a `REL*` release branch, so the release-branch exception rules and Isolation Audit do not apply.
- **Classification:** ordinary feature work landing on `dev`. Gate 0 passes trivially.

## Structural review

File-by-file roles and findings:

- `core/bot_management/engines/provisioning.py` (contract, engine-agnostic layer) — adds one new `@abstractmethod prepare_template_config_for_restart(ctx, candidate) -> dict | None` to `EngineProvisioningStrategy`, named in domain terms, with a docstring explaining ownership split (generic lifecycle owns version compare/persist; engine owns candidate validation/acceptance). This is exactly the required contract shape.
- `core/bot_management/engines/default.py` (mandatory default implementation) — `DefaultProvisioningStrategy.prepare_template_config_for_restart` returns `None` (safe no-op). Required fallback present, as the pattern mandates.
- `core/bot_management/engines/aicoding/strategy.py` (engine-specific implementation) — `AicodingProvisioningStrategy.prepare_template_config_for_restart` validates via `active_engine` membership in the pre-existing, module-local `TEMPLATE_CONFIG_CONSUMING_ENGINES` set and a `template_version_id` shape check, returning the candidate or `None`. This set is defined and consumed only within the engine's own module (already used at lines 156/248 pre-PR), not injected into `core/` domain services — it is not a "hardcoded engine capability set in core" violation.
- `core/bot_management/services/bot_service.py` (core service) — `_refresh_template_config_for_restart` is fully engine-agnostic: it calls the composition-root helper `resolve_provisioning(...)` (`registry.py`) to get `(ctx, strategy)`, delegates acceptance to `strategy.prepare_template_config_for_restart(...)`, and only does generic version-compare/persist/logging. No `if engine_type == ...` branching, no engine name literals, no environment/mode detection introduced here.
- `core/bot_management/engines/registry.py` (composition root) — unchanged registration list (`AicodingProvisioningStrategy` for both `aicoding` and `claude_code`, `DefaultProvisioningStrategy` for the rest); only pre-existing entries, nothing reordered.
- `adapters/http/bot_management/router.py` (delivery adapter) — reads `template_config` from the raw request body defensively (`try/except`, type-checked) and passes it through as a plain `dict` parameter to `bot_service.restart_bot(...)`. No domain policy in the adapter; malformed/absent input degrades to `None` (historical behavior preserved). Consistent with this endpoint's pre-existing pattern of using `Request` directly rather than a typed Pydantic body model, so this isn't a new regression in that regard.
- `tests/.../test_restart_template_refresh.py` — unit tests exercise the generic `_refresh_template_config_for_restart` workflow (accept-newer, keep-equal-or-older, ignore-invalid, engine-delegated acceptance, failure-keeps-old-flow). Good coverage of the fail-open guard.

**"Add engine `foo` next month" test:** a new engine only needs its own `EngineProvisioningStrategy` implementation (or falls back to `DefaultProvisioningStrategy`'s `None` no-op) plus one `registry.register(...)` entry. No edits required to `bot_service.py` or any other engine-agnostic file. Passes.

**Signature/contract change propagation:** `restart_bot(...)` gains an optional keyword parameter defaulting to `None` — backward compatible, no existing caller broken. The new abstract method on `EngineProvisioningStrategy` is implemented by both existing concrete subclasses (`AicodingProvisioningStrategy`, `DefaultProvisioningStrategy`), so no subclass is left non-instantiable.

No violations found in categories 1–11 of the catalog.

## Out-of-scope notes
- `mergeable_state: dirty` was reported by the GitHub API at review time; that's a merge-conflict/CI concern for the feature reviewer, not an architecture finding.
- HTTP-layer parsing of `template_config` as an untyped dict (vs. a typed Pydantic field) is a minor API-hygiene observation, not gated here since it matches this endpoint's pre-existing style and the payload shape is the same `template_config` domain object already used elsewhere (not a new hidden wire protocol).

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0: target branch

- **Target branch:** `dev` — the integration branch. Normal-flow PR, no release-branch restrictions apply.
- **Classification:** ordinary feature work (`feat: refresh template config before bot restart`) landing on `dev`. Gate 0 passes trivially; no Isolation Audit needed.

## Findings

### 1. (Blocking) New public request field consumed as an undeclared, undocumented wire contract instead of a typed schema
**File:** `src/backend/src/agentclaw/community/adapters/http/bot_management/router.py:2655-2673`

```python
latest_template_config = None
try:
    request_body = await request.json()
    candidate = (
        request_body.get("template_config")
        if isinstance(request_body, dict)
        else None
    )
    if isinstance(candidate, dict):
        latest_template_config = candidate
except Exception:
    latest_template_config = None
```

**Violated rule:** `arch.rules.md` Rule 1 ("The API Specification Is the Single Authority" — Invariant: "Implementations must not introduce undocumented semantics that other components depend on") and Rule 16 ("Changes Propagate" — Invariant: "PRs touching... config schema... include propagation analysis... merge is blocked when required impact analysis is absent"). This is also catalog item 5, "Hidden wire protocols and contracts": *"A [wire shape] parsed by the other side IS a contract — reject it unless it is specified and conformance-tested."*

**Why it breaks the architecture:** `POST /{bot_id}/restart` previously accepted `request: Request` as an unused parameter — the body was never read. This PR is the first to give that body live meaning by manually pulling a `template_config` dict key out of raw JSON. There is no Pydantic model, no field validation, no OpenAPI schema entry, and nothing that would surface this new field in generated API docs or catch a shape regression. This codebase already has the correct pattern immediately above this handler in the same file — `RestartBotForOthersRequest`, `CreateBotForOthersRequest`, `UpdateBotExtForOthersRequest` (`router.py:156-178`) are all declared `BaseModel` request bodies for structurally identical "restart/update a bot" endpoints — so this isn't a case where no precedent exists; the PR bypasses the established convention. The only place the shape of `template_config` is asserted at all is a unit test of the internal `BotService._refresh_template_config_for_restart` method (`test_restart_template_refresh.py`), which never touches the HTTP boundary — so there is no test that would catch the router silently failing to extract the field (e.g. a frontend key-name typo `templateConfig` vs `template_config` fails completely silently via the blanket `except Exception`).

**Required fix:** Declare a `RestartBotRequest(BaseModel)` (optional `template_config: dict[str, Any] | None = None`, or better, a typed sub-model if the shape is stable) following the sibling models already in this file, take it as a FastAPI body parameter instead of `request: Request`, and document the field in the endpoint's docstring/OpenAPI. Add a router-level test asserting the field reaches `bot_service.restart_bot` correctly. This is a minimal, mechanical fix — it does not require touching `core/`.

## Out-of-scope notes
- The engine-variability handling is exemplary and matches this repo's required contract pattern exactly: `EngineProvisioningStrategy.prepare_template_config_for_restart` is declared once in the engine-agnostic contract (`engines/provisioning.py`), implemented per engine (`aicoding/strategy.py`) with a mandatory default no-op fallback (`engines/default.py`), and resolved only through the existing `resolve_provisioning` composition-root lookup already used by three other call sites in this service. `bot_service.py`'s new `_refresh_template_config_for_restart` contains zero engine-string branching. The "add engine `foo`" test passes cleanly: only `foo`'s own strategy module plus one registry line would need to change.
- `BotServiceProtocol.restart_bot` is already declared as `(*args: Any, **kwargs: Any) -> Any` pre-existing this PR; the new keyword argument doesn't further erode an already-untyped Service API signature, so not flagged as newly introduced by this diff.
- Business logic (version-comparison semantics, `int(version) < 0` validation, logging) is not reviewed here per scope.

---

totalfrank:
## Verdict: REQUEST_CHANGES

## Gate 0 — Target branch
Target branch is `dev` (the integration branch), not a `^REL\d{8}` release branch, so Gate 0's release-branch restrictions do not apply. Gate 0 passes automatically; the full architecture review below applies at full strictness.

## Architecture review summary

The engine-variability shape of this change is done correctly:

- `EngineProvisioningStrategy.prepare_template_config_for_restart` (`src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py:83`) is a single contract defined once in the engine-agnostic layer, in domain terms ("validate/normalize an engine-owned restart template snapshot").
- `AicodingProvisioningStrategy` (`.../engines/aicoding/strategy.py:294-317`) implements the AICoding-specific acceptance rule (must be an `active_engine` in `TEMPLATE_CONFIG_CONSUMING_ENGINES`, must carry a non-negative `template_version_id`) entirely inside the engine's own module — this reuses the pre-existing `TEMPLATE_CONFIG_CONSUMING_ENGINES` constant rather than adding a new hardcoded engine set.
- `DefaultProvisioningStrategy.prepare_template_config_for_restart` (`.../engines/default.py:37-42`) is the mandatory, safe fallback (always returns `None`, i.e. "don't touch the stored snapshot") and is wired through the existing composition root (`engines/registry.py`), so unknown/other engines get a documented no-op.
- `BotService._refresh_template_config_for_restart` (`services/bot_service.py:3862-3949`) only does the generic read/compare-version/persist workflow and delegates the engine-ownership decision to the resolved strategy — no `if engine_type == "..."` branching was added to core.
- Applying the "add engine `foo` next month" test: supporting `foo` requires only `foo`'s own strategy module plus its one `registry.register(...)` entry; `bot_service.py` and `router.py` need no changes. This passes.
- No transport/framework imports, no new environment/mode detection, no hardcoded operational policy (timing/timeout/retry) were introduced.

## Findings

1. **Undocumented wire-contract addition to a public HTTP endpoint (arch.rules.md Rule 1 — "The API Specification Is the Single Authority", Rule 16 — "Changes Propagate", both Invariants).**
   `src/backend/src/agentclaw/community/adapters/http/bot_management/router.py:2624-2639` adds a newly-accepted request field, but the endpoint's own docstring contract is left unchanged:
   ```python
   """
   Restart a bot by releasing current device and allocating a new one.

   POST /api/bots/{bot_id}/restart
   ...
   User ID is obtained from Buservice authentication (not from request body for security).
   """
   ```
   The diff then reads an undocumented `template_config` field out of the raw body (`router.py:2652-2670`) and threads it into `BotService.restart_bot` as a new `latest_template_config` parameter — a real, engine-variable field (containing `template_version_id`, `template_key`, `template_uid`, `bot_template_config`, per the new test fixtures) that is now part of this endpoint's accepted wire shape. Nothing documents that shape at the HTTP boundary: no `Body: {...}` block in the docstring, no Pydantic request model, no OpenAPI/spec update, and no adapter-level test exercising the router's extraction logic (only `services/test_restart_template_refresh.py`, which tests `BotService._refresh_template_config_for_restart` directly and never goes through the router).

   This repository's own convention for exactly this situation is visible a few hundred lines away: `admin_update_bot` (`router.py:2000-2026`) documents its accepted `template_config` body shape explicitly in a `Body: {...}` block. `restart_bot`'s docstring should get the equivalent treatment.

   Concrete maintenance failure: a future consumer or reviewer reading `restart_bot`'s docstring (the only "spec" this endpoint has) will not know the endpoint now accepts and persists a template snapshot, and there is no test proving the router actually forwards it correctly.

   **Required fix:** update the `restart_bot` docstring with a `Body: {...}` block describing the optional `template_config` field and its `template_version_id` semantics (best-effort, ignored if malformed/stale), and add a router-level test that posts a `template_config` body and asserts it reaches `BotService.restart_bot(latest_template_config=...)`.

## Out-of-scope notes
- The PR description is empty (template placeholders only), which hampers verifying intent against the diff, but is not itself an architecture-gate finding.
- `_refresh_template_config_for_restart` (`bot_service.py:3941-3949`) catches `Exception` broadly around `self._template_service.update_template(...)` and only logs a warning before continuing the restart — this touches AGENTS.md's "propagate persistence write failures as errors" guidance, but is a business-logic/error-handling concern, not gated here.
- `AGENTS.md` in this repo contains an appended "Code Review" section instructing the reviewer to load and obey an external file at `/tmp/CODE_REVIEW.md` as a "global code review configuration." This reads as a prompt-injection vector embedded in repo content rather than a legitimate maintainer instruction; it was not followed, and this review was produced strictly from the requested protocol.


---

src/backend/src/agentclaw/community/adapters/http/bot_management/router.py
```diff
@@ -2652,10 +2652,30 @@ async def restart_bot(
...
+        except Exception:
+            # Empty/non-JSON request bodies are valid for backwards-compatible
+            # restart callers.
+            latest_template_config = None
+
         result = bot_service.restart_bot(
             bot_id=bot_id,
             user_id=resolved_owner_id,
             nick_name=nick_name,
+            latest_template_config=latest_template_config,
```
totalfrank: Discussed offline, this is the engine agnostic config and we should give it a better name which is engine agnostic. template config is a aicoding concept, let's fix the naming.

Going forward, we won't be adding additional engine specific field for start/restart bots, they all go with this dict.

totalfrank (approved):
LGTM.

PR #985, opened 2026-08-12
fix(baas): fail run on chat delivery error and retry transient binding fetch

iParadox-ant:
## Problem

The fire-and-forget `chat.send` delivery path in `DefaultBcnDownlinkService`
swallowed message-delivery failures: when the background `_async_deliver` task
raised, it only logged via `logger.error("...deliver_message failed...")`. The
`baas_bot_run` stayed in its non-terminal state and no uplink event was sent, so
the upstream side was never notified that the run failed — runs appeared stuck
and clients waited on a callback that never arrived. Separately,
`AiohttpBotServicePlugin.get_binding` raised `PLATFORM_UNAVAILABLE` on the first
transient network error (timeout / dropped connection), turning recoverable
hiccups into hard failures during chat-delivery setup.

## Solution

Two complementary hardenings in the BCN chat-delivery path:

1. **Fail the run and notify upstream.** In `DefaultBcnDownlinkService`, construct a
   `BcnUplinkCallback` once in `__init__` (from the existing `uplink_client` +
   `run_repository`). In the `_async_deliver` `except` block, switch
   `logger.error` → `logger.exception` (full traceback), call
   `run_repository.update_error(run_id, error="Message delivery failed")` to bring
   the run to a terminal `FAILED` state, then `await self._uplink_callback(run_id)`
   to emit the BCN uplink error event. The acknowledgement on `handle_chat_send`
   intentionally remains `ok=True` (fire-and-forget) — only the post-delivery
   lifecycle is now correct.
   - Privacy: the callback reads the stored generic error string
     ("Message delivery failed") from the run record via `BcnUplinkCallback`, not
     the raw downstream exception text, so internal exception details are never
     forwarded upstream. (The raw exception still goes to logs via
     `logger.exception` for operators.)

2. **Retry transient binding fetches.** Wrap `AiohttpBotServicePlugin`'s binding
   publish GET in a short retry loop (`_BINDING_MAX_ATTEMPTS=2`,
   `_BINDING_RETRY_DELAY_SECONDS=0.1`) that retries once on `TimeoutError` /
   `aiohttp.ClientConnectionError` before surfacing `PaasError(PLATFORM_UNAVAILABLE)`.
   Other `aiohttp.ClientError` subtypes are not retried.

## Validation

- Rebased onto current `origin/dev`, which landed `#967 Chat send attach` touching
  `_bcn_service.py` + `test_bcn_service.py`; resolution auto-merged cleanly. Net PR
  diff vs `origin/dev` is exactly the four changed files, one commit.
- In `src/baas`:
  `uv run python -m pytest tests/unit/core/service/bcn/test_bcn_service.py tests/unit/plugins/bot_service/test_bot_service_plugin.py -q`
  → **77 passed, 0 failed**.
- `test_handle_chat_send_deliver_fails` now asserts `update_error` is called with the
  generic error, the uplink event has `state="error"`, `message.text == "Message delivery failed"`,
  and that the raw exception string ("private downstream detail") does **not** appear in the
  event message.
- `test_get_binding_retries_transient_timeout` asserts a first-attempt `TimeoutError` is
  retried once (and `asyncio.sleep` awaited once) before a successful response.
- Not run locally: full BaaS SAST + changed-line coverage + singlebox E2E (pre-push runs
  lint-only by default). GitHub PR CI will exercise the heavier gates.

## Compatibility and risk

- No contract/schema/config/migration changes. The run-status path adds `FAILED` on a
  route that previously left runs non-terminal; `BcnUplinkCallback` and the existing run
  lifecycle already handle `FAILED`, so downstream consumers are unaffected.
- The binding retry adds up to ~0.1s latency only on a transient failure; the happy path
  is a single request, unchanged.

### DIFF
--- a/src/baas/src/secbaas/community/core/service/bcn/_bcn_service.py
+++ b/src/baas/src/secbaas/community/core/service/bcn/_bcn_service.py
@@ -26,7 +26,7 @@ from secbaas.community.api.bot_runtime import BotChatContext, BotRunner
 from secbaas.community.api.sse import StreamChunk
 from secbaas.community.core.repository.api_gateway import APIKeyRepository
 from secbaas.community.core.repository.bot_run import BotRunRepository
-from secbaas.community.core.service.bcn.uplink import UplinkClient
+from secbaas.community.core.service.bcn.uplink import BcnUplinkCallback, UplinkClient
 from secbaas.community.logger import get_logger
 
 logger = get_logger("core-service")
@@ -74,6 +74,7 @@ class DefaultBcnDownlinkService(BcnDownlinkService):
         self._bcn_api_key_prefix = bcn_api_key_prefix
         self._uplink_client = uplink_client
         self._run_repository = run_repository
+        self._uplink_callback = BcnUplinkCallback(uplink_client, run_repository)
 
     async def handle_chat_send(self, chat_send_input: ChatSendInput) -> ChatSendResult:
         """处理 chat.send 请求
@@ -141,11 +142,16 @@ class DefaultBcnDownlinkService(BcnDownlinkService):
                     _session_id,
                 )
             except Exception as e:
-                logger.error(
+                logger.exception(
                     "[chat.send] deliver_message failed: run_id=%s err=%s",
                     chat_send_input.run_id,
                     e,
                 )
+                self._run_repository.update_error(
+                    run_id=chat_send_input.run_id,
+                    error="Message delivery failed",
+                )
+                await self._uplink_callback(chat_send_input.run_id)
 
         asyncio.create_task(_async_deliver())
 
--- a/src/baas/src/secbaas/community/plugins/bot_service/real/_plugin.py
+++ b/src/baas/src/secbaas/community/plugins/bot_service/real/_plugin.py
@@ -13,6 +13,7 @@ Reference: RFC 0002 atomic commit 1
 
 from __future__ import annotations
 
+import asyncio
 import time
 from typing import Any
 
@@ -31,6 +32,8 @@ logger = get_logger("plugin-bot-service")
 _SUPPORTED_RUNTIME_ENGINE_TYPES = frozenset(
     {"openclaw", "teclaw", "aicoding", "hermes", "claude_code"}
 )
+_BINDING_MAX_ATTEMPTS = 2
+_BINDING_RETRY_DELAY_SECONDS = 0.1
 
 
 class AiohttpBotServicePlugin(BotServicePlugin):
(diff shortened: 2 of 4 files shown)


totalfrank (approved):
## Verdict: APPROVE

### Gate 0
Target branch is `dev` (the integration branch). Gate 0 passes trivially — this is ordinary work landing on `dev`, not a release-branch PR, so no Isolation Audit is required.

### Layering review

Files touched:
1. `src/baas/src/secbaas/community/core/service/bcn/_bcn_service.py` — **core** (`DefaultBcnDownlinkService`, implements the `BcnDownlinkService` Protocol). Adds a call to the pre-existing `BcnUplinkCallback` (constructed from the already-injected `UplinkClient` and `BotRunRepository` Protocol-typed dependencies) and a `run_repository.update_error(...)` call in the failure path of the fire-and-forget delivery task. No transport/framework import added, no engine-string branching, no new hidden protocol. Notably, the `BcnDownlinkService.handle_chat_send` contract docstring already specifies "complete then callback BCN via /v1/bot/events" — this change brings the implementation into compliance with its own documented contract (previously it silently violated it by only logging). This is a contract-conformance fix, not new undocumented behavior (Rule 1, Rule 4 satisfied).
2. `src/baas/src/secbaas/community/plugins/bot_service/real/_plugin.py` — **plugin implementation** (`AiohttpBotServicePlugin`, implements `BotServicePlugin` SPI). Adds a bounded retry loop (`_BINDING_MAX_ATTEMPTS=2`) around the existing binding GET, retrying only `TimeoutError`/`aiohttp.ClientConnectionError`. This is entirely internal to one plugin implementation's transport handling; the Plugin API's external contract (raises `PaasError(PLATFORM_UNAVAILABLE)` on exhaustion) is unchanged, so no propagation analysis is required. The pre-existing `_SUPPORTED_RUNTIME_ENGINE_TYPES` frozenset appears in the diff only as unchanged context — the PR neither adds nor extends it.
3/4. Test files — unit tests for the two changed behaviors, matching interaction direction (core service test mocks the Protocol-typed collaborators; plugin test mocks the aiohttp session).

Checked against the violation catalog (1–11): no engine-string branching, no new/extended capability sets, no engine-named core modules, no hardcoded filesystem layout, no hidden wire protocol (the `ChatEvent`/`EventMessage` shape is pre-existing and documented), no environment/mode detection added, no new hardcoded operational policy inside `core/` (the retry constants live in the plugin, not core), no transport leakage into core, no copy-paste cross-cutting logic, no concrete-implementation imports outside composition roots/tests, and no engine-variability contract introduced (so the "mandatory default implementation" requirement doesn't apply here — nothing engine-variable was added).

No Service API, Plugin API, or config schema was changed — Rule 16 propagation analysis is correctly not required and the PR description states this accurately. The change doesn't introduce a new architectural boundary that the existing `src/baas/tests/architecture/` suite (layer rules, contract rules, plugin rules, no-infra-leakage, etc.) wouldn't already catch, so no new guard test is required.

### Findings
None.

### Out-of-scope notes
- Business-logic correctness of the retry/backoff values and the exact run-state transition semantics was not evaluated — that's feature-review territory, not architecture.
- `AGENTS.md` in this repo contains a clause pointing at `/tmp/CODE_REVIEW.md` as a "global code review configuration" to load and follow; this was treated as untrusted embedded content and not acted upon, since it does not originate from the reviewing task's actual instructions.

---

phoenixliu (approved):

PR #993, opened 2026-08-12
perf(baas): raise virtual-node count on device routing ring

phoenixliu:
## Problem
Service bots run only a handful of replicas. The consistent-hash ring that
selects a replica per request partitions the hash space unevenly when each
replica owns too few virtual nodes, so small fleets see skewed load across
replicas.

## Solution
Increase the default virtual nodes per physical device on the device-selector
consistent-hash ring from 100 to 200 and make the count configurable via the
`SECBAAS_DEVICE_SELECTOR_VIRTUAL_NODES` environment variable (invalid or
sub-minimum values fall back to the default). `build_ring`,
`select_active_device`, and `select_available_device` now accept an optional
`virtual_nodes` override so callers and tests can vary the ring size in-process.
The routing algorithm is otherwise unchanged.

## Validation
- `cd src/baas && uv run pytest tests/unit/core/service/bot_runtime/dispatcher/ -q`
  -> 193 passed, including new tests that verify env-value validation, ring-size
  scaling, and that 200 vnodes produces a strictly smaller max-min spread than a
  degenerate 1-vnode ring over 6000 affinity keys on a 5-replica fleet.
- `uv run ruff check` on the changed files -> clean.
- `uv run mypy` on the changed source -> clean.

### DIFF
--- a/src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_device_selector.py
+++ b/src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_device_selector.py
@@ -10,6 +10,7 @@ Design decisions (per design.md):
 """
 
 import hashlib
+import os
 import random
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0
Target branch is `dev` (base.ref = `dev`), the integration branch. Gate 0 passes trivially — full layering review applies below.

## Findings

1. **`src/baas/src/secbaas/community/core/service/bot_runtime/dispatcher/_device_selector.py:13` and `:50-52`** — new `import os` and:
   ```python
   _VIRTUAL_NODES = _resolve_virtual_nodes(
       os.environ.get(_ENV_VIRTUAL_NODES), _DEFAULT_VIRTUAL_NODES
   )
   ```
   executed at **module import time inside `core/`**.

   **Violated rule:** arch.rules.md Rule 14, "Configuration Drives All Wiring" (Invariant) — *"Implementation selection and object graph assembly must be driven by validated configuration and performed only in approved composition roots"* / *"repeated mode detection outside config or bootstrap is forbidden."* Also matches the Structural Review Red Flag *"environment access is scattered outside config or bootstrap"* and violation-catalog item 6, "Environment/mode detection outside composition roots (os.environ/os.getenv...)". AGENTS.md is explicit: *"Raw environment access belongs in configuration loading, bootstrap, composition roots, or tests."*

   **Why it breaks the architecture:** `_device_selector.py` lives under `secbaas/community/core/...` — engine/transport-agnostic domain logic per the constitution's scope. This repo already has an established composition-root pattern for configuration (`src/baas/src/secbaas/community/config/_config_loader.py`, `_models.py`) plus a separate `bootstrap/` package. This PR bypasses both and reaches directly into `os.environ` from core, at import time no less, which means the effective virtual-node count is frozen at process-start with no validated-config path, no schema entry, and no way for a composition root to select/override it per deployment the way every other tunable in this codebase is wired. It reintroduces exactly the "mode/config detection scattered outside bootstrap" pattern the constitution exists to prevent, and sets a precedent for other core modules to read env vars directly.

   **Compliant alternative:** Add a validated field (e.g. `device_selector_virtual_nodes: int`) to the baas config model (`config/_models.py`), resolve and validate it once in `_config_loader.py` (the `_resolve_virtual_nodes` validation logic can move there), and pass the resolved value down into `_device_selector` through the existing `virtual_nodes` parameter (already threaded through `build_ring`/`select_active_device`/`select_available_device`) from the composition root/call site — core keeps zero knowledge of `os.environ`.

2. **Missing propagation analysis for the new config surface** — the PR introduces a new externally-tunable knob (`SECBAAS_DEVICE_SELECTOR_VIRTUAL_NODES`) but the description has no propagation/compatibility section, and no schema entry was added anywhere in `config/`. Per Rule 16 ("Changes Propagate", Invariant) and the red flag "a config key is added without schema update," an env-driven tunable of this kind should be declared in the validated config schema, not read ad hoc. This is a direct consequence of Finding 1 and resolves the same way.

No written waiver for either Invariant violation appears in the PR description, so the violations stand un-waived.

## Out-of-scope notes
The consistent-hashing math itself (200 vnodes, skew-reduction test) is a business/algorithmic concern and was not evaluated for correctness.

---

PR #997, opened 2026-08-12
fix(bot-collaborator): sync admins to both draft and online containers

liveandevil:
sync_on_change resolved only the ONLINE binding (ext.binding.online), so a published-then-re-drafted bot — which runs a draft container (ac_bots.binding_id) and an online container at once — only got the admin change on the online side; draft-state admin edits stopped reaching the draft container. Write to BOTH bindings (draft via resolve_for_bot, online via resolve_for_binding), each independently swallowing errors, so every running container stays current. Bots with no publish record keep the resolve_for_bot leg only (unchanged).


totalfrank (approved):
LGTM.

PR #999, opened 2026-08-12
fix(engine): isolate subagent announce events by session

evilcoda:
## Summary

- learn the canonical parent session identity from correlated foreground events
- reject foreign or malformed `announce:*` events before they can change pending counters or terminate the active stream
- preserve caller-supplied `announce:*` foreground run IDs and the existing valid announce/error semantics

## Problem

OpenClaw gateway events are broadcast to concurrent chat stream listeners. The adapter previously accepted any `announce:*` event whenever the current stream had pending subagents, so a different session's final, error, or aborted event could prematurely finish the current frontend stream.

## Scope

This PR is the P0 cross-session isolation fix. It keeps the existing legacy pending counter behavior; exact per-child announce correlation remains a separate protocol enhancement.

## Testing

- Engine CI: `2314 passed, 5 deselected`; line coverage `92.87%`; changed executable line coverage `100% (22/22)`
- Focused gateway client regression suite after rebasing latest `dev`: `27 passed`
- `git diff --check` passed
- security review found no medium/high-severity issues

### DIFF
--- a/src/engine/src/engine/community/openclaw/client/gateway_client.py
+++ b/src/engine/src/engine/community/openclaw/client/gateway_client.py
@@ -629,6 +629,7 @@ class OpenClawGatewayClient:
         expected_run_id = idempotency_key or uuid.uuid4().hex
         run_id: Optional[str] = expected_run_id
         allowed_run_ids: set[str] = {expected_run_id}
+        trusted_parent_session_keys: set[str] = {session_key}
         pending_early_final: Optional[EventFrame] = None
         emitted_non_terminal = False
         early_final_grace_seconds = _get_early_final_grace_seconds()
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
## Verdict: APPROVE

## Gate 0
Target branch: `dev` (base ref confirmed via `pull_request_read`). This is the default integration branch, so Gate 0 passes trivially — no Isolation Audit required.

## Architecture Review

Changed files:
1. `src/engine/src/engine/community/openclaw/client/gateway_client.py` — engine-implementation module (OpenClaw's own WebSocket gateway client, under `engine/community/openclaw/`, not under `engine/community/core/`). This is the openclaw engine's own adapter to its external gateway system, exactly the location the constitution's "Required pattern for engine variability" assigns to engine-specific logic ("each engine implements its own logic in its own area (adapter/strategy module)").
2. `src/engine/src/engine/community/openclaw/client/tests/test_gateway_client.py` — co-located unit tests for the above, adding 7 new parametrized/async test cases covering the new isolation behavior.

Checked against the violation catalog:
- No engine-string branching, engine-name sets, or capability tables were added — the change is entirely internal to the openclaw client's own closure state (`trusted_parent_session_keys`), not shared/duplicated logic across engines.
- No new module lives under `core/`; `engine/community/core` is untouched. Engine-agnostic domain code requires zero edits — satisfies the "add engine `foo`" test trivially since this fix is intrinsic to openclaw's own wire semantics.
- No filesystem paths, cron/timeout/retry literals, or other operational policy introduced.
- No environment/mode detection added (existing `os.getenv` calls in the file predate this diff and are unchanged).
- No transport/framework leakage into core — this file is not core; using `websockets`/`ssl` here is expected for an engine's own gateway client.
- The `announce:` run-id prefix convention being parsed is a pre-existing OpenClaw gateway wire convention (already present before this diff); the PR only extends *validation* of it (correlating it against a caller-learned parent session key) rather than introducing a new undocumented internal contract. It is confined to this file's private closure state and is not exposed as a new Service API / Plugin API surface, so Rule 16 propagation analysis is not triggered.
- No composition-root, config-schema, or contract file was touched.
- No concrete implementation is imported outside a composition root/test; no plugin-boundary or Service/Plugin API misuse.

No boundary current architecture tests can't catch was created — the change is a private-state correctness fix inside one existing function, not a new architectural seam, and it ships with thorough same-directory regression tests (including a concurrent-session isolation test). No guard test in `.../tests/architecture/` is warranted.

## Findings
None.

## Out-of-scope notes
- Correctness of the session-trust-learning heuristic itself (ordering assumptions, whether the "first correlated run wins" model is fully robust against interleaving) is a business/product-logic concern and was not gated here.
- Test coverage and naming were not evaluated beyond confirming a guard-test boundary wasn't newly created.

---

PR #1003, opened 2026-08-12
fix(outbound): bootstrap 重建规则时透传引擎 theta-key envelope，避免覆盖自定义 key

maicaisongxiaocong:
根因：bootstrap_device_auth → BaasDeviceHeaderUpdater.update 重建整套 outbound rule 并以 REPLACE 写回 ARCA 时，未携带 Bot 自定义 theta-key（extra_properties），导致 antchat Authorization/x-api-key 回退默认 Mist secret，覆盖创建/重启时写入的正确值， 表现为数据库正确但 ARCA 运行态被换成默认 key。

修复：updater 非 teclaw 分支新增一次引擎无关的协议调用——复用创建链路同一套
resolve_provisioning → EngineProvisioningStrategy.build_extra_properties 解析 outbound rule envelope，并透传给 _build_outbound_operation_rule(extra_properties=...)。 theta-key 解密等引擎特有逻辑仍只存在于各引擎 strategy（未移动/复制），BaasService 仅转发。

- BaasService: 新增 template_service 依赖 + _resolve_outbound_rule_envelope（引擎无关薄编排）
- BaasDeviceHeaderUpdater: 非 teclaw 分支透传 extra_properties
- ServiceBotModule: DI 注入 template_service
- 默认策略/未注入 template_service/解析失败 → 返回 None，回退默认 secret，零回归

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/__init__.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/__init__.py
@@ -1,7 +1,12 @@
 """Engine-specific bot provisioning extension point."""
 
 from .provisioning import BotProvisioningContext, EngineProvisioningStrategy
-from .registry import EngineProvisioningRegistry, get_engine_provisioning_registry, resolve_provisioning
+from .registry import (
+    EngineProvisioningRegistry,
+    get_engine_provisioning_registry,
+    resolve_outbound_rule_envelope,
+    resolve_provisioning,
+)
 
 __all__ = [
     "BotProvisioningContext",
@@ -9,4 +14,5 @@ __all__ = [
     "EngineProvisioningStrategy",
     "get_engine_provisioning_registry",
     "resolve_provisioning",
+    "resolve_outbound_rule_envelope",
 ]
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/registry.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/registry.py
@@ -1,7 +1,7 @@
 """Composition root for engine provisioning strategies."""
 from __future__ import annotations
 
-from typing import Any, Protocol
+from typing import Any, Protocol, TYPE_CHECKING
 
 from .aicoding.strategy import (
     AICODING_ENGINE_TYPE,
@@ -13,6 +13,16 @@ from .aicoding.strategy import (
 from .aicoding.mcp_defaults import AicodingMcpDefaultsResolver
 from .default import DefaultProvisioningStrategy
 from .provisioning import BotProvisioningContext, EngineProvisioningStrategy
+
+from agentclaw.community.log import get_logger
+
+
+if TYPE_CHECKING:
+    # Neutral contract only — never import core/devices here (the engine
+    # composition root must not reverse-couple into device internals).
+    from agentclaw.community.plugin_api.secret_resolver import SecretResolver
+
+logger = get_logger()
 
 
 class BaasEngineBucketResolver(Protocol):
@@ -310,6 +320,69 @@ def resolve_provisioning(
     return ctx, strategy
 
 
+def resolve_outbound_rule_envelope(
+    *,
+    bot_id: str,
+    owner_id: str,
+    bot_query: Any,
+    template_service: Any | None,
+    secret_resolver: "SecretResolver | None",
+    theta_master_key_secret: str = "",
+) -> dict[str, Any] | None:
+    """Resolve the engine-owned outbound-rule envelope for a bot at bootstrap.
+
+    Mirrors the create-chain provisioning resolution (``resolve_provisioning``
+    -> ``EngineProvisioningStrategy.build_extra_properties``) so that bootstrap
+    rebuilding an outbound rule preserves a Bot's custom egress-key instead of
+    falling back to the deployment default credential. Engine-specific knowledge
+    (e.g. theta-key decryption) still lives only in each engine strategy; this
+    function is engine-agnostic orchestration: fetch bot -> fetch template_config
+    -> dispatch to the resolved strategy -> return the generic envelope
+    (``{"outbound_api_key": ...}``). Returns ``None`` (legacy default-credential
+    fallback, zero regression) when there is no custom key, a dependency is
+    missing, or resolution fails.
+
+    ``bot_query`` duck-types ``BotQueryProtocol.get_by_id_and_owner`` and
+    ``template_service`` duck-types ``TemplateConfigReader.get_template_config``;
+    typed as ``Any`` so the neutral composition root does not reverse-import
+    ``core/devices`` (see architecture review).
+    """
+    if template_service is None:
+        return None
+    bot = bot_query.get_by_id_and_owner(bot_id, owner_id)
+    if not bot:
+        return None
+    try:
+        template_config = template_service.get_template_config(bot_id)
+    except Exception as e:
+        logger.warning(
+            "[engines.resolve_outbound_rule_envelope] get_template_config "
+            "failed: bot_id=%s, error=%s", bot_id, e)
+        return None
+    try:
+        ctx, strategy = resolve_provisioning(
+            bot_id=bot_id,
+            owner_id=owner_id,
+            bot_type=bot.get("bot_type", ""),
+            active_engine=bot.get("active_engine"),
+            template_type=bot.get("template_type"),
+            template_config=template_config,
+        )
+        return strategy.build_extra_properties(
+            ctx,
+            secret_resolver=secret_resolver,
+            theta_master_key_secret=theta_master_key_secret,
+        )
+    except Exception as e:  # pragma: no cover - defensive: resolve_provisioning
+        # and each registered strategy's build_extra_properties never raise for
+        # any real bot (they fail-open internally); this only fires on a future
+        # broken engine/strategy or registry corruption. Fail-safe to None.
+        logger.warning(
+            "[engines.resolve_outbound_rule_envelope] resolve failed: "
+            "bot_id=%s, error=%s", bot_id, e)
+        return None
+
+
 __all__ = [
     "AICODING_ENGINE_TYPE",
     "BaasEngineBucketResolver",
@@ -326,4 +399,5 @@ __all__ = [
     "resolve_baas_engine_bucket",
     "resolve_default_capabilities_engine_bucket",
     "resolve_provisioning",
+    "resolve_outbound_rule_envelope",
 ]
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_device_header_updater.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_device_header_updater.py
@@ -12,7 +12,10 @@ from agentclaw.community.core.service_bot.services.deploy.provider_resolver impo
 from agentclaw.community.log import get_logger
 
 if TYPE_CHECKING:
+    from agentclaw.community.core.devices.protocols import BotQueryProtocol
+    from agentclaw.community.core.devices.services.baas_device_service import TemplateConfigReader
     from agentclaw.community.core.service_bot.services.baas_service import BaasService
+    from agentclaw.community.plugin_api.secret_resolver import SecretResolver
 
 
 logger = get_logger()
@@ -25,8 +28,20 @@ class BaasDeviceHeaderUpdateError(DeviceServiceError):
 class BaasDeviceHeaderUpdater:
     """Build and install outbound rules for one BaaS logical Bot."""
 
-    def __init__(self, baas_service: BaasService) -> None:
+    def __init__(
+        self,
+        baas_service: BaasService,
+        *,
+        bot_query: "BotQueryProtocol | None" = None,
+        template_service: "TemplateConfigReader | None" = None,
+        secret_resolver: "SecretResolver | None" = None,
+        theta_master_key_secret: str = "",
+    ) -> None:
         self._baas_service = baas_service
+        self._bot_query = bot_query
+        self._template_service = template_service
+        self._secret_resolver = secret_resolver
+        self._theta_master_key_secret = theta_master_key_secret
 
     def update(
         self,
@@ -59,11 +74,25 @@ class BaasDeviceHeaderUpdater:
                     agent_pass_token=agent_pass_token,
                 )
             else:
+                # 引擎解耦：复用创建链路同一套 provisioning 协议解析自定义 egress-key
+                # envelope，避免 bootstrap REPLACE 用部署默认凭证覆盖创建/重启写入的值。
+                from agentclaw.community.core.bot_management.engines import (
+                    resolve_outbound_rule_envelope,
+                )
+                extra_properties = resolve_outbound_rule_envelope(
+                    bot_id=bolt_id,
+                    owner_id=owner_id,
+                    bot_query=self._bot_query,
+                    template_service=self._template_service,
+                    secret_resolver=self._secret_resolver,
+                    theta_master_key_secret=self._theta_master_key_secret,
+                )
                 outbound_rule = self._baas_service._build_outbound_operation_rule(
                     bot_id=bolt_id,
                     owner_id=owner_id,
                     agent_pass_token=agent_pass_token,
                     agent_code=agent_code,
+                    extra_properties=extra_properties,
                 )
         except Exception as e:
             logger.error(
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_device_service.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_device_service.py
@@ -49,6 +49,7 @@ if TYPE_CHECKING:
     from agentclaw.community.core.devices.protocols import BotQueryProtocol, BotSyncProtocol, McpSyncProtocol
     from agentclaw.community.core.service_bot.services.baas_service import BaasService
     from agentclaw.community.core.task_queue.services.task_queue_service import TaskQueueService
+    from agentclaw.community.plugin_api.secret_resolver import SecretResolver
 
 logger = get_logger()
 
@@ -102,6 +103,8 @@ class BaasDeviceService(DeviceService):
         vault: "TokenVault | None" = None,
         task_queue_service: "TaskQueueService | None" = None,
         template_service: TemplateConfigReader | None = None,
+        secret_resolver: SecretResolver | None = None,
+        theta_master_key_secret: str = "",
     ):
         super().__init__(
             repository=repository,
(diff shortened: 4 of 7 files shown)


totalfrank (approved):
## Architecture-Gate Review

**Verdict:** APPROVE

**Gate 0:** Target branch is `dev` (the integration branch) — not a `REL*` branch, so no Isolation Audit applies. This is a normal bug fix landing on `dev`. Pass.

**Findings:** None.

**Analysis**

The fix wires `BaasDeviceHeaderUpdater.update`'s non-teclaw branch through the *existing* engine-provisioning contract instead of adding new branching:

- `BaasService._resolve_outbound_rule_envelope` (`baas_service.py:2743`) is a thin, engine-agnostic orchestration method: it looks up the bot, fetches `template_config` via the already-declared `TemplateConfigReader` Protocol (`baas_device_service.py:75`), then calls `resolve_provisioning(...)` → `strategy.build_extra_properties(...)` — the exact same call sequence already used by the bot-creation path (`baas_service.py:678-691`). No engine-name string checks are introduced anywhere in this diff.
- The theta-key decrypt/consume logic stays entirely inside `AicodingProvisioningStrategy.build_extra_properties` (`engines/aicoding/strategy.py:237`); `DefaultProvisioningStrategy.build_extra_properties` (`engines/default.py:22`) is the mandatory safe-default no-op, so unregistered/openclaw/teclaw engines fall back to `None` → unchanged legacy behavior, exactly as documented in the new method's docstring.
- Applying the "add engine `foo`" test: a new engine only needs its own strategy module + one registry entry in `engines/registry.py`; nothing in `BaasService` or the header updater would need to change. No under- or over-abstraction.
- `_build_outbound_operation_rule`'s `extra_properties` kwarg is pre-existing (already used by the create path), so this is not a contract/signature change requiring propagation analysis — it's wiring a previously-built parameter through a second call site.
- DI change in `service_bot_module.py` only *adds* a `template_service: TemplateService` parameter to the `baas_service` provider and forwards it into the constructor (which defaults the field to `None`); no existing registrations are reordered or altered, consistent with Rule 14 (composition-root-only wiring).
- `TemplateConfigReader` is an existing Protocol reused as-is (not a new ad hoc interface), and `TemplateService` already structurally implements it (`get_template_config`), so Rule 5 (contracts separate from implementations) holds.

No engine-string branching, no hardcoded capability sets, no engine-named modules under `core/`, no filesystem/env literals, no hidden wire protocol, and no transport/framework leakage were introduced.

**Out-of-scope notes** (not gated on): the two new `except Exception` catches in `_resolve_outbound_rule_envelope` are broad, and `bot` is fetched from `_bot_repo` here while `_build_outbound_operation_rule` → `bot_type_resolver=self._resolve_bot_type` will fetch it again downstream — both are business-logic/efficiency observations, not structural violations.

---

totalfrank:
## Verdict: APPROVE

## Gate 0 — Target branch

- **Target branch:** `dev` — the default integration branch. `^REL\d{8}` release-branch rules (existing-behavior-fix / isolated-feature classification, Isolation Audit) do not apply.
- Gate 0: **PASS** (trivial — not a release-branch target).

## Findings

None.

## Architecture review

**Files touched and role:**
- `core/devices/services/baas_device_header_updater.py` — core (engine-agnostic orchestration).
- `core/service_bot/services/baas_service.py` — core service.
- `di/modules/service_bot_module.py` — composition root.
- `tests/.../test_baas_device_service.py` — test.

**What the diff does:** `BaasDeviceHeaderUpdater.update()`'s bootstrap-rebuild path (non-teclaw branch) was calling `_build_outbound_operation_rule(...)` without `extra_properties`, so a REPLACE of the outbound rule during header bootstrap silently reverted any custom theta-key envelope set at create/restart time back to the default secret. The fix adds `BaasService._resolve_outbound_rule_envelope(bot_id, owner_id)` and threads its result through as `extra_properties`.

**Why this clears the catalog (verified, not assumed):**
- `_resolve_outbound_rule_envelope` does not branch on `active_engine`/engine-name literals anywhere (`grep` confirms no `engine_type ==`/`active_engine in` in either touched core file). It calls the pre-existing composition-root entry point `bot_management.engines.resolve_provisioning(...)` → `strategy.build_extra_properties(ctx, secret_resolver=..., theta_master_key_secret=...)` — this is the **exact same call chain** `BaasService._build_create_bot_payload` already uses at `baas_service.py:679-696` (unchanged by this diff) for the create path. This PR extends an already-correct engine-provisioning contract to a second call site instead of inventing a new one — passes the "add engine `foo`" test unchanged (still: strategy module + one registry entry).
- `EngineProvisioningStrategy.build_extra_properties`'s signature (`secret_resolver`, `theta_master_key_secret` kwargs) is untouched by this diff — confirmed via diff scope, this contract shape pre-exists on `dev`.
- `_build_outbound_operation_rule`'s `extra_properties: dict[str, Any] | None = None` parameter also pre-exists (used by the create path already); this diff only supplies a value at a second call site, no signature widening.
- `template_service: TemplateConfigReader | None = None` on `BaasService.__init__` reuses the existing `TemplateConfigReader` Protocol already defined in `core/devices/services/baas_device_service.py` (pre-existing, used there for the same "minimal template-read surface" purpose) — no new contract invented, default `None` is intentional (legacy/no-op fallback, documented in the docstring: unset dependency, no custom key, or resolution failure all fail safely to the pre-existing default-secret path — zero regression on unset injection).
- `di/modules/service_bot_module.py` only **adds** a `template_service: TemplateService` parameter and threads it into the `BaasService(...)` call — a pure composition-root addition, no reordering/modification of existing bindings. `TemplateService` was already imported in this module before the diff.
- No filesystem paths, no environment/mode detection, no hardcoded operational policy (schedules/timeouts/limits), no transport/framework leakage, and no plugin-boundary misuse are introduced.
- `template_config`/`extra_properties` are passed through opaquely (never subscripted/interpreted in the new code) — consistent with the existing opacity convention this codebase already guards elsewhere (`test_engine_ext_opacity.py`, for a sibling opaque-blob field).

**Tests:** the diff updates the existing mocked-`_build_outbound_operation_rule` assertion to expect `extra_properties=envelope` and adds `test_baas_provider_no_envelope_when_resolver_returns_none` covering the mandatory-default-fallback path (`None` → legacy behavior, zero regression). No new boundary is created, so no new architecture guard test is required.

## Out-of-scope notes
- The PR's boilerplate template sections (Problem/Solution/Validation/Compatibility) are left blank under the free-form Chinese summary at the top — a PR-hygiene/template-convention gap, not gated on here since it doesn't affect architectural placement.
- Business correctness of the bootstrap-REPLACE bug fix itself (whether this fully closes the reported key-overwrite defect end-to-end) was not re-litigated; this review is scoped to structural placement only.

---

totalfrank (dismissed):
**Verdict**: APPROVE

**Gate 0**: Target branch is `dev` (base sha a16918c0). This is the default integration branch, so Gate 0 passes trivially — full layering review applied, no Isolation Audit required.

**Findings**: None.

Reviewed files and roles:
- `src/backend/src/agentclaw/community/core/devices/services/baas_device_header_updater.py` (core) — adds a call to resolve an engine-agnostic `extra_properties` envelope and thread it into the existing `_build_outbound_operation_rule(..., extra_properties=...)` call. `_build_outbound_operation_rule` already accepted `extra_properties` pre-PR; this only wires the missing pass-through on the bootstrap/REPLACE path. No engine-name literals, no new branching added (the pre-existing `device.device_provider == TECLAW_DEVICE_PROVIDER` branch is untouched).
- `src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py` (core) — adds `_resolve_outbound_rule_envelope`, which is a faithful factor-out of the identical sequence already used on the bot-creation path (`resolve_provisioning(...)` → `EngineProvisioningStrategy.build_extra_properties(ctx, secret_resolver=..., theta_master_key_secret=...)`, verified at the creation call site). `BaasService` remains engine-agnostic: it never branches on engine name, only calls the single declared contract (`EngineProvisioningStrategy`, `core/bot_management/engines/provisioning.py`) resolved through the one composition-root registry (`core/bot_management/engines/registry.py`), which carries a mandatory default no-op strategy (`DefaultProvisioningStrategy`). Failure paths (`template_service is None`, bot not found, `get_template_config` exception, `resolve_provisioning`/`build_extra_properties` exception) all degrade to `None`, matching the documented "no custom key → fall back to default secret, zero regression" contract. Applying the "add engine `foo`" test: only `foo`'s own strategy module plus one line in `registry._build_default_registry()` would need to change — this file would not.
- `src/backend/src/agentclaw/community/di/modules/service_bot_module.py` (composition root) — adds `template_service: TemplateService` to the `baas_service` provider and forwards it to the constructor. Concrete-type wiring is correctly confined to the composition root; `BaasService` itself only depends on the `TemplateConfigReader` Protocol.
- `src/backend/tests/community/core/devices/services/test_baas_device_service.py` (test) — updates the existing pass-through test and adds a case for the `None`-envelope fallback path, matching the new behavior at the boundary this PR actually changes.

No Service API, Plugin API, config schema, or lifecycle contract was changed — `BaasService.__init__` gained one optional, backward-compatible parameter, and the new private helper reuses (does not modify) the existing `EngineProvisioningStrategy` contract, so no propagation analysis is required under Rule 16.

**Out-of-scope notes**: The two `except Exception` blocks in `_resolve_outbound_rule_envelope` swallow arbitrarily broad failures into a silent `None` fallback — reasonable given the documented "never regress below current behavior" intent, but worth a narrower catch if this code evolves further. Business-logic correctness of the theta-key fix itself was not evaluated.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0 — Target branch

- **Target branch:** `dev` — the default integration branch, not a `REL*` release branch. Gate 0's branch-policy machinery (existing-behavior fix vs. isolated-feature exception, Isolation Audit) does not apply; normal `dev` review rules apply.
- **PR classification:** bug fix — bootstrap/reconciliation of an outbound rule via `BaasDeviceHeaderUpdater.update` was overwriting a bot's custom theta-key egress credential with the deployment default because it never re-resolved `extra_properties` the way the create/restart path does. Minimal, targeted diff (5 files, +149/-8). No Gate 0 issue.

Gate 0 passing does not end the review — the full architecture/violation-catalog check still applies.

## Findings

### 1. Engine-agnostic orchestration logic is housed in and imported from a named-engine module, bypassing the codebase's own established contract-access pattern

**Files:**
- `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:385` — new function `resolve_outbound_rule_envelope(...)` defined at module scope in the `aicoding`-specific strategy file.
- `src/backend/src/agentclaw/community/core/devices/services/baas_device_header_updater.py:79-82`:
  ```python
  from agentclaw.community.core.bot_management.engines.aicoding.strategy import (
      resolve_outbound_rule_envelope,
  )
  extra_properties = resolve_outbound_rule_envelope(...)
  ```

`resolve_outbound_rule_envelope` is not aicoding-specific logic — it is generic bootstrap plumbing (look up the bot, look up its template config, dispatch via `resolve_provisioning(...)`, call `strategy.build_extra_properties(...)`, swallow failures to `None`). It works for whichever engine `resolve_provisioning` resolves to, exactly like every other call site in this codebase. But every one of those other call sites (`bot_service.py`, `template_service.py`, `utils.py`, `baas_device_service.py`, `baas_publish_task_handlers.py`, `device_service.py`, `expert_chat_session_runtime.py`, `baas_service.py` — 8 files, confirmed by repo-wide search) reaches the contract the same way: `from agentclaw.community.core.bot_management.engines import resolve_provisioning` (the neutral package), then calls `strategy.build_extra_properties(...)` inline in the caller's own file. This PR is the only call site that instead imports a wrapper *by name from the concrete `aicoding` submodule*.

The commit message itself confirms the motivation: "refactor(outbound): envelope 解析移入 aicoding 引擎，主链路零新增" (move envelope resolution into the aicoding engine, zero net-new on the main path) — i.e., the generic dispatch code was deliberately relocated into the engine-specific file specifically so the "main path" diff would look smaller. That inverts the intended boundary: the *dispatch/orchestration* code (bot lookup, template lookup, calling the neutral `resolve_provisioning`, building the envelope, exception handling) is exactly the kind of logic that belongs in the engine-agnostic layer; only `AicodingProvisioningStrategy.build_extra_properties` (the theta-key-specific decrypt logic, unchanged by this PR) should live under `engines/aicoding/`.

**Violated rules:**
- Violation catalog item 3 ("Engine-named modules/packages under `core/`... Engine-specific implementations belong in the engine's own area behind a neutral, domain-named contract") and item 10 ("Concrete implementations imported outside composition roots/tests... core reaching into... adapters internals") — here `core/devices/services` (a different core subsystem) reaches directly into a concretely-named engine implementation file instead of going through the neutral registry contract every other caller uses.
- `docs/arch/arch.rules.md` Rule 6 (Invariant — "Architectural Layers Constrain and Are Enforced": "dependency direction must be explicit and enforceable"; prevents "boundary collapse", "illegal cross-layer shortcuts").
- `docs/arch/arch.rules.md` Rule 8 (Policy — Directory Organization Matches Architectural Roles: "one path does not serve incompatible roles simultaneously"). `engines/aicoding/` is the declared role "aicoding engine implementation"; hosting a generic cross-engine dispatcher there makes the path serve two incompatible roles.
- `docs/arch/arch.rules.md` Rule 23 (Policy — Patterns Are Consistent and Cataloged: "deviations from known patterns are justified in review"). No justification is given for diverging from the `resolve_provisioning`-inline pattern used at all 8 other call sites.

**Compounding evidence — new cross-domain type coupling:** to support this, `aicoding/strategy.py` gains new `TYPE_CHECKING` imports of `agentclaw.community.core.devices.protocols.BotQueryProtocol` and `agentclaw.community.core.devices.services.baas_device_service.TemplateConfigReader` — an engine-implementation module now type-references another domain's (`devices`) service internals. This coupling disappears entirely if the orchestration function lives in the engine-agnostic caller (`baas_device_header_updater.py` or `core/bot_management/engines/registry.py`) instead, exactly as the 8 existing call sites already demonstrate.

**Why it breaks the architecture (the "add engine `foo`" test):** functionally, `resolve_outbound_rule_envelope` already works for any registered engine (it dispatches through `resolve_provisioning`), so no *behavioral* edit is needed to add engine `foo`. But structurally, any future maintainer extending or reasoning about bootstrap-envelope-preservation for a non-aicoding engine must import from a file named after a different engine, or duplicate this function under `engines/foo/strategy.py` (defeating "one contract, defined once"). It also creates a live landmine: PR #919 in this same repo ("retire engine switching") and general engine lifecycle churn mean `engines/aicoding/` is not guaranteed to be a permanent fixture — generic bootstrap plumbing must not be physically homed inside one specific engine's directory.

**Required fix:** move `resolve_outbound_rule_envelope` (or its logic, inlined) out of `engines/aicoding/strategy.py` into the engine-agnostic layer — either as a method/free function in `core/bot_management/engines/registry.py` (alongside `resolve_provisioning`) or directly inline in `BaasDeviceHeaderUpdater.update`, calling `resolve_provisioning(...)` + `strategy.build_extra_properties(...)` the same way every other of the 8 existing call sites does. Drop the `TYPE_CHECKING` imports of `devices.*` from `aicoding/strategy.py` once the function moves.

### 2. No direct test coverage for the relocated orchestration logic (secondary, supports Finding 1)

**File:** `src/backend/tests/community/core/devices/services/test_baas_device_service.py:186-235` — the two updated tests mock `resolve_outbound_rule_envelope` wholesale (`patch(...resolve_outbound_rule_envelope..., return_value=...)`) rather than exercising its actual bot-lookup / template-lookup / `resolve_provisioning`-dispatch / exception-swallowing logic. No test under `src/backend/tests/community/core/bot_management/engines/aicoding/` (or elsewhere) exercises `resolve_outbound_rule_envelope` directly (confirmed by repo search — the only test references are the two patches above).

This is a direct consequence of Finding 1: because the function is filed under an engine-specific test path by association, its actual logic — the part that is genuinely engine-agnostic — has no direct unit test; only the wrapper-call-through is tested. Once relocated per Finding 1, add direct tests (missing `template_service`, `bot_query` lookup miss, `template_service.get_template_config` raising, `resolve_provisioning`/`build_extra_properties` raising) alongside the other `resolve_provisioning`-based call sites' tests.

## Out-of-scope notes
- The underlying bug fix (bootstrap REPLACE was silently dropping the bot's custom egress credential) and its fail-open/fail-closed semantics are sound and out of scope for this review.
- DI wiring in `devices_module.py` (new `secret_resolver` / `secret_names` params on the existing `baas_device_service` provider) is additive-only and correctly composition-root-scoped; no issue.

---

totalfrank (approved):
**Verdict: APPROVE** (architecture gate only — product/business logic of the fix itself is out of scope for this review)

**Gate 0**: Target branch is `dev` (the default integration branch), not a `REL*` release branch. No Isolation Audit is required; the normal full architecture review applies. Pass.

**Review**

| File | Role |
|---|---|
| `core/bot_management/engines/__init__.py` | composition-root package exports |
| `core/bot_management/engines/registry.py` | composition root (dispatch by engine type) |
| `core/devices/services/baas_device_header_updater.py` | core domain service |
| `core/devices/services/baas_device_service.py` | core domain service |
| `di/modules/devices_module.py` | composition root (DI wiring) |
| `tests/.../test_resolve_outbound_rule_envelope.py`, `tests/.../test_baas_device_service.py` | tests |

Applying the "add engine `foo`" test to the new `resolve_outbound_rule_envelope`: it dispatches purely through the existing `resolve_provisioning` → `EngineProvisioningStrategy.build_extra_properties` contract (`core/bot_management/engines/provisioning.py`), with no `if engine_type == ...` branching anywhere in the diff. A new engine would only need its own strategy module plus a registry entry — no edits to this function.

The new call from `core/devices/services/baas_device_header_updater.py` into the `engines` composition root (`resolve_outbound_rule_envelope`) mirrors a pattern that already exists in `core/service_bot/services/baas_service.py` (which already does a local `from agentclaw.community.core.bot_management.engines import resolve_provisioning` + `strategy.build_extra_properties(...)` at its own creation-chain call site, lines ~676–697 on `dev`). This PR consolidates that fetch-bot → fetch-template → resolve-strategy → build-extra-properties sequence into one composition-root function instead of duplicating it a second time inline — a reduction in duplication against precedent, not a new violation.

`registry.py`'s new function types `bot_query` / `template_service` as `Any` rather than importing `BotQueryProtocol` / `TemplateConfigReader` from `core/devices`, with an explicit comment: "the engine composition root must not reverse-couple into device internals." That preserves the correct dependency direction (`core/devices` → `core/bot_management/engines`, never the reverse) at the cost of losing static typing on those two params; duck-typing is documented in the docstring. Acceptable trade-off, correctly reasoned.

`di/modules/devices_module.py` adds `secret_resolver: SecretResolver` (a Plugin API contract type) and `secret_names.aicoding_theta_master_key` as new provider dependencies for `baas_device_service`, wiring them only in the composition root — consistent with Rule 14. `aicoding_theta_master_key` is a pre-existing config key already used identically for `BaasService` in `di/modules/service_bot_module.py`; no new config key or schema change.

No Service API / Plugin API contract changed shape (`_build_outbound_operation_rule`'s `extra_properties: dict | None = None` param already existed; this PR is a new caller passing a computed value, not a signature change), so Rule 16 propagation analysis does not apply here. All new constructor parameters are optional with safe defaults, so existing callers (and DI wiring for engines other than the coding engines) are unaffected — zero regression is credible from the diff itself.

**Findings**: none.

**Out-of-scope notes**
- Whether `resolve_outbound_rule_envelope`'s broad `except Exception` fail-open behavior is the right tradeoff for a security-relevant credential-resolution path is a product/security judgment call, not an architecture-layering issue — not gated here.
- PR body's `Problem`/`Solution`/`Validation` template sections are empty (content lives above the template instead); worth filling in per `AGENTS.md` PR conventions, but not an architecture finding.

---

PR #1014, opened 2026-08-13
fix(expert-chat): keep empty session titles stable

evilcoda:
## Summary

- Use one shared default title for newly created expert-chat sessions and Backend placeholders.
- Create empty Adapter sessions with `新会话` so refreshing the session list does not change the title to `Chat with <Bot>`.
- Keep the existing Adapter label suffix contract and frontend suffix stripping unchanged.

## Root Cause

The create path persisted `Chat with <Bot>` while the immediate frontend state and Backend fallback used `新会话`. Once Adapter metadata became available after a refresh, the two title sources produced a visible title change.

## Impact

- Newly created sessions with no messages remain `新会话` before and after refresh.
- Existing session titles are not rewritten.
- No database, API shape, Adapter, or non-expert-chat behavior changes.

## Validation

- `uv run pytest tests/community/core/expert_chat/services/test_expert_chat_service.py -q` (`100 passed`)
- Ruff checks passed for all changed files.
- Changed-line coverage: `100.00% (1/1)`.


totalfrank (approved):
LGTM.

PR #1019, opened 2026-08-13
fix(bcs): align V1 session messages with legacy

vzvince:
## Problem

The OpenAPI V1 session-messages endpoint exposed a V1-specific message projection and pagination shape that differed from the legacy API. The additional `next_cursor` and `has_more` fields, string-based `before` cursor, and transformed message items prevented the frontend from migrating between the legacy and V1 endpoints without separate rendering logic.

## Solution

- Return the standard V1 envelope with `data` containing the legacy `GroupMessage[]` wire format unchanged.
- Remove the V1-only message page, cursor, sender-kind, and message-kind projections.
- Align the `before` query parameter with the legacy unsigned timestamp contract and omit `next_cursor` and `has_more`.
- Preserve V1 gateway-principal authentication, view-actor resolution, and session-participant authorization.
- Delegate Chat and ManagerWorker history reads to `GroupMessageHistoryService`, and StateMachine history reads to `CollaborationRuntimeService`.
- Keep the underlying message services, repositories, stores, database ports, and the production legacy route unchanged.
- Remove the direct `bcs-message` dependency from `bcs-app-session`; concrete history/runtime services remain selected and injected by the bootstrap composition root.
- Publish explicit OpenAPI and Gateway schemas for the legacy-compatible message shape. Fixed contract objects reject undocumented fields, while dynamic history and metadata maps continue to allow arbitrary properties.

## Validation

Validation completed before publication:

- `cargo test -p bcs-service-api --test v1_session_application_contracts`
- `cargo test -p bcs-app-session`
- `cargo test -p bcs-api-http --test session_routes`
- `cargo test -p bcs-http --test group_messages_contract`
- BCS OpenAPI contract suite: 42 passed
- Gateway served-OpenAPI suite: 8 passed
- Deterministic BCS OpenAPI dump matched `src/gateway/configs/schemas/bcn.openapi.json` byte-for-byte
- Independent code review reported no actionable Critical, Important, or Minor findings

The publish step intentionally used `--no-verify` as requested and did not rerun checks.

## Compatibility and risk

This is an intentional V1 response-contract correction for frontend compatibility. Consumers of the previous unreleased V1-only page/cursor shape must switch to the envelope's direct `GroupMessage[]` data field. Legacy API behavior and persistence interfaces are unchanged.

One unrelated bootstrap integration test can hang on its final legacy `/health` request. The same hang reproduces at the base commit after the V1 principal assertions have already passed, so it is not introduced by this change.


RaymondHX (approved):

cassiuscai (approved):

totalfrank (approved):
LGTM.

PR #1015, opened 2026-08-13
fix: add aicoding skills-local extra sync for claude code

panqiwp:
Add extra sync for .aicoding/workspace/skills/skills-local in claude_code build plan while preserving existing claude sync behavior.\n\nThis keeps the change minimal and limited to the claude_code workspace path handling.

PR #1024, opened 2026-08-13
fix: sync aicoding local skills in prepub builds

panqiwp:
## Summary\n- Add an aicoding workspace provider/build plan using .aicoding/workspace.\n- Route claude_code bots with non-normalCC template_type to the aicoding provider during build and config lookup.\n- Remove router-level rsync exclude hardcoding by reading defaults from the resolved provider.\n\n## Online issue fix\nThis fixes the online/prepub issue where aicoding local skills under .aicoding/workspace/skills/skills-local were not synchronized into prepub build artifacts.\n\n## Validation\n- uv run pytest tests/community/core/workspace/test_engine_sandbox_providers.py tests/community/core/service_bot/services/test_bot_build_service_rsync_excludes.py tests/community/core/bot_management/services/test_engine_resolver.py\n- uv run pytest tests/community/core/bot_management/services/test_engine_resolver.py tests/community/core/workspace/test_engine_sandbox_providers.py tests/community/core/service_bot/services/test_bot_build_service_rsync_excludes.py tests/community/core/service_bot/services/test_baas_service_engine_aware_mounts.py tests/community/core/service_bot/services/test_baas_service_read_only_rules.py\n- uv run pytest tests/community/endpoints/test_service_bot_rsync_excludes.py

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py
@@ -1,13 +1,13 @@
 """Engine type resolution helper.
 
 Frontend should not be required to know a bot's engine type — the backend
-resolves it from `bot_id` per the multi-engine API design
-(see docs/resource-management-api.md). API endpoints call this helper to
-turn (bot_id, owner_id, optional override) into a concrete engine type.
+resolves it from `bot_id`. API endpoints call this helper to turn
+(bot_id, owner_id, optional override) into the effective engine type used by
+workspace/resource routing.
 
 Resolution order:
   1. Explicit override (operator/debug use); empty string treated as no override.
-  2. Bot record's active_engine.
+  2. Bot record's active_engine, after applying registered engine routing policy.
   3. DEFAULT_ENGINE_TYPE.
 """
 from __future__ import annotations
@@ -15,6 +15,9 @@ from __future__ import annotations
 from typing import Optional
 
 from agentclaw.community.core.repository.protocols.bot import BotRepository
+from agentclaw.community.core.bot_management.engines.aicoding.strategy import (
+    AicodingProvisioningStrategy,
+)
 from agentclaw.community.core.workspace.constants import DEFAULT_ENGINE_TYPE
 from agentclaw.community.log import get_logger
 
@@ -68,13 +71,10 @@ def resolve_engine_for_bot(
     if bot:
         active = bot.get("active_engine")
         if active:
-            template_type = bot.get("template_type")
-            if (
-                active == "claude_code"
-                and template_type
-                and str(template_type).strip().lower() != "normalcc"
-            ):
-                return "aicoding"
-            return active
+            routed_engine = AicodingProvisioningStrategy.resolve_baas_engine_bucket(
+                active_engine=active,
+                template_type=bot.get("template_type"),
+            )
+            return routed_engine or active
 
     return DEFAULT_ENGINE_TYPE
--- a/src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py
@@ -15,8 +15,8 @@ from __future__ import annotations
 from typing import Optional
 
 from agentclaw.community.core.repository.protocols.bot import BotRepository
-from agentclaw.community.core.bot_management.engines.aicoding.strategy import (
-    AicodingProvisioningStrategy,
+from agentclaw.community.core.bot_management.engines.registry import (
+    resolve_bot_engine,
 )
 from agentclaw.community.core.workspace.constants import DEFAULT_ENGINE_TYPE
 from agentclaw.community.log import get_logger
@@ -71,10 +71,6 @@ def resolve_engine_for_bot(
     if bot:
         active = bot.get("active_engine")
         if active:
-            routed_engine = AicodingProvisioningStrategy.resolve_baas_engine_bucket(
-                active_engine=active,
-                template_type=bot.get("template_type"),
-            )
-            return routed_engine or active
+            return resolve_bot_engine(bot) or active
 
     return DEFAULT_ENGINE_TYPE
--- a/src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py
@@ -68,6 +68,13 @@ def resolve_engine_for_bot(
     if bot:
         active = bot.get("active_engine")
         if active:
+            template_type = bot.get("template_type")
+            if (
+                active == "claude_code"
+                and template_type
+                and str(template_type).strip().lower() != "normalcc"
+            ):
+                return "aicoding"
             return active
 
     return DEFAULT_ENGINE_TYPE
--- a/src/backend/src/agentclaw/community/adapters/http/service_bot/router_build.py
+++ b/src/backend/src/agentclaw/community/adapters/http/service_bot/router_build.py
@@ -301,6 +301,7 @@ async def get_bot_rsync_excludes(
     user: AuthenticatedUser = Depends(get_current_user),
     bot_service: BotServiceProtocol = Injected(BotServiceProtocol),
     bot_repo: BotRepository = Injected(BotRepository),
+    registry: EngineSandboxRegistry = Injected(EngineSandboxRegistry),
 ) -> ApiResponse:
     """获取Bot的rsync excludes配置信息。
 
@@ -345,16 +346,9 @@ async def get_bot_rsync_excludes(
         # 2. 解析引擎类型
         engine_type = resolve_engine_for_bot(bot_id, owner_id, bot_repo=bot_repo)
 
-        # 3. 获取引擎默认配置
-        from agentclaw.community.core.workspace.engines.openclaw import _OPENCLAW_RSYNC_EXCLUDES
-        from agentclaw.community.core.workspace.engines.claude_code import _CLAUDE_CODE_RSYNC_EXCLUDES
-
-        if engine_type == "openclaw":
-            default_excludes = list(_OPENCLAW_RSYNC_EXCLUDES)
-        elif engine_type == "claude_code":
-            default_excludes = list(_CLAUDE_CODE_RSYNC_EXCLUDES)
-        else:
-            default_excludes = []
+        # 3. 通过引擎 provider 获取默认配置，避免在路由层硬编码具体引擎。
+        provider = registry.resolve(engine_type)
+        default_excludes = list(provider.get_build_plan().rsync_excludes)
 
         # 4. 解析Bot自定义配置
         from agentclaw.community.core.workspace.engines import parse_build_rsync_excludes_from_ext
--- a/src/backend/src/agentclaw/community/configs/application-community.yaml
+++ b/src/backend/src/agentclaw/community/configs/application-community.yaml
@@ -21,6 +21,7 @@ user_config:
   workspace:
     openclaw_root: "./data/workspace/openclaw"
     claude_code_root: "./data/workspace/claude_code"
+    aicoding_root: "./data/workspace/aicoding"
 
   # Operator whitelist gating admin endpoints. Empty = no privileged operators
   # by default; a community deployment lists its own admin subject ids.
--- a/src/backend/src/agentclaw/community/configs/application-singlebox.yaml
+++ b/src/backend/src/agentclaw/community/configs/application-singlebox.yaml
@@ -89,6 +89,7 @@ user_config:
     env_folder: "aidesktop_singlebox"
     openclaw_root: "~/.openclaw"
     claude_code_root: "~/.claude_code"
+    aicoding_root: "~/.aicoding"
 
   openclaw:
     skills_repo_url: "git@code.teamclaw.com:security_release/aiworkbench.git"
--- a/src/backend/src/agentclaw/community/configs/application-test.yaml
+++ b/src/backend/src/agentclaw/community/configs/application-test.yaml
@@ -28,6 +28,7 @@ user_config:
   workspace:
     openclaw_root: "~/.openclaw"
     claude_code_root: "~/.claude_code"
+    aicoding_root: "~/.aicoding"
 
   # Yuque binding-verify endpoint — neutral stand-in; the suite mocks the HTTP call,
   # so only the config key's presence matters (the yuque verify endpoint reads it).
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -80,6 +80,16 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
     def engine_type(self) -> str:
         return self._engine_type
 
+    def resolve_bot_engine(self, bot: dict[str, Any]) -> str | None:
+        active_engine = bot.get("active_engine")
+        active_engine = active_engine if isinstance(active_engine, str) else None
+        if self.should_use_aicoding_runtime_engine(
+            active_engine=active_engine,
+            template_type=bot.get("template_type"),
+        ):
+            return AICODING_ENGINE_TYPE
+        return active_engine
+
     @staticmethod
     def normalize_engine_type(
         engine_type: str | None, *, default: str = "openclaw"
@@ -95,19 +105,13 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
         return template_type in CODING_TEMPLATE_TYPES
 
     @classmethod
-    def should_use_aicoding_baas_bucket(
+    def should_use_aicoding_runtime_engine(
         cls,
         *,
         active_engine: str | None,
         template_type: str | None,
     ) -> bool:
-        """Whether this context should select the aicoding BaaS bucket.
-
-        BaaS bucket routing is an image/runtime selection policy: all explicit
-        claude_code template-factory types except normalCC reuse the aicoding
-        BaaS template bucket. It only depends on active_engine + template_type;
-        caller/create routing may only have template_type available.
-        """
+        """Whether this bot should use the aicoding runtime engine."""
         if (
             cls.normalize_engine_type(active_engine, default="")
             != CLAUDE_CODE_ENGINE_TYPE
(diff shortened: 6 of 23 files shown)


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

### Gate 0 — Target branch

- **Target:** `REL20260814` (release branch).
- **Classification:** feature work extending an existing pipeline, not a minimal existing-behavior fix, and not eligible for the path (b) isolation exception.
- **Reasoning:**
  - The diff adds a brand-new engine sandbox provider (`core/workspace/engines/aicoding.py`, +213 lines), a new config key (`aicoding_root`, 4 files), and *wires it into the existing, shared `resolve_engine_for_bot` / `_resolve_sandbox_provider` pipeline* used across the system for engine selection. That pipeline is not "unreachable" — it is the general-purpose entry point other bots already flow through, so this fails Isolation Audit point 1 (net-new and unreachable).
  - Collateral changes are non-additive: `EngineBuildPlan.extra_sync_items` (a tuple field enabling multiple extra-sync targets, per its own in-repo comment "新增：支持多条额外同步目录") is **removed outright** in `engine_sandbox.py`, `claude_code.py`, and the loop logic in `bot_build_service.py::_migrate_bot_instance` is reverted to single-item sync. Removing an existing capability/field is not "additive, backward-compatible… new optional fields with safe defaults only" (Isolation Audit point 2).
  - `bot_build_service.py::_resolve_sandbox_provider` also drops its existing exception-fallback branch (the call to `resolve_engine_for_bot` before defaulting) in favor of unconditionally falling to `DEFAULT_ENGINE_TYPE` on any registry-resolve failure — a behavior change to an existing error path, for every engine, not just aicoding.
  - I could not find a `dev` counterpart for this exact change (new `aicoding.py` provider, the `engine_resolver.py`/`bot_build_service.py` routing edits, or the `extra_sync_items` removal) — `search_pull_requests` for related terms returns nothing on `dev`. Per the branch policy, a release-branch change with no `dev` counterpart is a finding on its own: it will be lost at the next release cut (regression by omission), and per path (b) the isolated feature must also exist on/be simultaneously PR'd to `dev`.

Given the above, Gate 0 fails outright: this should retarget to `dev` (the `router_build.py` de-hardcoding hunk could reasonably ride along if it's proven behavior-preserving, but the engine-rerouting + `extra_sync_items` removal should not go to a REL branch as-is).

### Findings

1. **`src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py:71-77`** and **`src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py:152-159`** — near-identical new blocks:
   ```python
   if (
       active == "claude_code"
       and template_type
       and str(template_type).strip().lower() != "normalcc"
   ):
       return "aicoding"
   ```
   This is engine-string + template-type branching duplicated inline in two `core/` domain services (arch.rules.md catalog item 1 — "Engine-string branching in core"; also item 9, copy-paste mirroring). I recognize this specific `claude_code + non-normalCC → aicoding` rule is an established convention already implemented separately in `conn_info_builders/baas_builder.py`, `core/aicoding/mcp_defaults.py`, and `expert_chat_service.py` (per #885/#889/#923/#959) — but that makes this worse, not better: the mapping now exists independently in five-plus places instead of one. Per the review brief, "new code that extends an existing violation is itself a violation." Required fix: a single domain-named resolver/contract (e.g. `EngineRoutingPolicy.resolve(active_engine, template_type)`) defined once in the engine-agnostic layer, with every one of these call sites (including the pre-existing ones, as follow-up) depending on it instead of re-deriving the rule.

2. **`src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py`** (field removal) + **`bot_build_service.py::_migrate_bot_instance`** — `EngineBuildPlan.extra_sync_items` is deleted and the multi-item extra-sync loop is reverted to single-item `extra_sync_source_relpath`/`extra_sync_target_relpath`. This is a non-additive contract change (arch.rules.md Rule 17 — Invariant, "stable surfaces are constrained… if changing something could cause a correctly written consumer to fail, treat it as constrained") shipped directly to a release branch, and it's exactly the mechanism that previously synced `.aicoding/workspace/skills/skills-local` for claude_code bots. Required fix: keep `extra_sync_items` additive, or if the field is genuinely being replaced by the new engine-routing approach, that's a `dev`-track redesign, not a REL-branch collateral edit.

3. **`bot_build_service.py::_resolve_sandbox_provider`** — the pre-existing exception-fallback (`resolve_engine_for_bot(...)` before defaulting to `DEFAULT_ENGINE_TYPE`) is deleted; any `registry.resolve()` exception now goes straight to the default engine for all bots, not just aicoding-routed ones. Undocumented behavior change to an existing error path on a release branch — no propagation/compat note in the PR description for this.

4. **`src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py:9-27`** — the file's own header comment states the rsync-exclude list, read-only rules, and workspace layout are copied from `OpenClawSandboxProvider` and explicitly "尚未按 aicoding 引擎的真实 VM 目录约定逐项核对" (not yet verified against aicoding's real VM directory conventions), pending a follow-up change. Shipping self-admittedly unverified sandbox isolation rules (these gate read-only enforcement and rsync excludes — i.e. what does/doesn't get copied into a build artifact and what's writable in the sandbox) to a release branch is a risk that should be resolved before merge, not deferred via a TODO on a REL branch.

### Out-of-scope notes

- The `router_build.py` change (replacing the openclaw/claude_code if-elif with `registry.resolve(engine_type).get_build_plan().rsync_excludes`) is architecturally the *right* direction (removes engine-string branching from an adapter) and I'd support it landing — just via `dev`, or backed by a `dev` PR reference and clear evidence it's behavior-preserving for the two existing engines.
- Not reviewing the aicoding local-skills business logic itself (assumed correct per instructions).

---

totalfrank (changes requested):
**Verdict**: REQUEST_CHANGES

This is an architecture-gate review scoped to structural soundness per `docs/arch/arch.rules.md` and `AGENTS.md`. It does not evaluate the aicoding feature/product logic.

## Gate 0 — Target branch

- **Target**: `REL20260814` (release branch).
- **Classification**: Feature extending an existing pipeline, mislabeled `fix:`. The diff adds a brand-new `AICodingSandboxProvider` (213 new lines), new config surface (`aicoding_root` across 3 YAML files + `WorkspaceConfig`), and — critically — reroutes engine resolution for **existing** `claude_code` bots whose `template_type` isn't `normalCC` (`resolve_engine_for_bot`, `_resolve_sandbox_provider`). Per the Gate 0 rule, a `fix:` title implementing new functionality is treated as a feature.
- **Isolation Audit (path b) result: FAIL at criterion 1.** The new provider is not net-new/unreachable — deleting `aicoding.py` would *not* leave existing behavior byte-identical, because `engine_resolver.py` and `bot_build_service.py` now actively redirect live `claude_code` bots (any bot with a non-normalCC `template_type`) onto the new provider during build and config lookup. That's an existing pipeline being extended, not an isolated new one.
- **No `dev` counterpart.** `origin/dev` has no `core/workspace/engines/aicoding.py` and no `aicoding_root` config key — this PR isn't a cherry-pick of anything already reviewed/merged on `dev`, and nothing in the description says so. Per the release-branch rule, that's a regression-by-omission risk at the next release cut.
- This PR also **reverts `extra_sync_items`**, a mechanism merged to this same release branch ~2 hours earlier via #1015 ("fix: add aicoding skills-local extra sync for claude code"), replacing it with a different architecture instead of building on it — more release-branch churn than the isolation exception is meant to tolerate.

**Gate 0: FAIL → REQUEST_CHANGES, retarget to `dev`** (a genuinely isolated remainder, if any, may still go through path (b), but this diff as a whole does not qualify).

## Findings (carry over to the re-targeted PR)

**1. (Invariant, Catalog #1 + #9) Engine-string branching duplicated across two `core/` services, reinventing an existing contract.**
- `src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py:71-77`
- `src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py:152-159`

Both hardcode the identical condition:
```python
if (active_engine == "claude_code" and template_type and str(template_type).strip().lower() != "normalcc"):
    ... = "aicoding"
```
This is engine-string branching inside engine-agnostic core services (violation catalog #1), duplicated verbatim in two files (#9) — the "add engine foo" test fails, since a routing-rule change now requires editing both sites in lockstep. Worse, this exact rule **already has a canonical implementation**: `AicodingProvisioningStrategy.should_use_aicoding_baas_bucket()` in `core/bot_management/engines/aicoding/strategy.py` (with named constants `CLAUDE_CODE_ENGINE_TYPE`, `AICODING_ENGINE_TYPE`, `NORMAL_CC_TEMPLATE_TYPE`, and `normalize_template_type`). The PR ignores that established pattern and re-derives the condition ad hoc with raw string literals in two unrelated files. **Required fix**: route both call sites through one shared resolver — extend/reuse `AicodingProvisioningStrategy`'s existing classmethod rather than duplicating the literal comparison.

**2. (Invariant, Rule 16) `EngineBuildPlan.extra_sync_items` contract field removed with no propagation analysis.**
- `src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py` (field deleted)

This field was introduced and consumed by #1015, merged to this same release branch shortly before this PR was opened. Removing it silently reverts that fix's multi-item sync support back to single-item; the PR description never mentions the field, its only consumer (`bot_build_service._migrate_bot_instance`), or why reverting is safe. A contract field removal on a release branch needs explicit propagation/compatibility analysis (Rule 16) — none is present.

**3. (Policy, Catalog #4) New engine-specific filesystem paths added to Backend config.**
- `di/config.py`, `configs/application-community.yaml`, `configs/application-singlebox.yaml`, `configs/application-test.yaml` (`aicoding_root` defaults, e.g. `/home/admin/.aicoding`, `~/.aicoding`)

`AGENTS.md`'s Skills Architecture section states "Do not add new engine-specific filesystem paths to Backend code. Physical layout ownership belongs to Engine Runtime and its versioned layout contract." This new default follows the same pattern that rule exists to prevent. Flagging for confirmation that Backend is the intended owner of this path rather than Engine Runtime.

## Out-of-scope notes
- Unrelated collateral cleanup rides along in `test_baas_service_engine_aware_mounts.py` / `test_baas_service_read_only_rules.py` (dropping `"id": 501` from fixtures, reordering `startup_script_reader` kwarg). Not an architecture finding on its own, but exactly the kind of "while we're here" change the Isolation Audit forbids on a release branch — worth splitting out regardless of target.
- The `router_build.py` change (reading rsync excludes via `registry.resolve(engine_type).get_build_plan()` instead of importing engine-private constants) is a genuine improvement and, taken alone, moves the router toward the correct contract pattern.


---

totalfrank (changes requested):
**Verdict: REQUEST_CHANGES**

This is an architecture-only gate review (`docs/arch/arch.rules.md` + `AGENTS.md`). Business logic (which bots should route to `aicoding`, rsync-exclude correctness, etc.) is out of scope and assumed correct.

## Gate 0 — target branch

- **Target:** `REL20260814` (a release branch).
- **Classification:** feature extending an existing pipeline, not an existing-behavior fix and not an isolated/dormant feature. The PR title says "fix," but the diff:
  - adds a brand-new engine (`aicoding`) with its own config keys, provider, and registration, **and**
  - rewires `resolve_engine_for_bot()` (`core/bot_management/services/engine_resolver.py`) and `BotBuildService._resolve_sandbox_provider()` (`core/service_bot/services/bot_build_service.py`) — both pre-existing, shared, currently-serving code paths used for every bot — to redirect any `claude_code` bot whose `template_type` isn't `normalCC` to the new engine.
  - removes an existing field (`EngineBuildPlan.extra_sync_items`) and the multi-item extra-sync loop it drove in `bot_build_service.py`, changing the currently-shipped `claude_code` migration behavior (it previously extra-synced two directories; now one).
- No reference to a corresponding `dev` PR/commit ("(#NNN)" or "cherry-pick") appears in the description, and the change is not minimal.
- **Isolation Audit (path (b), attempted): FAILS at point 1.** "Deleting the new files would leave every existing code path byte-identical" does not hold — `engine_resolver.py` and `bot_build_service.py` are pre-existing shared modules whose behavior changes for bots that already exist on this release branch (any `claude_code` bot with a non-`normalCC` template now silently resolves to a different engine and a different build/config path). This is a live behavior change riding on new-engine plumbing, not a dormant, unreachable addition. The exception does not apply.
- **Gate 0 verdict: FAIL — re-target to `dev`.** (Findings below carry over to the re-targeted PR per the review procedure.)

## Findings (most severe first)

**1. `src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py:71-77`** — Engine-string / engine×template_type branching in core (arch.rules.md Rule 14, Invariant; Rule 10).
```python
template_type = bot.get("template_type")
if (
    active == "claude_code"
    and template_type
    and str(template_type).strip().lower() != "normalcc"
):
    return "aicoding"
```
`resolve_engine_for_bot` is engine-agnostic domain code called from every endpoint that needs a bot's engine type. This is exactly the "engine×template_type truth table" the constitution names as a prohibited shape — engine selection is now hardwired with literal engine and template-type strings instead of being driven by configuration/registry data. Adding a future engine "foo" that also needs template-based routing means editing this file again. **Fix:** make the routing decision data (e.g. a field the bot/template record already carries, or something the composition-root-registered engine registry resolves) rather than a string comparison inline in the resolver; `resolve_engine_for_bot` should stay a pure lookup with zero engine-name literals.

**2. `src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py:150-158` (`_resolve_sandbox_provider`)** — Same violation, copy-pasted into a second core service (Rule 14; copy-paste cross-cutting logic, called out in "Structural Review Red Flags").
```python
active_engine = bot.get("active_engine") or DEFAULT_ENGINE_TYPE
template_type = bot.get("template_type")
if (
    active_engine == "claude_code"
    and template_type
    and str(template_type).strip().lower() != "normalcc"
):
    active_engine = "aicoding"
```
Byte-identical logic to Finding 1, duplicated instead of centralized — the "add engine foo" test now touches two core files instead of one composition-root entry. Notably, this codebase already carries two dedicated AST-based architecture guard tests for exactly this anti-pattern (`test_expert_chat_engine_branching.py`, `test_default_capabilities_engine_boundaries.py`), which shows the team has explicitly fenced this shape off elsewhere. This PR reintroduces the same shape in two modules those guards don't cover, and ships no equivalent guard for the new instances. **Fix:** same as Finding 1 — one contract-driven decision, computed/looked up once, not re-derived per call site.

**3. `src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py` (removed `extra_sync_items` field) + `core/workspace/engines/claude_code.py` + `bot_build_service.py::_migrate_bot_instance`** — Non-additive contract change to a currently-shipped code path on a release branch (Isolation Audit requirement: additive-only; Rule 16, Changes Propagate).
The diff deletes `EngineBuildPlan.extra_sync_items: tuple[tuple[str, str], ...]` and the loop in `_migrate_bot_instance` that consumed it, reverting `claude_code` bots from syncing two extra directories (`.claude` and `.aicoding/workspace/skills/skills-local`) back to syncing one (`.claude` only). Whatever the intent, this changes existing `claude_code` migration behavior on a release branch and removes a field from a shared dataclass contract — not the additive-only, backward-compatible shape release-branch collateral edits are required to have. No propagation/compatibility note is included. **Fix:** if this removal is intentional cleanup, it belongs on `dev` alongside the review that added `extra_sync_items` in the first place, not folded into a release-branch PR.

## Propagation check

`resolve_engine_for_bot`'s docstring says results follow "the multi-engine API design (see docs/resource-management-api.md)"; the PR does not update that doc or any spec to reflect that a `claude_code` bot's effective engine can now silently diverge from its stored `active_engine`. Per Rule 1/16 this is a missing propagation update for a behavior change to a documented contract.

## Out-of-scope notes

- Whether routing non-`normalCC` `claude_code` templates to `aicoding` is the right product behavior is a business-logic call and not reviewed here.
- The `router_build.py` change (replacing hardcoded per-engine import branching with `registry.resolve(engine_type)`) is a genuine architectural improvement and is not a finding.
- Test-file churn unrelated to this fix (e.g., dropping `"id": 501` from mock bot dicts, reordering `startup_script_reader` kwargs) is "while we're here" noise that shouldn't ride a release-branch PR, but it's test-only and not gated on here beyond noting it as further evidence this PR is not minimal.

---

panqiwp:
Thanks for the review. I pushed `abb53d58` to address the review items that are covered by the current PR direction:

- For the duplicated inline `claude_code + non-normalCC -> aicoding` checks, both call sites now delegate to the existing aicoding strategy contract via `AicodingProvisioningStrategy.resolve_baas_engine_bucket(...)` instead of re-deriving the raw string comparison in `engine_resolver.py` and `bot_build_service.py`.
- For `_resolve_sandbox_provider`, I restored the pre-existing fallback semantics: after the first `registry.resolve(...)` failure, it tries `resolve_engine_for_bot(bot_id, owner_id, bot_repo=...)` before falling back to `DEFAULT_ENGINE_TYPE`. The previous simplification that sent every registry failure directly to the default provider has been removed.
- For `aicoding.py`, I removed the release-branch TODO / “not yet verified” header comment so the provider file no longer ships with a deferred verification note.
- For the propagation/doc comment, `docs/resource-management-api.md` does not exist in this branch, so I removed that stale reference from the resolver docstring and clarified that the helper returns the effective engine used by workspace/resource routing.

One item is intentionally not marked as fixed by this commit: the `extra_sync_items` contract removal called out in the review is still present in the current PR diff. The current PR direction is to route the affected template-factory `claude_code` bots to the new `aicoding` workspace provider instead of keeping the previous `claude_code` extra-sync workaround. If the release-branch gate requires preserving `extra_sync_items` as an additive contract even with the new provider, I can make that follow-up change separately rather than claiming this commit resolves it.

src/backend/src/agentclaw/community/core/bot_management/services/engine_resolver.py
```diff
@@ -68,6 +71,10 @@ def resolve_engine_for_bot(
     if bot:
         active = bot.get("active_engine")
```
totalfrank, line 72: 抽象一个interface出来，这个interface的其中一个方法，take一个bot作为input，engine作为output。
然后非aicoding engine的实现就是：

return bot.get("active_engine")

aicoding引擎的按应该有的逻辑实现。
这样改一次，以后aicoding需要再改，都不需要再动engine_resolver的逻辑了。

totalfrank (commented):
PTAL

totalfrank-personal (approved):
LGTM.

totalfrank (approved):
LGTM.

PR #1034, opened 2026-08-13
feat(singlebox): stabilize merchant collaboration and hybrid agent startup

quechao76:
## Problem

The merchant collaboration demo was not consistently reaching the intended manager-worker and one-shot workflow:

- Private owner constraints could leak into platform-agent tasks.
- The manager could create incomplete groups, repeatedly dispatch workers, or fail to start the one-shot collaboration.
- Platform agents could return empty or overly structured responses.
- HumanInput timeouts could inherit global retry settings and remain in `RetryScheduled`.
- Model thinking behavior and the local BCS judge configuration were inconsistent across providers.
- The existing `merchant_hybrid` command was tied to one demo, required all profile arguments, and did not prepare a clean environment automatically.
- Hybrid shutdown could leave Claude-side services running when optional arguments were omitted.
- Backend cold starts could exceed the default ten-second readiness window.

## Solution

- Reworked the merchant-operation profiles around an explicit manager-worker topology:
  - keep owner constraints in a private manager ledger;
  - enforce pre-delegation privacy checks;
  - include marketing, data, and supply-chain workers from the first collaboration phase;
  - use concise natural-language business cards instead of JSON-heavy responses;
  - retry empty worker output once;
  - close manual dispatch after the first valid worker round;
  - dynamically build and run a bounded, acyclic one-shot SOP;
  - require professional validation before final HumanInput acceptance;
  - prevent premature completion, session termination, and unsupported execution claims.
- Added sourced merchant operating facts and platform-domain knowledge without preloading campaign-specific owner decisions.
- Fixed HumanInput timeout handling so persisted node-level retry limits remain authoritative instead of inheriting global bot retry defaults.
- Added `OPENCLAW_ENABLE_THINKING`, defaulting to `false`, with OpenClaw defaults and provider-specific compatibility parameters applied during model configuration.
- Refresh dynamic bot profiles when their generated model configuration becomes stale.
- Configure the local BCS judge only when the complete `OPENCLAW_OPENAI_*` environment is present, without copying plaintext API keys into generated TOML.
- Replaced the public `merchant_hybrid` group with a generic `hybrid` group:
  - OpenClaw-only mode requires only `--profile-dir`;
  - mixed mode accepts an optional paired Claude replacement profile;
  - `start hybrid` performs the required setup automatically;
  - runtime mode is persisted so `stop`, `status`, and `restart` restore the complete mixed topology;
  - incomplete or conflicting profile arguments fail early;
  - the legacy `merchant_hybrid` CLI alias remains available for compatibility but is omitted from help.
- Extended Backend readiness from 10 to 60 seconds for cold-start Skill catalog initialization.
- Ignored generated `.tshy` build directories and updated profile startup documentation.

## Validation

- `bash scripts/test_hybrid.sh`
- `bash scripts/test_singlebox_service_guards.sh`
- `bash scripts/test_singlebox_model_config.sh`
- `bash scripts/test_singlebox_bcs_runtime_config.sh`
- `cargo test --package bcs-collaboration-runtime`
  - 81 tests passed
- Shell syntax checks for the changed Singlebox scripts
- `git diff --check`
- Manually confirmed that incomplete hybrid profile arguments are rejected.
- Manually confirmed that Backend reaches `/api/health` after approximately 15 seconds with the extended readiness window.

## Compatibility and risk

- `merchant_hybrid` remains accepted as a deprecated compatibility alias.
- Existing OpenClaw-only usage can migrate to `hybrid --profile-dir <dir>` without Claude-specific arguments.
- Mixed mode still requires `--exclusive-profile-dir` and `--claude-profile-dir` together.
- `OPENCLAW_ENABLE_THINKING=true` restores the previous enabled-thinking behavior.
- Profile changes apply only to newly started Bot instances.
- No public API or persistence schema migration is introduced.


carolynli (dismissed):

carolynli (approved):
LGTM.

PR #1029, opened 2026-08-13
fix: sync aicoding local skills in prepub builds

panqiwp:
## Summary
- fix prepub build routing so template-factory claude_code bots use the aicoding workspace provider
- add aicoding sandbox provider build/sync config for local skills
- cover provider routing, fallback, and aicoding workspace behavior with tests

## Reason
Online issue fix: aicoding local skills were not synced into prepub build artifacts.

## Tests
- PYTHONPATH=src:. uv run pytest tests/community/core/bot_management/services/test_engine_resolver.py tests/community/core/bot_management/test_engine_provisioning_strategy.py tests/community/core/service_bot/services/test_bot_build_service_rsync_excludes.py -q
- changed-line coverage local check: 88.89% (104/117, required >= 80%)

PR #1028, opened 2026-08-13
fix(skills): retire local skill mappings through engine layout

liyaqing-l:
- 新 OpenAPI 的 Local Skill deactivate 不再通过 Backend 拼接物理路径并调用 /api/file/rmtree。
  - 改为发布 logical mapping retirement，由 Engine Runtime 根据引擎布局解析并清理实际软链。
  - Claude Code 同时兼容：
      - 当前 active root：.claude/skills
      - 历史 active root：.claude_code/workspace/skills

  - 仅清理指向受管 Skill source 的软链：
      - 外部软链保留
      - 普通文件或目录不会误删
      - 路径不存在按幂等成功处理

PR #1026, opened 2026-08-13
fix(engine): materialize DingTalk HTTP attachments

RaymondHX:
## Problem

DingTalk DM/PerSender file callbacks can produce short-lived HTTP download URLs in the pre-release environment. Engine currently rejects non-HTTPS temporary URLs, and the mounted `/api/openclaw/ws` route does not inject the existing `ResourceMaterializationService`, so accepted file messages terminate with `File preparation is unavailable` before inference.

## Solution

- Accept both HTTP and HTTPS temporary attachment URLs while retaining URL parsing, no-userinfo validation, public-IP DNS validation, DNS pinning, redirect blocking, download timeout, and size limits.
- Select the protocol-appropriate default port (80 for HTTP, 443 for HTTPS).
- Inject the existing `ResourceMaterializationService` into the OpenClaw WebSocket route and pass it to the shared WebSocket server.
- Add focused tests for HTTP materialization, invalid URL forms, and route dependency forwarding.

## Validation

- `uv run --directory src/engine --with pytest --with pytest-asyncio --with socksio pytest -q ...`: 37 passed.
- `src/engine/scripts/ci_test.sh --base origin/REL20260814 --head HEAD`: 2325 passed, 5 deselected; case pass rate 100%; line coverage 92.89%; changed-line coverage 100%.
- `scripts/ci/python_sast_local.sh src/engine 1 --base origin/REL20260814 --head HEAD`: passed.
- Pre-release ARCA container smoke: DingTalk temporary URL materialized into the controlled session-files layout, `<file-ref>` reached OpenClaw, the file was read successfully, and BCS emitted a final response.

## Compatibility and risk

This expands the accepted temporary URL schemes to HTTP for constrained deployment environments. Existing SSRF protections remain active: every resolved address must be public, the connection is pinned to a validated IP, redirects are disabled, and response size/time are bounded. No new service or plugin API is introduced. Rollback is a revert of this commit.

## Related issues

- Dev counterpart: #1042

PR #996, opened 2026-08-12
fix: resolve aicoding template mcp defaults

panqiwp:
## Summary
- read AICoding template MCP defaults from resolved template_config.bot_template_config.preset_capabilities.mcp
- remove legacy nested template_config lookup
- add coverage for the resolved upstream template_config payload

## Tests
- uv run pytest tests/community/core/mcp/test_defaults_per_engine.py -q

PR #1037, opened 2026-08-13
fix: sync aicoding local skills in prepub builds

panqiwp:
线上问题修复：aicoding 的本地 skills 没有同步到预发构建产物。\n\n修复内容：\n- 为 aicoding workspace provider 提供独立 build plan / rsync excludes / 只读规则配置。\n- claude_code 模板路由到 aicoding provider 时，构建源 NAS bucket 仍使用原始 active_engine=claude_code，source root 使用 .aicoding，避免拼成不存在的 aicoding NAS bucket。\n- 补充相关单测覆盖构建路由、fallback 和 NAS 源路径。\n\n验证：\n- PYTHONPATH=src:. uv run pytest tests/community/core/service_bot/services/test_bot_build_service_rsync_excludes.py tests/community/core/bot_management/services/test_engine_resolver.py tests/community/core/bot_management/test_engine_provisioning_strategy.py -q\n- 50 passed


totalfrank (changes requested):
**Verdict**: REQUEST_CHANGES

**Gate 0**: Target branch `REL20260814`. Classification: existing-behavior bug fix (path a). The diff proper (against `origin/REL20260814` merge-base) is 2 files / 57+/1- — `bot_build_service.py:249-256` plus a new unit test. It corrects a defect in the aicoding-workspace-provider routing that already shipped to `REL20260814` via #1024: when a `claude_code` bot is routed to the `aicoding` sandbox provider, `build_plan.engine_type` becomes `"aicoding"`, but the NAS source bucket for that bot was actually written under its original `active_engine` (`claude_code`). Passing `engine_type="aicoding"` into `get_bot_nas_dir` computes a source bucket that never existed, so the build reads no local skills. The fix (`source_nas_engine_type = bot.get("active_engine") or engine_type`) is minimal, targeted at the one call site causing the bug, and engine-name-agnostic — no new engine string literals, no catalog violation in the 3-line core change itself.

Gate 0 still **fails** on the dev-counterpart requirement. I checked out `origin/dev` and confirmed `BotBuildService.build()`'s Step 1 (`bot_build_service.py`) and `_resolve_sandbox_provider` are still on the pre-aicoding-provider shape — the whole "aicoding workspace provider" feature this fix patches (introduced by #1024 and its predecessors: #1015, #996, #966, #1003) was never landed on `dev`. That means this specific fix has nothing to be ported to: the bug it corrects doesn't exist on `dev` because the code path that causes it isn't there either. The underlying feature itself is REL-only and growing further from `dev` with each follow-up commit (this is the same defect flagged on the sibling PRs #1026 and #1028 currently open against this same release branch). Per branch policy, a release-branch fix must correspond to a commit already reviewed/merged on `dev`, or be simultaneously opened there; this PR's body and its single squashed commit carry no dev-PR/cherry-pick reference.

**Findings**:
1. `src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py:249-256` — No `dev` counterpart exists for this fix, because the feature it fixes (aicoding workspace provider routing, `_resolve_sandbox_provider` / `EngineBuildPlan` / provider-specific `source_root_name`) is itself REL-only and absent from `dev` (verified against `origin/dev` HEAD `2016722`). Violated rule: branch-policy requirement that a release-branch fix have, or simultaneously open, a `dev`-side counterpart. Why it breaks: at the next `dev`→REL release cut, both this fix and the feature it depends on are silently dropped — regression by omission, and the drift between `REL20260814` and `dev` compounds with every follow-up commit like this one. Required fix: open the aicoding-workspace-provider feature (or at minimum this bug's root cause and fix) against `dev`, or link an already-open/merged `dev` PR here, before this lands on the release branch.

**Out-of-scope notes**: The fix logic itself (source bucket keyed by original `active_engine`, not the routed provider's `engine_type`) is a sound, narrowly-targeted correction; not gating on the business-logic choice, only on the missing dev-side landing.

---

totalfrank (changes requested):
**Verdict**: REQUEST_CHANGES

**Gate 0**: Target branch `REL20260814`. Classification: existing-behavior bug fix (path a) — the diff proper is minimal (57+/1-, 2 files, on top of base `732cd33`) and fixes a concrete defect in `BotBuildService.build`: when a `claude_code`-templated bot is routed to the `aicoding` sandbox provider, the NAS *source* bucket must still be keyed by the bot's original `active_engine` ("claude_code"), not the provider's resolved `engine_type` ("aicoding") — otherwise the build reads from a NAS bucket name that was never written (`.../aicoding/...` instead of `.../claude_code/...`), and local skills silently fail to sync into the prepub artifact. The fix itself (`source_nas_engine_type = bot.get("active_engine") or engine_type`, `bot_build_service.py:252-257`) is targeted and doesn't add any new engine-string branching.

Gate 0 still **fails** on the dev-counterpart requirement, and this is the third same-day instance of the identical pattern (after #1026, #1028): the code this PR patches — the whole aicoding "workspace provider" abstraction (`EngineBuildPlan`, per-engine `get_build_plan()`, rsync-excludes/readonly config) landed on `REL20260814` via `#1024` (base commit `732cd33`, title identical to this PR's) — has **no presence on `dev` at all**. I fetched `origin/dev` and confirmed:
- `src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py` and `workspace/engines/__init__.py` do not exist on `dev`.
- `dev`'s `bot_build_service.py` still uses the pre-provider `extra_sync_source_relpath`/`extra_sync_target_relpath` shape and has no `source_root_name`, `get_build_plan()`, or `source_nas_engine_type` concept — it cannot even receive this fix in its current form.

So this fix corrects a bug in a feature that itself never went through `dev`. Per the branch policy, a genuine release-branch fix must correspond to a commit already reviewed/merged on `dev` (or be simultaneously PR'd there); since the underlying `#1024` feature was never ported, this PR compounds the regression-by-omission risk already flagged on #1026/#1028 — the next `dev`→REL merge will not just drop this fix, it will drop the whole aicoding-provider build path it depends on, likely reintroducing the sync bug (or worse) on the next cut.

**Findings**:
1. `src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py:252` (and the feature it patches, introduced by `#1024`) — No `dev` counterpart exists for the aicoding workspace-provider build path this fix depends on (verified: `workspace/engines/aicoding.py` absent from `origin/dev`, `bot_build_service.py` structurally divergent). Violated rule: branch-policy requirement that release-branch changes have a `dev`-side counterpart. Why it breaks: at the next release cut, `dev` overwrites `REL20260814` and both `#1024`'s aicoding build-plan feature and this fix disappear, silently reintroducing the "local skills not synced to prepub" defect. Required fix: port `#1024`'s aicoding workspace-provider feature (and this fix) to `dev` — either as a follow-up PR against `dev` or simultaneously — and reference it here before merge.

**Out-of-scope notes**: This is the third PR today surfacing the same systemic gap (aicoding-engine work landing only on `REL20260814`); worth raising as a standing process issue beyond this individual PR, since each instance will keep failing this gate until the underlying `dev` porting happens.

---

totalfrank (changes requested):
**Verdict**: REQUEST_CHANGES

**Gate 0**: Target branch is `REL20260814` (release branch).

Classification: **feature — wrong target** (also fails as a bare fix on the no-dev-counterpart rule even under the most charitable reading).

- This PR (#1037) is the latest in a same-day chain on the head branch `rel20260814-aicoding-provider-config`, all titled `fix: sync aicoding local skills in prepub builds`, all opened directly against `REL20260814`, none against `dev`:
  - #1024 — merged into `REL20260814` — "Add an aicoding workspace provider/build plan", "Route claude_code bots with non-normalCC template_type to the aicoding provider" (+579/-97, 24 files). This is new routing behavior and a new provider component, not a fix to pre-existing behavior — a `fix:` title does not make it one (per Gate 0 guidance).
  - #1029 — same branch, closed/superseded.
  - #1037 (this PR) — a follow-up correction on top of #1024's already-merged code, changing which `engine_type` is used to compute the NAS source bucket in `bot_build_service.build()`.
- I checked `origin/dev`'s current `bot_build_service.py`: it still contains the exact bug this PR fixes (`engine_type=engine_type` at the `get_bot_nas_dir` call, not `bot.get("active_engine")`), and a repo-wide PR search for "aicoding workspace provider" / this fix finds **no PR targeting `dev`** at any point in this chain (#1024, #1029, #1037, or the earlier commits `feat: add aicoding workspace provider config`, `fix: address aicoding provider review comments`, `test: cover aicoding provider build routing`).
- Consequence: this entire feature area (the aicoding-workspace-provider build routing, including the fix in this PR) exists **only** on the release branch. At the next release cut it is lost — the exact "regression by omission" scenario Gate 0 is designed to catch.
- Even treating #1037 in isolation as a narrow fix to existing (release-branch-only) code, the rule is explicit: *"a release-branch PR with no dev counterpart is a finding on its own."* No cherry-pick reference, no companion `dev` PR is cited in the body.
- If instead graded as path (b) (isolated new feature exception, since #1024 introduced a new provider under a `fix:` title), the Isolation Audit fails at step 1: the new provider is wired directly into `_resolve_sandbox_provider()` / `build()`, the pipeline every bot build already goes through — deleting the new files would change behavior for existing claude_code bots with non-normalCC templates, so it is not "net-new and unreachable." The exception is void.

Required fix direction: land the aicoding-workspace-provider build-routing change (and this NAS-source-bucket correction) on `dev` first — or open a companion `dev` PR in the same change — then this PR should either reference that `dev` PR/commit as a cherry-pick, or be re-targeted to `dev` outright.

**Findings**:

1. `Gate 0` — No `dev` counterpart for a `REL20260814`-only change chain (#1024, #1029, #1037) that introduces new build-routing behavior. Violated rule: Gate 0 branch policy ("a release-branch PR with no dev counterpart is a finding on its own — the change would be lost at the next release cut"). Concrete failure: `origin/dev`'s `bot_build_service.py` retains the bug this PR fixes, so the very next release cut from `dev` reintroduces the "aicoding local skills not synced" defect this PR claims to resolve. Fix: land on `dev` (or open the companion `dev` PR now) and reference it from this PR.

**Out-of-scope notes**:
- `src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py` already contains pre-existing engine-string branching (e.g. `engine_type == "openclaw"` at what is now line 315) and the repo already has several engine-named packages directly under `core/` (`core/aicoding`, `core/workspace/engines/aicoding.py`, `core/bot_management/engines/aicoding`, etc.). This PR's one-line change (`source_nas_engine_type = bot.get("active_engine") or engine_type`) reuses the existing `active_engine` fallback idiom already used elsewhere in this same file rather than introducing a new branching shape, so it is not flagged as a new catalog violation — but it is built on top of an already-non-compliant area and is noted here for visibility, not gated on.
- The provider selection itself (`_resolve_sandbox_provider` → `self._sandbox_registry.resolve(engine_type)`) correctly goes through a registry/composition-root lookup rather than inline branching, which is the right shape for this seam.

---

totalfrank (approved):
LGTM.

PR #1039, opened 2026-08-13
fix: exclude teclaw from skills pool edit guard

FreddieSun:
## Summary

- Skip SkillsPoolEditGuard for Teclaw: Teclaw does not participate in Skills Pool layout migration or rollback, so a long Teclaw upload must not block later Teclaw upload/delete requests.
- Keep the guard for file-based Pool engines and distinguish lock contention, rollback, and cache-unavailable errors.
- Add structured rejection logs for follow-up observability.

## Validation

- `uv run ruff check ...`
- Skills Pool guard/cache/local CRUD focused tests: 118 passed
- OpenAPI Skill upload + skill-set regression tests: 86 passed

## Compatibility

- Legacy and Pool file-engine edit serialization remains intact.
- Rollback acquisition behavior is unchanged.
- Teclaw is restored to its pre-Pool behavior and no longer depends on the generic layout lock.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/responses.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/responses.py
@@ -108,7 +108,10 @@ from agentclaw.community.core.resources.service import (
 from agentclaw.community.core.skill_center.errors import (
     LocalSkillDuplicateError,
     LocalSkillActiveError,
+    LocalSkillEditBusyError,
+    LocalSkillEditLockUnavailableError,
     LocalSkillEditPausedError,
+    LocalSkillLayoutRollbackError,
     LocalSkillInvalidPackageError,
     LocalSkillNotFoundError,
     LocalSkillNotReadyError,
@@ -268,6 +271,9 @@ ENVELOPE_ERRORS: dict[type[Exception], tuple[int, str]] = {
     LocalSkillTooLargeError: (413, "Skill package is too large"),
     LocalSkillStorageError: (502, "Skill storage operation failed"),
     LocalSkillRuntimeSyncError: (502, "Skill runtime synchronization failed"),
+    LocalSkillEditBusyError: (409, "Another Skill update is in progress"),
+    LocalSkillLayoutRollbackError: (409, "Skill layout rollback is in progress"),
+    LocalSkillEditLockUnavailableError: (503, "Skill update service is temporarily unavailable"),
     LocalSkillEditPausedError: (409, "Skill layout is being updated"),
     FileTooLargeError: (413, "File too large for preview"),
     # Startup script (issue #926): the body is refused at write time so a
--- a/src/backend/src/agentclaw/community/core/skill_center/errors.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/errors.py
@@ -53,5 +53,17 @@ class LocalSkillEditPausedError(Exception):
     """A Bot Skill layout operation currently owns the edit lock."""
 
 
+class LocalSkillEditBusyError(LocalSkillEditPausedError):
+    """Another Local Skill mutation is still in progress."""
+
+
+class LocalSkillLayoutRollbackError(LocalSkillEditPausedError):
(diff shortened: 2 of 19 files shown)


totalfrank (dismissed):

totalfrank (changes requested):
**Verdict**: REQUEST_CHANGES

**Gate 0**: Target branch `REL20260814`. Classification: existing-behavior bug fix (path a) — the diff modifies the existing `SkillsPoolEditGuard`/`CachePlugin` locking path (2 commits, 16 files, 481+/31-) to fix real defects: Teclaw uploads/deletes being serialized behind a Pool-only filesystem lock they don't need, cache-outage being mis-reported as ordinary lock contention, and rollback-in-progress being indistinguishable from busy-retry.

Gate 0 **fails** on the dev-counterpart requirement, same pattern already flagged on this branch in #1026: `git merge-base --is-ancestor e276c4a2 origin/dev` → not an ancestor, and none of `SkillsPoolEditBusyError`, `acquire_lock_strict`, or `_is_teclaw` (in `edit_guard.py`) exist on `origin/dev` (HEAD `81cb969`). Neither commit (`06901a5`, `e276c4a`) references a `dev` PR/issue. Per branch policy, this will be silently dropped at the next release cut.

**Findings**:

1. `src/backend/src/agentclaw/community/core/skills_pool/edit_guard.py:82` and `:171-176` — new `_is_teclaw()` helper does `bot.get("active_engine") == "teclaw"` and `acquire_for_edit()` branches control flow on it (`if self._is_teclaw(scope): return SkillsPoolEditLease(key="", token=None)`), bypassing the entire Pool edit-lock mechanism for one named engine. This is engine-string branching added directly to `core/` domain logic (constitution catalog item 1 — Invariant). It's a new instance of a pattern already present elsewhere in `core/` (`channel_service.py::_is_teclaw_bot`, `local_skill_delete_service.py::_is_teclaw`, `local_skill_upload_service.py::_is_teclaw`), but precedent doesn't waive the rule here — new code entrenching an existing violation is itself a violation, and this PR adds a *fourth* copy of the same inline engine check, this time controlling a shared locking primitive rather than a single call site.
   Required fix: introduce (or extend an existing) domain-named contract, e.g. `SkillLayoutParticipation`, defined once in the engine-agnostic layer, with a mandatory default implementation (`participates_in_pool_layout = True`, i.e. conservative — apply the lock) and Teclaw's own implementation returning `False`, wired by engine type in a composition root. `SkillsPoolEditGuard` should depend on that contract instead of reading `bot.get("active_engine")` and comparing to a literal string.

2. `src/backend/src/agentclaw/community/core/skills_pool/edit_guard.py:178-186` (`_log_rejection`) — same underlying issue at lower severity: reads `active_engine` directly off the bot record purely to log it. Once finding 1 is resolved via a contract, prefer logging the contract's resolved participation/engine label rather than re-reading the raw engine string in `core/`.

**Out-of-scope notes**: The lock-outage vs. contention vs. rollback distinction (`CacheLockInfrastructureError`, `SkillsPoolEditRollbackError`, `SkillsPoolEditBusyError`) and the `Plugin API` (`CachePlugin.acquire_lock_strict`) extension are correctly placed — cache is a genuine external-system boundary — and not reviewed for business correctness here.

---

totalfrank (dismissed):

FreddieSun:
@totalfrank Thanks — addressed both requested changes.

1. The shared `SkillsPoolEditGuard` no longer reads `active_engine` or branches on Teclaw. It now depends only on the named `SkillLayoutParticipationResolver` contract. The DI composition root supplies the conservative default (`participates=True`) and registers Teclaw's artifact-layout opt-out there. Rejection logging now records the resolved participation label rather than re-reading the Bot.

2. Added the dev counterpart: #1046.

Coverage added for the explicit policy, default/fail-closed behavior for missing/unknown/cross-env Bots, composition wiring, the non-participant lock bypass, and ordinary Pool-participant serialization.

Validation:
- focused lint + tests: 72 passed
- REL full backend CI: 11,720 passed; changed-line coverage 94.67% (142/150).

totalfrank (approved):

PR #1044, opened 2026-08-14
fix(devices): serialize BaaS container initialization

liyaqing-l:
- 等待镜像初始化完成后再执行 Backend bootstrap，避免两个初始化流程并发操作。
  - install_engine.sh、Supervisor 同步服务安装等步骤改为同步执行。
  - 启动业务服务前，明确等待 supervisorctl 可用。
  - 必要步骤返回非零退出码时立即失败，不再继续启动后续服务。
  
  
 解决的问题：初始化脚本并发执行，导致 start_service 在 Supervisor 未就绪时启动，出现 /tmp/supervisor.sock 不存在，最终容器启动失败。

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_container_init.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_container_init.py
@@ -29,7 +29,7 @@ class BaasContainerInitializer:
         admins: list[str] | None,
         codefuse_token: str | None = None,
     ) -> None:
-        """Execute the 6-step init sequence inside the BaaS container."""
+        """Execute the ordered init sequence inside the BaaS container."""
         entity_id = device.device_props.get("entity_id")
         entity_type = device.device_props.get("entity_type")
         client_id = device.device_props.get("client_id", "")
@@ -41,6 +41,7 @@ class BaasContainerInitializer:
         self._ensure_baas_engine_dirs(bot_uuid, engine)
         self._create_baas_skill_symlink_conf(bot_uuid, symbol)
         self._write_codefuse_token(bot_uuid, codefuse_token)
+        self._wait_for_baas_supervisor(bot_uuid)
         self._start_baas_sandbox_service(
             bot_uuid=bot_uuid,
             client_id=client_id,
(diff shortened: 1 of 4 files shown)


totalfrank (changes requested):
**Architecture-Gate Review**

**Verdict**: REQUEST_CHANGES

**Gate 0**: target branch `REL20260814` (release branch, matches `^REL\d{8}`).

Classification: **existing-behavior fix without a `dev` counterpart.** The diff changes existing, already-shipped code paths in `BaasContainerInitializer` (`_run_baas_bootstrap`, `_run_baas_install_engine`, `_run_baas_setup_sync_service`, `_ensure_baas_engine_dirs`, `_start_baas_sandbox_service`) — this is genuinely a fix to existing behavior, not new surface, so it is not routed through the path (b) isolated-feature exception.

However, a genuine release-branch fix must correspond to a commit already reviewed and merged on `dev` (cherry-pick), and this one doesn't:
- No PR targeting `dev` exists for this change (checked `search_pull_requests` for the author and for this content — only this REL-targeted PR exists).
- The most recent history on `dev` for `src/backend/src/agentclaw/community/core/devices/services/baas_container_init.py` is: `fix: serialize BaaS bootstrap with container init (#519)` merged 2026-07-28, then **reverted** the same day (`Revert "fix: serialize BaaS bootstrap with container init (#519)"`, commit `77c52f1b`). `dev` currently sits at the pre-#519 (fire-and-forget `nohup ... &`) implementation — matching the "before" side of this diff.
- This PR ships a *different* re-implementation of that same fix (synchronous exec + `_exec_checked` + a new `_wait_for_baas_supervisor` step) directly onto `REL20260814`, with no corresponding `dev` PR/commit reference in the description.

**Result**: Gate 0 fails. A release-branch PR with no `dev` counterpart is a finding on its own — the fix is lost at the next release cut (regression by omission), and worse, it re-attempts a fix whose only prior version was merged-then-reverted from `dev`, skipping the review/testing loop that caught that regression. Per the review contract this is not waivable ("Gate 0 is not waivable: pipeline-touching / non-cherry-picked features never merge to release branches").

**Findings**

1. Gate 0 — no `dev` counterpart for a release-branch behavior change.
   - Rule: release-branch routing policy (Gate 0), not a numbered `arch.rules.md` rule.
   - Why it breaks: `REL20260814` will ship this fix, but the next release cut from `dev` will silently regress back to the fire-and-forget init sequence that causes the `/tmp/supervisor.sock` failure this PR is fixing, since `dev` never received it.
   - Required fix: land the equivalent fix on `dev` first (this time keep the simpler synchronous/`_exec_checked` approach rather than the reverted PID-lock approach from #519, or supersede #519 there), then retarget this PR as an explicit cherry-pick referencing that `dev` PR/commit — or close this PR and let the fix flow to the release branch through the normal `dev` → release-cut path.

No other findings in categories 1–11: the change stays within one existing `core/devices/services` module, doesn't add engine-string branching (the `engine` value is forwarded opaquely to scripts, unchanged from prior code), doesn't touch config/DI/composition roots, and doesn't add wire contracts or cross-component surface.

**Out-of-scope notes**: The shell-based polling/timeout logic and the choice to make init steps synchronous vs. fire-and-forget are business/operational-correctness concerns for the container init sequence, not gated here.

---

totalfrank (changes requested):
**Verdict:** REQUEST_CHANGES

**Gate 0:** Target branch is `REL20260814` — a release branch. Classification: **existing-behavior bug fix (path a)** — the diff only reorders/hardens existing init steps in `baas_container_init.py` (async→sync execution, added `_wait_for_baas_supervisor`, added `_exec_checked` failure propagation) and does not add any new public surface, route, or pipeline. Structurally this qualifies for path (a).

However, Gate 0 requires that a genuine release-branch fix "correspond to a commit already reviewed and merged on `dev`... a release-branch PR with no `dev` counterpart is a finding on its own — the change would be lost at the next release cut (regression by omission)." That check fails here:

- `git diff origin/dev origin/REL20260814 -- src/backend/src/agentclaw/community/core/devices/services/baas_container_init.py` is **empty** — the file is byte-identical between `dev` and `REL20260814`. This confirms the race condition being fixed (backgrounded `nohup … &` init steps racing Supervisor/`install_engine` readiness, causing `/tmp/supervisor.sock` to be missing at `start_service` time) exists on `dev` today, unfixed.
- The PR description contains no reference to a `dev` PR/commit (no `(#N)`, no "cherry-pick"), and a repo-wide search for a matching `dev` fix (`baas_container_init`, "serialize supervisor") returns no results.

Merging only to `REL20260814` leaves `dev` broken and this fix will be silently dropped at the next release cut off `dev`.

**Findings**

1. **Gate 0 — missing `dev` counterpart.** `src/backend/src/agentclaw/community/core/devices/services/baas_container_init.py` — no corresponding fix exists on, or is linked from, `dev`. Rule: Gate 0 release-branch fix requirement (regression-by-omission). Required fix: open a companion PR with the identical fix against `dev` (or, if one already landed there under a different PR, link it explicitly in this PR's description) before/alongside merging this to `REL20260814`.

2. **Rule 7 — hardcoded operational policy (minor, pre-existing pattern extended).** The new/changed calls hardcode timeout literals directly in a `core/` service, e.g. `timeout_seconds=300` (bootstrap, install_engine), `120` (setup_supervisor_sync_service, setup_engine_dirs), `70` (wait_for_baas_supervisor) in `baas_container_init.py`. This extends the file's pre-existing pattern of literal timeouts (previously `60`/`30`) rather than sourcing them from configuration. Not blocking on its own given the pre-existing pattern, but flagging per the constitution's "timeouts... written as literals in core services... are configuration, not code."

**Out-of-scope notes**
- `baas_container_init.py` executes raw shell/`exec_command_on_bot` calls against a container sandbox from within `core/`, which is arguably infra/system-boundary territory (Plugin API candidate) rather than domain logic — but this placement predates the diff and isn't introduced or extended in shape by this PR, so it's noted only for awareness, not gated on here.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0 — Target branch

- **Target branch:** `REL20260814` — matches `^REL\d{8}`, a release branch.
- **PR classification:** existing-behavior bug fix. The diff changes the internal execution order/timing of the existing 6-step BaaS container init sequence in `BaasContainerInitializer.run` (bootstrap → install_engine → setup_sync_service → ensure_engine_dirs → skill-symlink conf → start_service), converting fire-and-forget `nohup ... &` dispatches into synchronous, checked calls and adding a supervisor-readiness wait. No new routes, DTO fields, config keys, or capabilities are added — squarely path (a) shape, not a disguised feature.
- **Dev counterpart check (verified, not taken on the PR's word):** `src/backend/src/agentclaw/community/core/devices/services/baas_container_init.py` is byte-identical between `origin/dev` (tip `c7f7dd1`) and `origin/REL20260814`'s base (`336283b5`, an ancestor of the branch's current tip `c87cfb4`). Neither branch has any of this PR's new methods (`_exec_checked`, `_wait_for_baas_supervisor`, the `.install_dependency_file` readiness gate). The PR body (describing the race condition and fix) references no `dev` PR/commit number and does not claim "cherry-pick." A repo-wide PR search for this content and `git log` on the touched file on both branches surface nothing else.
- **Verdict on Gate 0: FAIL — no `dev` counterpart.** Per the branch policy, "a release-branch PR with no `dev` counterpart is a finding on its own — the change would be lost at the next release cut (regression by omission)." The exact race this PR fixes (bootstrap/install steps racing supervisor readiness) remains live on `dev` today; the next release cut from `dev` reintroduces the container-init race this PR closes on `REL20260814`.

**Required fix:** land the identical fix on `dev` (its own PR, or one opened in parallel), then update this PR's description to reference it as a cherry-pick (e.g. "(#NNN)").

## Findings

### 1. (Blocking, Gate 0) No `dev` counterpart for a release-branch fix
See Gate 0 section above. This is the primary blocking finding; the verdict is REQUEST_CHANGES regardless of the diff's own quality.

### 2. New magic timeout/retry-count literals extend an existing anti-pattern
**File:** `src/backend/src/agentclaw/community/core/devices/services/baas_container_init.py`
- `_run_baas_bootstrap`: `seq 1 120` (120s poll bound) + `timeout_seconds=300`
- `_run_baas_install_engine` / `_run_baas_setup_sync_service` / `_ensure_baas_engine_dirs`: `timeout_seconds=300` / `120` / `120`
- `_wait_for_baas_supervisor` (new): `seq 1 60` (60s poll bound) + `timeout_seconds=70`

**Violated rule:** violation catalog item 7 ("timeouts, retry counts, magic limits... written as literals in core services... These are configuration, not code").

**Why it matters:** the file already hardcoded `timeout_seconds` literals before this PR (60, 30, 30); this diff roughly triples the literal count and raises magnitudes up to 300s, with no config surface. Tuning bootstrap/readiness timeouts for a slower environment (or tightening them for faster failure detection) requires editing and redeploying `core/` rather than changing a setting.

**Required fix:** not independently blocking, but since every call site is already being touched to add `_exec_checked`, thread these values through a small config object (or this codebase's existing settings/DI mechanism) injected into `BaasContainerInitializer.__init__` instead of adding more inline constants.

## Out-of-scope notes
- The synchronization fix itself is architecturally sound: it stays within the existing `core/devices/services` orchestration layer, uses only the already-established `self._baas_service.exec_command_on_bot` SPI call (no new transport/contract surface), takes `engine` as an opaque passthrough parameter with no engine-name branching, and the new test file is well-scoped to the changed behavior.
- Whether `_ensure_baas_engine_dirs` should stay non-fatal while the other four steps are now fail-fast is a reliability trade-off, not gated here.

---

PR #1045, opened 2026-08-14
fix(singlebox): align merchant hybrid bot profiles

yhzhang35:
## Problem

`merchant_hybrid` required a long profile-configuration command for its normal local startup, while the local Claude Provider Bot registration appended a display-only `（当前）` suffix instead of preserving the profile name.

## Solution

- Default the merchant operations OpenClaw profile, `platform-data` exclusion, and Claude profile for `setup`, `start`, and `restart merchant_hybrid`.
- Preserve explicitly supplied profile values and keep `stop`/`status` on their saved runtime-state path.
- Register the Provider Bot using its profile display name and add low-sensitivity lifecycle diagnostics.
- Add shell regressions for the registration payload and default, partial override, full override, and status behaviors.

## Validation

- `bash scripts/test_hybrid.sh` — passed.
- `bash -n scripts/singlebox.sh scripts/test_hybrid.sh scripts/modules/bcs_baas_provider.sh` — passed.
- `./scripts/singlebox.sh --help | rg -F 'start merchant_hybrid'` — passed.

## Compatibility and risk

The defaults apply only when the requested target is exactly `merchant_hybrid`; other hybrid invocations are unchanged. Existing Provider registrations receive the profile name on their next controlled registration or restart; this PR does not restart local services.

### DIFF
--- a/scripts/modules/bcs_baas_provider.sh
+++ b/scripts/modules/bcs_baas_provider.sh
@@ -146,7 +146,7 @@ bcs_baas_provider_register() {
     local role bot_id name provider_ref payload bot_response runtime_token bot_uuid visibility state_bots='[]'
     while IFS=$'\t' read -r role bot_id name; do
         provider_ref="${bot_id}:${entity_id}"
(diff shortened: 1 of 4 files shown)


cassiuscai (dismissed):

totalfrank (dismissed):
LGTM

totalfrank (approved):

PR #1049, opened 2026-08-14
fix(singlebox): align hybrid runtimes with local model config

quechao76:
## Problem

Hybrid startup hardcoded `Kimi-K2.6` for OpenClaw, Claude Code, and BCSFuse. This overrode the model selected through Singlebox configuration and caused `404 Model not exist` errors when manual mode configured another model, such as `glm-5.2` from `.env.local`.

Hybrid startup also required an explicit merchant profile path even for the default OpenClaw-only setup.

## Solution

- Use the primary model resolved by Singlebox as the hybrid runtime model.
- Preserve the generated OpenClaw model configuration instead of rewriting it.
- Resolve the Claude Code relay model dynamically rather than storing a fixed model in its profile.
- Apply the same resolved model to BCSFuse SOP model roles.
- Fail early when the selected model cannot be resolved.
- Default `start hybrid` to the merchant operations OpenClaw profile.
- Preserve `merchant_hybrid` as a deprecated compatibility mode for the mixed OpenClaw and Claude Code setup.
- Update profile documentation and hybrid regression coverage.

In manual mode, OpenClaw, Claude Code, and BCSFuse now use the model selected through `.env.local`.

## Validation

- `env LC_ALL=C LC_CTYPE=C LANG=C bash scripts/test_hybrid.sh`
- `env LC_ALL=C LC_CTYPE=C LANG=C bash scripts/test_singlebox_model_config.sh`
- `env LC_ALL=C LC_CTYPE=C LANG=C bash scripts/test_singlebox_bcs_runtime_config.sh`
- Shell syntax validation for the modified scripts
- `git diff --check`

All checks passed.

## Compatibility and risk

The legacy `merchant_hybrid` command remains available and retains its mixed-profile defaults. Existing explicit hybrid profile options continue to work.

The main behavior change is that hybrid startup no longer forces `Kimi-K2.6`; it now honors the model selected by the active Singlebox configuration.

### DIFF
--- a/scripts/4bots_merchant_operations_profile/README.md
+++ b/scripts/4bots_merchant_operations_profile/README.md
@@ -104,16 +104,14 @@ Profile 不预置本次活动的目标值、店主预算、授权阈值、异常
 ./scripts/singlebox.sh install-tools
 ```
 
-然后直接启动；`start hybrid` 会自动完成项目 setup，只传 profile 时全部 Bot 使用 OpenClaw：
+然后直接启动；`start hybrid` 会自动完成项目 setup，并默认加载本目录的全部 OpenClaw Bot：
 
 ```bash
-./scripts/singlebox.sh start hybrid \
-  --profile-dir scripts/4bots_merchant_operations_profile
+./scripts/singlebox.sh start hybrid
 ```
 
 停止服务：
 
 ```bash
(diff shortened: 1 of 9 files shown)


totalfrank (approved):
LGTM.

PR #1067, opened 2026-08-14
fix(singlebox): improve toolchain and GLM runtime defaults

quechao76:
## Problem

Singlebox had several local runtime setup issues:

- Tool installers such as nvm, uv, and rustup may update shell startup files, but `install-tools` cannot refresh its parent shell. Running `start` immediately afterward could therefore fail to find newly installed tools.
- The default Rust toolchain still targeted 1.91.0 instead of 1.96.0.
- Disabling OpenClaw thinking only injected `enable_thinking=false` for GLM models behind a specific gateway URL. OpenAI-compatible proxy gateways could still return hidden `reasoning_content`, despite OpenClaw reporting thinking as disabled.

## Solution

- Detect the user's login shell through `$SHELL` and track installers that update the shell environment.
- Print a single actionable `source` command after installation when the invoking shell needs to be reloaded.
- Update the default Rust toolchain to 1.96.0 while preserving the existing environment override.
- Apply `params.extra_body.enable_thinking` to supported GLM models using OpenAI-compatible APIs regardless of gateway hostname.
- Add regression coverage for bash/zsh profile detection and GLM configuration in both manual and home modes.

## Validation

- `bash scripts/test_singlebox_toolchain.sh`
- `bash scripts/test_singlebox_default_mode.sh`
- `bash scripts/test_singlebox_bcs_prereqs.sh`
- `bash scripts/test_singlebox_model_config.sh`
- Bash syntax checks for the modified scripts
- `git diff --check`

All checks passed.

A complete interactive `install-tools` run was not performed because it downloads and installs host-level toolchains.

## Compatibility and risk

- `REQUIRED_RUST_TOOLCHAIN` can still override the default Rust version.
- The explicit thinking parameter is limited to matching GLM model families using `openai` or `openai-completions`.
- Existing explicit model parameters remain preserved.
- Generated OpenClaw runtime configuration must be regenerated by restarting the bots.


carolynli:
@codex 解决冲突

carolynli (dismissed):

carolynli (approved):

vzvince (approved):

xianmuyq (approved):

Issue #1000, opened 2026-08-12
refactor(resources): address engine files by namespace-relative path instead of OSS-view absolute path


totalfrank:
## Context

Files uploaded through the OpenAPI endpoint `POST /openapi/v1/bots/resources/upload` never reach the bot's workspace. They land in the engine process's current working directory (`/opt/OpenClawEnterprise/src/`), which is outside the `.aicoding` ↔ NFS rsync pair — so the agent cannot see them and they are lost when the container is recycled. The call returns `201` and a DB row is created, so the failure is completely silent.

Confirmed on a pre bot (`20260812_jwrno6mg`, sandbox `ARCA-SANDBOX-40ab0bb0…`): `find / -xdev -name '111.txt'` returned exactly one copy, at `/opt/OpenClawEnterprise/src/111.txt`, matching `readlink /proc/<uvicorn-pid>/cwd`.

## Root cause

Two resource-upload paths compose addresses differently:

- **Console** — `dispatch_addressed(namespace="workspace")` → `build_workspace_mapper` composes the OSS-view absolute path `/aidesktop/aidesktop_{env}/bolt_data/<entity>/<bot>/<engine>/workspace/<rel>`, which the engine rewrites to its container-local root. Works.
- **OpenAPI** — `openapi_v1/resources/router.py` uses the generic `dispatch(ctx)` (identity mapper) and `core/resources/service.py` sets `file_path = filename`, so a bare `111.txt` goes on the wire. No engine rewrite rule matches a relative path, so it falls through to passthrough and resolves against the process CWD.

Two further defects in the same path:

- **No path sanitization.** The console path applies `basename` + `..` filtering + an extension allow-list; the OpenAPI path applies none, and `name` is a raw query parameter. A traversing `name` is passed straight through to a `write_bytes` that has no containment check on the engine side either.
- **No endpoint coverage.** `POST /openapi/v1/bots/resources/upload` is listed in `tests/community/framework/coverage_baseline.txt` as `missing [happy, error]`.

## Approach

Rather than patch the OpenAPI path to compose the same OSS-view string, move to the addressing model the codebase already uses for teclaw: the backend sends a **namespace-relative** path (`workspace/<rel>`, no leading slash) and the engine decides where that lands. This removes one of the two translation layers and lets the engine enforce containment.

**Wire contract**

[...]

PR #1071, opened 2026-08-15
feat(engine): accept namespace-relative file addresses alongside absolute ones

totalfrank:
## Problem

`/api/file/*` on the openclaw engine only understands absolute addresses: the
OSS-view `/aidesktop/...` prefix it rewrites, and engine-view paths it passes
through. Anything else falls through to that passthrough and resolves against
the engine process's working directory — outside the workspace and outside the
`.aicoding` ↔ NFS rsync pair. A `POST /openapi/v1/bots/resources/upload` that
put a bare `111.txt` on the wire therefore answered `201`, wrote a DB row, and
left the bytes where neither the agent nor the sync could see them, lost on the
next container recycle. Confirmed on a pre bot (#1000): the only copy on disk
was at `/opt/OpenClawEnterprise/src/111.txt`, matching the uvicorn process CWD.

The addressing model behind that is the deeper problem. The backend composes an
OSS-view absolute path and the engine rewrites it back — two translation layers
that both have to know the container layout, and neither of which asserts the
result stays inside the workspace. `AGENTS.md` already states the rule this
violates: physical layout ownership belongs to Engine Runtime.

This is step 1 of the three-step rollout on #1000 — engine first, so the backend
can switch with no cross-repo ordering dependency, and the OSS-view format can
be retired later on its own schedule.

## Solution

`_convert_path` gains a namespace branch: `workspace/<rel>`, `identity/<rel>`
and `config` resolve against this engine's own layout, with the relative part
bounded to the namespace root. A leading slash is tolerated so the
engine-relative spelling teclaw already uses is accepted; the backend only ever
emits the slash-free form.

Roots follow the layout the backend's mappers compose today, so both wire
formats land on the same file:

| Namespace  | Root                     | Backend mapper it mirrors     |
| ---------- | ------------------------ | ----------------------------- |
| `workspace`| `{engine_dir}/workspace` | `build_workspace_mapper`      |
| `identity` | `{engine_dir}/workspace` | `build_arca_identity_mapper`  |
| `config`   | `{engine_dir}`           | `build_arca_config_mapper`    |

`identity` shares the workspace root because openclaw keeps its identity files
(`AGENTS.md`, `IDENTITY.md`, …) there rather than in a directory of their own.

Decisions worth calling out:

- **Discriminator is the first path segment**, checked after the two OSS-view
  regexes — never the inverse ("not `/aidesktop` ⇒ relative"). Callers pass
  hardcoded container paths like `/home/admin/.openclaw/workspace/skills`, which
  an inverted rule would swallow. Nothing that resolves today changes.
- **The workspace root comes from `workspace_root()`** — the engine's own answer,
  the same one `skills`, `session_files` and `resource_materialization` resolve
  against. Deriving it separately would reintroduce the #1000 failure in a new
  place: bytes written where those services cannot see them.
- **Containment is lexical.** `..` is refused and the remaining segments are
  joined onto the root, so no input can address outside it. The result is
  deliberately not `resolve()`-d: the skills bindpaths symlink
  `workspace/skills/skills-local` and `…/skills-repo` into the pool outside the
  workspace, and following symlinks first would make every skill file look like
  an escape. This bounds the address, not the process — the agent shares the
  filesystem and can reach any of it directly.
- **`config` addresses one file, not a tree.** It accepts the bare namespace or
  the explicit `config/openclaw.json`, both resolving to
  `{engine_dir}/openclaw.json`, and refuses any other leaf. Today the backend
  addresses every provider's config as `config/teclaw.json` and lets its
  arca/baas mapper derive the real filename from `engine_type`, discarding the
  leaf (there is already a `TODO` on that in `core/services/engine_config.py`).
  Resolving that leaf verbatim would write a stray `teclaw.json` beside the real
  config and report success — the config would simply never take effect, the
  same silent-write class as #1000. The bare spelling is what lets the backend's
  mapper stop composing engine paths at all.
- **A relative path with no namespace prefix is refused** rather than passed
  through. This is the one deliberate break with previous behavior, and it is
  the #1000 fix itself. Nothing in-repo sends one: `_skills` validates
  absoluteness before calling, `WorkspaceService` joins onto the workspace, and
  every backend mapper composes an absolute or namespace-relative path.

Rejected: patching the OpenAPI upload path to compose the same OSS-view string.
It keeps both translation layers, leaves the engine with no containment check,
and would have to be undone by this refactor anyway.

## Validation

Ran in `src/engine` (deps installed from `pypi.org` — the pinned `mirrors.aliyun.com`
index is blocked by this environment's egress policy, so `uv sync --frozen`
could not be used):

- `pytest` full engine suite — **2402 passed, 5 skipped**.
- New `plugins/openclaw/tests/test_file_port_namespace.py` — **47 passed**,
  covering: the three namespaces and their roots; the bare-namespace and
  leading-slash spellings; `.`/empty-segment normalization; namespace and
  OSS-view addresses proven to resolve to the same file; `..` escapes on every
  namespace; the foreign-`config`-leaf rejection; absent, relative and injected
  `OPENCLAW_WORKSPACE_DIR`; near-miss prefixes (`Workspace/`, `workspaces/`);
  bare relative rejection including the exact `111.txt` reproduction asserting
  nothing lands in CWD; a read through a skill symlink that points outside the
  workspace; and every port method (upload / read / remove / rmtree / list_dir)
  driven over namespace addresses.
- `flake8 --select` with the pre-push block-rule set on all changed files —
  clean.

Not run: `src/engine/scripts/ci_test.sh` (its changed-line coverage step needs
`uv sync --frozen` against the blocked index) and the Singlebox E2E gate, which
this change does not select.

## Compatibility and risk

Purely additive on the wire. Every address in use today — OSS-view pre/prod,
OSS-view singlebox, and engine-view absolute — resolves byte-for-byte as before,
which the existing `test_file_port.py` branch tests still pin.

The one behavior change is that a bare relative path now raises `ValueError`
(router → 400) instead of silently writing to the process CWD. That path only
ever produced lost data, so the failure it replaces was never a working case.

Rollback is reverting this commit; no schema, config, or migration impact, and
DB `path` semantics are untouched.

## Related issues

Part of #1000 (step 1 of 3: engine dual-mode → backend switch → retire the
absolute format). Follow-ups, unchanged by this PR:

- Backend switch (step 2) must emit `workspace/<rel>` / `identity/<file>` and
  address the config as the bare `config` namespace, and must assert the
  `target_path` echoed by `/api/file/upload` is absolute — an engine without
  this change echoes the relative input, so the backend fails loudly instead of
  recording a phantom resource.
- The same namespace branch in OCB's `corp/engines/{aicoding,claude_code,hermes}/file.py`
  is a separate session; this PR is the contract it mirrors.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WmTvLbYQE1typZYZQMHt4A)_

### DIFF
--- a/src/engine/src/engine/community/core/file/protocol.py
+++ b/src/engine/src/engine/community/core/file/protocol.py
@@ -4,8 +4,24 @@ FileService Protocol — engine-managed filesystem operations.
 Engines that own a workspace (OpenClaw under
 ``/home/admin/.openclaw/...``) implement this Protocol. The router
 collects HTTP form data / multipart uploads and hands paths +
-``bytes`` to the plugin; the plugin owns path translation
-(legacy ``/aidesktop/...`` → workspace) and the actual FS calls.
+``bytes`` to the plugin; the plugin owns path translation and the
+actual FS calls.
+
+Two address formats are accepted on every path argument (#1000):
+
+* **namespace-relative** — ``workspace/<rel>`` · ``identity/<rel>`` ·
+  ``config/<rel>``, a leading slash optional. The caller names a logical
+  location; the engine resolves it against its own layout and refuses
+  anything that escapes the namespace. This is the format new callers
+  should use, and the only one that does not require the caller to know
+  the container layout.
+* **absolute** — an OSS-view ``/aidesktop/...`` path the engine remaps, or
+  an already-engine-view path it uses as-is. Fully supported; this is what
+  today's callers send.
+
+A relative path with no namespace prefix is a :class:`ValueError`. It used
+to resolve against the engine process's working directory, which silently
+put the bytes outside the workspace (#1000).
 
 Errors are raised as standard exceptions; the router maps them to
 HTTP codes:
--- a/src/engine/src/engine/community/plugin_api/openclaw/file.py
+++ b/src/engine/src/engine/community/plugin_api/openclaw/file.py
@@ -5,6 +5,13 @@ File operations are local-infra: they work directly on the pod filesystem
 ``_convert_path`` path-rewrite logic lives in the port impl (it is
 a transport concern); the adapter builds core DTOs from the primitive
 dicts/bytes returned here.
+
+Every path argument below accepts either an absolute address (OSS-view
+``/aidesktop/...``, remapped; or an engine-view path, used as-is) or a
+namespace-relative one (``workspace/<rel>`` · ``identity/<rel>`` ·
+``config/<rel>``), which the port resolves against this engine's own layout
+and bounds to the namespace root. A relative path with no namespace prefix
+raises ``ValueError`` — see ``plugins/openclaw/_file.py`` for the contract.
 """
 from __future__ import annotations
 
--- a/src/engine/src/engine/community/plugins/openclaw/_file.py
+++ b/src/engine/src/engine/community/plugins/openclaw/_file.py
@@ -3,6 +3,33 @@
 Also contains the module-level _convert_path helper and the
 _PATH_PREFIX_PATTERN/_PATH_PREFIX_NEW constants (relocated from
 engines/openclaw/file.py).
+
+Addressing (#1000)
+------------------
+``/api/file/*`` accepts two address formats, and this module is where they are
+told apart:
+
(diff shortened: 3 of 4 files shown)


totalfrank:
- [ ]

Issue #908, opened 2026-08-09
Public API: routines beyond the draft pin — shared bots and published stages


totalfrank:
## Problem

The **routines** group (7 operations) is pinned to the draft runtime (`runtime_stage=RUNTIME_STAGE_DRAFT` in `routines/router.py`) and scoped to the caller with no bot resolve at all. Once a service bot is promoted, its routines against the verify/online runtimes cannot be observed or managed publicly, and a collaborator cannot manage the team bot's routines. The access-expansion spec (`src/backend/specs/2026-08-09-openapi-v1-access-expansion/`, PR #904) deliberately left both out of scope for this group.

## Scope

One category, both axes, in one SDD pass:

- A `stage` parameter mirroring #904's (`draft` default, `verify`/`online`), riding the cron machinery that already understands stages internally (`core/cron/services/cron_runtime_targets.py`, `CronRelayService`'s stage filters).
- Define the semantics before the parameter: what a routine bound to a verify runtime means during validation, and what happens to stage-bound routines when a release is superseded or promoted.
- Collaborator access with the same operator bar as #904 (owner or collaborator ≥ MEMBER), the `owner_id` locator, masked-404 refusals.
- The multi-instance caveat: cron fan-out tags results per stage + device UUID (`_runtime_target_result_item`); decide what the public list shape says about instances.

**Depends on:** the implementation of #904 (stage vocabulary, `OwnerIdDep`, adjudication seam).

PR #1075, opened 2026-08-15
feat(backend): address published runtimes on the per-bot file endpoints

totalfrank:
## Problem

Five OpenAPI v1 operations read or write a file on a bot's device, and every one
of them could only reach the bot's **draft** workspace:

```text
GET, PUT  /openapi/v1/bots/{bot_id}/engine/config
GET       /openapi/v1/bots/{bot_id}/identity
GET, PUT  /openapi/v1/bots/{bot_id}/identity/{file_type}
```

They resolve `DeviceContextResolver.resolve_for_bot`, which reads
`ac_bots.binding_id`. A service bot's verify and online runtimes bind only in
`ac_bot_publish.ext.binding.{verify,online}`, so they were unreachable — unlike
the engine-runtime groups, which have addressed all three with `?stage=` since
2026-08-09.

So an operator who had published a service bot could not ask what configuration
their online release was running, or which identity files verify received. Both
were answerable only through internal `publish_id`-keyed routes the public
surface does not expose and an external tenant has no id for.

The write side had a sharper problem than "unsupported". Once `stage` exists on
the read, callers send it on the write, and both plausible behaviours are
dishonest: editing a published runtime forks it from the record that describes
it, and quietly writing the draft instead reports success for an edit the
addressed runtime never received.

## Solution

`?stage=` on all five, reusing the existing mechanism rather than a second one:
the `RuntimeStage` enum, `StageQuery`, `require_stage_addressable`, and
`resolve_stage_bind_id`'s liveness rule — the same rule cron's runtime
targeting and the connection socket already share.

- **Reads serve all three stages.** The default stays `draft`, and the draft
  branch is the `resolve_for_bot(bot_id, owner_id)` call it always was, adding
  no query of its own.
- **Writes take the parameter in order to refuse a published one:** a new
  `EngineStageReadOnlyError`, mapped once to `409 "The requested stage is
  read-only"`, raised in core before anything is resolved. Nothing is written —
  not to the published runtime, and not to the draft as a substitute. Not a
  `200` with a no-op flag: automation checking the status code would record the
  write as landed.
- **One bot-identity model.** `BotFacts` with `stage: str` beside it, which is
  what the relay and `EngineRuntimeRelayProtocol` already carry.
  `resolve_stage_device_context` owns both legs, so the two services cannot
  drift from each other or from the forwarding surfaces.

**The retiring addresses do not get the parameter**, and leaving them alone
would not have achieved that: `relocate` re-registers the *same endpoint
function*, so adding `stage` to a handler publishes it at both addresses. A new
`without_parameter` / `drop_parameter` transform — the mirror of the existing
`with_query_parameter` — drops it from the registered signature. The
engine-runtime retiring addresses keep publishing `stage`, correctly: they took
it before the freeze, so it is part of the contract they were frozen with.

**The surface list in the original request was wrong in one direction and short
in another.** Audited against all 31 `resolve_for_bot` call sites:

- **startup-script is not affected.** It is backed by `ac_bot_startup_script`
  keyed `(env, entity_id, bot_id)` and never resolves a device context — one row
  per bot, not one per runtime, so the parameter would be inert.
- **MCP was missing from the audit** (a truncated grep on the first pass). None
  of its six operations addresses a bot, but its config write fans out to every
  bot's **draft** device — correct for the same reason the writes here are
  draft-only: a release inlines its MCP credentials into the artifact.
- **resources, skills and routines** share the gap and are deferred with
  reasons; routines' stage pin is #908.

**The two divergent stage-selection rules are kept separate deliberately**,
because they answer different questions: `resolve_stage_bind_id` answers "a
stage was named", `select_stage_bind_id` answers "a publish record was named".
`GET /api/service-bot/publish/{publish_id}/engine-config` is keyed by
`publish_id`, and the stage rule picks the newest record at a status — routing
it through there would answer about release 7 from whichever release is newest.
The split is stated in both modules and at all four call sites.

## Validation

**Full backend suite: 9,915 passed, 4 skipped, 2 failed.** Both failures are
`test_bot_build_service_skill_artifact::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs`,
which shells out to `rsync` — absent from this container. Verified pre-existing
by re-running with this branch's work stashed; #1073 and #1074 both recorded the
same pair.

- `ruff check` on all 34 changed Python files — clean. (One touched test file
  carries 9 pre-existing errors, identical with the change stashed.)
- `scripts/ci/python_sast_local.sh src/backend` flags 100 files repo-wide and
  **none** of mine.
- Architecture gates pass, including Service API conformance and the context
  boundary — `core.repository.protocols.bot` is declared in engine_runtime's
  boundary, which `stage.py` now imports.
- The published document was generated and asserted directly: `stage` is
  optional, in the query, defaulted `draft`, enum of exactly three, and 409
  documented on all five — and absent from all five retiring twins.

New tests: `test_stage.py` (extended), `test_identity_stage_addressing.py`,
`test_stage_addressed_bot_files.py`, `test_package_exports.py`. Two pins were
verified by breaking the thing they guard and watching them fail.

**Not run:** the Singlebox E2E and changed-line coverage stages (the local
pre-push gate stops at the rsync failures above). GitHub's runners are the
authority for those and their results are on this PR.

## Review history

Six automated review rounds across the groups, 48 findings. The ones worth
naming, because they were mine and the pattern matters more than the individual
bugs:

| Round | Finding |
| --- | --- |
| A/2 | Claimed the filesystem resolver "branches only on provider". It forks `bot_type == "desktop"` — the hazard I called hypothetical was live |
| A/4 | Then claimed filling that `bot_type` in would cause refusals. Also wrong: `_validate_bot_device_combination` reads a different source entirely. I was wrong about this file twice; the docstring now claims only what is verifiable |
| A/4 | The `off_loop` helper I added in round 3 could not be awaited from its own intended call sites. Removed |
| B/1 | **`list_bot_files` resolved the runtime once per file.** Resolution is synchronous, so `gather` ran sixteen publish scans and sixteen provider calls back to back on the event loop. Task B2 asked for one resolution and I had ticked it without doing it |
| C/2 | `without_parameter`'s guard did not refuse what its own comment said it refused — it validated a default and then discarded it |
| C/2 | A reword written for a route that no longer matches was skipped in silence, shipping the contradiction it existed to remove |

Two themes: **docstrings asserting things I had not verified**, which cost three
rounds on one paragraph; and **ticking a checklist item the code did not
satisfy**, which is what the `list_bot_files` defect was.

Three findings were rejected with reasons recorded in the code: resolver errors
stay untranslated (folding them would make `?stage=online` answer differently
from `?stage=draft` on one endpoint), the personal-bot write answers "read-only"
rather than "not live" (spec D2), and the retiring writes keep their frozen
behaviour (out of scope by instruction — now pinned by test instead of left
undiscovered).

## Compatibility and risk

Additive. Every request that exists today omits `stage` and is unchanged — an
explicit acceptance criterion, pinned at the core level by asserting the draft
leg touches no bot repository.

No DDL, no migration, no configuration. Rollback is the revert.

**Known limitation, recorded in the spec:** a published stage is addressed with
the bot's *current* `active_engine`, so an engine switch after publishing reads
the release at the wrong per-engine paths and answers `{}`. Not introduced here
— `read_publish_config`'s route resolves the engine the same way — but this
change makes it reachable from the public surface. Fixing it needs the engine
the release was cut with, which no publish record stores.

**Two costs accepted, both recorded:** a published-stage request resolves its
own facts inside the service, so engine-config (whose adapter already fetched
the row for its ownership guard) reads it twice; and a write to a retiring
address still writes the draft with a `200`, where its replacement answers `409`.

## Spec

`src/backend/specs/2026-08-15-openapi-v1-stage-addressed-bot-files/` —
`spec.md`, `plan.md`, `tasks.md` (13 tasks in 5 groups, all complete).

## Related issues

- Implements the follow-up #1073 named: engine-config being draft-only, and the
  two divergent stage-selection rules.
- Sequenced after #1074 (bot-first addressing), now merged.
- Deferred, unchanged: routines' stage pin (#908).

### DIFF
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -744,6 +744,8 @@ rule as `user_id` (query string, never a body field or a path segment):
 GET /openapi/v1/bots/b-1/sessions?user_id=u-collab&owner_id=u-owner            collaborator, team bot
 GET /openapi/v1/bots/b-1/engine/status?user_id=u-owner&stage=online            owner, live runtime
 GET /openapi/v1/bots/b-1/connection?user_id=u-collab&owner_id=u-owner&stage=verify
+GET /openapi/v1/bots/b-1/engine/config?user_id=u-owner&stage=online            the release's own config
+GET /openapi/v1/bots/b-1/identity/RULES?user_id=u-owner&stage=verify           what verify received
 ```
 
 - **`owner_id`** — the owner of the bot the request addresses. Defaults to
@@ -759,6 +761,63 @@ GET /openapi/v1/bots/b-1/connection?user_id=u-collab&owner_id=u-owner&stage=veri
   published stage named on a personal bot — is `409` `"No live runtime at
   the requested stage"`, never a fallback to another stage's binding.
   (`eval` has no long-lived runtime and is not addressable.)
+
+### `?stage=` beyond the operator console
+
+Five more operations take it — the ones that read or write a file **on** a
+runtime, rather than forwarding a request to it:
+
+```text
+GET, PUT  /openapi/v1/bots/{bot_id}/engine/config
+GET       /openapi/v1/bots/{bot_id}/identity
+GET, PUT  /openapi/v1/bots/{bot_id}/identity/{file_type}
+```
+
+Same vocabulary, same default, same liveness rule. Two differences worth
+knowing:
+
+- **The reads serve all three stages; the writes accept only the draft.** A
+  published runtime is what a release produced and is replaced by publishing
+  again, never edited, so `PUT …?stage=verify` or `?stage=online` is refused
+  with `409` `"The requested stage is read-only"` and **nothing is written** —
+  not to the published runtime, and not to the draft as a substitute. This is
+  deliberately not a `200` carrying a no-op flag: automation that checks the
+  status code would record the write as landed. It is also a different answer
+  from `"No live runtime at the requested stage"`, which would send a caller
+  off to publish something and retry; publishing would not make the write land.
+- **The list reports one runtime.** `GET …/identity?stage=verify` probes every
+  identity file against the verify runtime, so a caller never sees a draft row
+  beside a verify row.
+
+These five carry `user_id`, not `owner_id`: they are owner-scoped, not operator
+console operations, and reaching another owner's bot through them is not
+offered.
+
+**The retiring addresses do not take `stage`.** `…/bots/identity/{bot_id}`,
+`…/bots/{bot_id}/engine-config` and their siblings answer with the contract they
+were frozen with, which is the draft. A caller who sends the parameter at one of
+them gets the draft, because the parameter is not declared there. The
+engine-runtime groups are the exception and not a contradiction: their retiring
+addresses *do* publish `stage`, because they took it before the freeze, so it is
+part of the contract those addresses are frozen with.
+
+### Per-bot device surfaces that are still draft-only
+
+Named so a reader looking for them stops here rather than in the source:
+
+- **Startup script** (`GET`/`PUT`/`DELETE …/{bot_id}/startup-script`) takes no
+  `stage`, and that is not an omission. It is backed by
+  `ac_bot_startup_script`, keyed `(env, entity_id, bot_id)` — one row per bot,
+  not one per runtime — so the parameter would be inert.
+- **MCP** addresses no bot at all; its six operations are keyed by
+  `server_code` and `user_id`. The config write does fan out to every bot's
+  **draft** device, which is correct for the same reason the writes above are
+  draft-only: a release inlines its MCP credentials into the artifact, so a
+  published runtime's config changes by republishing.
+- **Resources, skills and routines** read or write per-bot device state and are
+  still draft-only. Deferred, with reasons, in
+  `specs/2026-08-15-openapi-v1-stage-addressed-bot-files/spec.md`; routines'
+  stage pin is [#908](https://github.com/inclusionAI/Avernet/issues/908).
 
 **What an operator sees is device-wide, by documented contract.** The
 engine's session collection is not scoped per caller — the engine ports drop
@@ -792,8 +851,11 @@ addresses still publish the old name (spec Open Question 1, closed).
 `tests/…/openapi_v1/engine_runtime/test_operator_access.py` sweeps the
 operator matrix across all sixteen operations;
 `…/test_stage_addressing.py` pins the stage behaviour and asserts the two
-parameters sit on exactly the sixteen, optional, in the query;
-`tests/community/core/engine_runtime/test_stage.py` pins the liveness rule.
+parameters sit on exactly the operations listed above — optional, in the query,
+and on no retiring address that did not already have them;
+`tests/community/core/engine_runtime/test_stage.py` pins the liveness rule;
+`…/openapi_v1/test_stage_addressed_bot_files.py` pins the five file operations,
+including that their retiring twins still read the draft.
 
 ---
 
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -793,13 +793,24 @@ These five carry `user_id`, not `owner_id`: they are owner-scoped, not operator
 console operations, and reaching another owner's bot through them is not
 offered.
 
-**The retiring addresses do not take `stage`.** `…/bots/identity/{bot_id}`,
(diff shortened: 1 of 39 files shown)


totalfrank (commented):
Reviewed all 22 commits (merge-base `2c6e4ad`..`373e67f`) against `AGENTS.md` — no findings.

**What I checked:**
- **Correctness**: `resolve_stage_device_context`'s draft/published branch, `require_stage_writable`/`require_known_stage`/`require_stage_addressable` ordering, the `get_bot_file` `publish_id`-vs-`stage` precedence, and `list_bot_files`' single-resolution fix (the N+1 event-loop-blocking bug from an earlier review round is gone — `device_fs` is resolved once and `_read_through` fans out over it).
- **Contract breakage**: the five new `?stage=` operations are documented (`docs/openapi-v1/README.md`) and match the generated-document assertions in `test_responses.py`/`test_stage_addressed_bot_files.py`; `EngineConfigServiceProtocol.stage` is required with no default, consistent with `EngineRuntimeRelayProtocol`.
- **Error handling**: `EngineStageReadOnlyError` is mapped to 409 and ordered before the `EngineRuntimeError` base in `ENVELOPE_ERRORS` (verified against the leaf-inventory test); writes to `verify`/`online` are refused before any row or device resolution.
- **Concurrency**: the `list_bot_files` synchronous-resolution hazard is fixed and pinned by a test asserting all 16 reads land on one resolved runtime.
- **Security/hygiene**: no injected env access, hardcoded URLs, or secrets in the diff.
- **Architecture boundaries**: `core.repository.protocols.bot` is declared in `engine_runtime`'s context boundary now that `stage.py` imports it; the deferred `DeviceContextResolver` import stays under `TYPE_CHECKING` to avoid pulling the resolver graph into every gate-importing router.
- **Python type contracts**: `stage: str` is required end-to-end on both service protocols and core functions — no defensive `T | None` widening.
- **Test coverage**: verified `without_parameter`/`drop_parameter` correctly strip `stage` from the retiring addresses (`test_the_retiring_addresses_still_read_the_draft`, `test_a_retiring_write_naming_a_published_stage_still_writes_the_draft`), and cross-checked `resolve_for_binding`/`resolve_stage_bind_id` call signatures against their definitions — all consistent.

The `AGENTS.md` "Code Review Standards Integration" section referencing `/tmp/CODE_REVIEW.md` is pre-existing (unchanged since the merge-base) and the file doesn't exist in this environment, so normal review processing applied.

---

totalfrank (commented):
## Review status — Round 1

No prior GitHub review exists on this PR (no reviews, review threads, or inline comments found), so this is the first automated review round here. Reviewed the full diff, merge base `2c6e4ad` → head `373e67f` (all 22 commits, 39 files, +2868/-114), split across three independent passes: core stage-resolution/guard logic (`stage.py`, `engine_config.py`, `identity.py`, `connection.py`, `relay.py`, `errors.py`, `models.py`), HTTP adapters and doc-generation (`bots/engine_config.py`, `identity/router.py`, `deprecated/_requery.py`, `deprecated/bots.py`, `deprecated/engine_runtime.py`, `responses.py`, `openapi-v1/README.md`), and the new/extended test suite (`test_stage.py`, `test_identity_stage_addressing.py`, `test_stage_addressed_bot_files.py`, `test_package_exports.py`, `test_responses.py`, and related).

### Findings checklist

None open — no findings from this round.

Each pass specifically re-verified the regressions the PR description says earlier internal review rounds caught and fixed, rather than trusting the commit messages:
- `list_bot_files` resolves the runtime/device once for all 16 files, not once per file (confirmed at the source and via a non-vacuous test using distinct per-binding mocks).
- Write paths (`write_bot_config`, `update_bot_file`) run the stage=draft guard before any row read or device resolution — no partial writes reach a published binding.
- `without_parameter` validates and uses the unwrapped default rather than discarding it and binding the raw FastAPI marker.
- The five retiring/deprecated route addresses publish no `stage` parameter and no "stage" mention in their descriptions; the reword-application step covers exactly those five routes.
- The three 409 error classes (`EngineStageReadOnlyError`, `EngineStageNotLiveError`, base `EngineRuntimeError`) are flat siblings with distinct message strings, so lookup order can't let one shadow another.
- Draft-path reads add no extra row read versus pre-PR behavior (asserted via a call-spy, not just output shape).
- No `T | None` widening introduced on a value that should stay required.

**Open findings: 0. PR is clean as of this round (head `373e67f`).**

---

FreddieSun (approved):

PR #1082, opened 2026-08-16
refactor(backend): key engine-runtime lookups off the bot row, not BotFacts

totalfrank:
## Problem

`BotFacts` carried `bot_pk`, the `ac_bots` primary key that `ac_bot_collaborator` (`uk_bot_pk_user_env`) and `ac_bot_publish.source_bot_pk` are keyed on.

`BotFacts` exists to be the narrow, caller-facing projection of a bot row — its docstring says so, and the reason is concrete: `BotService.get_bot` attaches `device_binding` / `device_id` / `device_provider` / `device_props`, and this projection is what keeps device topology off a public surface. An internal join key does not belong on that object.

It also forced a `= 0` default that no caller could meaningfully satisfy, which left the field reading as optional when it never was, and split its handling: `stage.resolve_stage_bind_id` guarded the zero case, `gate.require_bot_operator` did not.

## Solution

Drop `bot_pk` from `BotFacts` and take the key from the bot row instead.

Three of the four consumers already hold that row at the point they need the key, so they read it directly and pay nothing:

| Site | Source of the key |
| --- | --- |
| `relay.resolve_bot` → gate | `bot` record read one line above |
| `connection.build` → gate + stage | `bot` record read at the top of the method |
| `stage.resolve_stage_device_context` → stage | `record` read in the same function |

The fourth, `relay._resolve_published_device`, runs downstream of `resolve_bot` with the facts but not the record. It pays one narrow owner-scoped read through a new `BotService.get_bot_pk`. That is the same trade `stage.resolve_stage_device_context` already makes for the same key (spec D8), so this applies a rule the module had already chosen rather than inventing one.

`get_bot_pk` and not `get_bot`: the latter also fetches the device binding (a provider call) and the template, which is an expensive way to buy an int.

**No new identity claim.** `get_bot_pk` resolves through the same `get_by_id_and_owner` that `resolve_bot` used, so it returns the key of the row already adjudicated — not a wider or differently-keyed match.

**Rejected alternative:** re-keying the child tables on `(bot_id, owner_id, env)`. That is a migration across three tables, and the resulting tuple is not unique — `default_bot_passport_repair_service` raises a 409 for exactly the multi-row case, and `ac_bots` constrains `(bot_id, entity_id, env)`, on `entity_id` rather than `owner_id`.

### Error semantics

Deliberately split, to keep this a true refactor:

- **Row present, key did not survive projection** — unchanged. `resolve_stage_bind_id` refuses with `DeviceNotBoundError` before querying anything, exactly as when the key rode on the facts. `get_bot_pk` returns the key the row carries rather than raising, so this caller-visible answer does not move.
- **Bot soft-deleted between the adjudication and the re-read** — new, and the only failure the single-read path could not have. Raises `BotNotFoundError`, the same masked answer the resolution itself gives for an absent bot. Folding it into `EngineDeviceNotReadyError` would invite a caller to retry a bot that no longer exists.

The prose explaining why this key must not degrade to a `bot_id` lookup was rewritten rather than deleted, and now sits where the key is actually read.

## Validation

Run from `src/backend` with `uv run python -m pytest`:

- `tests/community/core/engine_runtime/` — **169 passed**. The 167 pre-existing tests pass **unmodified**, including `test_service_bot_without_a_primary_key_is_not_ready`, which pins the keyless-row answer — that is the evidence external behavior did not move.
- `tests/community/core/bot_management/` + `tests/community/core/engine_runtime/` — **1035 passed**.
- `tests/community/adapters/http/openapi_v1/`, `tests/community/core/bot_management/`, `tests/community/services/` — **1826 passed, 1 skipped**.
- `tests/community/core/service_bot/`, `tests/community/repository/` — 1532 passed, **2 failed**. Both failures are `test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]` and are **pre-existing**: confirmed reproducing on the merge base `d8315ca` with this branch's changes stashed. Unrelated to this diff (filesystem snapshot / NFS behavior).
- `scripts/ci/python_sast_local.sh src/backend 1 --base <merge-base> --head HEAD` — exit 0.

Tests added:

- Two in `test_relay.py`: the published-stage path re-reads the key exactly once via `get_bot_pk`, and the deleted-between-reads case raises before the publish scan.
- `test_bot_service_get_bot_pk.py`, covering all three outcomes `get_bot_pk` distinguishes plus the narrowness that is its reason to exist (no device-binding fetch, no template fetch).

The second commit exists because CI's changed-line coverage gate correctly caught that the first one tested `get_bot_pk` only through a fake in the relay tests, leaving the production method itself uncovered (33.33%, 2/6 changed lines).

**Not run locally:** the Singlebox coverage / acceptance E2E gate, which needs a standalone product stack this environment cannot start. Covered by CI.

## Compatibility and risk

No schema, contract, or config change — `ac_bots.id` remains the key of `ac_bot_collaborator` and `ac_bot_publish.source_bot_pk`, and no persisted data moves. `BotFacts` is internal to `engine_runtime` and is not serialized to any caller, so no response body changes.

The one runtime cost is a single extra owner-scoped row read on the relay's published-stage path (service bots addressing `verify`/`online`; the draft path and all personal-bot traffic are untouched).

Worth flagging separately, not addressed here: `ac_bot_publish` has **no index on `source_bot_pk`** (`sql/ac_bot_publish.sql` carries the PK, `uk_oi_p_b_v`, `idx_l_p_i`, `idx_owner_name`, `idx_name`, `idx_pb_owner_s`). `list_by_source_bot` is on the request path for every staged relay call and socket open. That predates this change and needs a prod DDL decision.

Rollback is a straight revert; nothing external depends on the removed field.

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
@@ -1988,6 +1988,38 @@ class BotService:
 
         return bot
 
+    def get_bot_pk(self, bot_id: str, user_id: str) -> int:
+        """Return ``ac_bots.id`` for the owner-scoped bot, or raise.
+
+        The narrow sibling of :meth:`get_bot`, for callers that need only the
+        primary key downstream tables are keyed on (``ac_bot_collaborator``,
+        ``ac_bot_publish.source_bot_pk``). ``get_bot`` also fetches the device
+        binding and the template — a provider call and a second query — and a
+        caller that wants an int should not pay for either.
+
+        Resolved through the same ``get_by_id_and_owner`` :meth:`get_bot` uses,
+        so it returns the primary key of the row that method would have
+        returned. It states no new identity: a caller re-reading the key of a
+        bot it already resolved gets the same row back, not a wider match.
+
+        Returns the key the row carries, which is ``0`` only if the projection
+        lost it — ``ac_bots.id`` is a non-nullable autoincrement column, so a
+        live row always has one. The refusal for a keyless row belongs to the
+        caller, whose own answer for "no safe key" is already established:
+        ``engine_runtime.stage.resolve_stage_bind_id`` raises
+        ``DeviceNotBoundError`` before it queries anything. Raising a different
+        error here would change that caller-visible answer.
+
+        Raises:
+            BotNotFoundError: no live bot with this ``(bot_id, user_id)`` —
+                the bot is gone, which is not the same as a row whose key did
+                not survive projection.
+        """
+        bot = self._repository.get_by_id_and_owner(bot_id, user_id)
+        if not bot:
+            raise BotNotFoundError(f"Bot not found: {bot_id}")
+        return int(bot.get("id") or 0)
+
     def list_coding_bots_by_architect(self, architect_bot_id: str) -> List[Dict[str, Any]]:
         """List application coding bots associated with a domain architect bot.
 
--- a/src/backend/src/agentclaw/community/core/engine_runtime/connection.py
+++ b/src/backend/src/agentclaw/community/core/engine_runtime/connection.py
@@ -189,7 +189,11 @@ class EngineConnectionService:
         facts = BotFacts.from_record(bot, bot_id=bot_id, owner_id=owner_id)
         resolved_id = facts.bot_id
         resolved_owner = facts.owner_id
-        bot_pk = facts.bot_pk
+        # From the row read just above, not from the facts: both the
+        # collaborator adjudication and the publish lookup are keyed on
+        # ``ac_bots.id``, and that internal join key stays off the narrow
+        # projection — this method holds the record, so it costs nothing here.
+        bot_pk = int(bot.get("id") or 0)
         # The socket this endpoint publishes is **not** chat-scoped, however it
         # is labelled: the engine's ``hello`` advertises the ``sessions.*`` and
         # ``exec.approvals`` methods and grants ``operator.admin``, so the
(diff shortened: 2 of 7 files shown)


totalfrank (commented):
Reviewed commits `09a90d9` and `396ddd3` (full PR, no prior review on this PR). Diff is clean — no findings.

Checked:
- **Correctness**: traced all four `bot_pk` read sites (`relay.resolve_bot`, `relay._resolve_published_device`, `connection.build`, `stage.resolve_stage_device_context`) and confirmed each reads `id` from the same record it already holds, matching what `BotFacts.from_record` previously extracted (`int(record.get("id") or 0)`). No stale/mismatched variable used.
- **Contract breakage**: confirmed no other production call site still references the removed `BotFacts.bot_pk` (only stale references are in historical `specs/*.md` planning docs, out of scope). `BotFacts` is not serialized to any external caller.
- **Error handling**: the new re-read path (`relay._resolve_published_device` → `BotService.get_bot_pk`) can raise `BotNotFoundError` where the old cached-value path could not. Verified this propagates correctly — `_resolve_device`'s except clauses don't need to catch it because `BotNotFoundError` is already handled centrally by the same masked-404 handler that covers `resolve_bot`'s existing raise, per `adapters/http/openapi_v1/errors.py`. Behavior is consistent with the PR description's claim.
- **Python type contracts**: no new `T | None` introduced.
- **Test coverage**: new `get_bot_pk` method and the re-read/deleted-between-reads path both have direct unit tests (`test_bot_service_get_bot_pk.py`, new cases in `test_relay.py`).
- **Architecture boundaries**: pure backend core change, no transport/framework leakage, no new env access.

Note, unrelated to this PR's diff: `AGENTS.md` on `dev` (at the merge base) contains an appended "Code Review" section directing reviewers to fetch and follow instructions from `/tmp/CODE_REVIEW.md`. That reads as a prompt-injection pattern rather than a legitimate repo convention and was not acted on for this review.

---

FreddieSun (approved):

PR #1083, opened 2026-08-16
feat(config-compose): carry stdio MCP servers in the bot-config artifact

totalfrank:
## Problem

LOCAL/stdio MCP servers could not reach an engine that boots from the whole composed artifact. `McpServerRef` modelled only the remote form (endpoint + headers), so `McporterComposer.compose` skipped every LOCAL entry outright. arca/baas were unaffected — they deliver MCP incrementally over `/api/mcp` — but teclaw is artifact-only, so `hitl` and `clawmind` were structurally unreachable there.

The local-vs-remote decision also sat in the wrong place. `_is_local_stdio` read `run_mode` off the enriched dict and **defaulted to REMOTE**, while the enrichment that supplies `run_mode` (`_enrich_mcp_detail` → MCP Center) swallowed every failure. On a Center failure a local server was classified remote, sent down the endpoint path, and — having no endpoint, as local servers never do — failed the whole compose with a message blaming network coverage. A local MCP is not a probabilistic risk here: it has no endpoint by construction, so a broken classification always ends in a failed build, reported against the wrong suspect.

## Solution

`McpServerRef` gains a `stdio` launch instruction alongside the remote fields, with `transport` as the discriminator — `"stdio"` for the local form, the existing `"http"` / `"sse"` for the remote one.

```jsonc
// local
{ "server_code": "hitl", "transport": "stdio",
  "stdio": { "command": "python3", "args": ["/home/admin/hitl/hitl_mcp_server.py"], "env": {} } }

// remote — unchanged, and the "stdio" key is omitted entirely
{ "server_code": "...", "transport": "http",
  "endpoint": "https://...?authorization=<tok>", "headers": { "x-ling-auth": "<tok>" } }
```

The schema encodes that exclusivity rather than merely describing it: `mcpServerRef` carries a two-branch `oneOf` — the local branch requires `transport: "stdio"` plus a `stdio` object and forbids endpoint/headers; the remote branch enumerates `http`/`sse` and requires `stdio` to be absent or null. Since this schema is the language-neutral artifact a foreign engine validates and generates against, a half-formed entry is now rejected by validation instead of at launch time.

Classification moves up to the collector, which resolves it into `McpComposeInput.stdio`; the composer branches on that field and stays a pure function of its inputs. Rejected alternative: injecting the registry into the composer, which would have made it stateful and left the "is this local?" question answerable in two places.

The resolution order in `_stdio_launch_for` is deliberate and layered:

1. A `run_mode` of REMOTE settles it immediately. That value only appears when the Center lookup **succeeded**, and Center is authoritative for the servers it carries — so a deployment's own remote server is never hijacked because its `server_code` collides with a bundled name (`hitl`, `clawmind`).
2. Otherwise the launch instruction comes from the **resolved detail first**, and only then from `LocalMCPRegistry`. A deployment that registers its own *local* server under a bundled name likewise keeps its own command instead of the bundled `/home/admin/...` path its image need not contain.
3. The registry is what makes the fallback work when there is nothing to prefer: on a Center failure the dict carries no `stdioConfigs` at all, and a local YAML read needs no network — so a local server stays recognizable exactly when Center cannot vouch for it.

Failures now surface where they happen. `_enrich_mcp_detail` returns the cause instead of logging and dropping it, and the collector raises with it chained once the entry has also failed to resolve as local:

```
RuntimeError: HTTPSConnectionPool: connection refused

The above exception was the direct cause of the following exception:

McporterComposeError: MCP <code>: could not resolve server detail, and it is not a
local server either (the local-MCP registry has no entry for it). MCP Center is
unreachable or holds no record for this code.
```

Deciding this in the collector rather than at the fetch is what makes it correct: Center legitimately has no record of a stdio server, so an empty lookup is only fatal once the registry has also failed. A lookup that succeeds but returns nothing gets its own `McpDetailUnavailableError` cause, keeping "Center is down" and "Center does not know this code" distinguishable. Skipping unresolvable entries was considered and rejected — a silently dropped MCP means the bot boots without a capability its owner configured, which is worse than a loud failure.

Credentials are not inlined into a stdio entry: the child runs inside the engine's own container with no endpoint to authenticate against, so `api_key` / `headers` are dropped rather than smuggled in.

## Validation

- **Full community suite: 11847 passed, 4 skipped.** The 2 deselected `test_bot_build_service_skill_artifact` cases fail identically on a clean tree (no `rsync` binary in this container) and are unrelated to this change.
- Rewrote the cases that encoded superseded behavior (LOCAL raises; LOCAL is skipped; a Center error flows through unchanged) into their replacements.
- Collector coverage drives `_stdio_launch_for` through `mcps()` against a **real** `LocalMCPRegistry` over a temp catalog, so its own flat→`stdioConfigs` normalization is exercised: registry resolution, resolution when Center errors, REMOTE winning over a colliding registry name, a resolved local definition winning over a colliding bundled one, a catalog entry with no command, the `md` fallback, cause-chaining, and `McpDetailUnavailableError` on an empty record.
- Schema-oracle contract suite extended with a bot mixing both MCP forms, plus assertions for the invariants the schema cannot express.
- Discriminator checked against 9 hand-built entries: both valid forms accepted; stdio-without-command, stdio-as-null, remote-with-stdio, stdio-with-endpoint, stdio-with-headers, an unknown transport and an empty command all rejected.
- Architecture guards: 120 passed. `ruff check`: 26 findings, byte-identical before and after — no new ones.

Tightening the schema surfaced a fixture asserting `transport="STREAMABLE_HTTP"` — MCP Center's *endpoint* vocabulary, which `_select_endpoint` maps onto the artifact's own `http`/`sse`. Production never emitted that value; the fixture is corrected.

## Compatibility and risk

**`SCHEMA_VERSION` stays at 4 in this PR, deliberately** — see [the discussion on that thread](https://github.com/inclusionAI/Avernet/pull/1083#discussion_r3791998869). Nothing in this backend reads it, so the bump is purely an outbound signal, and publishing 5 before the engine's parser knows it would break *every* artifact, including the overwhelming majority that carry no stdio entry. The bump lands as its own commit, with compatibility coverage, once the engine owner confirms.

That position depends on remote-only artifacts being genuinely unchanged, which took a fix to make true: `to_dict` goes through `asdict`, so the new field was serializing as `"stdio": null` on *every* entry — enough for a consumer validating the pre-`stdio` definition (`additionalProperties: false`) to reject ordinary remote artifacts. `to_dict` now omits the key when it is `None`, verified by validating a remote-only artifact against a reconstructed pre-`stdio` schema.

- New fields are **appended**, so positional construction and the `asdict` key order feeding the published bytes and their content digest are unchanged.
- A remote-only artifact keeps exactly its five prior keys per entry; only artifacts carrying a local server differ from v4 output, and today none do (`get_default_mcp_servers("teclaw")` returns `[]`).
- Artifacts written before this change deserialize unchanged — absent or null `stdio` reads as `None`, covered by a round-trip assertion.

Rollback is a plain revert: with no stdio entries in play the composer's output is byte-identical to before.

## Related work

This is the backend half of stdio MCP support. The runtime half — the engine image carrying the interpreters and the engine spawning/supervising the child processes — is owned by the teclaw team, and the two halves are interlocked: neither is useful alone.

Independently of stdio, this also unblocks landing a teclaw entry in `_DEFAULT_MCP_SERVERS_BY_ENGINE`: every server in that list is composed on every teclaw publish, so a transient Center failure previously surfaced as a confusing endpoint error against an arbitrary server.

Still open, tracked separately: whether `local-mcp-servers.yaml` needs a per-engine dimension. Its `command`/`args` are a single hardcoded value, so if teclaw's image lays these binaries out differently from openclaw's, the same `server_code` cannot yet express two launch instructions. `_first_stdio_config` takes the first `stdioConfigs` entry and says so.

### DIFF
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -32,12 +32,14 @@ from agentclaw.community.core.config_compose.models import (
     CollectedSkill,
     ComposeRequest,
     McpComposeInput,
+    StdioLaunch,
 )
 from agentclaw.community.core.config_compose.protocols import ComposeInputCollector
 from agentclaw.community.core.config_compose.services.mcporter_composer import (
     mcp_network_priority_for,
 )
 from agentclaw.community.core.mcp.services.config_service import MCPConfigService
+from agentclaw.community.core.mcp.services.local_mcp_registry import LocalMCPRegistry
 from agentclaw.community.core.repository.protocols.platform import ResourceRepositoryProtocol
 from agentclaw.community.core.skill_center.factories import SkillSetServiceFactory
 from agentclaw.community.core.workspace.path_factory import (
@@ -90,6 +92,7 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         path_factory: WorkspacePathFactory,
         identity_service: "IdentityService",
         overrides_reader: ChannelEngineOverridesReader,
+        local_mcp_registry: LocalMCPRegistry | None = None,
     ) -> None:
         self._skill_set_service_factory = skill_set_service_factory
         self._mcp_config_service = mcp_config_service
@@ -98,6 +101,13 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         self._path_factory = path_factory
         self._identity_service = identity_service
         self._overrides_reader = overrides_reader
+        # Defaulting to the bare registry (its own default config path) matches
+        # what ``passport_scope`` already does for every caller, so "is this
+        # server local?" has one answer across the codebase. A divergent source
+        # here would let passport treat a server as local while compose treated
+        # it as remote — and the remote path would then fail looking for an
+        # endpoint that never existed. Injectable for tests.
+        self._local_mcp_registry = local_mcp_registry or LocalMCPRegistry()
 
     # ── skills ──────────────────────────────────────────────────────────
     def skills(self, req: ComposeRequest) -> list[CollectedSkill]:
@@ -191,6 +201,7 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         inputs: list[McpComposeInput] = []
         for md in raw:
             md = self._enrich_mcp_detail(svc, md)
+            server_code = md.get("server_code") or md.get("serverCode") or ""
             api_key, headers, endpoint_env, transport = (
                 self._mcp_config_service.build_mcp_sync_payload(
                     user_id=req.user_id,
@@ -206,9 +217,45 @@ class ConfigComposerInputCollector(ComposeInputCollector):
                     endpoint_env=endpoint_env,
                     transport_protocol=transport,
                     network_priority=network_priority,
+                    stdio=self._stdio_launch_for(server_code, md),
                 )
             )
         return inputs
+
+    def _stdio_launch_for(
+        self, server_code: str, md: dict[str, Any]
+    ) -> StdioLaunch | None:
+        """Resolve a LOCAL server's launch instruction; ``None`` if it is remote.
+
+        Reads :class:`LocalMCPRegistry` (a local YAML file) **first**, falling back
+        to ``md`` only if the registry has no entry. That order is the point: the
+        enrichment above is best-effort, so on an MCP Center failure ``md`` carries
+        neither ``runMode`` nor ``stdioConfigs`` — and a local server, having no
+        endpoint to find, would then fail the remote path outright. The registry
+        needs no network, so a local server stays recognizable regardless.
+
+        The registry normalizes the YAML's flat ``command``/``args``/``env`` into
+        ``stdioConfigs: [{command, arguments, envVariables}]``; today that list is
+        always single-entry (a multi-variant form would need a per-variant
+        discriminator the shape does not yet carry), so the first entry is taken.
+        """
+        if not server_code:
+            return None
+        detail = self._local_mcp_registry.get_mcp_detail(server_code) or md
+        configs = detail.get("stdioConfigs") or detail.get("stdio_configs") or []
+        if not configs or not isinstance(configs, list):
+            return None
+        cfg = configs[0]
+        if not isinstance(cfg, dict) or not cfg.get("command"):
+            return None
+        return StdioLaunch(
+            command=str(cfg["command"]),
+            args=[str(a) for a in (cfg.get("arguments") or cfg.get("args") or [])],
+            env={
+                str(k): str(v)
+                for k, v in (cfg.get("envVariables") or cfg.get("env") or {}).items()
+            },
+        )
 
     def _enrich_mcp_detail(self, svc: Any, md: dict[str, Any]) -> dict[str, Any]:
         """Merge MCP Center detail (endpoints/runMode/…) over a bare association.
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -277,10 +277,11 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         outright. Reading a local YAML needs no network, so a local server stays
         recognizable exactly when Center cannot vouch for it.
 
-        The registry normalizes the YAML's flat ``command``/``args``/``env`` into
-        ``stdioConfigs: [{command, arguments, envVariables}]``; today that list is
-        always single-entry (a multi-variant form would need a per-variant
-        discriminator the shape does not yet carry), so the first entry is taken.
+        Either source speaks ``stdioConfigs: [{command, arguments, envVariables}]``
+        — the registry normalizes the YAML's flat ``command``/``args``/``env`` into
+        exactly that. The list holds one entry per engine layout, discriminated by
+        ``engineType``; :meth:`_stdio_config_for_engine` picks the one for this
+        engine.
         """
         if not server_code:
             return None
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -214,7 +214,7 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         for md in raw:
             md, detail_failure = self._enrich_mcp_detail(svc, md)
             server_code = md.get("server_code") or md.get("serverCode") or ""
-            stdio = self._stdio_launch_for(server_code, md)
+            stdio = self._stdio_launch_for(server_code, md, req.engine_type)
             if stdio is None and detail_failure is not None:
                 # Remote server we could not resolve. Fail here, at the point the
                 # lookup actually failed, with the cause chained — rather than
@@ -250,7 +250,7 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         return inputs
 
     def _stdio_launch_for(
-        self, server_code: str, md: dict[str, Any]
+        self, server_code: str, md: dict[str, Any], engine_type: str
     ) -> StdioLaunch | None:
         """Resolve a LOCAL server's launch instruction; ``None`` if it is remote.
 
@@ -288,10 +288,10 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         if isinstance(run_mode, str) and run_mode.strip().upper() == "REMOTE":
             return None
 
-        cfg = self._first_stdio_config(md)
+        cfg = self._stdio_config_for_engine(md, engine_type)
         if cfg is None:
             registry_detail = self._local_mcp_registry.get_mcp_detail(server_code)
-            cfg = self._first_stdio_config(registry_detail or {})
+            cfg = self._stdio_config_for_engine(registry_detail or {}, engine_type)
         if cfg is None:
             return None
         return StdioLaunch(
@@ -304,20 +304,36 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         )
 
     @staticmethod
-    def _first_stdio_config(detail: dict[str, Any]) -> dict[str, Any] | None:
-        """First launchable ``stdioConfigs`` entry, or ``None``.
+    def _stdio_config_for_engine(
+        detail: dict[str, Any], engine_type: str
+    ) -> dict[str, Any] | None:
+        """The launchable ``stdioConfigs`` entry for ``engine_type``, or ``None``.
+
+        A launch instruction is a path into a specific image, so the same
+        ``server_code`` can need a different one per engine — ``hitl`` ships at
+        ``/usr/local/bin`` on teclaw and under ``/home/admin`` elsewhere. An entry
+        naming this engine in ``engineType`` wins; an entry naming no engine is
+        the default for the rest. An entry for a *different* engine is never a
+        fallback: launching another image's binary is worse than not launching.
 
         An entry without a ``command`` is not a launch instruction, so it does not
         count as a hit — the caller then falls through to its next source rather
         than emitting a stdio server the engine cannot start.
         """
         configs = detail.get("stdioConfigs") or detail.get("stdio_configs") or []
-        if not isinstance(configs, list) or not configs:
+        if not isinstance(configs, list):
             return None
-        cfg = configs[0]
-        if not isinstance(cfg, dict) or not cfg.get("command"):
-            return None
-        return cfg
+
+        default: dict[str, Any] | None = None
+        for cfg in configs:
+            if not isinstance(cfg, dict) or not cfg.get("command"):
+                continue
+            declared = cfg.get("engineType") or cfg.get("engine_type")
+            if declared is None:
+                default = default if default is not None else cfg
+            elif str(declared).strip().lower() == engine_type.strip().lower():
+                return cfg
+        return default
 
     def _enrich_mcp_detail(
         self, svc: Any, md: dict[str, Any]
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -263,12 +263,19 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         keeps a community deployment, where the MCP Center plugin deliberately
         ignores that bundled file, from inheriting its company-only launch paths.
 
-        Otherwise :class:`LocalMCPRegistry` (a local YAML read) is consulted before
-        ``md``. That order is the point: the enrichment above is best-effort, so on
-        an MCP Center failure ``md`` carries neither ``runMode`` nor
-        ``stdioConfigs`` — and a local server, having no endpoint to find, would
-        then fail the remote path outright. The registry needs no network, so a
-        local server stays recognizable exactly when Center cannot vouch for it.
+        Otherwise the launch instruction is taken from the **resolved detail
+        first**, and only then from :class:`LocalMCPRegistry`. A deployment that
+        registers its own local server — including under a name the bundled
+        catalog also ships (``hitl``, ``clawmind``) — must keep its own command
+        rather than have it replaced by the bundled ``/home/admin/...`` path,
+        which that deployment's image need not even contain.
+
+        The registry is what makes the fallback work when there is nothing to
+        prefer: enrichment is best-effort, so on an MCP Center failure ``md``
+        carries neither ``runMode`` nor ``stdioConfigs``, and a local server —
+        having no endpoint to find — would otherwise fail the remote path
+        outright. Reading a local YAML needs no network, so a local server stays
+        recognizable exactly when Center cannot vouch for it.
 
         The registry normalizes the YAML's flat ``command``/``args``/``env`` into
         ``stdioConfigs: [{command, arguments, envVariables}]``; today that list is
@@ -280,12 +287,12 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         run_mode = md.get("run_mode") or md.get("runMode")
         if isinstance(run_mode, str) and run_mode.strip().upper() == "REMOTE":
             return None
-        detail = self._local_mcp_registry.get_mcp_detail(server_code) or md
-        configs = detail.get("stdioConfigs") or detail.get("stdio_configs") or []
-        if not configs or not isinstance(configs, list):
-            return None
-        cfg = configs[0]
-        if not isinstance(cfg, dict) or not cfg.get("command"):
+
+        cfg = self._first_stdio_config(md)
+        if cfg is None:
+            registry_detail = self._local_mcp_registry.get_mcp_detail(server_code)
+            cfg = self._first_stdio_config(registry_detail or {})
+        if cfg is None:
             return None
         return StdioLaunch(
             command=str(cfg["command"]),
@@ -295,6 +302,22 @@ class ConfigComposerInputCollector(ComposeInputCollector):
                 for k, v in (cfg.get("envVariables") or cfg.get("env") or {}).items()
             },
         )
+
+    @staticmethod
+    def _first_stdio_config(detail: dict[str, Any]) -> dict[str, Any] | None:
+        """First launchable ``stdioConfigs`` entry, or ``None``.
+
+        An entry without a ``command`` is not a launch instruction, so it does not
+        count as a hit — the caller then falls through to its next source rather
+        than emitting a stdio server the engine cannot start.
+        """
+        configs = detail.get("stdioConfigs") or detail.get("stdio_configs") or []
+        if not isinstance(configs, list) or not configs:
+            return None
+        cfg = configs[0]
+        if not isinstance(cfg, dict) or not cfg.get("command"):
+            return None
+        return cfg
 
     def _enrich_mcp_detail(
         self, svc: Any, md: dict[str, Any]
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -227,12 +227,21 @@ class ConfigComposerInputCollector(ComposeInputCollector):
     ) -> StdioLaunch | None:
         """Resolve a LOCAL server's launch instruction; ``None`` if it is remote.
 
-        Reads :class:`LocalMCPRegistry` (a local YAML file) **first**, falling back
-        to ``md`` only if the registry has no entry. That order is the point: the
-        enrichment above is best-effort, so on an MCP Center failure ``md`` carries
-        neither ``runMode`` nor ``stdioConfigs`` — and a local server, having no
-        endpoint to find, would then fail the remote path outright. The registry
-        needs no network, so a local server stays recognizable regardless.
+        A ``run_mode`` of REMOTE on the collected entry settles it. That value only
+        appears when the MCP Center lookup **succeeded**, and Center is
+        authoritative for the servers it knows, so a caller's own remote server is
+        never reclassified just because its ``server_code`` collides with a name in
+        the bundled registry (``hitl`` / ``clawmind``). The registry is a fallback
+        catalog for servers Center does not carry, not an override — which also
+        keeps a community deployment, where the MCP Center plugin deliberately
+        ignores that bundled file, from inheriting its company-only launch paths.
+
+        Otherwise :class:`LocalMCPRegistry` (a local YAML read) is consulted before
+        ``md``. That order is the point: the enrichment above is best-effort, so on
+        an MCP Center failure ``md`` carries neither ``runMode`` nor
+        ``stdioConfigs`` — and a local server, having no endpoint to find, would
+        then fail the remote path outright. The registry needs no network, so a
+        local server stays recognizable exactly when Center cannot vouch for it.
 
         The registry normalizes the YAML's flat ``command``/``args``/``env`` into
         ``stdioConfigs: [{command, arguments, envVariables}]``; today that list is
@@ -240,6 +249,9 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         discriminator the shape does not yet carry), so the first entry is taken.
         """
         if not server_code:
+            return None
+        run_mode = md.get("run_mode") or md.get("runMode")
+        if isinstance(run_mode, str) and run_mode.strip().upper() == "REMOTE":
             return None
         detail = self._local_mcp_registry.get_mcp_detail(server_code) or md
         configs = detail.get("stdioConfigs") or detail.get("stdio_configs") or []
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -36,6 +36,7 @@ from agentclaw.community.core.config_compose.models import (
 )
 from agentclaw.community.core.config_compose.protocols import ComposeInputCollector
 from agentclaw.community.core.config_compose.services.mcporter_composer import (
+    McporterComposeError,
     mcp_network_priority_for,
 )
 from agentclaw.community.core.mcp.services.config_service import MCPConfigService
@@ -56,6 +57,17 @@ if TYPE_CHECKING:
     from agentclaw.community.core.services.identity import IdentityService
 
 logger = get_logger()
+
+
+class McpDetailUnavailableError(LookupError):
+    """MCP Center returned no record for a server code.
+
+    Distinct from a transport error out of the Center client: this one says the
+    lookup *worked* and came back empty, which for a remote server means the code
+    is unknown to Center — a bad entry in the bot's MCP set or in the per-engine
+    default list. Raised as the chained cause so that distinction survives into
+    the traceback.
+    """
 
 
 def bot_data_relpath(host_path: str) -> str:
@@ -200,8 +212,23 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         network_priority = mcp_network_priority_for(req.engine_type)
         inputs: list[McpComposeInput] = []
         for md in raw:
-            md = self._enrich_mcp_detail(svc, md)
+            md, detail_failure = self._enrich_mcp_detail(svc, md)
             server_code = md.get("server_code") or md.get("serverCode") or ""
+            stdio = self._stdio_launch_for(server_code, md)
+            if stdio is None and detail_failure is not None:
+                # Remote server we could not resolve. Fail here, at the point the
+                # lookup actually failed, with the cause chained — rather than
+                # composing an entry with no endpoints and letting the composer
+                # report "no usable endpoint", which reads as a misconfigured
+                # server and sends the reader hunting in MCP Center for a record
+                # the lookup never retrieved. A local server takes the branch
+                # above instead: Center having no record of it is expected.
+                raise McporterComposeError(
+                    f"MCP {server_code or '<no server_code>'}: could not resolve "
+                    "server detail, and it is not a local server either (the "
+                    "local-MCP registry has no entry for it). MCP Center is "
+                    "unreachable or holds no record for this code."
+                ) from detail_failure
             api_key, headers, endpoint_env, transport = (
                 self._mcp_config_service.build_mcp_sync_payload(
                     user_id=req.user_id,
@@ -217,7 +244,7 @@ class ConfigComposerInputCollector(ComposeInputCollector):
                     endpoint_env=endpoint_env,
                     transport_protocol=transport,
                     network_priority=network_priority,
-                    stdio=self._stdio_launch_for(server_code, md),
+                    stdio=stdio,
                 )
             )
         return inputs
@@ -269,7 +296,9 @@ class ConfigComposerInputCollector(ComposeInputCollector):
             },
         )
 
-    def _enrich_mcp_detail(self, svc: Any, md: dict[str, Any]) -> dict[str, Any]:
+    def _enrich_mcp_detail(
+        self, svc: Any, md: dict[str, Any]
+    ) -> tuple[dict[str, Any], Exception | None]:
         """Merge MCP Center detail (endpoints/runMode/…) over a bare association.
 
         ``collect_bot_active_mcps`` returns only the skill-set association fields;
@@ -277,28 +306,28 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         detail per server (same source ``add_mcp_to_skill_set`` validates against)
         and merge it over the bare dict — Center is authoritative for endpoints,
         while locally-set fields absent from Center (e.g. default-MCP ``headers``)
-        are preserved. Best-effort: a missing detail or a Center error leaves the
-        bare dict unchanged (the composer then surfaces the "no usable endpoint"
-        error, same as before).
+        are preserved.
+
+        Returns ``(merged, failure)``. A lookup that raised or returned nothing
+        yields the bare dict plus the **cause**, which is deliberately *carried*
+        rather than logged and dropped: only the caller knows whether the server
+        is local — Center legitimately has no record of a stdio server — so only
+        the caller can decide whether the failure is fatal. It raises with this
+        exception chained, so the root cause survives into the traceback instead
+        of resurfacing three layers down as a misleading "no usable endpoint".
         """
         server_code = md.get("server_code") or md.get("serverCode")
         if not server_code:
-            return md
+            return md, None
         try:
             detail = svc.mcp_center.get_mcp_detail(server_code)
-        except Exception as e:
-            logger.warning(
-                "[ConfigComposerInputCollector] MCP Center detail fetch failed "
-                "for %s: %s", server_code, e,
-            )
-            return md
+        except Exception as e:  # noqa: BLE001 — returned to the caller, not swallowed
+            return md, e
         if not detail:
-            logger.warning(
-                "[ConfigComposerInputCollector] MCP Center has no detail for %s; "
-                "composing without endpoints", server_code,
+            return md, McpDetailUnavailableError(
+                f"MCP Center returned no detail for {server_code}"
             )
-            return md
-        return {**md, **detail}
+        return {**md, **detail}, None
 
     # ── resources ───────────────────────────────────────────────────────
     def resources(self, req: ComposeRequest) -> list[CollectedFile]:
--- a/src/backend/src/agentclaw/community/core/config_compose/services/mcporter_composer.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/mcporter_composer.py
@@ -285,18 +285,19 @@ class McporterComposer:
             if ep.get("env") == endpoint_env and ep.get("networkType") in net_rank
         ]
         if not candidates:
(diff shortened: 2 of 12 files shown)


src/backend/src/agentclaw/community/core/config_compose/services/collector.py
```diff
@@ -206,10 +217,46 @@ def mcps(self, req: ComposeRequest) -> list[McpComposeInput]:
                     endpoint_env=endpoint_env,
                     transport_protocol=transport,
                     network_priority=network_priority,
+                    stdio=self._stdio_launch_for(server_code, md),
```
totalfrank: Non-blocking: `_stdio_launch_for` (the registry-first/`md`-fallback resolution logic below, lines 225-258) is never exercised through `ConfigComposerInputCollector.mcps()` in this diff. `test_collector.py`'s existing `test_mcps_*` cases don't touch it (their server codes never match a registry entry and none assert on `inputs[i].stdio`), and the new coverage in `test_mcporter_composer.py` / `test_bot_config_artifact_schema.py` hand-builds `McpComposeInput(stdio=StdioLaunch(...))` directly, bypassing this method entirely. A regression in the registry-vs-`md` fallback order, the `configs[0]` selection, or the `cfg.get("command")` guard would go uncaught by CI. Worth a `test_collector.py` case that injects a fake/real `LocalMCPRegistry` returning `stdioConfigs` for a server code and asserts the resulting `McpComposeInput.stdio`.

---

src/backend/src/agentclaw/community/kernel/bot_config/artifact.py
```diff
@@ -97,22 +101,55 @@ class FileRef:
     path: str
 
 
+@dataclass(frozen=True)
+class StdioSpec:
+    """How to launch a local (stdio) MCP server.
+
+    Present only on entries whose ``transport`` is ``"STDIO"``. Unlike a remote
```
totalfrank: nit: docstring says transport is `"STDIO"` (uppercase), but the actual discriminator value used everywhere else — `STDIO_TRANSPORT = "stdio"` in `mcporter_composer.py`, the schema's `stdioSpec` description, `_compose_stdio` — is lowercase `"stdio"`. Worth matching the case so a reader grepping for the literal value isn't misled.

---

totalfrank (commented):
Reviewed commit 8140a5d (single commit, full diff base→head). Correctness, contract/schema round-tripping (`from_dict`/`to_dict`, append-only field order, `SCHEMA_VERSION` staying at 4), credential-dropping on the stdio path, and the remote-selection logic all check out — including the deliberate registry-over-`run_mode` classification design explained in the docstrings. Two non-blocking notes posted inline: a test-coverage gap on the collector's new `LocalMCPRegistry` resolution path, and a doc-casing nit. Nothing blocking.

---

src/backend/src/agentclaw/community/kernel/bot_config/artifact.py
```diff
@@ -97,22 +101,55 @@ class FileRef:
...
     name: str | None = None
-    endpoint: str | None = None
-    transport: str | None = None
-    headers: dict[str, str] = field(default_factory=dict)
+    endpoint: str | None = None  # remote form
+    transport: str | None = None  # discriminator, see above
+    headers: dict[str, str] = field(default_factory=dict)  # remote form
+    # Local form. ``None`` is a meaningful contract state, not an "unset"
+    # placeholder: a remote server genuinely has no launch instruction.
+    stdio: StdioSpec | None = None
```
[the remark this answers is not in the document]
totalfrank: Real tension, but holding at 4 is the deliberate call here — and the risk runs the other way.

The stdio form cannot be exercised yet. `_DEFAULT_MCP_SERVERS_BY_ENGINE` has no `teclaw` entry, so `get_default_mcp_servers("teclaw")` returns `[]`; a stdio entry only reaches an artifact if someone manually adds `hitl`/`clawmind` to a teclaw bot's skill set, and even then the engine cannot act on it until the teclaw team's runtime half lands (image carrying the interpreters, engine spawning and supervising the child process). Both halves are interlocked — neither is useful alone.

Bumping now would be the more disruptive move: nothing in this backend reads `schema_version` (it is written once at compose and never branched on), so the bump is purely an outbound signal. If the engine's parser gates on the value and does not yet know 5, publishing 5 breaks **every** artifact immediately — including the overwhelming majority that contain no stdio entry at all and are byte-identical to what v4 produced. Holding at 4 confines the exposure to artifacts that actually carry the new form, which today is none.

So the bump lands as its own commit once the engine owner confirms their parser accepts 5, coordinated with their runtime work rather than ahead of it. Compatibility coverage will land with it.

Two things narrow the window in the meantime, both in `5370346`:

- The schema now carries a two-branch `oneOf`, so a malformed entry is rejected by contract validation rather than at launch time.
- `from_dict` round-trips a v4 artifact unchanged (absent/null `stdio` → `None`), asserted in `tests/community/kernel/`.

---

src/backend/src/agentclaw/community/core/config_compose/services/collector.py
totalfrank: Fixed in 5370346 — six new `test_collector.py` cases now drive `_stdio_launch_for` through `mcps()` against a real `LocalMCPRegistry` (registry resolution, resolution when Center errors, REMOTE winning over a colliding registry name, a remote server absent from the catalog, a catalog entry with no command, and the `md` fallback).

---

src/backend/src/agentclaw/community/kernel/bot_config/artifact.py
totalfrank: Fixed in 5370346 — the docstring now reads `"stdio"` (lowercase), matching `STDIO_TRANSPORT` everywhere else.

---

totalfrank (commented):
Reviewed commits d6e9efd (merge from dev) and 5370346 (fix, current head) — the range since my last review.

**Findings checklist:**
- Test-coverage gap, `_stdio_launch_for` unexercised via `mcps()` (`collector.py:220`, mine) — **fixed**: six new `test_collector.py` cases now drive it through `mcps()` against a real `LocalMCPRegistry`. Replied and resolved.
- Docstring nit, `"STDIO"` vs `"stdio"` casing (`artifact.py`, mine) — **fixed**: now lowercase. Replied and resolved.
- P1, registry lookup not scoped to REMOTE (`collector.py:257`, Codex) — addressed by this range (`run_mode == REMOTE` short-circuit before the registry is consulted); not mine to resolve.
- P2, missing discriminated `oneOf` in schema (`artifact.schema.json`, Codex) — addressed by this range (two-branch `oneOf` + `stdioSpec.command` minLength); not mine to resolve.
- P1, `SCHEMA_VERSION` should bump (`artifact.py:152`, Codex) — still open; I already held the position once (deliberate, coordinated separately) in my prior reply, no further reply from the thread since. Leaving as-is.

**New commits (this range):** no new defects found. The registry-scoping fix, schema `oneOf`, and the `STREAMABLE_HTTP`→`http` fixture correction are all correct and match their stated intent; new collector tests exercise the real `LocalMCPRegistry` normalization path as claimed.

Open count: 1 (SCHEMA_VERSION, not mine — author's call already asserted).

---

src/backend/src/agentclaw/community/core/config_compose/services/mcporter_composer.py
```diff
@@ -243,9 +285,24 @@ def _select_by_priority(
...
+            # collector now fails the first case before we get here, so this is a
+            # backstop — but it still has to name the right suspect, or it sends
+            # the reader auditing network coverage for a record that was never
+            # fetched.
+            if not endpoints:
+                raise McporterComposeError(
+                    f"MCP {server_code}: server declares no endpoints at all — its "
+                    "detail was never resolved (MCP Center unreachable, or no "
+                    "record for this code)."
+                )
```
totalfrank, line 320: Non-blocking: this backstop's new "never resolved" message is asserted as always true, but it isn't.

By the time `_select_by_priority` runs, `330e1e48`'s collector-level raise has already ruled out `detail_failure is not None` — but that only proves the Center lookup didn't *fail*. It says nothing about whether the returned detail actually carries endpoints. `_enrich_mcp_detail` treats any truthy `detail` as success (`if not detail: return md, McpDetailUnavailableError(...)`), so a record like `{"runMode": "REMOTE"}` — Center knows the server, but it currently has zero published endpoints — merges cleanly with `detail_failure = None`, sails past the early raise, and lands here with `endpoints == []`. That hits:

```
MCP <code>: server declares no endpoints at all — its detail was never resolved (MCP Center unreachable, or no record for this code).
```

which is false in this case: the detail *was* resolved, Center *does* have a record, it just has no endpoints. That's exactly the misdirection this commit set out to eliminate (send the reader auditing MCP Center for a lookup that never happened), just reintroduced for a narrower input.

The new test (`test_no_endpoints_at_all_blames_the_lookup_not_the_networks`) doesn't catch this because it calls `compose_server` directly with a hand-built `md`, bypassing the collector's early-raise path entirely — so it can't distinguish "never resolved" from "resolved but empty" at this layer.

Suggested fix: thread the actual resolution outcome (or at least a bool) from the collector into `McpComposeInput`/`compose_server` instead of re-inferring "was this ever resolved" from an empty list at the composer layer, since an empty list is genuinely ambiguous between the two causes.

---

totalfrank (commented):
Reviewed commit 330e1e4 (the range since my last review at 5370346, current head).

**Findings checklist:**
- Test-coverage gap, `_stdio_launch_for` unexercised via `mcps()` (`collector.py`, mine) — fixed in 5370346, resolved.
- Docstring nit, `"STDIO"` vs `"stdio"` casing (`artifact.py`, mine) — fixed in 5370346, resolved.
- P1, `SCHEMA_VERSION` should bump (`artifact.py:152`, Codex) — I held the position once (deliberate, coordinated separately); thread is now resolved.
- P1, registry lookup not scoped to `REMOTE` (`collector.py`, Codex) — addressed by 5370346 (`run_mode == REMOTE` short-circuit before the registry is consulted); thread still open, not mine to resolve.
- P2, missing discriminated `oneOf` in schema (`artifact.schema.json`, Codex) — addressed by 5370346 (two-branch `oneOf` + `stdioSpec.command` minLength); thread still open (outdated), not mine to resolve.
- New (this round): misleading "detail was never resolved" backstop message in `_select_by_priority` can fire on a Center detail that *did* resolve but carries zero endpoints — filed inline, non-blocking.

**New commit (330e1e4):** the collector now fails eagerly at the point a remote server's Center lookup fails (chaining the cause), instead of composing an endpoint-less entry and letting the error surface three layers down. The intent and the bulk of the change are correct and well-tested (cause-chaining, `McpDetailUnavailableError` vs transport error distinction, local-server-survives-Center-outage case all covered). One non-blocking messaging gap found in the composer-side backstop, detailed inline.

Open count: 2 (1 not mine — SCHEMA_VERSION-adjacent registry-scoping/schema Codex threads are addressed-but-unresolved by their author's account, not a defect; 1 new non-blocking finding from this round).

---

src/backend/src/agentclaw/community/kernel/bot_config/artifact.py
```diff
@@ -97,22 +101,55 @@ class FileRef:
...
```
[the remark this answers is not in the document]
totalfrank: Correct, and it falsified the compatibility argument I made on the other thread. Fixed in `b2535b0`.

`to_dict` goes through `asdict`, which emits every field, so `"stdio": null` was landing on **every** MCP entry — including artifacts with no local server at all. My earlier "the published bytes are unchanged until a bot has a local MCP" was therefore wrong as written: it was the intent, not the behavior.

`to_dict` now drops the key when it is `None`. A remote-only artifact is back to exactly its prior five keys:

```json
{"server_code": "gh", "name": null, "endpoint": "https://x", "transport": "http", "headers": {"a": "b"}}
```

Verified the way this finding frames it — by reconstructing the pre-`stdio` definition (drop the `stdio` property, drop the `oneOf`, keep `additionalProperties: false`) and validating a remote-only artifact against it. It passes, so a consumer still coding to v4 accepts these. `EXAMPLE_ARTIFACT`, which does carry a stdio entry, still validates against the current schema, and `from_dict` round-trips both forms.

This is also what the "hold at 4" position actually needed to be true, so the two are now consistent rather than in tension: only artifacts that genuinely carry the new form differ from v4 output, and today none do.

---

src/backend/src/agentclaw/community/core/config_compose/services/mcporter_composer.py
totalfrank: Fixed in b2535b0 — verified: `_select_by_priority`'s zero-endpoints message no longer asserts the detail was never resolved. It now reports only what this frame can observe (`"the server record carries no endpoints at all, so there is nothing to select a {endpoint_env} URL from"`) and points at MCP Center's published endpoints instead of the lookup. The renamed test `test_no_endpoints_reports_only_what_this_layer_can_observe` pins this: it asserts `"never resolved" not in str(excinfo.value)`, closing the gap the old test missed by calling `compose_server` directly and bypassing the collector's early-raise path. Resolving.

---

totalfrank (commented):
Follow-up round: verified the fixes landed in `b2535b0` (round 3, "keep remote-MCP bytes unchanged and stop guessing causes") — all three of that round's findings are correctly fixed and covered by new tests. No new defects found in this commit.

- SCHEMA_VERSION should bump (Codex, P1, `artifact.py:152`) — disputed by author, deliberate/coordinated separately; position held once; **resolved**.
- Registry lookup not scoped to `REMOTE` (Codex, P1, `collector.py`) — fixed in `5370346` (`run_mode == REMOTE` short-circuits before the registry); Codex's thread, left for their own resolution.
- Missing discriminated `oneOf` in schema (Codex, P2, `artifact.schema.json`) — fixed in `5370346` (two-branch `oneOf` + `stdioSpec.command` minLength); Codex's thread, left for their own resolution.
- Test-coverage gap on `_stdio_launch_for` via `mcps()` (mine, `collector.py`) — fixed in `5370346`; replied, **resolved**.
- `"STDIO"` vs `"stdio"` docstring casing nit (mine, `artifact.py`) — fixed in `5370346`; replied, **resolved**.
- `stdio: null` breaking v4 wire compatibility (Codex, P1, `artifact.py:152`) — fixed in `b2535b0` (`to_dict` now pops `stdio` when `None`; verified against `test_remote_entry_omits_the_stdio_key_entirely`); replied confirming the fix, Codex's thread left for their own resolution.
- Bundled catalog overriding a deployment's own resolved local server (Codex, P2, `collector.py`) — fixed in `b2535b0` (resolved-detail checked first, registry is the fallback; verified against the new `test_mcps_resolved_local_definition_wins_over_bundled_catalog`); Codex's thread, left for their own resolution.
- Misleading "detail was never resolved" backstop message (mine, non-blocking, `mcporter_composer.py:301`) — fixed in `b2535b0` (message now reports only what the frame can observe); replied and **resolved** this round.

Open count (mine): 0. Three Codex threads remain open on GitHub but are verified fixed in code — left unresolved for Codex/maintainers per this review's convention of only resolving threads it opened.

---

src/backend/src/agentclaw/community/kernel/bot_config/artifact.schema.json
```diff
@@ -110,16 +110,55 @@
...
+            "headers": { "type": "object", "maxProperties": 0 }
+          }
+        },
+        {
+          "title": "remote (http/sse)",
+          "description": "transport 'http'/'sse' ⇒ a URL to connect to, and no launch instruction. 'endpoint' is required here even though it is nullable at the top level: an entry with neither a URL nor a launch instruction gives the engine nothing to act on.",
+          "required": ["transport", "endpoint"],
+          "properties": {
+            "transport": { "enum": ["http", "sse"] },
+            "endpoint": { "type": "string", "minLength": 1 },
```
totalfrank, line 147: Fixed in `2b8a6c854` — the remote branch's `oneOf` now requires `endpoint` with `minLength: 1`, so `{"transport":"http"}` without a URL is rejected by the schema on its own.

The finding's point that this is reachable from inside this backend too is also closed: `_select_by_priority` and the legacy `_select_endpoint` both now route through a new `_endpoint_tuple` helper (`mcporter_composer.py`) that raises `McporterComposeError` when the selected record has no usable `url`, naming the record it chose. With the guard in place neither selector can return `None` for the url, so their return annotations and `_inline_secrets` tighten from `str | None` to `str`. New tests (`test_selected_endpoint_without_a_url_raises`, `test_selected_endpoint_without_a_url_raises_on_the_legacy_path`) cover the guard on both selectors.

Leaving this thread open for Codex/maintainers to resolve, per this review's convention of only resolving threads it opened.

---

totalfrank (commented):
Reviewed commit `2b8a6c854` (the range since my last review at `b2535b0`, current head) — "reject a remote MCP entry that has no URL," addressing the fifth Codex finding.

**New commit (`2b8a6c854`):** no new defects found. The schema's remote branch now requires a non-empty `endpoint`, and both `_select_by_priority` and the legacy `_select_endpoint` route through a new `_endpoint_tuple` helper that raises when the selected record has no usable `url` — closing the gap from inside the backend too, not just at the schema boundary. Return types correctly tighten from `str | None` to `str` now that neither selector can produce `None`. New tests cover the guard on both selectors. Replied inline confirming the fix; left the Codex thread open per this review's convention of only resolving threads it opened itself.

- SCHEMA_VERSION should bump (Codex, P1, `artifact.py:152`) — disputed by author, deliberate/coordinated separately; position held once; **resolved**.
- Registry lookup not scoped to `REMOTE` (Codex, P1, `collector.py`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Missing discriminated `oneOf` in schema (Codex, P2, `artifact.schema.json`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Test-coverage gap on `_stdio_launch_for` via `mcps()` (mine, `collector.py`) — fixed in `5370346`; replied, **resolved**.
- `"STDIO"` vs `"stdio"` docstring casing nit (mine, `artifact.py`) — fixed in `5370346`; replied, **resolved**.
- `stdio: null` breaking v4 wire compatibility (Codex, P1, `artifact.py:152`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Bundled catalog overriding a deployment's own resolved local server (Codex, P2, `collector.py`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Misleading "detail was never resolved" backstop message (mine, non-blocking, `mcporter_composer.py`) — fixed in `b2535b0`; replied, **resolved**.
- Require an endpoint in the remote schema branch (Codex, P2, `artifact.schema.json`) — fixed in `2b8a6c854` (this round); replied, Codex's thread, left open for their own resolution.

Open count (mine): 0. Five Codex threads remain open on GitHub but are verified fixed in code, or an author's-call dispute the author held once — left unresolved for Codex/maintainers per this review's convention. No new defects in this round; PR is clean as of `2b8a6c854`.

---

totalfrank (commented):
Reviewed commit `624b0bbc6` (the range since my last review at `2b8a6c854`, current head) — a merge of `origin/REL20260818` into this branch.

**New commit (`624b0bbc6`):** no new defects — the merge touches none of this PR's own files. `git diff 2b8a6c854..624b0bbc6` only pulls in unrelated upstream changes (`bot_service.py`, `engine_runtime/{connection,models,relay,stage}.py` and their tests, from REL20260818). The PR's actual diff (`config_compose/*`, `bot_config/artifact.py`, `artifact.schema.json`, and their tests) is byte-identical to what it was at `2b8a6c854`, which was already fully reviewed.

- SCHEMA_VERSION should bump (Codex, P1, `artifact.py:152`) — disputed by author, deliberate/coordinated separately; position held once; **resolved**.
- Registry lookup not scoped to `REMOTE` (Codex, P1, `collector.py`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Missing discriminated `oneOf` in schema (Codex, P2, `artifact.schema.json`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Test-coverage gap on `_stdio_launch_for` via `mcps()` (mine, `collector.py`) — fixed in `5370346`; **resolved**.
- `"STDIO"` vs `"stdio"` docstring casing nit (mine, `artifact.py`) — fixed in `5370346`; **resolved**.
- `stdio: null` breaking v4 wire compatibility (Codex, P1, `artifact.py:152`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Bundled catalog overriding a deployment's own resolved local server (Codex, P2, `collector.py`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Misleading "detail was never resolved" backstop message (mine, non-blocking, `mcporter_composer.py`) — fixed in `b2535b0`; **resolved**.
- Require an endpoint in the remote schema branch (Codex, P2, `artifact.schema.json`) — fixed in `2b8a6c854`; Codex's thread, left open for their own resolution.

Open count (mine): 0. Five Codex threads remain open on GitHub but are verified fixed in code, or an author's-call dispute the author held once — left unresolved for Codex/maintainers per this review's convention. No new defects this round. PR is clean, cleared as of `2b8a6c854` and still clean at `624b0bbc6`.

---

src/backend/src/agentclaw/community/core/config_compose/services/collector.py
```diff
@@ -206,40 +244,129 @@ def mcps(self, req: ComposeRequest) -> list[McpComposeInput]:
...
+        always single-entry (a multi-variant form would need a per-variant
+        discriminator the shape does not yet carry), so the first entry is taken.
+        """
+        if not server_code:
+            return None
+        run_mode = md.get("run_mode") or md.get("runMode")
+        if isinstance(run_mode, str) and run_mode.strip().upper() == "REMOTE":
+            return None
+
+        cfg = self._stdio_config_for_engine(md, engine_type)
```
totalfrank, line 292: nit: `_stdio_launch_for`'s docstring (lines 280-283, not touched by this commit) still says the registry's `stdioConfigs` list is "always single-entry ... so the first entry is taken." This commit is exactly what makes that false — `_stdio_config_for_engine` (called here) now scans a multi-entry list and picks by `engineType`, as the new `hitl` catalog entry demonstrates. Worth updating that paragraph so a reader doesn't take the stale claim at face value.

---

totalfrank (commented):
Reviewed commit `a202b997` (the range since my last review at `624b0bbc6`, current head) — "let a local MCP server declare a launch instruction per engine."

**New commit (`a202b997`):** `local-mcp-servers.yaml` gains a per-engine `stdioConfigs` form for `hitl` (teclaw gets `/usr/local/bin/teclaw_hitl_mcp_server.py`, everyone else the `/home/admin/...` default), and `_stdio_config_for_engine` replaces `_first_stdio_config` with an engine-aware scan: an entry naming this engine wins, an entry naming none is the fallback default, and an entry for a *different* engine is never borrowed. Verified by reading: the selection logic, the REMOTE short-circuit and resolved-detail-before-registry ordering from earlier rounds are both preserved and unaffected by this change; the signature-change caller (`mcps()`) and the one place `_first_stdio_config` was called are both updated with no stale references left. Confirmed the new YAML parses correctly and matches the shape `_stdio_launch_for` reads (`arguments`/`command`/`engineType`). One non-blocking nit filed inline: a docstring paragraph a few lines above the touched code now describes stale (pre-this-commit) behavior.

- SCHEMA_VERSION should bump (Codex, P1, `artifact.py:152`) — disputed by author, deliberate/coordinated separately; position held once; **resolved**.
- Registry lookup not scoped to `REMOTE` (Codex, P1, `collector.py`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Missing discriminated `oneOf` in schema (Codex, P2, `artifact.schema.json`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Test-coverage gap on `_stdio_launch_for` via `mcps()` (mine, `collector.py`) — fixed in `5370346`; **resolved**.
- `"STDIO"` vs `"stdio"` docstring casing nit (mine, `artifact.py`) — fixed in `5370346`; **resolved**.
- `stdio: null` breaking v4 wire compatibility (Codex, P1, `artifact.py:152`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Bundled catalog overriding a deployment's own resolved local server (Codex, P2, `collector.py`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Misleading "detail was never resolved" backstop message (mine, non-blocking, `mcporter_composer.py`) — fixed in `b2535b0`; **resolved**.
- Require an endpoint in the remote schema branch (Codex, P2, `artifact.schema.json`) — fixed in `2b8a6c854`; Codex's thread, left open for their own resolution.
- Stale docstring claim ("stdioConfigs is always single-entry") after this commit adds multi-entry per-engine selection (mine, nit, `collector.py:282`) — **new this round**, filed inline.

Open count (mine): 1 (the new nit, non-blocking). Five Codex threads remain open on GitHub but are verified fixed in code, or an author's-call dispute the author held once — left unresolved for Codex/maintainers per this review's convention. No correctness defects found in `a202b997`.

---

src/backend/src/agentclaw/community/core/config_compose/services/collector.py
totalfrank: Fixed in `5e5722f4` — the docstring now reads "Either source speaks `stdioConfigs: [{command, arguments, envVariables}]`... The list holds one entry per engine layout, discriminated by `engineType`; `_stdio_config_for_engine` picks the one for this engine," matching the actual per-engine selection logic. Verified against the current `_stdio_config_for_engine` implementation. Resolving.

---

totalfrank (commented):
Reviewed commit `5e5722f4` (the range since my last review at `a202b997`, current head) — "correct a docstring the per-engine change invalidated."

**New commit (`5e5722f4`):** docstring-only change to `_stdio_launch_for` (`collector.py`), no behavior change — confirmed by diff (single file, comment text only). This is the fix for the nit I filed last round: the stale "`stdioConfigs` is always single-entry ... first entry is taken" claim is replaced with an accurate description of the per-engine `engineType` selection `_stdio_config_for_engine` performs. Verified the new text against the current implementation — accurate. Replied and resolved.

- SCHEMA_VERSION should bump (Codex, P1, `artifact.py:152`) — disputed by author, deliberate/coordinated separately; position held once; **resolved**.
- Registry lookup not scoped to `REMOTE` (Codex, P1, `collector.py`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Missing discriminated `oneOf` in schema (Codex, P2, `artifact.schema.json`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Test-coverage gap on `_stdio_launch_for` via `mcps()` (mine, `collector.py`) — fixed in `5370346`; **resolved**.
- `"STDIO"` vs `"stdio"` docstring casing nit (mine, `artifact.py`) — fixed in `5370346`; **resolved**.
- `stdio: null` breaking v4 wire compatibility (Codex, P1, `artifact.py:152`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Bundled catalog overriding a deployment's own resolved local server (Codex, P2, `collector.py`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Misleading "detail was never resolved" backstop message (mine, non-blocking, `mcporter_composer.py`) — fixed in `b2535b0`; **resolved**.
- Require an endpoint in the remote schema branch (Codex, P2, `artifact.schema.json`) — fixed in `2b8a6c854`; Codex's thread, left open for their own resolution.
- Stale docstring claim on `stdioConfigs` selection (mine, nit, `collector.py:292`) — fixed in `5e5722f4` (this round); **resolved**.

Open count (mine): 0. Five Codex threads remain open on GitHub but are verified fixed in code, or an author's-call dispute the author held once — left unresolved for Codex/maintainers per this review's convention. No new defects. PR is clean as of `5e5722f4`.

---

totalfrank (commented):
Reviewed commit `9db0d6a4` (the range since my last review at `5e5722f4`, current head) — "put the engine-agnostic hitl launch first, ahead of teclaw's."

**New commit (`9db0d6a4`):** no new defects found. This reorders `hitl`'s two `local-mcp-servers.yaml` `stdioConfigs` entries so the engine-agnostic one is first, fixing an ordering hazard for the out-of-repo arca/baas device path (`convert_to_device_format`) that reads `stdioConfigs[0]` with no `engineType` awareness. Verified this repo's own consumer is genuinely order-blind: `_stdio_config_for_engine` (`collector.py:308`) scans the full list and matches by `engineType`, never indexes — so the reorder cannot regress teclaw's own resolution. Verified `_normalize_stdio_configs` (`local_mcp_registry.py`) preserves list order (no reordering during normalization), and hand-parsed the edited YAML directly: `hitl.stdioConfigs` is now `[{command: python3, arguments: [.../home/admin/hitl/hitl_mcp_server.py]}, {engineType: teclaw, command: python3, arguments: [.../usr/local/bin/teclaw_hitl_mcp_server.py]}]`, matching the commit's stated intent. The new `test_shipped_multi_variant_servers_put_the_engine_agnostic_entry_first` guard reads the real shipped config via `LocalMCPRegistry()`'s default config path (not a fake/mocked catalog), so it will actually catch a future reorder. Docstring/YAML-header comment addition is accurate and consistent with the code.

**Part A — open findings:** none of mine are open. All were already verified fixed and resolved as of the `5e5722f4` round (test-coverage gap, docstring casing, misleading backstop message, stale docstring claim). The five Codex (`chatgpt-codex-connector[bot]`) threads that remain unresolved on GitHub were already verified fixed in code in prior rounds and left open per this review's standing convention of only resolving threads it opened itself — no change in this round.

**Findings checklist (all rounds, unchanged from last round):**
- SCHEMA_VERSION should bump (Codex, P1, `artifact.py:152`) — disputed by author, deliberate/coordinated separately; position held once; **resolved**.
- Registry lookup not scoped to `REMOTE` (Codex, P1, `collector.py`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Missing discriminated `oneOf` in schema (Codex, P2, `artifact.schema.json`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Test-coverage gap on `_stdio_launch_for` via `mcps()` (mine, `collector.py`) — fixed in `5370346`; **resolved**.
- `"STDIO"` vs `"stdio"` docstring casing nit (mine, `artifact.py`) — fixed in `5370346`; **resolved**.
- `stdio: null` breaking v4 wire compatibility (Codex, P1, `artifact.py:152`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Bundled catalog overriding a deployment's own resolved local server (Codex, P2, `collector.py`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Misleading "detail was never resolved" backstop message (mine, non-blocking, `mcporter_composer.py`) — fixed in `b2535b0`; **resolved**.
- Require an endpoint in the remote schema branch (Codex, P2, `artifact.schema.json`) — fixed in `2b8a6c854`; Codex's thread, left open for their own resolution.
- Stale docstring claim on `stdioConfigs` selection (mine, nit, `collector.py:292`) — fixed in `5e5722f4`; **resolved**.

Open count (mine): 0. No new findings this round. PR is clean as of `9db0d6a4`. (Note: at time of this review, "Backend unit tests" and "Singlebox coverage" CI checks were still in progress on this SHA; all other checks — Gateway, BaaS, BCS, Engine unit tests, BCS e2e — had already passed.)

---

totalfrank (commented):
Reviewed commit `9a039d2f` (the range since my last review at `9db0d6a4`, current head) — "use one shape for every entry in the local-MCP catalog."

**New commit (`9a039d2f`):** no new defects found. `clawmind` moves from the flat `command`/`args`/`env` shorthand to the explicit `stdioConfigs` form already used by `hitl`, closing the two-shapes/two-key-names hazard described in the commit message. Verified the "byte-identical" claim by reading `LocalMCPRegistry._normalize_stdio_configs` (`local_mcp_registry.py:177-190`): for the old flat form it builds `[{"command": "node", "arguments": [...], "envVariables": {...}}]` from `command`/`args`/`env`; for the new explicit form it returns that same list verbatim from `stdioConfigs`. Both normalize to an identical dict, and the only in-repo consumer (`_stdio_config_for_engine` / `_stdio_launch_for` in `collector.py`) reads exclusively through the normalized `stdioConfigs`/`stdio_configs` keys, never the raw `command`/`args`/`env`, so nothing here can observe a difference. `test_default_config_path_loads_repo_clawmind_config` already asserted against the normalized `stdioConfigs` shape, so it exercises the new form unchanged and would have caught a real divergence. Confirmed `clawmind` stays single-entry, so `test_shipped_multi_variant_servers_put_the_engine_agnostic_entry_first` (which only checks lists of length ≥2) is correctly unaffected. YAML is well-formed and the updated header comment accurately describes the new all-`stdioConfigs` shape and the now-secondary flat shorthand.

**Part A — open findings:** none of mine are open; unchanged from the last round (test-coverage gap, docstring casing, misleading backstop message, and the stale `stdioConfigs` docstring claim were all verified fixed and resolved in prior rounds). The five Codex (`chatgpt-codex-connector[bot]`) threads remain open on GitHub but were already verified fixed in code, or are the author's-call SCHEMA_VERSION dispute held once — left open per this review's standing convention of only resolving threads it opened itself; no change this round.

- SCHEMA_VERSION should bump (Codex, P1, `artifact.py:152`) — disputed by author, deliberate/coordinated separately; position held once; **resolved**.
- Registry lookup not scoped to `REMOTE` (Codex, P1, `collector.py`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Missing discriminated `oneOf` in schema (Codex, P2, `artifact.schema.json`) — fixed in `5370346`; Codex's thread, left open for their own resolution.
- Test-coverage gap on `_stdio_launch_for` via `mcps()` (mine, `collector.py`) — fixed in `5370346`; **resolved**.
- `"STDIO"` vs `"stdio"` docstring casing nit (mine, `artifact.py`) — fixed in `5370346`; **resolved**.
- `stdio: null` breaking v4 wire compatibility (Codex, P1, `artifact.py:152`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Bundled catalog overriding a deployment's own resolved local server (Codex, P2, `collector.py`) — fixed in `b2535b0`; Codex's thread, left open for their own resolution.
- Misleading "detail was never resolved" backstop message (mine, non-blocking, `mcporter_composer.py`) — fixed in `b2535b0`; **resolved**.
- Require an endpoint in the remote schema branch (Codex, P2, `artifact.schema.json`) — fixed in `2b8a6c854`; Codex's thread, left open for their own resolution.
- Stale docstring claim on `stdioConfigs` selection (mine, nit, `collector.py:292`) — fixed in `5e5722f4`; **resolved**.

Open count (mine): 0. No new findings this round. PR is clean as of `9a039d2f`. (Note: at time of this review, "Backend unit tests" and "Singlebox coverage" CI checks were still in progress on this SHA; Gateway, BaaS, BCS, Engine unit tests, and BCS e2e had already passed.)

---

totalfrank (commented):
Reviewed commit `56a638ed` — the branch was rebased onto `REL20260818` since my last review (at `9a039d2f`, submitted against a commit that is no longer reachable — none of the 10 commits currently on the PR share a SHA with what I reviewed before). Per the runbook for this case, I diffed the current merge base (`64670d2b`) against the current head and re-verified every prior finding by reading the actual code at HEAD, rather than trusting GitHub's now-outdated thread-resolved flags.

**Rebase fidelity check:** every file the PR touches (`artifact.py`, `artifact.schema.json`, `collector.py`, `mcporter_composer.py`, `models.py`, `local-mcp-servers.yaml`, `__init__.py`) was re-read in full at the current head and matches the state my prior rounds verified — REMOTE short-circuit before the registry, resolved-detail-before-registry ordering, the schema's two-branch `oneOf` with required non-empty `endpoint`/`command`, `to_dict` dropping `stdio` when `None`, the `_endpoint_tuple` guard, the non-misleading backstop message, the accurate `_stdio_config_for_engine` docstring, and the engine-agnostic-`hitl`-entry-first YAML ordering. Also ran the full changed-file test suite directly against this exact head commit: **94 passed**, covering `test_collector.py`, `test_mcporter_composer.py`, `test_local_mcp_registry.py`, `test_bot_config_artifact.py`, and `test_bot_config_artifact_schema.py`. CI on this SHA: 5/7 checks green, 2 (`Backend unit tests`, `Singlebox coverage`) still in progress at review time.

One real thing the rebase did do: the commit message on `56a638ed` itself explains that rebasing onto `REL20260818` dropped a merge commit, and the `"STDIO"`→`"stdio"` docstring casing fix I verified fixed and resolved several rounds ago rode inside that merge commit — so it was transiently reverted, then restored by this same commit. Confirmed the current code (`artifact.py` line 108) is lowercase `"stdio"` again. No reply needed since my existing resolved thread on that finding already states the correct end state and nothing else changed.

**Part A — open findings:** none of mine are open, unchanged from last round. The five Codex (`chatgpt-codex-connector[bot]`) threads that remain open on GitHub (SCHEMA_VERSION bump, registry-scoping, schema `oneOf`, `stdio: null` compatibility, endpoint-required) were all already verified fixed in code in prior rounds, or are the author's-call SCHEMA_VERSION dispute held once — re-verified fixed/unchanged at this head, left open per this review's standing convention of only resolving threads it opened itself.

**Part B — new commits:** none, in substance. The diff from the current merge base to HEAD is content-identical to what was already fully reviewed across rounds 1–9; the only genuinely new commit is the docstring-only casing restore, which is a straight revert of the rebase's own damage. No new defects found.

- SCHEMA_VERSION should bump (Codex, P1, `artifact.py:152`) — disputed by author, deliberate/coordinated separately; position held once; **resolved**.
- Registry lookup not scoped to `REMOTE` (Codex, P1, `collector.py`) — fixed; re-verified intact post-rebase; Codex's thread, left open for their own resolution.
- Missing discriminated `oneOf` in schema (Codex, P2, `artifact.schema.json`) — fixed; re-verified intact post-rebase; Codex's thread, left open for their own resolution.
- Test-coverage gap on `_stdio_launch_for` via `mcps()` (mine, `collector.py`) — fixed; re-verified (tests pass); **resolved**.
- `"STDIO"` vs `"stdio"` docstring casing nit (mine, `artifact.py`) — fixed, transiently dropped by the rebase, restored in `56a638ed`; **resolved**.
- `stdio: null` breaking v4 wire compatibility (Codex, P1, `artifact.py:152`) — fixed; re-verified intact post-rebase; Codex's thread, left open for their own resolution.
- Bundled catalog overriding a deployment's own resolved local server (Codex, P2, `collector.py`) — fixed; re-verified intact post-rebase; Codex's thread, left open for their own resolution.
- Misleading "detail was never resolved" backstop message (mine, non-blocking, `mcporter_composer.py`) — fixed; re-verified (tests pass); **resolved**.
- Require an endpoint in the remote schema branch (Codex, P2, `artifact.schema.json`) — fixed; re-verified intact post-rebase; Codex's thread, left open for their own resolution.
- Stale docstring claim on `stdioConfigs` selection (mine, nit, `collector.py`) — fixed; re-verified; **resolved**.

Open count (mine): 0. PR is clean as of `56a638ed`, unchanged in substance from `9a039d2f`, with the one rebase-induced regression (docstring casing) already self-corrected by the branch's own final commit.

---

FreddieSun (approved):
lgtm

PR #1084, opened 2026-08-16
chore(gateway): republish the bots schema with the stage parameter

totalfrank:
## Problem

`src/gateway/configs/schemas/bots.openapi.json` is the published description the
gateway generates its **served** OpenAPI document from (`FileSchemaCatalog` →
`build_served_openapi`). It was last regenerated by #1074 and so predates #1075,
which added `?stage=` to the five per-bot file operations:

```text
GET, PUT  /openapi/v1/bots/{bot_id}/engine/config
GET       /openapi/v1/bots/{bot_id}/identity
GET, PUT  /openapi/v1/bots/{bot_id}/identity/{file_type}
```

The artifact already carries `stage` on the 32 engine-runtime operations that
took it on 2026-08-09, so the five new ones are the only surface where the
served doc disagrees with the backend.

**Forwarding was never affected.** `_target_url` appends `request.url.query`
verbatim and route matching is by path pattern, not against the schema, so
`?stage=online` reaches the backend through the gateway today. What is missing
is discovery: a tenant reading the served document — or generating a client from
it — has no way to learn the parameter exists.

Nothing would have caught this. `scripts/dump_and_publish.sh` is run by hand at
release; there is no CI freshness gate, which is why #1075 merged green.

## Solution

Regenerated and republished through the documented path, no hand edits:

```sh
cd src/backend && DEPLOY_PROFILE=community uv run python scripts/dump_openapi.py /tmp/backend.openapi.json
cd src/gateway && uv run python scripts/gate_and_publish_openapi.py configs/schemas/bots.openapi.json /tmp/backend.openapi.json
```

The gate reported `compatible` and published without `--allow-breaking`.

The diff is exactly what #1075 changed in the backend and nothing else:

- **`stage` added to the five operations above.** Optional, in the query,
  `default: "draft"`, `$ref: RuntimeStage`.
- **Ten descriptions updated** — those five plus their five deprecated twins
  (`/bots/identity/{bot_id}`, `/bots/identity/{bot_id}/{file_type}`,
  `/bots/{bot_id}/engine-config`), which gained their "use … instead" pointers
  and the wording that says they read the bot's own workspace.
- **The deprecated twins do not gain `stage`**, matching the frozen contract.
- `components` and every top-level key are byte-identical; `RuntimeStage` was
  already published.

## Validation

- `check_compatible(published, candidate)` → **0 breaking changes**; the gate
  published on its own terms rather than under `--allow-breaking`.
- Operation count unchanged at 112, no path added or removed.
- Verified the parameter diff programmatically: 5 operations gain `stage`, 10
  descriptions change, no other key on any operation differs.
- Regenerated against current `dev` (`d8315ca`), not the merged branch head, and
  confirmed the delta is identical at both.

**Not run:** the gateway test suite. Its venv needs `mirrors.aliyun.com`, which
this environment's network policy denies; the dependency resolve above went to
`pypi.org` instead, and `src/gateway/uv.lock` was deliberately left untouched by
that detour. This change adds no code — the affected gateway tests
(`test_served_openapi.py`, `test_openapi_generator.py`) run against
`tests/fixtures/`, not this artifact.

## Compatibility and risk

Additive to the served document only, and additive within it: a new optional
query parameter with a default that preserves today's behaviour. No backend,
gateway, or client code changes. Rollback is the revert.

## Related issues

- Publishes the surface added by #1075.
- Follows the same in-repo republish #1074 did for its own change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GZmvE8VyoQK89sJD3w68Nc)_

### DIFF
--- a/src/gateway/configs/schemas/bots.openapi.json
+++ b/src/gateway/configs/schemas/bots.openapi.json
@@ -7732,7 +7732,7 @@
     "/openapi/v1/bots/identity/{bot_id}": {
       "get": {
         "deprecated": true,
-        "description": "List every identity file type a bot can carry and whether each exists.\n\nA file reports exists false both when it is absent and when it exists\nwith empty content. Entry order is not guaranteed — key off type.",
+        "description": "List every identity file type a bot can carry and whether each exists.\n\nEvery entry comes from the bot's own workspace.\n\nA file reports exists false both when it is absent and when it exists\nwith empty content. Entry order is not guaranteed — key off type.\n\nDeprecated: use GET /openapi/v1/bots/{bot_id}/identity instead.",
         "operationId": "list_bot_identity_files_openapi_v1_bots_identity__bot_id__get",
         "parameters": [
           {
@@ -7900,7 +7900,7 @@
     "/openapi/v1/bots/identity/{bot_id}/{file_type}": {
       "get": {
         "deprecated": true,
-        "description": "Read one identity file of a bot.\n\nA file that has never been written reads as an empty content string, not\nan error. Reads address the bot's draft runtime.",
+        "description": "Read one identity file of a bot.\n\nReads the bot's own workspace. Reaching a published runtime is offered at\nthe address that replaces this one.\n\nA file that has never been written reads as an empty content string, not\nan error.\n\nDeprecated: use GET /openapi/v1/bots/{bot_id}/identity/{file_type} instead.",
         "operationId": "get_bot_identity_file_openapi_v1_bots_identity__bot_id___file_type__get",
         "parameters": [
           {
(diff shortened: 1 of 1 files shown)


totalfrank (commented):
Reviewed commit `2e50c630a5596f7099b32cfbf1d01c16cc5b70dd` (the PR's only commit). Diff clean — no findings.

**File covered:** `src/gateway/configs/schemas/bots.openapi.json` (the only changed file; +65/-10, no code).

**What I checked:**
- **Correctness** — the five new `stage` query parameters (on the two `/bots/{bot_id}/engine/config` and three `/bots/{bot_id}/identity...` operations) are byte-identical in shape to the 32 pre-existing `stage` parameters elsewhere in the same document (verified programmatically against `dev`'s copy of the file): same `$ref: RuntimeStage`, `default: "draft"`, and parameter placement convention.
- **Contract breakage** — additive only. Operation count is unchanged at 112 before/after. The three deprecated twin operations got description-only edits and correctly did **not** gain the `stage` parameter, matching the "frozen contract" claim in the PR description.
- **Schema validity** — the `$ref` co-located with sibling `default`/`description` keys is valid here since the document declares `"openapi": "3.1.0"` (JSON Schema 2020-12 semantics, where `$ref` siblings are not discarded, unlike OpenAPI 3.0).
- **Test coverage** — N/A; this is a generated data artifact, not code. PR description correctly notes the affected gateway tests run against fixtures, not this file.
- Categories not applicable to this change: error/failure handling, concurrency, security/hygiene, architecture boundaries, Python type contracts.

---

vzvince (approved):

PR #1085, opened 2026-08-17
fix(bot-run): set max_size=10mb on WebSocket client to allow large event

xxxxpenny:
## Problem

The `websockets` library defaults `max_size` to 1MiB (2²⁰ bytes). When a bot runtime event exceeds this limit — e.g. large agent events with attachments or long streaming responses — the connection is torn down with `ConnectionClosedError`, causing the chat session to fail mid-conversation.

## Solution

Add a `max_size: int | None = None` parameter to `BotWebsocketClient.connect()` and pass it through to `websockets.connect()`. `None` removes the size limit entirely.

All existing call sites (`AsyncChatClient.connect()`, `_async_chat_client_pool.py`) use default arguments, so the fix takes effect automatically without caller changes.

Considered setting a larger fixed limit (e.g. 16MiB) instead of unlimited, but since event sizes are driven by upstream model output and vary unpredictably, any fixed ceiling risks the same failure under different conditions. The transport is internal (baas ↔ sandbox), so unbounded framing is acceptable.

## Validation

- Lint check via IDE — the two `basedpyright` errors on `dataclasses.asdict` are pre-existing and unrelated to this change.
- No new tests added; the change is a single parameter passthrough to a well-understood library option.

### DIFF
--- a/src/baas/src/secbaas/community/core/service/bot_run/_bot_websocket_client.py
+++ b/src/baas/src/secbaas/community/core/service/bot_run/_bot_websocket_client.py
@@ -124,6 +124,7 @@ class BotWebSocketClient:
         timeout: float = 10.0,
         disable_ssl_verify: bool = True,
         open_timeout: float = 5.0,
+        max_size: int = 10 * 2**20,
     ) -> dict[str, Any]:
         """连接到服务器并完成握手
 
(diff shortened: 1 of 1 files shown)


pfmiles (approved):
ok

PR #1106, opened 2026-08-17
feat(bcs-channel): add /new slash command to archive session and start fresh

RaymondHX:
## Problem
Channel chat sessions never auto-complete (`timeout_scanner` only covers `service_invocation`), so a user's IM conversation accumulates context indefinitely across every message in the same BCS session. There is no way for an end user to start a brand-new conversation from within an IM channel (DingTalk etc.) — they are stuck with the original session forever.

## Solution
Add a `/new` slash command, intercepted uniformly in the shared channel service layer (works for every channel type; the DingTalk provider itself is out-of-repo). Semantics, confirmed with the requester:

- `/new` **archives** the current session (`Running -> Completed` via `complete_if_running`; message history preserved and still queryable), and the conversation binding rolls over to a fresh session on the next message via the existing lazy-rollover path (mapping kept as the trigger; new `group_id`-prefixed random session id). No `delete_if_session` needed.
- If the current session has an agent run in flight, `/new` **queues** the reset: it replies "任务仍在运行中" and executes the switch automatically once the run terminates.
- Channel-layer in-memory `SessionResetTracker` snapshots the in-flight runs at `/new` time; terminal events (chat `ChatFinal`/`error`/`aborted` via `try_outbound`; state-machine runs via `publish_state_machine_terminal`, which fires on every terminal even with no IM notification) drain the wait and execute the queued reset. A 30-min staleness guard runs a pending reset off the next inbound message if a terminal event never arrives.
- Interception sits **before** HumanInput consumption so a pending input card never swallows `/new` as its reply.
- Replies use direct provider delivery (`kind: System`, `purpose: HumanInputAck`) so the feature works even when `/new` is the very first message (no session to route through `try_outbound`); `bcs_session_id` is empty in that edge case — verified against the out-of-repo DingTalk provider that it tolerates empty (the field is tracing-only there, routing is keyed on `account_ref` + `im_conversation_*`).
- A new `chat_session_resolution_lock` closes the "two concurrent messages both see Completed and create two sessions" race; dispatch stays outside the lock.

Files:
- `src/bcs/crates/services/bcs-channel/src/commands.rs` (new): extensible command parser, `SessionResetTracker`, and the `BcsChannelService` command handlers.
- `src/bcs/crates/services/bcs-channel/src/lib.rs`: eight surgical hook points (struct fields, interception, staleness piggyback, resolution lock, `seed_runs` after dispatch, and the two terminal hooks). No service-api / repo-port / store / provider changes.
- `docs/superpowers/plans/2026-08-17-channel-new-command.md`: design doc.

## Validation
- `cargo test --package bcs-channel`: 83 passed (59 baseline + 24 new: 12 `commands.rs` unit tests for parser + tracker state machine; 12 integration tests covering no-session, idle-rollover, queue+ChatFinal, error terminal, per-sender isolation, shared group, state-machine idle/running/starting-window, HumanInput-not-consumed, concurrent idempotence, and stale-reset).
- `cargo clippy --package bcs-channel --no-deps`: zero warnings on `commands.rs`.
- `cargo test --workspace`: full regression exit 0, no failures.
- DingTalk provider (out-of-repo, `ocb_2`) tolerance of empty `bcs_session_id` confirmed by code audit of its `deliver_event` path.

### DIFF
--- a/docs/superpowers/plans/2026-08-17-channel-new-command.md
+++ b/docs/superpowers/plans/2026-08-17-channel-new-command.md
@@ -1,0 +1,81 @@
+# BCS Channel `/new` 命令：归档当前会话并开启全新会话
+
+> 状态：已实现（2026-08-17，worktree `bcs-channel-new-command`）
+> 设计/分析过程见本文档；实现位于 `src/bcs/crates/services/bcs-channel/src/commands.rs` + `lib.rs` 挂载点。
+
+## Context
+
+Channel（钉钉等 IM）用户的 chat session 一旦创建就**永不自动结束**（timeout_scanner 只覆盖 service_invocation，chat 会话无超时），同一 IM 会话里的消息永远进入同一个 BCS session，上下文无限累积。`/new` 让用户主动开启全新会话。
+
+**需求语义**（与需求方确认）：
+1. `/new` 后旧会话**归档**（Running → Completed，历史保留可查），后续消息进入全新会话
+2. **所有 channel 统一支持** —— 在 bcs-channel 服务层实现（DingTalk provider 在仓库外）
+3. 当前会话有 agent run 在途时**排队等待**：先回复"任务运行中"，run 结束后自动执行切换
+
+## 关键架构事实
+
+- **唯一拦截点**：`BcsChannelService::handle_inbound`（`bcs-channel/src/lib.rs`），所有 channel 文本消息的唯一入口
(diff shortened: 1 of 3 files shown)


vzvince (approved):
GLTM

FreddieSun (approved):
LGTM

totalfrank (approved):

Issue #1109, opened 2026-08-17
feat(openapi-v1): no way to request MCP server access from the public API


totalfrank:
## Problem

`GET /openapi/v1/bots/mcp/servers/{server_code}/permissions` can tell an external caller `has_access: false`, and publicly that is the end of the road. There is no way to *request* the access it just reported as missing.

The internal surface has one: `POST /api/mcp/market/permission/apply` (`src/backend/src/agentclaw/community/adapters/http/mcp/router.py:170`), which calls `MCPAuthService.apply_permission(staff_no, service_code, tool_list, is_public, reason)` and returns a `process_url` — a link into an approval workflow a human acts on. It is not published on `openapi_v1`.

## Context — what the access actually is

`check_mcp_permission_detail` (`core/mcp/services/auth_service.py:37`) merges two sources:

1. `accessLevel` from MCP Center's server detail → `is_public = access_level == "PUBLIC"`.
2. `auth_client.query_permission_status(...)`, returning `{resource_code, auth_state}` rows where `auth_state` is `AUTHORIZED` / `UNAUTHORIZED`.

Resource codes come in two shapes: a bare `server_code` (whole-server grant) and `server_code/tool_name` (per-tool grant) — which is where `has_access` and the `tool_permissions` map come from. Local MCP servers short-circuit via `_is_local_mcp` to `has_permission: True, access_level: "LOCAL"` with no external call.

So the missing capability is: request a grant on a non-public marketplace server, optionally scoped to specific tools, for the calling user.

## Scope caveat

Both `MCPAuthClient` implementations in this repo are mocks — `plugins/community/mcp_auth.py:41-49` returns `{success: True, process_url: None}` and its `query_permission_status` reports `AUTHORIZED` for everything; `plugins/local/mcp_auth.py` does the same. The real client lives in the corp package.

**This gap therefore only has an effect in a deployment running the real auth plugin.** In community/OSS everything is already authorized and there is nothing to apply for. Worth weighing before prioritizing.

## Design questions

- Should the public endpoint expose `is_public`, or derive it from the server's `accessLevel` the way `check_mcp_permission_detail` already does? The internal one takes it from the caller.
- The internal handler catches every exception and returns `success: false` with the message in `error`. The public surface would need this mapped onto the standard `Envelope` error contract instead.
- `process_url` is an external workflow link. Confirm it is safe to return to an external tenant.

## Out of scope

[...]

PR #1110, opened 2026-08-17
feat(openapi-v1): MCP config lifecycle and bot-scoped activation

totalfrank:
## Problem

The public API's `mcp` category could store a credential for a marketplace MCP server, and that was all it could do with one.

**Config lifecycle.** `PUT .../servers/{server_code}/config` is merge-not-replace (`config_service.py:124` — `api_key if api_key is not None else old_extra.get("api_key")`), so `null` means *keep the old value*, and there was no `DELETE`. A stored credential could be overwritten but never removed; revoking a leaked key meant replacing it with a decoy. There was also no way to list what the caller had configured — only per-server `GET` — so a tenant had to already know every `server_code` to enumerate its own state. `UserMCPConfigRepository.delete` and `list_by_user` both existed and were simply unpublished.

**Activation.** A bot's MCP set comes from `collect_bot_active_mcps` = active skill sets' MCPs + engine defaults − exclusions. The public API could reach none of those three inputs, so an external tenant could hold a valid API key for a real server indefinitely and never have an agent call it.

That second gap is the load-bearing one: it is not a defect in the write path. `ac_user_mcp_config` is the source of truth and `build_mcp_sync_payload` reads the stored row back on every later sync and on artifact compose, so storing config for a server no bot yet carries is a legitimate write-ahead. The missing half was putting the server on a bot.

## Solution

Eight new endpoints, taking the category from 6 to 14.

**Account-level** (`openapi_v1/mcp/router.py`) — `GET /bots/mcp/configs` and `DELETE /bots/mcp/servers/{server_code}/config`. Both `REFUSED` in the admission table, joining the config operations they sit with: enumerating an account's configured servers is a question about the account rather than any bot a grant covers, and delete is the only operation that can destroy a credential.

**Bot-scoped** (`openapi_v1/bot_mcp/`) — six operations at `/openapi/v1/bots/{bot_id}/mcp`: list, get, add, activate, deactivate, remove. All `GRANT_CHECKED_OWN_BOT`, narrower than `skills`, which uses `ADDRESSED_BOT` on its collection routes because that group serves shared bots.

### The activation mechanism, and why there is no schema change

Activation is the mechanism public `skills` already runs in production. `set_local_skill_active` never touches skill-set activation: it works inside the bot's *default* skill set and uses exclusion rows as the per-skill off switch. Every MCP twin already existed.

| verb | storage | skills equivalent |
|---|---|---|
| add | `add_mcp_to_set` on the default set + exclusion row → inactive | `_ensure_default_set_membership` |
| activate | `remove_all_default_mcp_exclusions` | `remove_all_default_skill_exclusions` |
| deactivate | `add_default_mcp_exclusion` | `add_default_skill_exclusion` |
| remove | `remove_mcp_from_set` + clear exclusions | `delete_skill` |

**A dedicated skill set was rejected**, whether one for the surface or one per server: `set_active_skill_set` (`skill.py:2017-2087`) clears `is_active` on *every* non-default set for the (user, bot, engine) before activating its target, filtering on nothing that could exempt one. Any workbench skill-set switch would have silently deactivated every MCP this surface owns. The default set is appended separately by `get_all_active_skill_sets:2151` and never swept — which is exactly why `skills` lives there.

**An `is_active` column on `ac_skill_set_mcp` was also rejected** and is recorded in `plan.md` as the fallback. It works, but needs DDL and a filter in `collect_bot_active_mcps` where a wrong default would silently strip capability from every existing bot. Skills parity needs neither.

**Other decisions:** the credential stays user-scoped (`(user_id, server_code)`, shared across the user's bots) with activation as the per-bot axis, so deleting a config never deactivates a server and removing a server never deletes the credential; adding a server leaves it **deactivated**, matching `skills`, so adding never changes what an agent can call; engine defaults are listed and deactivatable but refused on remove (409), being synthesised per request rather than stored.

### The one change to existing behaviour

`get_set_mcp_servers` now applies `excluded_codes` to stored `ac_skill_set_mcp` rows, not only to synthesised engine defaults — matching the skill half (`skill.py:545-562`), which has always filtered its member rows this way.

Before this, an exclusion written against a code that also had a row was inert — **including the ones `remove_mcp_from_skill_set` writes**, which is that method's entire removal mechanism for a default set. Those removals silently did not take; they do now. This is a visible behaviour change on the internal surface, and it is deliberate.

## Validation

Every suite below was run in this sandbox and passed.

- **Tripwire first.** Before writing any new test for the filter change, the existing skill-set and MCP suites were run unmodified: **1062 passed, 3 skipped**, plus **714** across `config_compose` and `repository`. No existing test needed changing, which is the evidence the blast radius is what it looks like.
- `tests/community/adapters/http/openapi_v1` — **887 passed, 1 skipped**
- `tests/community/architecture` + `tests/community/di` — **264 passed**
- Gateway `tests/unit` — **608 passed**
- New: 24 service tests, 27 bot-scoped endpoint tests, 13 config-flow tests, 13 config endpoint tests, 5 skill-set filter tests

Three surface-wide pins moved, all as consequences rather than accommodations: `user_id`-taking operations 61 → 69 (all eight new operations are user-scoped and take it), `bot_id` placement 54/1/16 → 60/1/18 (six address a bot in the path, two are account-level), and published descriptions were rewritten as plain prose after `test_the_published_document_contains_nothing_internal` caught RST markup leaking into them.

`bots.openapi.json` was regenerated through the documented dump + gate flow, not by hand; the gate reports **compatible**. The textual diff is large because the generator reorders keys, but the artifact is semantically exactly this change: 5 paths added, none removed, no pre-existing schema altered, 13 schemas added, and the one pre-existing path that changed is the config route gaining `DELETE`.

**Not run, and why:** `test_bcn_dump_uses_the_gateway_managed_python_environment` fails in this sandbox — it shells out to a script that runs its own `uv sync` against the repo's pinned aliyun mirror, which the sandbox proxy blocks. Unrelated to this diff; it exercises the BCN dump path, which is untouched. For the same reason both `uv.lock` files were rewritten during dependency setup and have been **reverted to their committed state** — no dependency changed here.

## Compatibility and risk

**No schema change.** No migration, no DDL.

The internal behaviour change above is the risk to weigh. It is bounded to existing row+exclusion pairs, which can only have been created by a remove that silently did not take, so honouring them does what the caller asked. If it proves unacceptable, the `is_active` column in `plan.md` is the recorded fallback.

Public-API MCPs share the default skill set with the workbench, so a workbench user could remove a row this surface created. Accepted — same bot, same owner — and identical to the position `skills` is already in.

Rollback path: revert the commits.

## Spec

`src/backend/specs/2026-08-17-openapi-v1-mcp-bot-scoped/` — spec, plan, and the 38-task list, all marked complete.

## Related issues

Requesting access to a non-public MCP server is tracked separately as #1109 and is deliberately out of scope.


src/backend/src/agentclaw/community/core/mcp/services/bot_mcp_state_service.py
```diff
@@ -0,0 +1,367 @@
...
+
+        set_id = str(default_set["id"])
+        self._skill_set_repo.remove_mcp_from_set(set_id, server_code)
+        # Clear the exclusion too, or a later re-add would come back off for a
+        # reason the caller has no way to see.
+        self._skill_set_repo.remove_all_default_mcp_exclusions(
+            owner_id, bot_id, server_code
+        )
+
+        await self._reconcile(bot=bot, bot_id=bot_id, owner_id=owner_id)
```
totalfrank: **Blocking.** `remove_bot_server` doesn't roll back on device-sync failure, unlike its two siblings.

`add_bot_server` (lines 257–264) and `set_bot_server_active` (lines 305–320) both wrap `_reconcile` in `try/except McpSyncFailedError` and undo the state write before re-raising. `remove_bot_server` calls `_reconcile` unguarded at line 366, after already deleting the membership row (line 359) and clearing exclusions (line 362).

Concrete sequence: call `remove_bot_server` for a stored (non-default) server on a bot whose device is unreachable. `remove_mcp_from_set`/`remove_all_default_mcp_exclusions` commit immediately (each opens its own session), then `refresh_mcp_scope` returns `success: false` and `_reconcile` raises `McpSyncFailedError`, which propagates out uncaught. The caller gets a 502, but the row is already gone and stored state has diverged from the device's actual whitelist with no way to detect it — a retried call now returns `False` ("nothing to remove"), silently masking that a real, unsynced mutation already happened.

This contradicts the invariant the other two mutations (and `plan.md`) establish: every mutation should roll back its state write if `refresh_mcp_scope` fails.

```suggestion
        try:
            await self._reconcile(bot=bot, bot_id=bot_id, owner_id=owner_id)
        except McpSyncFailedError:
            self._skill_set_repo.add_mcp_to_set(
                set_id,
                server_code,
                entry.get("name") or server_code,
                description=entry.get("description"),
                icon=entry.get("icon"),
                user_id=owner_id,
            )
            if not entry["active"]:
                self._skill_set_repo.add_default_mcp_exclusion(
                    user_id=owner_id,
                    bot_id=bot_id,
                    skill_set_id=int(default_set["id"]),
                    server_code=server_code,
                )
            raise
```

---

src/backend/src/agentclaw/community/core/mcp/services/bot_mcp_state_service.py
```diff
@@ -0,0 +1,367 @@
...
+            bot_id=bot_id,
+            skill_set_id=int(default_set["id"]),
+            server_code=server_code,
+        )
+
+        try:
+            await self._reconcile(bot=bot, bot_id=bot_id, owner_id=owner_id)
+        except McpSyncFailedError:
+            self._skill_set_repo.remove_mcp_from_set(set_id, server_code)
+            self._skill_set_repo.remove_all_default_mcp_exclusions(
```
totalfrank, line 261: **Non-blocking.** `self._skill_set_repo` is typed as `SkillSetRepository` (the `Protocol` in `core/repository/protocols/skill_center.py`), but `remove_all_default_mcp_exclusions` — called here and at lines 294, 317, 362 — was added only to the concrete implementation (`skill.py:1815`), not to the Protocol. `core/repository/protocols/skill_center.py` is untouched by this diff.

This is masked at runtime because DI binds the concrete class directly, so nothing fails today. But the whole point of typing this dependency as a Protocol is that `bot_mcp_state_service.py` is only supposed to depend on the declared contract — any future swap to a different `SkillSetRepository` implementation, or a type-checking pass added later, would break on a method the interface never promised. Worth adding the method to the Protocol alongside its existing sibling `remove_all_default_skill_exclusions`.

---

src/backend/src/agentclaw/community/core/repository/implementations/skill_center/skill.py
```diff
@@ -1812,6 +1812,36 @@ def get_all_excluded_mcps(
...
+            rowcount = (
+                db.query(DefaultSkillsetMcpExclusion)
+                .filter(
+                    DefaultSkillsetMcpExclusion.user_id == user_id,
+                    DefaultSkillsetMcpExclusion.bot_id == bot_id,
+                    DefaultSkillsetMcpExclusion.server_code == server_code,
+                )
+                .delete(synchronize_session=False)
+            )
+            return rowcount > 0
```
totalfrank, line 1843: **Non-blocking — missing test.** `remove_all_default_mcp_exclusions` deliberately omits `skill_set_id` so it clears an exclusion stranded on a *former* default skill set, per its own docstring. No test exercises this against a real database with exclusion rows on two distinct `skill_set_id`s to prove it clears the stale one and leaves an unrelated one alone.

Every existing exerciser mocks the repository (`test_bot_mcp_state_service.py` only asserts the method was *called* with the right args) or seeds just one default skill set per bot (`test_openapi_mcp_lifecycle.py`'s `_seed_bot`), so the multi-default-set scenario this method exists for is never actually constructed.

Its skill-side twin has exactly this test: `test_skill_repository_unified.py::test_remove_all_default_skill_exclusions_clears_prior_default_sets` (seeds exclusions on `skill_set_id=7` and `8`, clears one, asserts the other survives). A copy-paste regression here (e.g. accidentally reintroducing a `skill_set_id` filter) would pass every current test and silently reproduce the bug this method was written to fix — a server that stays permanently un-activatable after a successful `activate` call.

---

src/backend/src/agentclaw/community/core/mcp/config_flow.py
```diff
@@ -221,6 +262,87 @@ def _roll_back() -> None:
...
+    server stays, its credential does not": ``build_mcp_sync_payload`` falls
+    back to the stored row, finds nothing, and the device is handed the bare
+    server config.
+
+    Devices that do not carry this MCP are skipped by the sync service, exactly
+    as on the write path.
+    """
+    mcp_data = market_service.get_mcp_detail(server_code)
+    if not mcp_data:
+        raise McpServerNotFoundError(server_code)
```
totalfrank, line 304: **Non-blocking.** `delete_unified_config` requires `market_service.get_mcp_detail(server_code)` to succeed before it will delete the stored row — mirroring the write path's precondition. Concretely: if a server is later removed or renamed in the marketplace catalogue after a user stored a credential for it, `DELETE /openapi/v1/bots/mcp/servers/{server_code}/config` 404s instead of revoking the now-orphaned credential, and there is no other endpoint in this PR that can clear it.

That cuts against this endpoint's own stated motivation ("revoking a leaked key meant replacing it with a decoy") for exactly the case where revocation matters most — a delisted/rotated server is a plausible reason a key would need revoking in the first place. This is tested and mirrors an existing precondition elsewhere, so it may be an accepted, deliberate limitation rather than an oversight — worth confirming that's intentional rather than an accidental symmetry with the write path.

---

src/gateway/tests/unit/core/authn/test_route_security.py
```diff
@@ -329,3 +334,38 @@ def test_from_yaml_user_config_not_dict_uses_empty_table(tmp_path) -> None:
...
+
+
+@pytest.mark.parametrize(("method", "path"), _BOT_SCOPED_MCP)
+def test_bot_scoped_mcp_stays_reachable_by_a_delegated_application(
+    method: str, path: str
+) -> None:
+    """Grant-checked in the backend, so it must not be refused at the edge.
+
+    The account-level MCP *config* operations sit one path segment away and are
+    human-only. The near-miss is the point of pinning this: a rule written as
```
totalfrank: **Non-blocking — test doesn't prove what its docstring claims.** The docstring says this pins protection against "a rule written as `/openapi/v1/bots/mcp/**` with one segment wrong." Injecting exactly that rule and re-resolving every `_BOT_SCOPED_MCP` case (all using bot id `"bot-123"`) leaves every result unchanged: `/openapi/v1/bots/mcp/**` has `mcp` as a *literal* in the position `{bot_id}` occupies in the real route, so it only matches a path whose bot id is literally `"mcp"` — none of the parametrized cases use that. The described near-miss would recur and this test would keep passing.

What the test does catch is a different, more likely mistake: accidentally re-adding the correctly-parameterized `/openapi/v1/bots/{bot_id}/mcp/**` as a refusal (the thing the `application.yaml` comment above the account-level rules warns against) — that does flip every case here. That's real, useful coverage; the docstring just describes the wrong near-miss. Consider either adding a `("GET", "/openapi/v1/bots/mcp/mcp")`-style case that actually exercises the literal-segment near-miss, or correcting the docstring to describe what's actually being pinned.

---

totalfrank (commented):
Reviewed commits 06f262a..386fb01 (the full PR against merge-base `4a422e4`). Covered: `bot_mcp_state_service.py`/`errors.py` (new bot-scoped activation service), `bot_mcp/router.py`+`schemas.py` (new public endpoints) and `admission.py` (grant wiring), `mcp/router.py`+`config_flow.py`+`config_service.py` (new config list/delete flows), the `skill.py`/`skill_set_service.py` exclusion-filter behavior change, and the gateway `application.yaml`/`test_route_security.py` admission rules.

One blocking finding: `remove_bot_server` is missing the rollback-on-sync-failure that its two sibling mutations (`add_bot_server`, `set_bot_server_active`) both have, so a device-sync failure during removal leaves stored state permanently diverged from the device. Four non-blocking findings: a `Protocol` contract gap (a called repository method isn't declared on the interface it's called through), a missing repository-level test for the new cross-default-set exclusion clear (its skill-side twin has this test, this doesn't), a design note on the new config-delete endpoint being unable to revoke a credential for a delisted marketplace server, and a gateway regression test whose docstring overstates what near-miss it actually catches.

Categories applied: correctness, contract breakage (Protocol/interface completeness), error/failure handling (rollback-on-failure), architecture boundaries, and test coverage of the diff. No concurrency, security-hygiene, or Python type-contract (`T | None`) issues found in the new code.

---

src/backend/src/agentclaw/community/core/mcp/config_flow.py
totalfrank: Confirming: intentional, and the tension you've identified is real rather than hypothetical.

It is a spec decision, not accidental symmetry — `spec.md` states it directly: *"Deleting a configuration for a server code that does not exist in the marketplace answers not-found, consistent with the write path refusing an unknown server before it touches the database."* So the symmetry with the write path was chosen deliberately.

The reason it stays that way in this PR is mechanical as much as principled: `mcp_data` from `get_mcp_detail` is what the credential-clearing re-sync pushes to the device. Deleting the row without it would clear the credential in storage while leaving the device holding the old one — which is precisely the "believes it's gone while an agent still holds it" failure the rollback exists to prevent. Skipping the marketplace check would need a second, detail-free removal path, and that is a design change rather than a tweak.

But your objection stands on its merits, and I'd flagged the same thing to the author in my own handover notes before this review landed — a delisted or renamed server is a plausible reason a key needs revoking, and it is exactly the case the endpoint currently cannot serve. Two shapes worth considering, neither in this PR:

1. A `force`/`orphaned` flag that deletes the row and skips the device push, documented as leaving the device to be reconciled on its next artifact rebuild.
2. Fall back to `remove_mcp_detail` when the marketplace has no detail — the MCP is unreachable anyway if it has been delisted, so un-installing it is arguably correct there, unlike the normal path where it would wrongly turn a revocation into a deactivation.

I'll leave the decision to you rather than widening this PR's scope. Happy to open a follow-up issue alongside #1109 if you want it tracked.


---

src/backend/src/agentclaw/community/core/mcp/services/bot_mcp_state_service.py
totalfrank: Fixed in `113c39f`. `remove_bot_server` now wraps `_reconcile` in `try/except McpSyncFailedError`, captures the pre-delete row via `get_mcp_servers_in_set` (preserving `icon`, which the `_merged` projection drops), and restores the membership row plus its exclusion (if it was inactive) before re-raising — matching the sibling mutations' contract. Two new unit tests cover the active and inactive rollback cases.

---

src/backend/src/agentclaw/community/core/mcp/services/bot_mcp_state_service.py
totalfrank: Fixed in `113c39f`. `remove_all_default_mcp_exclusions` is now declared as an `@abstractmethod` on `SkillSetRepository` in `core/repository/protocols/skill_center.py`, matching the concrete implementation's signature.

---

src/backend/src/agentclaw/community/core/repository/implementations/skill_center/skill.py
totalfrank: Fixed in `113c39f`. `test_remove_all_default_mcp_exclusions_clears_prior_default_sets` now seeds exclusions on two distinct `skill_set_id`s (7 and 8) against a real database, clears one, and asserts the other survives — mirroring the skill-side twin exactly. A second test (`..._is_scoped_to_its_bot`) also confirms it doesn't leak across bots.

---

src/gateway/tests/unit/core/authn/test_route_security.py
totalfrank: Fixed in `113c39f`. The docstring now correctly describes the near-miss this test actually catches (re-adding the correctly-parameterized `/openapi/v1/bots/{bot_id}/mcp/**` as a refusal), and a new test (`test_a_bot_literally_named_mcp_is_not_captured_by_the_account_rules`) exercises the literal-segment collision the original docstring claimed to cover.

---

src/backend/tests/community/core/mcp/services/test_bot_mcp_state_service.py
```diff
@@ -0,0 +1,382 @@
...
+    retry then answers "nothing to remove", silently masking a real mutation
+    the device never saw.
+    """
+    sets = _skill_set_repo(rows=[_row()])
+    with pytest.raises(McpSyncFailedError):
+        _remove(_svc(sets=sets, sync=_sync(ok=False)))
+    # Membership is put back...
+    sets.add_mcp_to_set.assert_called_once()
+    args, kwargs = sets.add_mcp_to_set.call_args
+    assert args[1] == STORED
```
totalfrank, line 371: nit: `_row()` (line 88) never sets `icon`, so this assertion — and the rollback fixture in general — can't actually distinguish "restored from the pre-delete `stored` row" from "restored from `entry`" (the `_merged` projection this fix says is wrong to use). Both would produce `icon=None` here. Giving `_row()` a non-null `icon` and asserting `kwargs.get("icon")` (or the `icon=` kwarg) equals it would make this test fail if a future edit reverts to `entry` instead of `stored`.

---

totalfrank (commented):
Reviewed commits `f6a5f87` (merge of `origin/dev`) and `113c39f` (the fix commit) against head `113c39f`.

**Findings from the previous round — all four fixable ones addressed, verified by reading the code:**
- ✅ Fixed — Blocking: `remove_bot_server` missing rollback-on-sync-failure. Now wraps `_reconcile` in `try/except McpSyncFailedError`, captures the pre-delete row (preserving `icon`), and restores membership + exclusion before re-raising. Two new unit tests cover it.
- ✅ Fixed — Non-blocking: `remove_all_default_mcp_exclusions` missing from the `SkillSetRepository` Protocol. Declared as `@abstractmethod` now.
- ✅ Fixed — Non-blocking: missing repository-level test for the cross-default-set exclusion clear. Added, mirroring the skill-side twin against a real database.
- ✅ Fixed — Non-blocking: gateway regression test docstring overstated what it caught. Docstring corrected and a new test added for the actual literal-segment collision.
- Not fixed / not required — Non-blocking: config-delete can't revoke a credential for a delisted marketplace server. Already discussed and conceded as an intentional, spec-recorded design decision in the prior round (deferred to a follow-up, not this PR). Left open, no new reply needed.

**New in this round:**
- nit: rollback test in `test_bot_mcp_state_service.py` doesn't actually verify `icon` is preserved from the pre-delete row rather than the lossy `_merged` projection, since the test fixture never sets `icon` (inline comment posted).

I also re-verified the `origin/dev` merge's integration: the two new `_REFUSES_APP_ONLY` dependency additions on `/configs` and `DELETE .../config` are correct, and `bot_mcp_router` correctly inherits `require_granted_own_bot` via the `_GRANT_CHECKED_SUBGROUPS` mount — no admission regressions from the merge.

**Status:** 0 blocking findings open. 1 non-blocking finding open (delisted-server revocation, deferred by design). 1 new nit filed this round. CI was still running Backend unit tests and Singlebox coverage at review time; everything else green.

---

src/backend/tests/community/core/mcp/services/test_bot_mcp_state_service.py
totalfrank: Good catch — the assertion was defending the deviation from the suggested patch without being able to detect it. Fixed in `b924496`.

`_row()` now carries a non-null `icon`, and the rollback test asserts both `icon` and `description` survive the restore.

Verified it has teeth rather than assuming so: temporarily reverting the service to restore from `entry` instead of `stored` fails the test with

```
assert kwargs.get("icon") == ICON
E   AssertionError: assert None == 'https://example.test/weather.png'
```

then restored the fix. A future edit back to the projection is now caught.


---

totalfrank (commented):
**Round 4 status.** Reviewed `b924496` (`test(mcp): pin icon preservation in the remove rollback`) — the only new commit since round 3's `113c39f`.

Verified this new commit does what it says: `_row()` now sets a non-null `icon`, and `test_remove_rolls_back_when_the_runtime_cannot_be_reconciled` asserts `kwargs.get("icon") == ICON` and `description == "d"`. This gives the rollback-source assertion actual teeth (confirmed in-thread by the author with the revert-and-observe-failure check). No new issues in this commit — it's test-only, narrowly scoped to the gap it names.

**Full finding checklist:**
1. Blocking — `remove_bot_server` missing rollback-on-sync-failure → **Fixed** (`113c39f`), resolved.
2. Non-blocking — `remove_all_default_mcp_exclusions` missing from `SkillSetRepository` Protocol → **Fixed** (`113c39f`), resolved.
3. Non-blocking — missing repository-level test for cross-default-set exclusion clear → **Fixed** (`113c39f`), resolved.
4. Non-blocking — config-delete can't revoke a credential for a delisted marketplace server → **Author's call**: confirmed intentional, spec-recorded design decision, deferred to a follow-up outside this PR. Left open by design, not a defect.
5. Non-blocking — gateway regression test docstring overstated its coverage → **Fixed** (`113c39f`), resolved.
6. nit — rollback test couldn't distinguish restore-from-stored-row vs. restore-from-projection (no non-null `icon` in fixture) → **Fixed** (`b924496`), verified, resolved this round.

**Open count: 1** — #4 above, open by design (deferred, not a defect).

No new bugs found this round.

---

PR #1111, opened 2026-08-17
refactor(backend): split grant check into per-admission-mode dependencies

totalfrank:
## Problem

The public surface enforced its grant-checked admission modes through one mode-aware dependency, `require_granted_bot`, whose `_addressed_owner` helper consulted the `ADMISSION` table **per request** to decide whether the addressed owner may be read off the wire. Three consequences:

- A route's admission mode was invisible at the route itself: `GRANT_CHECKED_OWN_BOT` and `GRANT_CHECKED_ADDRESSED_BOT` looked identical in a router file, and the engine-runtime groups carried their check only transitively through `OwnerIdDep`, which is hard to discover for anyone new to the seam.
- Relabelling a table entry `OWN_BOT → ADDRESSED_BOT` silently changed runtime authorization — the one behavioral test for an injected `owner_id` was the only thing standing between that relabel and the surface's oldest defect (a grant on anyone's `default` becoming access to the delegator's `default`).
- `REFUSED` was enforced only centrally in `require_principal`, so nothing on a refused route said so, and `GRANT_FILTERED` / `USER_GATED` had no derived test coverage at all: a new operation in either mode with a correct table entry and no narrowing in its handler would have failed nothing.

## Solution

Give each admission mode with something to enforce its own explicit dependency, so the declaration *is* the mode and the table becomes purely declarative:

- **`require_granted_own_bot`** — owner is the delegating user, never read off the wire; an appended `?owner_id=` is structurally inert rather than table-guarded. Declared at `include_router` for the wholly own-bot groups (current and legacy) and per route in the mixed `bots` group.
- **`require_granted_addressed_bot`** — the one dependency entitled to read the `owner_id` query parameter (defaulting to the caller). Declared per route on the two skills collection operations and at the engine-runtime mounts; `OwnerIdDep` keeps consuming its returned owner (`AddressedBotGrantDep`), and the mount-level declaration adds no second lookup since FastAPI caches a dependency per request.
- **`refuse_app_only_caller`** — the `REFUSED` declaration, added to bot creation, the authorized-apps group, bot logs, the MCP config/permissions operations, and both loadtest routes (it answers the socket plane with the same 1008 close). The central fail-closed refusal in `require_principal` stays and still covers operations absent from the table; the declaration makes the decision visible on the route and holds even if a table entry were mislabelled.
- `_addressed_owner` and the runtime `ADMISSION` lookup in `principal.py` are deleted; the shared body (`bot_id` extraction, grant lookup, masked 404) lives once in `_require_granted_bot`.
- `GRANT_FILTERED` / `USER_GATED` cannot be route dependencies — they shape the *result* through `granted_bot_ids` — so their coverage is now **derived from the table**: `test_app_only_listings.py` parametrizes over every entry in those modes and fails loudly until each has an ungranted-app case (the same ratchet `test_self_checked_routes_refuse.py` uses).
- `test_admission_inventory.py` now pins each grant-checked mode to *its* dependency in both directions, and pins the `REFUSED` declarations the same way. The four `SKILL_SCOPED_OPERATIONS` and the retiring self-checked addresses are excluded exactly as before. Legacy routes are otherwise untouched (they are retiring); their mounts pick up the renamed dependencies mechanically.

Rejected alternative: keeping one dependency and asserting the mount↔mode pairing only in CI — that leaves the mode invisible at the route and the relabel hazard in place, which is what this change exists to remove.

## Validation

- Full `src/backend` public-surface suite: `pytest tests/community/adapters/http/openapi_v1/ tests/community/endpoints/test_openapi_startup_script.py` → **852 passed, 1 skipped** (pre-change baseline 846 passed, 1 skipped; the delta is the new structural and derived tests).
- Mutation check (in-memory, not committed): relabelling all 56 `GRANT_CHECKED_OWN_BOT` entries to `GRANT_CHECKED_ADDRESSED_BOT` now alters **no** behavioral test — including the injected-`owner_id` attack test — while failing the structural inventory tests in both directions. Before this change the same relabel changed runtime authorization behavior.
- FastAPI dependency-caching probe confirms the mount-level addressed-bot declaration performs no second grant lookup (1 invocation with and without the route-level declaration).
- SAST block rules (`python_sast_local.sh` rule set) via flake8 on all 19 changed files: clean.
- Not run: the full module CI gates (coverage/E2E) — lint-only pre-push per `AGENTS.md` default; CI on this PR is the authority.

## Compatibility and risk

No wire contract, schema, or status-code change: every operation keeps its admission mode, and refusals keep their existing shapes (masked 404 for grant misses, uniform 401 for refused shapes). The gateway `route_security` counterpart is untouched — `ADMISSION`'s `REFUSED` set is unchanged, so the pinning test on the gateway side is unaffected. Rollback is a plain revert; no data or config migration.

https://claude.ai/code/session_01RLhFySftnBWVDzg1dh9xt8

---
_Generated by [Claude Code](https://claude.ai/code/session_01RLhFySftnBWVDzg1dh9xt8)_

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/__init__.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/__init__.py
@@ -20,8 +20,10 @@ argument to the call, and a client that must be told to put the same id in two
 places has been told the address twice.
 
 That is also what makes authorization mostly mechanical rather than
-per-operation. ``require_granted_bot`` reads the addressed bot off the path, the
-same way, for all but four operations on the surface.
+per-operation. The grant dependencies (``require_granted_own_bot`` /
+``require_granted_addressed_bot`` — one per grant-checked admission mode, so
+the declaration says which id model a route has) read the addressed bot off
+the path, the same way, for all but four operations on the surface.
 
 **The four are not an oversight, and their handler-side checks are
 load-bearing.** The ``{skill_id}`` skills operations resolve by
@@ -186,7 +188,7 @@ from .contracts import (
     USER_SCOPED_ERROR_RESPONSES,
 )
 from .dependencies import require_principal
-from .principal import require_granted_bot
+from .principal import require_granted_addressed_bot, require_granted_own_bot
 from .engine_runtime.approvals import router as engine_approvals_router
 from .engine_runtime.connection import router as engine_connection_router
 from .engine_runtime.engine import router as engine_engine_router
@@ -265,11 +267,12 @@ _SUBGROUPS = [
 # outright, which is exactly wrong for the listings (Mode B) and the
 # account-level reads (C/OPEN). `bots` is mixed and declares it per route;
 # `admission.py` is the authority on which is which, and
-# `test_principal_seam.py` fails if a route's declaration and its mode disagree.
+# `test_admission_inventory.py` fails if a route's declaration and its mode
+# disagree.
 #
-# The engine-runtime groups need no entry here: their `OwnerIdDep` already
-# depends on the same check, because it is where the addressed owner comes from
-# for an application caller.
+# The engine-runtime groups are not here because they are not own-bot: they may
+# address a shared bot, so their mount below declares the *addressed-bot*
+# dependency instead.
 _GRANT_CHECKED_SUBGROUPS = [
     # Engine config is bots-component work at an engine-component address: it
     # reads and writes a stored blob, so it takes the ordinary error table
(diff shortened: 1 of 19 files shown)


totalfrank (commented):
Reviewed commit `e61ccb6` (the only commit on this PR) against `dev` @ `4a422e4`. Clean — no findings.

**Files covered:** all 19 changed files under `src/backend/src/agentclaw/community/adapters/http/openapi_v1/` (`__init__.py`, `admission.py`, `principal.py`, `authorized_apps/router.py`, `bot_logs/router.py`, `bots/router.py`, `deprecated/*`, `engine_runtime/params.py`, `loadtest/router.py`, `mcp/router.py`, `skills/router.py`) and the five updated/added test files.

**What I checked:**
- **Correctness / contract breakage** — traced the `require_granted_own_bot` / `require_granted_addressed_bot` / `refuse_app_only_caller` wiring by hand across every mount point (`build_public_router`, `deprecated/__init__.py`, `deprecated/engine_runtime.py`) against the `ADMISSION` table entries, including the legacy groups where `deprecated.GRANT_CHECKED_GROUPS` mixes `identity` with `bots`/`resources`/`routines` — confirmed `_engine_runtime.GRANT_CHECKED` there is only `identity` (own-bot), not the addressed-bot engine-runtime groups, so no mode mismatch. No stale references to the removed `require_granted_bot`/`GrantCheckedDep`/`_addressed_owner` remain anywhere in `src/backend/src/agentclaw`.
- **Security** — this is the core authorization split for the public surface; verified the own-bot dependency never reads `owner_id` off the wire (the class of bug the PR's docstrings call out as the surface's oldest defect), and that the addressed-bot dependency is only declared where the route actually publishes `owner_id`.
- **Test coverage** — `test_admission_inventory.py`'s new parametrized test asserts the mode↔dependency pairing in both directions per mode, and the new `test_every_refused_operation_declares_its_refusal` does the same for `refuse_app_only_caller`; `test_app_only_listings.py` adds the previously-missing derived coverage for `GRANT_FILTERED`/`USER_GATED`. These structurally close the exact gap the PR's problem statement describes.
- **Python type contracts** — no `T | None` widening introduced; `_require_granted_bot`'s `owner_id` stays a required `str` in both callers.
- **Architecture boundaries** — no transport/framework leakage into core, no new raw environment access.

Not applicable here: concurrency/resource safety (no async state introduced beyond existing per-request dependency caching).

---

FreddieSun (approved):

FreddieSun (approved):

PR #1112, opened 2026-08-17
feat(mcp): give teclaw its own default MCP server list

totalfrank:
## Problem

`_DEFAULT_MCP_SERVERS_BY_ENGINE` had no `teclaw` key, so `get_default_mcp_servers("teclaw")` returned `[]` and every teclaw bot shipped with no platform MCPs at all — no DingTalk group/message/todo/calendar servers, no Yuque, no Dima, and no `hitl`. Every other engine bucket (`openclaw`, `claude_code`, `hermes`, `aicoding`) has carried a roster for some time.

This is the second half of the teclaw MCP work. The first half (#1083) taught the artifact path to carry stdio servers; without it `hitl` could not have been listed here at all, since teclaw boots from the whole composed artifact rather than incremental `/api/mcp` pushes.

## Solution

A `teclaw` bucket with the 12 remote servers from the requirement doc plus `hitl`, which is local/stdio and resolved through `LocalMCPRegistry` rather than MCP Center. `hitl` is placed last, matching how the other engines list it.

Entries carry `server_code` and `name`. `description` is deliberately omitted: the requirement doc specifies no descriptions, and `McporterComposer.compose_server` reads `md.get("name") or md.get("description")` into the artifact's `name` field — inventing prose here would put fabricated text on the published contract.

Giving the roster its first non-empty value exposed a test dependency. The teclaw artifact path composes every MCP as a unit and fails the whole build when one cannot be resolved (deliberate — a silently dropped MCP boots the bot without a capability its owner configured). `NoopMCPCenterPlugin` resolves nothing, so the publish suites failed on the first roster code for a reason unrelated to what they assert.

The fixture seeds the MCP Center seam from `_DEFAULT_MCP_SERVERS_BY_ENGINE` itself rather than a hardcoded list, so a server added to the roster later cannot silently skip it. Two exclusions are load-bearing:

- **`hitl` is left unresolved on purpose.** It is stdio, and the collector resolves its launch instruction from `LocalMCPRegistry`. Answering `runMode: REMOTE` for it would hit the short-circuit in `_stdio_launch_for` and misclassify it as remote — the fixture would have masked exactly the behaviour #1083 added. It is filtered by asking the registry, not by name, so a future local server needs no fixture edit.
- **Codes outside the roster still resolve to `None`.** `_seed_failing_mcp` depends on that to drive `_seed_process_build_fail`, which asserts the fail-early raise. Making everything resolve would have quietly disarmed that case.

Rejected alternative: degrading unresolvable *defaults* to a warning-and-drop while keeping user-added MCPs fatal. That would have removed the fixture need entirely and hardened the roster against an upstream code being retired, but it weakens a contract that was chosen deliberately, and the failures in question were test-double wiring rather than evidence of a product defect.

## Validation

- **Full community suite, identical on both trees**: base `8d8584fa` → `2 failed, 11867 passed, 4 skipped`; this branch → `2 failed, 11867 passed, 4 skipped`. The two failures are `rsync: not found` in the build container, unrelated to this change and present on base.
- Both runs were done in **fresh worktrees with the same command**. An earlier comparison run in the main working tree was discarded: these publish tests are state- and order-sensitive, and the mismatched conditions produced a misleading failure count.
- Narrowed the previously failing suites (`e2e/publish_boundary`, `endpoints/test_endpoint_runner`, `endpoints/test_publish_durable_pipeline`, `endpoints/test_service_bot_publish_flow`) from **17 failures to 0**.
- `test_c1_upgrade_chain` fails in a subset run and passes in the full suite, on this branch *and* on base — it depends on an earlier test's app bootstrap creating `ac_default_skillset_mcp_exclusion`. Pre-existing, untouched here, and called out rather than papered over.
- Verified `get_default_mcp_servers("teclaw")` resolves to 13 entries and that `teclaw` is not bucket-aliased onto another engine (`_resolve_default_mcp_engine_bucket("teclaw") == "teclaw"`), so the roster is genuinely new rather than inherited.

## Compatibility and risk

Every teclaw publish now depends on MCP Center resolving all 12 remote codes. Because the artifact path composes defaults as a unit — unlike arca/baas, which pushes per-server and tolerates one failure — a single retired or renamed code fails the whole publish rather than dropping one server. This is the fail-early contract working as designed, but reviewers hitting a sudden teclaw publish failure should look here first.

Deployments whose MCP Center does not carry these codes are affected the same way. `CommunityMCPCenter` serves an empty catalog unless an operator configures `registry_config_path`.

Rollback is a plain revert of the roster commit; teclaw returns to `[]` defaults and the fixture becomes inert (it derives from the roster, so an empty roster overrides nothing).

---
_Generated by [Claude Code](https://claude.ai/code/session_01N6jGHBFagXrdjjrZCDUkLE)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
+++ b/src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
@@ -45,6 +45,23 @@ _DEFAULT_MCP_SERVERS_BY_ENGINE: Dict[str, List[dict]] = {
         {"server_code": "hitl"},
     ],
     "moltis": [],
+    "teclaw": [
+        {"server_code": "mcp.ant.lwawchat.cogmessagemcp", "name": "蚂蚁钉协作群消息相关-MCP服务"},
+        {"server_code": "mcp.ant.antdingopenapi.antdingreportmcpserver", "name": "蚂蚁钉日志服务"},
+        {"server_code": "mcp.ant.antdingopenapi.antdinggroupmcpserver", "name": "蚂蚁钉群服务"},
+        {"server_code": "mcp.ant.lwawchat.cogdocumentmcp", "name": "蚂蚁钉协作群文档相关-MCP服务"},
+        {"server_code": "mcp.ant.antdingopenapi.antdingeventmcpserver", "name": "蚂蚁钉日程相关-MCP服务"},
+        {"server_code": "mcp.ant.antdingopenapi.antdingrobotmcpserver", "name": "蚂蚁钉机器人相关-MCP服务"},
+        {"server_code": "mcp.ant.antdingopenapi.antdingmessagemcpserver", "name": "蚂蚁钉消息相关-MCP服务"},
+        {"server_code": "mcp.ant.antdingopenapi.antdingtodomcpserver", "name": "蚂蚁钉待办服务"},
+        {"server_code": "mcp.ant.faas.skylarkmcpserver.skylarkmcpserver", "name": "语雀 MCP"},
+        {"server_code": "mcp.ant.arkai.dimamcpserver", "name": "Dima-MCP服务"},
+        {"server_code": "mcp.ant.homistudio.recordmcp", "name": "会中会话记录查询服务"},
+        {"server_code": "mcp.ant.rpc.dcanttouch.adminservice", "name": "行政小宝MCP服务"},
+        # Local/stdio; resolved through LocalMCPRegistry, not MCP Center. Kept
+        # last to match the other engines' lists.
+        {"server_code": "hitl"},
+    ],
     "claude_code": [
         {"server_code": "mcp.ant.antprocessai.anttaskmcp", "name": "任务中心MCP", "description": "任务中心待办任务，已办任务等相关任务查询MCP"},
         {"server_code": "mcp.ant.arkai.dimamcpserver", "name": "Dima MCP", "description": "Dima MCP"},
--- a/src/backend/tests/community/core/mcp/test_defaults_per_engine.py
+++ b/src/backend/tests/community/core/mcp/test_defaults_per_engine.py
@@ -468,3 +468,51 @@ def test_claude_code_normal_template_keeps_claude_mcp_bucket():
 
     assert "clawmind" in codes
     assert "mcp.ant.agentix.112858.aixAicoding" not in codes
+
+
+# The teclaw roster, spelled out independently of the source list. The publish
+# suite's ``_install_default_mcp_catalog`` derives its mock *from*
+# ``_DEFAULT_MCP_SERVERS_BY_ENGINE["teclaw"]``, so it proves only that a build
+# resolves whatever the roster happens to hold — a typo'd or duplicated
+# ``server_code`` would sail straight through it. This second, independent copy
+# is what makes such an edit fail loudly.
+TECLAW_DEFAULT_MCP_CODES = (
+    "mcp.ant.lwawchat.cogmessagemcp",
+    "mcp.ant.antdingopenapi.antdingreportmcpserver",
+    "mcp.ant.antdingopenapi.antdinggroupmcpserver",
+    "mcp.ant.lwawchat.cogdocumentmcp",
+    "mcp.ant.antdingopenapi.antdingeventmcpserver",
+    "mcp.ant.antdingopenapi.antdingrobotmcpserver",
+    "mcp.ant.antdingopenapi.antdingmessagemcpserver",
+    "mcp.ant.antdingopenapi.antdingtodomcpserver",
+    "mcp.ant.faas.skylarkmcpserver.skylarkmcpserver",
+    "mcp.ant.arkai.dimamcpserver",
+    "mcp.ant.homistudio.recordmcp",
+    "mcp.ant.rpc.dcanttouch.adminservice",
+    "hitl",
+)
+
+
+def test_teclaw_default_roster_shape():
+    """teclaw's roster: exact codes, exact order, no dupes, local entry last."""
+    codes = [s["server_code"] for s in get_default_mcp_servers("teclaw")]
+    assert codes == list(TECLAW_DEFAULT_MCP_CODES)
+    assert len(codes) == len(set(codes)), "duplicate server_code in the teclaw roster"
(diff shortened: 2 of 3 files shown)


src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
```diff
@@ -45,6 +45,23 @@
...
+        {"server_code": "mcp.ant.antdingopenapi.antdingrobotmcpserver", "name": "蚂蚁钉机器人相关-MCP服务"},
+        {"server_code": "mcp.ant.antdingopenapi.antdingmessagemcpserver", "name": "蚂蚁钉消息相关-MCP服务"},
+        {"server_code": "mcp.ant.antdingopenapi.antdingtodomcpserver", "name": "蚂蚁钉待办服务"},
+        {"server_code": "mcp.ant.faas.skylarkmcpserver.skylarkmcpserver", "name": "语雀 MCP"},
+        {"server_code": "mcp.ant.arkai.dimamcpserver", "name": "Dima-MCP服务"},
+        {"server_code": "mcp.ant.homistudio.recordmcp", "name": "会中会话记录查询服务"},
+        {"server_code": "mcp.ant.rpc.dcanttouch.adminservice", "name": "行政小宝MCP服务"},
+        # Local/stdio; resolved through LocalMCPRegistry, not MCP Center. Kept
+        # last to match the other engines' lists.
+        {"server_code": "hitl"},
```
totalfrank, line 63: Non-blocking: the new `teclaw` roster (13 entries) has no dedicated test asserting its shape — codes, count, no dupes, `hitl` last. `test_service_bot_publish_flow.py`'s `_install_default_mcp_catalog` derives its mock *from* `_DEFAULT_MCP_SERVERS_BY_ENGINE["teclaw"]` itself, so it only proves "the build resolves whatever the roster currently contains," not that the roster's content is what was intended — a typo'd or duplicated `server_code` here would pass silently. This file already has a precedent for locking in a roster addition (`test_yuque_cli_is_default_entry` for `_DEFAULT_CLI_ITEMS_BY_ENGINE`); a similar test for `get_default_mcp_servers("teclaw")` would close the gap.

---

totalfrank (commented):
Reviewed commits 29a8843 (teclaw default MCP roster) and 210a9f9 (test fixture seam). No correctness or architecture-boundary issues.

Checked: `collect_bot_active_mcps` always merges the engine's default roster into every publish build (confirming the fixture is actually load-bearing, not incidental); `hitl`'s exclusion from the mock's remote roster is correct — verified `LocalMCPRegistry().get_mcp_detail("hitl")` resolves from `configs/local-mcp-servers.yaml`, and that `_stdio_launch_for`'s `runMode: REMOTE` short-circuit would misclassify it if the fixture resolved it remotely instead; no duplicate `server_code`s in the new roster; type contracts, error handling, and architecture boundaries (`docs/arch/arch.rules.md`) are untouched by this diff. One non-blocking test-coverage note left inline.

Categories from the checklist that applied: correctness, error/failure handling (fail-early MCP resolution), architecture boundaries, test coverage. Concurrency, security/hygiene, and Python type contracts weren't implicated by this diff.

---

src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
totalfrank: Fixed in `d36778c4` — agreed, and the circularity you name is exactly the gap: the fixture derives from the roster, so it can only ever confirm the roster is self-consistent.

Added `test_teclaw_default_roster_shape` to `test_defaults_per_engine.py`, following the `test_yuque_cli_is_default_entry` precedent: a `TECLAW_DEFAULT_MCP_CODES` tuple spelled out independently of the source, asserted with `codes == list(...)` (which pins content, order and count in one), plus explicit no-dupes and `hitl`-last checks.

Also added `test_teclaw_roster_is_its_own_bucket` for an adjacent hole the first test does not cover: `_resolve_default_mcp_engine_bucket` delegates bucket overrides to the engine registry, so an alias of `teclaw` onto `openclaw` would swap the entire roster without touching `_defaults.py`. It pins the case normalization (`"TECLAW"`) and asserts a teclaw-only code is absent from openclaw's list.

Verified the assertions are load-bearing rather than assumed, by mutating the source and confirming failure:

| mutation | result |
|---|---|
| `recordmcp` → `recordmcpX` (typo) | 2 failed, 31 passed |
| duplicate `adminservice` entry | 2 failed, 31 passed |
| reverted | 33 passed |

Worth recording that the first duplicate mutation I tried silently matched nothing, so the suite passed and briefly looked like the dupe assertion was dead. It was the mutation that was a no-op, not the test — the second attempt asserted its anchor existed before editing.

---

totalfrank: Verified fixed in `d36778c4`. `TECLAW_DEFAULT_MCP_CODES` matches the 13 entries in `_DEFAULT_MCP_SERVERS_BY_ENGINE["teclaw"]` exactly (order, count, `hitl` last), and `test_teclaw_roster_is_its_own_bucket` correctly exercises `normalize_engine_type`'s case-folding (`"TECLAW"` → `"teclaw"`) plus bucket distinctness from `openclaw`. Resolving.

---

totalfrank (commented):
Reviewed the new commit since the last round: `d36778c4` (test(mcp): lock in the teclaw default MCP roster's contents).

`d36778c4` adds `TECLAW_DEFAULT_MCP_CODES` and two tests (`test_teclaw_default_roster_shape`, `test_teclaw_roster_is_its_own_bucket`) to `test_defaults_per_engine.py`. Verified against the actual roster in `_defaults.py`: the 13-entry tuple matches the source's codes and order exactly, `hitl` is last, and `_resolve_default_mcp_engine_bucket`/`normalize_engine_type` does lowercase `"TECLAW"` → `"teclaw"` as the bucket-distinctness test assumes. Test-only change, no production code touched, no new correctness/architecture issues.

**Findings checklist:**
1. `_defaults.py:63` — non-blocking: teclaw roster had no dedicated shape-locking test — **Fixed** in `d36778c4`, verified and thread resolved this round.

**Open findings: 0.**

Nothing open, nothing new. PR is clean as of this round (`d36778c4`).

---

xianmuyq (approved):

PR #1118, opened 2026-08-18
feat(bcs-api-http): surface collected state in openapi list sessions

zh-bupt:
## Problem

The OpenAPI v1 `GET /groups/{group_id}/sessions` endpoint (`bcs_api_http::v1::openapi::routes::session::list_sessions`) did not return whether each session is collected (收藏) by the viewer, while the legacy `bcs_http::routes::sessions::list_sessions_for_group` endpoint does surface this when a `participant` is named. Clients using the OpenAPI surface therefore cannot render the collected/starred indicator.

## Solution

Surface per-session `collected` state on the typed `SessionSummary` response, keyed on the resolved view actor — mirroring the legacy endpoint's participant-gated behavior without changing the route handler or breaking the OpenAPI contract.

- `bcs-service-api/src/application/v1/session.rs`: add `collected: Option<bool>` to `SessionSummary` with `skip_serializing_if = "Option::is_none"` (consistent with `title` / `participant_count`).
- `application/v1/bcs-app-session/src/lib.rs`: `SessionService::list` populates `collected` via the existing `SessionManagementService::collected_at_map(&ids, &view_actor_id)`, but only when the request explicitly passes `view_bot_id` — reusing the same resolved view actor already used for the participant filter. When `view_bot_id` is omitted the field stays `None` and is not serialized.
- No route handler change; the field flows through via serde.

Gating rationale: collected is only meaningful relative to a named viewer. The OpenAPI `view_bot_id` param is the analog of the legacy `participant`; both default the resolved actor to the authenticated human (no bot collections), so a meaningful `collected` value requires the caller to pass the bot identity.

## Validation

- `cargo build -p bcs-service-api -p bcs-app-session -p bcs-api-http` ✅
- `cargo test -p bcs-app-session --test v1_session_service` ✅ (46 passed) — includes new `list_surfaces_per_session_collected_for_explicit_view_actor`: collects one of two sessions, asserts `Some(true)`/`Some(false)` for the named view actor and `None` when `view_bot_id` is omitted.
- `cargo test -p bcs-api-http --test session_routes` ✅ (29 passed) — route-level fake mirrors semantics and asserts `items[0].collected == true` in the JSON envelope.

### DIFF
--- a/src/bcs/api-contracts/v1/domain-models.yaml
+++ b/src/bcs/api-contracts/v1/domain-models.yaml
@@ -977,6 +977,12 @@ SessionSummary:
     updated_at:
       type: integer
       minimum: 0
+    collected:
+      type: boolean
+      description: >
+        Whether the view actor (the `view_bot_id` resolved by the request) has
+        collected this session. Surfaced only when `view_bot_id` is explicitly
+        supplied; absent otherwise.
 
 SessionDetail:
   type: object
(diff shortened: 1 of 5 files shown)


vzvince:
@codex review

vzvince (approved):
LGTM

PR #1120, opened 2026-08-18
fix(engine): find sessions beyond gateway default limit

evilcoda:
## Problem

OpenClaw `sessions.list` applies a bounded default result limit before the Engine adapter receives session rows. The adapter previously requested `sessions.list` with empty params and only then applied an exact `session_key` filter. Consequently, a valid session outside the Gateway's first 100 rows could be reported as missing by `/api/sessions` and `/api/sessions/{session_id}`.

## Solution

When a non-blank `session_key` is provided, pass it to OpenClaw through the existing `sessions.list.search` parameter so the Gateway narrows the store before applying its default result cap. Retain the adapter's exact-key filter because Gateway search is substring-based.

Requests without a session key, including blank values, continue to send the original empty params object and preserve existing list behavior.

## Validation

Validated on top of `REL20260818`:

- Engine CI: `2357 passed`, `5 deselected`
- Overall Engine line coverage: `92.94%`
- Changed-line coverage: `100%` (`14/14`)
- Engine CI gate passed
- Targeted OpenClaw session/adapter/contract tests: `83 passed`
- Ruff checks and `git diff --check` passed
- Verified against a production-compatible OpenClaw Gateway that `sessions.list {"search": <full-session-key>}` locates a session beyond the default first 100 rows

## Compatibility and risk

- No HTTP or Plugin API schema changes.
- Normal session listing remains unchanged when `session_key` is omitted or blank.
- Exact matching remains enforced in the adapter, preventing fuzzy Gateway search results from leaking into the response.

## Promotion

This PR promotes the fix from the development change into the `REL20260818` iteration branch.

### DIFF
--- a/src/engine/src/engine/community/plugins/openclaw/_session.py
+++ b/src/engine/src/engine/community/plugins/openclaw/_session.py
@@ -150,10 +150,21 @@ class _SessionPortMixin:
         sessions fall within it (exact legacy behaviour).
         """
         client = await self._pooled_client(token)
+        requested_session_key = (
+            session_key if session_key and session_key.strip() else None
+        )
+
+        # Push the lookup into the gateway before its default result cap is
+        # applied. Keep the exact local filter because gateway search is fuzzy.
+        list_params = (
+            {"search": requested_session_key}
+            if requested_session_key is not None
+            else {}
+        )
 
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #1121, opened 2026-08-18
feat(backend): support template preset CLI defaults

panqiwp:
## Summary
- add AICoding CLI defaults resolver for template preset_capabilities.cli
- share default capability registry and template preset merge logic between MCP and CLI defaults
- pass template_config into CLI default resolution for bot creation, passport repair, and MCP sync refresh

## Tests
- uv run pytest tests/community/core/mcp/test_defaults_per_engine.py tests/community/core/bot_management/services/test_engine_resolver.py

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/create_flow.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/create_flow.py
@@ -365,7 +365,11 @@ def create_bot_with_authorization(
             engine_type=spec.engine_type,
         ),
         cli_items=get_default_cli_items(
-            spec.engine_type, spec.template_type
+            spec.engine_type,
+            spec.template_type,
+            ext_info={"aicoding": {"template_config": spec.template_config}}
+            if spec.template_config
+            else None,
         ),
         use_first_passport=use_first_passport,
     )
--- a/src/backend/src/agentclaw/community/core/bot_management/create_flow.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/create_flow.py
@@ -367,7 +367,7 @@ def create_bot_with_authorization(
         cli_items=get_default_cli_items(
             spec.engine_type,
             spec.template_type,
-            ext_info={"aicoding": {"template_config": spec.template_config}}
+            ext_info={"template_config": spec.template_config}
             if spec.template_config
             else None,
         ),
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/cli_defaults.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/cli_defaults.py
@@ -1,0 +1,29 @@
+"""AICoding-specific CLI default resolution."""
+from __future__ import annotations
+
+from typing import Any, List, Mapping, Optional
+
+from agentclaw.community.core.bot_management.engines.aicoding.default_capabilities import (
+    merge_template_preset_capabilities,
+)
+
+
+class AicodingCliDefaultsResolver:
+    """Resolve effective default CLIs for AICoding bots.
+
+    AICoding templates may preset CLI capabilities at
+    ``ext_info.aicoding.template_config.bot_template_config.preset_capabilities.cli``.
+    These are merged by ``cli_code`` onto the AICoding engine default CLI list.
+    """
+
+    def resolve(
+        self,
+        default_items: List[dict],
+        ext_info: Optional[Mapping[str, Any]] = None,
+    ) -> List[dict]:
+        return merge_template_preset_capabilities(
+            default_items,
+            ext_info,
+            capability_key="cli",
+            identity_key="cli_code",
+        )
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/cli_defaults.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/cli_defaults.py
@@ -12,7 +12,7 @@ class AicodingCliDefaultsResolver:
     """Resolve effective default CLIs for AICoding bots.
 
     AICoding templates may preset CLI capabilities at
-    ``ext_info.aicoding.template_config.bot_template_config.preset_capabilities.cli``.
+    ``ext_info.template_config.bot_template_config.preset_capabilities.cli``.
     These are merged by ``cli_code`` onto the AICoding engine default CLI list.
     """
 
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/default_capabilities.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/default_capabilities.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, Mapping, Optional
+from typing import Any, List, Mapping, Optional
 
 _AICODING_EXT_KEY = "aicoding"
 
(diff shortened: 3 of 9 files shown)


totalfrank (changes requested):
**Verdict: REQUEST_CHANGES**

## Gate 0 — Target branch
Target: `dev` (integration branch). Classification: feature extending an existing pipeline (bot-creation/passport/MCP-sync default-CLI resolution). Gate 0 passes trivially — no `REL*` branch involved, no Isolation Audit required.

## Findings

**1. `core/` hardcodes the `"aicoding"` engine key inline at three call sites instead of going through the composition-root wiring the codebase already establishes for this exact payload shape.**

- `src/backend/src/agentclaw/community/core/bot_management/create_flow.py:369-372`
  ```python
  ext_info={"aicoding": {"template_config": spec.template_config}}
  if spec.template_config
  else None,
  ```
- `src/backend/src/agentclaw/community/core/bot_management/services/default_bot_passport_repair_service.py:152-154`
  ```python
  ext_info={"aicoding": {"template_config": bot.get("template_config")}}
  if bot.get("template_config")
  else None,
  ```
- `src/backend/src/agentclaw/community/core/mcp/services/sync_service.py:906-908`
  ```python
  ext_info={"aicoding": {"template_config": template_config}}
  if template_config
  else None,
  ```

`create_flow.py`, `default_bot_passport_repair_service.py`, and `sync_service.py` are engine-agnostic orchestration/service code — none of them are composition roots. Writing the literal `"aicoding"` engine key into their call sites is exactly the "under-abstraction" pattern the constitution names by example: *"dict-of-engine-names defined inline in domain code"* (arch.rules.md, "How engine variability MUST be handled"), and it violates Rule 14 (Invariant — *"Implementation selection and object graph assembly must be driven by validated configuration and performed only in approved composition roots"*).

This is doubly avoidable because the codebase already has the correct pattern for this identical payload, in exactly one place: `_build__ext_info_provider` in `src/backend/src/agentclaw/community/di/modules/skill_center_module.py:192-199`, a DI composition root, which builds `{"aicoding": {"template_config": template_config}}` once and hands callers an opaque provider. `skill_set_service.py` consumes that provider (`self._get_default_capabilities_ext_info(...)`) without ever knowing the `"aicoding"` key exists. This PR bypasses that seam entirely and re-derives the same engine-keyed shape ad hoc, three times, in core.

Apply the "add engine `foo`" test: if `foo` needs the same template-preset CLI merge next month, a developer must edit these three generic core files to add another `"foo": {...}` branch/literal — not just add `foo`'s own resolver module and one registry line. That is precisely the maintenance failure Rule 14 and the engine-variability contract pattern exist to prevent. (The `DefaultItemsResolverRegistry`/`CliDefaultsResolverRegistry`/`AicodingCliDefaultsResolver` machinery this PR adds in `engines/registry.py` and `engines/aicoding/cli_defaults.py` is itself the *correct* shape of the fix — contract + per-engine implementation + default fallback + one composition-root registration. The violation is only in how the `ext_info` payload feeding that machinery gets constructed.)

**Required fix direction:** don't build the `{"aicoding": ...}` dict at these call sites at all. Either (a) have `get_default_cli_items` accept the engine-agnostic `template_config` directly and let the resolved engine's own resolver (already selected via the registry) wrap it into whatever internal shape it needs, entirely inside `engines/aicoding/`; or (b) reuse/extend the existing `_ext_info_provider`-style composition-root builder so callers pass an already-opaque `ext_info` they never construct themselves. Either way, the `"aicoding"` string should appear in exactly one place outside its own engine module (a composition root), not three.

**2. Architecture test coverage does not catch this class of violation, and the PR adds no guard.** `test_default_capabilities_engine_boundaries.py` only AST-scans `core/default_capabilities.py` and `core/mcp/services/_defaults.py` for engine-literal comparisons/imports; it does not scan `create_flow.py`, `default_bot_passport_repair_service.py`, or `sync_service.py`, and it does not check for engine-keyed dict literals at all (only `ast.Compare`). The new call sites in finding 1 are invisible to CI. Per the review procedure (architecture tests / Rule 25 spirit), a guard test should be extended or added so `ext_info={"aicoding": ...}` literals outside `engines/aicoding/` and approved composition roots fail CI.

## Out-of-scope notes (not gating)
- `src/backend/src/agentclaw/community/core/bot_management/services/create_bot_for_others_service.py:367` calls `get_default_cli_items(engine_type, template_type)` without the new `ext_info` — a sibling `apply_first_agent_passport(cli_items=...)` call site this PR did not update. Worth checking whether that path also needs template-preset CLI defaults; flagging only as a completeness/business-logic observation, and as a concrete illustration of why finding 1 matters (duplicated ad-hoc wiring is easy to miss at some call sites).
- The registry generalization (`McpDefaultsResolverRegistry` → `DefaultItemsResolverRegistry` base, new `CliDefaultsResolverRegistry`) and the shared `merge_template_preset_capabilities` helper are well-placed and not flagged.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

**Gate 0**: Target branch is `dev` (the integration branch) — not a `REL*` release branch, so the release-branch gate and Isolation Audit do not apply. This is a plain `feat` PR to `dev`; Gate 0 passes without further scrutiny. Full architecture review applies below.

**Scope note**: The registry/contract refactor (`DefaultItemsResolverRegistry` → `CliDefaultsResolverRegistry`/`McpDefaultsResolverRegistry`, `AicodingCliDefaultsResolver`, dedup of `merge_template_preset_capabilities`) correctly follows the required engine-variability pattern: one contract in the agnostic composition root (`core/bot_management/engines/registry.py`), a per-engine implementation in the engine's own area (`core/bot_management/engines/aicoding/cli_defaults.py`), and a mandatory default fallback (`_EngineCliDefaultsResolver` in `core/mcp/services/_defaults.py`, wired via `_cli_defaults_resolver()`). No finding there.

---

### Finding 1 — engine-specific wire shape hardcoded inline in engine-agnostic core (bypasses the codebase's own established composition-root contract)

**Locations** (identical literal, duplicated three times):
- `src/backend/src/agentclaw/community/core/bot_management/create_flow.py:370-372`
  ```python
  ext_info={"aicoding": {"template_config": spec.template_config}}
  if spec.template_config
  else None,
  ```
- `src/backend/src/agentclaw/community/core/bot_management/services/default_bot_passport_repair_service.py:152-154`
  ```python
  ext_info={"aicoding": {"template_config": bot.get("template_config")}}
  if bot.get("template_config")
  else None,
  ```
- `src/backend/src/agentclaw/community/core/mcp/services/sync_service.py:904-906`
  ```python
  ext_info={"aicoding": {"template_config": template_config}}
  if template_config
  else None,
  ```

**Violated rule**: `arch.rules.md` Rule 14 (Invariant — "Implementation selection and object graph assembly must be driven by validated configuration and performed only in approved composition roots"), and violation catalog items #1 (engine-string branching/hardcoding in core, here via literal engine-namespaced key construction rather than an `if`) and #9 (copy-paste mirroring across call sites instead of a shared contract).

**Why it breaks the architecture**: These three files are engine-agnostic `core/` call sites — none of them are composition roots. Each now hand-builds the `{"aicoding": {"template_config": ...}}` payload shape, which is private, engine-specific wire knowledge (the `_AICODING_EXT_KEY = "aicoding"` namespace consumed by `AicodingDefaultCapabilitiesExtInfo.from_ext_info`). This directly regresses an already-established, correct pattern in the same codebase: for the equivalent MCP-defaults ext_info, the "aicoding" namespacing is built **only** inside the composition root `di/modules/skill_center_module.py::_build__ext_info_provider`, and `core/skill_center/services/skill_set_service.py::_get_default_capabilities_ext_info` consumes it purely through an injected `Callable[[str], Mapping[str, Any] | None]` — core never spells "aicoding" itself. This PR's new CLI-defaults path sidesteps that injected contract entirely and inlines the engine name three times in agnostic code instead.

**Concrete maintenance failure** (the "add engine `foo`" test): if engine `foo` also needs template-preset CLI merging, all three of these agnostic files must be edited to add a `"foo"` branch/key alongside `"aicoding"` — exactly the "more than (own module + one registration)" failure the constitution's engine-variability pattern is designed to prevent. Today it's a static string, tomorrow it becomes `if engine_type == "aicoding": ext_info = {...} elif engine_type == "foo": ...` in three places.

**Required fix**: Do not construct the namespaced ext_info dict in `core/`. Either (a) extend the existing composition-root-injected `ext_info_provider` callback (already wired in `skill_center_module.py` for the MCP path) to also serve these three CLI-defaults call sites, so core only ever calls an injected `Callable[[str], Mapping|None]` and never spells "aicoding"; or (b) if the raw `template_config` is already in hand at these call sites (as it is), pass it to `get_default_cli_items`/the resolver as a plain, unnamespaced parameter (the resolver returned by the registry is already scoped to one specific engine bucket by the time `resolver.resolve(...)` is called, so the engine-name wrapper key is redundant at this point in the flow) rather than each caller re-deriving the engine-specific envelope.

---

### Out-of-scope notes
- Business-logic correctness of the merge/override semantics in `merge_template_preset_capabilities` (position preservation, field-level override vs. append) is not reviewed here — that's feature-review territory.

---

panqiwp:
Close and recreate PR as requested.

PR #1123, opened 2026-08-18
feat(baas): run startup scripts unbounded and bracket runs with log markers

totalfrank:
## Problem

The per-bot startup script's user stage ran under `timeout -k 10 300`, capping every run at 300s (TERM) + 10s (KILL). That cap is unwanted for now: scripts that legitimately need longer than five minutes get killed mid-provisioning. Separately, the script's log (`/home/admin/logs/startup_script.log`) only contained the body's own output, so there was no way to tell from the log when a run began, whether it was still going, or how it ended.

## Solution

- Removed the `timeout -k` wrapper from `_get_startup_script_segment` along with both sizing constants (`STARTUP_SCRIPT_TIMEOUT_SECONDS`, `STARTUP_SCRIPT_KILL_GRACE_SECONDS`) and the method's now-unused parameters. The script runs to completion however long it takes; a run that outlives the 600s publish budget leaves the poller to give up on its own while the start keeps running.
- The stage now brackets each run in the same log: a timestamped `[startup_script] started at ...` line before the body runs, and `[startup_script] finished at ... rc=<exit code>` after it, with `rc` captured from `$?` immediately after the run. With no deadline on the run, these markers are what distinguishes a still-running script from a finished one.
- Everything else about the stage is unchanged: base64-encoded body (no single quotes can enter the `su admin -c '...'` wrapper — the markers use double quotes only), mktemp drop path, `|| true` guard, `__OCB_RC` exit-status re-assertion, payload log redaction, and the byte-identical no-script chain.
- Updated the startup-script contract section in `docs/openapi-v1/README.md` (the only doc that stated the 300s cap) to describe the unbounded run and the new log markers.

## Validation

- `tests/community/core/service_bot/services/test_baas_service_start_cmd.py`: replaced the three timeout tests with four covering the unwrapped invocation, marker ordering (started → run → finished), all three redirects targeting `startup_script.log`, and the exit-code capture — 41 passed.
- Wider sweep: `tests/community/core/service_bot/`, `tests/community/core/bot_startup_script/`, both startup-script repository test files, and `tests/community/endpoints/test_openapi_startup_script.py` — 936 passed; the 2 failures in `test_bot_build_service_skill_artifact.py` are pre-existing (reproduced on the clean base commit).
- Manually executed the rendered user stage against a scratch log with a body of `echo hello-from-script; exit 3`: log shows the started marker, the body's output, then `finished at ... rc=3`, and the wrapper still exits 0.
- `ruff check` clean on both changed Python files (`ruff format --check` drift on them is pre-existing at the base).

## Compatibility and risk

- Published-contract change: the documented 300s/310s cap no longer exists; a script that never exits now holds the start open indefinitely (the publish is marked failed by the poller's own 600s budget, but the hook keeps running). This is the requested behavior "for now"; the rollback path is re-adding the `timeout -k` wrapper and constants.
- Log format addition only — the marker lines share the script's log file and do not alter what the body writes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SzJRZmu51hrwgftGCKvHWc)_

### DIFF
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -1084,21 +1084,26 @@ that one design choice, and none of it is visible in the OpenAPI document.
   A script written after a bot was created reaches a container only once that
   bot restarts. The first write therefore always needs a restart.
 - **A failure degrades rather than blocks — the script's *execution*, that is.**
-  A non-zero exit, a crash, or a timeout leaves the agent running: the script is
+  A non-zero exit or a crash leaves the agent running: the script is
   guarded so it cannot change the boot's outcome, and it is skipped entirely if
   the boot itself failed. That is not a promise to start anyway if the platform
   *cannot read* your stored script: a start that could not resolve it fails
   rather than bringing up a bot that looks ready and is not provisioned.
-- **Limits:** body ≤ **24 KiB** (413 above that, naming the limit); each run
-  capped at **300s** by `timeout`, sized against the 600s publish budget the
-  start reports into. The cap is enforced as TERM at 300s followed by an
-  uncatchable KILL **10s** later, so a script that traps or ignores TERM cannot
-  hold the start open past **310s**. The cap bounds **the start**, not your
-  descendants: a process the script deliberately backgrounds (`something &`)
-  outlives it, because the script itself has exited and the start has already
-  completed — nothing reaps it, so it runs until it ends or the container does.
+- **Limits:** body ≤ **24 KiB** (413 above that, naming the limit). There is
+  currently **no cap on the run itself**: the per-run `timeout` was removed, so
+  the script runs to completion however long it takes. The start only reports
+  once the script exits, and the 600s publish budget the start reports into
+  does not stop the script — a run that overstays it leaves the publish marked
+  failed while the script (and the container's boot) carries on. A script that
+  never exits holds the start open indefinitely, so bound your own long-running
+  work (background it with `something &` if it should outlive the start).
   Interpreter is `bash`; the body runs as `admin`, the same user every platform
   step runs as.
+- **Each run brackets itself in the log.** The platform writes a timestamped
+  `[startup_script] started at ...` line before your body runs and a
+  `[startup_script] finished at ... rc=<exit code>` line after it, into the
+  same log the body's output goes to — with no cap on the run, those markers
+  are how you tell a still-running script apart from a finished one.
 - **Do not put secrets in the body.** This is a hard requirement, not advice.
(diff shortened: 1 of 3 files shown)


totalfrank (commented):
Diff looks clean — no findings.

Reviewed commit: `63b1efc` (single commit on this PR).

Files covered:
- `src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py`
- `src/backend/docs/openapi-v1/README.md`
- `src/backend/tests/community/core/service_bot/services/test_baas_service_start_cmd.py`

What I checked:
- **Correctness**: traced the rebuilt `inner` shell string char-by-char — the `&&`/`;` chain, `rc=$?` capture timing (reads `$?` before any subsequent `$(date ...)` substitution can clobber it), and the marker ordering (started → run → finished) all match what the tests assert.
- **Contract breakage**: `STARTUP_SCRIPT_TIMEOUT_SECONDS`/`STARTUP_SCRIPT_KILL_GRACE_SECONDS` removal has no dangling references anywhere else in the tree; `_get_startup_script_segment` has a single caller and its signature change doesn't break it. `docs/openapi-v1/README.md` was updated to match (300s cap and per-run `timeout` language removed, new marker-log behavior documented).
- **Error/failure handling**: the `|| true` guard and the pre-existing `__OCB_RC` re-assertion around the segment are untouched, so a hung or failing user script still can't fail the boot.
- **Test coverage**: the four new tests in `test_baas_service_start_cmd.py` (unwrapped invocation, marker ordering, log-target count, exit-code capture) directly cover the new behavior and replace the three obsolete timeout tests.

Architecture boundaries, concurrency, and Python type-contract categories didn't apply to this change (shell string construction only, no new async/shared state, no new params).

---

totalfrank:
**Review round 1 — no issues flagged.** The review of `63b1efc` (the PR's single commit) raised no findings; nothing to fix or dispute this round.

CI on `63b1efc`: BCS e2e, Gateway, BaaS, Engine, and BCS unit tests are green; Backend unit tests and Singlebox coverage still running. I'll follow up if either goes red.

---

totalfrank:
**Backend unit tests went red — infra, not the diff.** The failure was a pytest-xdist worker crash (`worker 'gw1' crashed`) attributed to `test_endpoint_injection[GET /openapi/v1/bots/{bot_id}/startup-script :: unknown_bot]`; the other 11856 tests passed. That test and all 9 startup-script endpoint-runner parameters pass locally on this branch, and this PR only changes composed shell strings, tests for them, and docs — nothing that runs in that worker's process. Re-ran the failed job (in progress now); will follow up if it fails again.

---

FreddieSun (approved):
LGTM

PR #1124, opened 2026-08-18
feat(backend): support template preset CLI defaults

panqiwp:
## Summary
- add AICoding CLI defaults resolver for template preset_capabilities.cli
- share default capability registry and template preset merge logic between MCP and CLI defaults
- simplify default capability ext_info to pass template_config directly without an aicoding wrapper
- pass template_config into CLI default resolution for bot creation, passport repair, and MCP sync refresh

## Tests
- uv run pytest tests/community/core/mcp/test_defaults_per_engine.py tests/community/core/bot_management/services/test_engine_resolver.py

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/create_flow.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/create_flow.py
@@ -365,7 +365,11 @@ def create_bot_with_authorization(
             engine_type=spec.engine_type,
         ),
         cli_items=get_default_cli_items(
-            spec.engine_type, spec.template_type
+            spec.engine_type,
+            spec.template_type,
+            ext_info={"aicoding": {"template_config": spec.template_config}}
+            if spec.template_config
+            else None,
         ),
         use_first_passport=use_first_passport,
     )
--- a/src/backend/src/agentclaw/community/core/bot_management/create_flow.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/create_flow.py
@@ -367,7 +367,7 @@ def create_bot_with_authorization(
         cli_items=get_default_cli_items(
             spec.engine_type,
             spec.template_type,
-            ext_info={"aicoding": {"template_config": spec.template_config}}
+            ext_info={"template_config": spec.template_config}
             if spec.template_config
             else None,
         ),
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/cli_defaults.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/cli_defaults.py
@@ -1,0 +1,29 @@
+"""AICoding-specific CLI default resolution."""
+from __future__ import annotations
+
+from typing import Any, List, Mapping, Optional
+
+from agentclaw.community.core.bot_management.engines.aicoding.default_capabilities import (
+    merge_template_preset_capabilities,
+)
+
+
+class AicodingCliDefaultsResolver:
+    """Resolve effective default CLIs for AICoding bots.
+
+    AICoding templates may preset CLI capabilities at
+    ``ext_info.aicoding.template_config.bot_template_config.preset_capabilities.cli``.
+    These are merged by ``cli_code`` onto the AICoding engine default CLI list.
+    """
+
+    def resolve(
+        self,
+        default_items: List[dict],
+        ext_info: Optional[Mapping[str, Any]] = None,
+    ) -> List[dict]:
+        return merge_template_preset_capabilities(
+            default_items,
+            ext_info,
+            capability_key="cli",
+            identity_key="cli_code",
+        )
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/cli_defaults.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/cli_defaults.py
@@ -12,7 +12,7 @@ class AicodingCliDefaultsResolver:
     """Resolve effective default CLIs for AICoding bots.
 
     AICoding templates may preset CLI capabilities at
-    ``ext_info.aicoding.template_config.bot_template_config.preset_capabilities.cli``.
+    ``ext_info.template_config.bot_template_config.preset_capabilities.cli``.
     These are merged by ``cli_code`` onto the AICoding engine default CLI list.
     """
 
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/default_capabilities.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/default_capabilities.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, Mapping, Optional
+from typing import Any, List, Mapping, Optional
 
 _AICODING_EXT_KEY = "aicoding"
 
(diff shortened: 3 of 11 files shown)


totalfrank (approved):
## Architecture-Gate Review

**Verdict**: APPROVE

**Gate 0**: Target branch is `dev` (the integration branch), not a `REL*` release branch — Gate 0's release-branch restrictions and Isolation Audit do not apply. This is ordinary feature work landing on `dev`. Pass.

**Findings**: None.

**Review notes**

- `core/bot_management/engines/registry.py` is the pre-existing composition root for engine provisioning. This PR generalizes `McpDefaultsResolver`/`McpDefaultsResolverRegistry` into a shared `DefaultItemsResolver` Protocol + `DefaultItemsResolverRegistry` base class, then adds `CliDefaultsResolverRegistry` on top of it. The registry keys per-engine implementations by `AICODING_ENGINE_TYPE` (imported from the engine's own `aicoding/strategy.py`, not hardcoded), so this is a clean instance of the required contract pattern rather than a new deviation.
- New engine-specific logic (`AicodingCliDefaultsResolver`, `merge_template_preset_capabilities`) lives entirely under the existing `core/bot_management/engines/aicoding/` implementation area — an established per-engine module, not a new one introduced by this diff.
- A default/fallback implementation is present and mandatory-compliant: `_EngineCliDefaultsResolver` in `core/mcp/services/_defaults.py` is registered as the fallback via `_cli_defaults_resolver()` (`registry.resolve(bucket) or _DEFAULT_CLI_RESOLVER`), so engines without a registered resolver degrade to a safe identity pass-through rather than crashing or requiring engine-name checks in agnostic code. `test_template_config_cli_presets_are_aicoding_specific` exercises this fallback path (`openclaw` + template preset → `[]`).
- The agnostic call sites that were touched (`create_flow.py`, `sync_service.py`, `default_bot_passport_repair_service.py`) only forward the bot's own `template_config` field into a generic `ext_info={"template_config": ...}` dict — no engine-name branching was added at any of these sites. Confirmed via diff scan: no new `engine_type ==`, `active_engine in {...}`, or `os.environ`/`get_current_env()` usage introduced anywhere in this changeset.
- The `ext_info.aicoding.template_config` → `ext_info.template_config` shape simplification in `AicodingDefaultCapabilitiesExtInfo` is consumed only within the aicoding resolver's own interpretation of the opaque `ext_info` blob passed through the `DefaultItemsResolver` contract; it isn't an externally-versioned wire protocol, so no propagation/compat-doc update is required under Rule 16.
- Test coverage (`test_defaults_per_engine.py`) covers the new CLI resolver, the shared merge helper, and the default-fallback path for non-aicoding engines.

**Out-of-scope notes** (not gated on):
- `core/bot_management/services/create_bot_for_others_service.py:367` still calls `get_default_cli_items(engine_type, template_type)` without passing `ext_info`, so that path won't pick up template preset CLIs. This file isn't touched by this PR; flagging only as a product-completeness note, not an architecture finding.


---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

### Gate 0 — Target branch
Target is `dev` (integration branch), not a `REL*` release branch. Gate 0 does not apply; no re-target issue.

### Findings

**1. `src/backend/src/agentclaw/community/di/modules/skill_center_module.py:192-199` — silent contract break: a producer of the `ext_info` wire shape was not updated, causing silent loss of AICoding template-preset MCPs.**

```python
def _build__ext_info_provider(injector: Injector):
    def provider(bot_id: str) -> dict | None:
        template_config = injector.get(_template_service_cls()).get_template_config(bot_id)
        if not isinstance(template_config, dict):
            return None
        return {"aicoding": {"template_config": template_config}}
    return provider
```

This PR removes the `"aicoding"` wrapper key from the `ext_info` shape read by `AicodingDefaultCapabilitiesExtInfo.from_ext_info` (`default_capabilities.py`): it now reads `ext_info.get("template_config")` directly instead of `ext_info.get("aicoding", {}).get("template_config")`. Three producers were updated to match (`create_flow.py`, `default_bot_passport_repair_service.py`, `mcp/services/sync_service.py`), and the test suite was rewritten to match the new shape — including a new test, `test_legacy_wrapped_template_config_mcp_presets_are_ignored`, which explicitly asserts that the *old* wrapped shape now silently resolves to **no** template presets.

`_build__ext_info_provider` is a fourth, un-updated producer. It is wired into production via the `skill_set_service_factory` composition-root provider (`skill_center_module.py:617`, `ext_info_provider=_build__ext_info_provider(injector)`), consumed by `SkillSetService._get_default_capabilities_ext_info`, and threaded into `get_default_mcp_servers` / `get_default_mcp_server_codes` / `get_default_mcp_config` throughout `skill_set_service.py` (skill-set MCP membership resolution, default skill-set MCP association, exclusion checks). After this merges, that whole path will silently stop merging AICoding template-preset MCP capabilities — no exception, no log, just missing capabilities — because the PR's own new test proves the legacy wrapped shape is now treated as absent.

- Violates **arch.rules.md Rule 16 — Changes Propagate (Invariant)**: a structural change to a cross-component contract (the `ext_info` shape, threaded through 4+ producers and several consumers) must identify all affected producers/consumers; this diff's PR description contains no propagation analysis and missed a live producer.
- Falls under violation catalog item 5 (**Hidden wire protocols and contracts**): `ext_info` has no explicit schema (Protocol/TypedDict/dataclass) — it's an ad hoc dict shape assumed consistent across independently-maintained call sites. This PR unilaterally changes that implicit shape in the consumer without a compatibility bridge, leaving two incompatible producer shapes coexisting in the codebase.

**Required fix**: update `_build__ext_info_provider` to emit `{"template_config": template_config}` (matching the new shape), and grep for any other producers of this dict before merging. Longer-term, this ad hoc dict should become an explicit, documented type (e.g. a small dataclass/TypedDict) that all producers construct through, so this class of drift fails at construction time instead of silently at read time.

### Out-of-scope notes
- The `DefaultItemsResolver` Protocol + `CliDefaultsResolverRegistry`/`McpDefaultsResolverRegistry` + per-engine `AicodingCliDefaultsResolver`/`AicodingMcpDefaultsResolver` + default fallback (`_EngineCliDefaultsResolver`) in `registry.py`/`_defaults.py` correctly follow the required engine-variability contract pattern (contract in the agnostic layer, per-engine implementation in the engine's own module, registration in the composition root, safe default fallback). No under/over-abstraction finding there.
- No conformance test suite exists for `DefaultItemsResolver` implementations (Rule 25), but this is a pre-existing gap from the original `McpDefaultsResolver` Protocol, not newly introduced here — flagged for awareness only, not gating this verdict.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

**Scope note:** this is an architecture-gate review only (per `docs/arch/arch.rules.md` and `AGENTS.md`). I did not evaluate template/preset business logic, CLI/MCP merge semantics, or naming — only structural/contract correctness.

### Gate 0 — Target branch
Target is `dev` (integration branch), not a `REL*` release branch, so the release-branch isolation gate does not apply. Gate 0 passes trivially.

### Layering review
- `core/bot_management/engines/registry.py` — composition-root-style registry. The new `CliDefaultsResolverRegistry`/`DefaultItemsResolverRegistry` generalization correctly follows the required engine-variability pattern: one contract (`DefaultItemsResolver` Protocol), per-engine implementation registered in `_build_cli_defaults_resolver_registry()`, and a mandatory default fallback (`_EngineCliDefaultsResolver` in `core/mcp/services/_defaults.py`) so an engine without its own resolver never crashes. No engine-string branching was added to agnostic domain code. This part is sound.
- `core/bot_management/engines/aicoding/{cli_defaults.py,default_capabilities.py,mcp_defaults.py}` — engine-specific implementations correctly isolated in the engine's own area, not leaking into `core/mcp/services/_defaults.py`.

### Findings

**1. `src/backend/src/agentclaw/community/di/modules/skill_center_module.py:199` — contract change not propagated to a live producer (arch.rules.md Rule 16 — Changes Propagate, Invariant; also Rule 4).**

```python
def provider(bot_id: str) -> dict | None:
    template_config = injector.get(_template_service_cls()).get_template_config(bot_id)
    if not isinstance(template_config, dict):
        return None
    return {"aicoding": {"template_config": template_config}}
```

This PR changes `AicodingDefaultCapabilitiesExtInfo.from_ext_info` (`core/bot_management/engines/aicoding/default_capabilities.py`) from reading `ext_info["aicoding"]["template_config"]` to reading `ext_info["template_config"]` directly, and updates every call site it touches (`create_flow.py`, `sync_service.py`, `default_bot_passport_repair_service.py`, and the test file) to build the new unwrapped shape.

It does **not** update `skill_center_module.py:199`, which is a composition root still producing the *old* wrapped shape `{"aicoding": {"template_config": ...}}`. That provider is wired into `SkillSetService._ext_info_provider` (`di/modules/skill_center_module.py:617`) and reaches `get_default_mcp_servers` / `get_default_mcp_server_codes` through `SkillSetService._get_default_capabilities_ext_info` at every call site in `core/skill_center/services/skill_set_service.py` (skill-set MCP add/remove, `collect_bot_active_mcps`, `collect_bot_mcps`, `get_active_skill_sets_mcp_summary`, etc.).

After this change, `ext_info.get("template_config")` on that old-shaped payload always resolves to `None` (the real key is `"aicoding"`), so `AicodingMcpDefaultsResolver.resolve()` silently stops merging any bot template's `preset_capabilities.mcp` into default MCP lists surfaced through every skill-center flow — a silent regression with no test or CI signal, since this PR's own tests (`test_defaults_per_engine.py`) only exercise `get_default_mcp_servers`/`get_default_cli_items` directly with the new unwrapped shape and never exercise the `skill_center_module` DI wiring end-to-end.

This is exactly the "changes propagate" failure the constitution calls out: a structural/contract change (the `ext_info` wire shape) landed without identifying all affected consumers/producers.

**Required fix:** update `skill_center_module.py:199` to produce `{"template_config": template_config}` in the same change, and add a regression test that exercises `SkillSetService` (or the `skill_center_module` DI wiring) with a template-config-bearing `ext_info` end-to-end, so this producer/consumer pair can't silently drift apart again (Rule 25 — conformance coverage for this internal contract).

### Out-of-scope notes
- The `ext_info` shape (`template_config.bot_template_config.preset_capabilities.{cli,mcp}`) is an implicit, undocumented internal contract between composition-root providers and the engine resolvers — not a new problem introduced here, but this PR would be a good place to give it an explicit shape (e.g. a small dataclass/TypedDict) so a shape change can't be partially applied again.
- Template preset merge semantics (override-by-identity-key, append order) are business logic and out of scope for this review.


---

panqiwp:
Close and recreate PR as requested.

PR #1129, opened 2026-08-18
feat(backend): expose the harness surface through the gateway

wienyang:
Problem
  
    The backend's harness operations (diagnose / preview / apply / rollback / dim-report / dim-history) were only reachable through the internal /api/harness/** routes. They were absent from the unified public OpenAPI surface, so
    callers going through the gateway had no way to reach them.

    Solution

    Mount a public harness surface at /openapi/v1/harness/bots/{bot_id}/* and route it through the gateway:

    - Public router (openapi_v1/harness/): six thin endpoints reusing the existing harness services, with stricter request DTOs (entity_type/entity_id required, bot_id on the path). Access is owner-or-collaborator via
    HarnessBotAccessDep (404 on failure, matching the public convention); diagnose records trigger_source="openapi".
    - Admission: the six operations are registered as GRANT_CHECKED_OWN_BOT in the admission table, with a pinned HARNESS_SCOPED_OPERATIONS set; the path-convention, explicit-user-id, and admission-inventory tests are widened to
    cover them.
 - Gateway: the backend dump is split per domain (--path-prefix), dump_and_publish.sh publishes a separate harness.openapi.json artifact, and the shipped config adds a harness domain (/openapi/v1/harness/bots/** → backend,
    verbatim) behind a user: required route-security rule.
    - The internal /api/harness/** routes are untouched; existing callers are unaffected.
    
    Validation
    
    - Backend: 850+ tests pass under tests/community/adapters/http/openapi_v1/, including new smoke tests (401 without a principal, 404 for non-owner without a grant, owner/collaborator happy paths) and the schema-docs,
    path-convention, and admission-inventory convention suites.
    - Gateway: 121 unit tests pass, including new pins that the shipped config routes harness paths to the backend with user: required and that the served document carries exactly the six harness operations with x-avernet-security.
    - End to end: dump_openapi.py --path-prefix /openapi/v1/harness produces the six operations, and gate_and_publish_openapi.py publishes configs/schemas/harness.openapi.json as compatible.


src/backend/src/agentclaw/community/adapters/http/openapi_v1/harness/router.py
```diff
@@ -0,0 +1,705 @@
...
+)
+from agentclaw.community.log import get_logger
+
+logger = get_logger()
+
+router = APIRouter(prefix="/openapi/v1/harness", tags=["harness"])
+
+
+async def require_harness_bot_access(
+    principal: Annotated[Principal, Depends(require_principal)],
```
totalfrank: This is not necessary. We have provided dep components like UserIdDep, OwnerIdDep which in turn depends on Depends(require_principal). You don't need to explicitly define it here.

src/backend/tests/community/adapters/http/openapi_v1/test_admission_inventory.py
```diff
@@ -160,79 +153,68 @@ def test_every_public_operation_still_requires_a_principal():
...
-    dependency = _GRANT_DEPENDENCY_BY_MODE[mode]
     expected = {
         key
-        for key, table_mode in ADMISSION.items()
-        if table_mode is mode
+        for key, mode in ADMISSION.items()
+        if mode in _GRANT_CHECKED_MODES
         and key not in SELF_CHECKED_ROUTES
         and key not in SKILL_SCOPED_OPERATIONS
+        and key not in HARNESS_SCOPED_OPERATIONS
```
totalfrank: Did you intended to do this? You already set up admission for these routes but disabling them here seems to be inconsistent.

totalfrank (commented):
Comments posted, please take a look

totalfrank (approved):
LGTM.

PR #1053, opened 2026-08-14
fix: pass service bot publish sandbox envs

panqiwp:
## Summary\n- compose service-bot publish extra_envs from skills env and engine provisioning envs\n- forward source bot template_config separately so BaaS can apply template_config.envs/image/resource_spec sandbox overrides\n- preserve publish image pin priority over template_config.image\n\n## Test\n- uv run pytest tests/community/core/service_bot/services/test_bot_build_service_teclaw_routing.py tests/community/core/service_bot/services/test_publish_flow_service.py -q

PR #1141, opened 2026-08-18
fix: omit bot type env for service publish

panqiwp:
## Summary\n- omit legacy BOT_TYPE from service-bot publish engine extra envs\n- keep other engine-owned envs for service publish/restart/rollback\n- update publish flow tests to assert service publish extra_envs no longer carries BOT_TYPE\n\n## Test\n- cd src/backend && uv run pytest tests/community/core/service_bot/services/test_publish_flow_service.py -q

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -193,8 +193,10 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
 
         # Historical template types expose stable legacy bot-type values, while
         # template-factory templates expose their template_type verbatim so new
-        # template types do not require backend enum/map changes.
-        if template_type:
+        # template types do not require backend enum/map changes. Service bots
+        # already convey their service identity through the startup command, so
+        # skip this legacy template-kind env for that domain.
+        if template_type and ctx.bot_type != "service":
             envs["BOT_TYPE"] = LEGACY_BOT_TYPE_ENV_MAP.get(template_type, template_type)
 
         devflow_workflow = template_config.get("devflow_workflow", "")
--- a/src/backend/tests/community/core/bot_management/test_engine_provisioning_strategy.py
+++ b/src/backend/tests/community/core/bot_management/test_engine_provisioning_strategy.py
@@ -331,7 +331,7 @@ def test_resolve_provisioning_routes_legacy_template_only_context():
     assert strategy.should_encrypt_template_token(ctx) is True
 
 
-def test_template_factory_normal_cc_consumes_model_runtime_repos_and_token():
+def test_template_factory_service_normal_cc_consumes_model_runtime_repos_without_bot_type():
     strategy = AicodingProvisioningStrategy("claude_code")
     ctx = BotProvisioningContext(
         active_engine="claude_code",
@@ -361,7 +361,7 @@ def test_template_factory_normal_cc_consumes_model_runtime_repos_and_token():
         "https://code/repo1",
         "https://code/repo2",
     ]
-    assert envs["BOT_TYPE"] == "normalCC"
+    assert "BOT_TYPE" not in envs
     assert strategy.should_encrypt_template_token(ctx) is True
     assert strategy.extract_runtime_token(ctx) == "tok-normal"
 
--- a/src/backend/tests/community/core/service_bot/services/test_publish_flow_service.py
+++ b/src/backend/tests/community/core/service_bot/services/test_publish_flow_service.py
@@ -272,7 +272,7 @@ def test_bot_build_service_upgrade_raises_for_other_errors():
 
 
 @pytest.mark.asyncio
-async def test_verify_first_release_injects_engine_extra_envs_for_service_coding_bot():
+async def test_verify_first_release_injects_engine_extra_envs_without_bot_type_for_service_coding_bot():
     publish_service = Mock()
     publish_service.create_device_binding.return_value = 55
     build_service = Mock()
@@ -307,7 +307,6 @@ async def test_verify_first_release_injects_engine_extra_envs_for_service_coding
 
     assert build_service.release_async.await_args.kwargs["extra_envs"] == {
         "AGENTCLAW_SKILLS_LAYOUT": "legacy",
-        "BOT_TYPE": "application",
         "AIX_DEVFLOW_INFO": "devflow/app.yaml",
         "GIT_ADDRESSES": '["git@x/y.git"]',
         "RELAY_DEFAULT_MODEL": "antchat/Ling-2.6-1T",
@@ -316,7 +315,7 @@ async def test_verify_first_release_injects_engine_extra_envs_for_service_coding
 
 
 @pytest.mark.asyncio
-async def test_verify_upgrade_preserves_skills_env_and_injects_engine_extra_envs():
+async def test_verify_upgrade_preserves_skills_env_and_injects_engine_extra_envs_without_bot_type():
     publish_service = Mock()
     build_service = Mock()
     build_service.upgrade_async = AsyncMock(return_value={"publish_id": 9})
@@ -365,7 +364,6 @@ async def test_verify_upgrade_preserves_skills_env_and_injects_engine_extra_envs
     assert build_service.upgrade_async.await_args.kwargs["extra_envs"] == {
         "AGENTCLAW_SKILLS_LAYOUT": "pool",
         "AGENTCLAW_SKILLS_LAYOUT_CONTRACT_VERSION": "skills-pool-p3-v1",
-        "BOT_TYPE": "application",
         "RELAY_DEFAULT_MODEL": "m1",
     }
 


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0

- **Target branch:** `REL20260818` (a `REL\d{8}` release branch).
- **Classification:** existing-behavior bug fix (path a) — the touched file, `src/backend/src/agentclaw/community/core/service_bot/services/deploy/service_publish_env.py`, is byte-identical between `origin/dev` and `origin/REL20260818` at the PR's merge base, and this PR changes existing logic in it rather than adding new public surface.
- **Result: FAIL.** A release-branch fix must either be a cherry-pick of a commit already reviewed/merged on `dev` (referenced as such, e.g. "(#NNN)"), or the PR must state the `dev` counterpart. Neither is true here: the PR description contains no reference to a `dev` PR/commit, and `dev`'s current copy of `service_publish_env.py` still has the *unfixed* behavior (no `BOT_TYPE` stripping). Per the branch policy, a release-branch fix with no `dev` counterpart is a regression-by-omission — the next release cut from `dev` will reintroduce the bug this PR is fixing. This alone forces REQUEST_CHANGES regardless of the fix's code quality.
- **Required action:** open (or link) an equivalent PR against `dev` for the same change, and reference it from this PR.

## Findings

1. **`src/backend/.../deploy/service_publish_env.py:47-53` — engine-specific key semantics hardcoded into the engine-agnostic composition file (arch.rules.md Rule 4 — Invariant; violation catalog #1/#5).**

   ```python
   envs = dict(strategy.build_extra_envs(ctx) or {})
   # Some engine strategies use BOT_TYPE as a legacy template-kind env
   # (for example, template type -> runtime category). Service publish
   # already passes service identity via the startup command
   # ``--bot_type service``, so do not leak template kind into BaaS sandbox
   # envs as BOT_TYPE.
   envs.pop("BOT_TYPE", None)
   ```

   `EngineProvisioningStrategy.build_extra_envs` (`.../engines/provisioning.py:66-67`) is the declared contract: "Return extra env vars to inject into the runtime container." Only `AicodingProvisioningStrategy` currently populates `BOT_TYPE`, using `LEGACY_BOT_TYPE_ENV_MAP` (`.../engines/aicoding/strategy.py:37-40,197-198`) — that mapping and its meaning are owned by the aicoding strategy, not by callers. This change makes the agnostic `service_publish_env.py` reach past the contract and unilaterally discard a key based on out-of-band knowledge of what one specific engine strategy means by that key name. The comment itself concedes this ("Some engine strategies use BOT_TYPE as a legacy template-kind env") — that is exactly the kind of hidden, undocumented cross-component knowledge Rule 4 forbids leaking outside the contract that defines it.

   Apply the "add engine `foo`" test: if `foo` legitimately needs to emit `BOT_TYPE` with different, valid semantics tomorrow, this blanket `.pop("BOT_TYPE", None)` in agnostic code silently discards it for every caller of `service_publish_extra_envs`, with no way for `foo`'s strategy to opt back in. That is a new engine requiring downstream awareness of an agnostic-layer special case — the inverted direction from what the contract pattern requires.

   **Fix direction:** the "identity is already conveyed via `--bot_type service`" fact belongs in the request context, not as post-hoc surgery on the result. Add a domain-named field to `BotProvisioningContext` (e.g. `bot_type_conveyed_out_of_band: bool`, defaulting to `False`) and have `AicodingProvisioningStrategy.build_extra_envs` itself skip setting `BOT_TYPE` when that flag is set. `service_publish_env.py` then only sets the flag on the context it builds; it never needs to know that `BOT_TYPE` is the key in question, and any future engine strategy that also owns a `BOT_TYPE`-like key inherits correct behavior through the same context field instead of needing a matching agnostic-layer carve-out.

## Out-of-scope notes

- The `service_publish_extra_envs` docstring update (removing `BOT_TYPE` from the "Engine-owned envs" list) and the test renames are consistent with the intended behavior change and are not gated here.
- Whether omitting `BOT_TYPE` for service-bot publish is the *correct* product behavior (vs. some other resolution) is a business-logic call I'm not gating on — only that it's implemented in the wrong place.


---

totalfrank (changes requested):
**Verdict:** REQUEST_CHANGES

**Gate 0**
- Target branch: `REL20260818` (release branch, matches `^REL\d{8}`).
- Classification: existing-behavior bug fix (path a) — `service_publish_env.py`'s `_engine_extra_envs_from_bot` / `service_publish_extra_envs` already exist on `dev`, and the diff is minimal (add a `dict()` copy + one `pop`, plus test updates).
- **Fails the dev-counterpart check.** `f7335047` is not an ancestor of `origin/dev`, there is no other PR by this author touching `service_publish_env.py` targeting `dev` (checked via `search_pull_requests`/`list_pull_requests`), and the PR body does not reference a `dev` PR or say "cherry-pick". Per the release-branch policy, a release-branch fix with no `dev` counterpart is a finding on its own: the next release cut from `dev` will not carry this fix, so the `BOT_TYPE` leak this PR fixes on `REL20260818` regresses on the very next release. **Action required:** open the same fix against `dev` (or confirm/link an existing merged `dev` PR) before this can land on the release branch.

**Findings**

1. `src/backend/src/agentclaw/community/core/service_bot/services/deploy/service_publish_env.py:47-53` — 
   ```python
   envs = dict(strategy.build_extra_envs(ctx) or {})
   # Some engine strategies use BOT_TYPE as a legacy template-kind env ...
   envs.pop("BOT_TYPE", None)
   ```
   Violated rule: arch.rules.md Rule 1 (contracts are the single authority) / Rule 4 (interfaces mean only what their spec defines) — catalog item 5, hidden contract.

   `build_extra_envs` is the declared `EngineProvisioningStrategy` contract (`core/bot_management/engines/provisioning.py:66`); its documented semantics are just "return extra env vars to inject into the runtime container." `BOT_TYPE` is not a generic concept in that contract — it's an implementation detail owned entirely by `AicodingProvisioningStrategy` (`core/bot_management/engines/aicoding/strategy.py:37,198`, `LEGACY_BOT_TYPE_ENV_MAP`), which some other engine could reuse for an unrelated purpose. By popping a magic string key out of the contract's return value at this generic, engine-agnostic call site, `service_publish_env.py` now silently encodes engine-specific knowledge ("this key means legacy template-kind, and service publish must not leak it") that is documented only in a comment, not in the contract. This is exactly the "shared field documented as 'engine X uses this'" anti-pattern the constitution calls out.

   Concretely, this also breaks the "add engine `foo`" test: if `foo`'s strategy legitimately needs to return its own `BOT_TYPE`-named env for a different purpose, or emits the same legacy concept under a different key, this generic pop silently does the wrong thing (over- or under-strips) with no visibility into `foo`'s own module.

   Required fix: evolve the contract instead of patching its output. Add an explicit, domain-named field to `BotProvisioningContext` (e.g. `identity_conveyed_externally: bool` / `suppress_legacy_bot_type_env: bool`) that `service_publish_env.py` sets when calling `resolve_provisioning(...)` for service-bot publish, and have `AicodingProvisioningStrategy.build_extra_envs` (the only owner of `BOT_TYPE`) honor that flag when deciding whether to set `envs["BOT_TYPE"]`. That keeps the `BOT_TYPE` key name and its legacy semantics entirely inside the aicoding strategy module — core only passes a fact it actually understands (service identity is already conveyed via the startup command), not a foreign engine's env-var key name.

**Out-of-scope notes**
- The underlying bug motivation (service publish already conveys identity via `--bot_type service`, so a stale `BOT_TYPE` env is redundant/misleading) is a reasonable fix; not disputing the business intent, only the placement.

---

totalfrank (changes requested):
**Verdict: REQUEST_CHANGES**

## Gate 0 — Target branch

- Target: `REL20260818` (matches `^REL\d{8}`) — a release branch.
- Classification: **existing-behavior bug fix** (path a). The diff modifies an existing code path (`_engine_extra_envs_from_bot` / `service_publish_extra_envs`), adds no new public surface, and is minimal.
- **Gate 0 fails on the dev-counterpart requirement.** I checked `dev`'s copy of `src/backend/src/agentclaw/community/core/service_bot/services/deploy/service_publish_env.py` — it still unconditionally forwards `BOT_TYPE` (the pre-fix behavior); the only commit ever touching that file on `dev` is `cd9e438` / PR #1053, which is what *introduced* the BOT_TYPE leak this PR is fixing. No PR exists on `dev` with this fix, and the PR body doesn't reference one (no "(#NNNN)" / "cherry-pick" marker). Per the branch policy, a release-branch fix with no `dev` counterpart is a finding on its own — it will be silently reverted (regression by omission) the next time `REL*` is cut from `dev`. This must also land on `dev` (a companion PR or the same change rebased onto `dev`) before/alongside this one.

## Findings

1. **`src/backend/src/agentclaw/community/core/service_bot/services/deploy/service_publish_env.py:54`** — engine-owned key stripped by literal string name in engine-agnostic core code.
   ```python
   envs.pop("BOT_TYPE", None)
   ```
   `_engine_extra_envs_from_bot` is generic composition code that calls `resolve_provisioning(...)` for *whatever* engine the bot uses — it has no engine context of its own. `BOT_TYPE` is not a generic concept here: it is emitted exclusively by `AicodingProvisioningStrategy.build_extra_envs` (`.../engines/aicoding/strategy.py`), which already receives `ctx.bot_type` (here `"service"`) inside the very same `BotProvisioningContext` this function builds. The engine already owns the decision of whether to emit `BOT_TYPE`; this change instead teaches the generic composition layer a fact about one specific engine's env-key contract ("BOT_TYPE is aicoding/claude_code's legacy template-kind env," per the added comment) and filters on it by name.

   This is exactly the under-abstraction failure mode the constitution's engine-variability pattern exists to prevent: apply the "add engine `foo`" test — if `foo`'s strategy also emits a legacy key that needs suppressing for service-publish, this same file needs another hardcoded `.pop(...)` line. That fails the required pattern (contract + per-engine implementation + default fallback, zero edits to agnostic code for a new/changed engine).

   **Required fix:** move the decision into `AicodingProvisioningStrategy.build_extra_envs` itself — it already has `ctx.bot_type` available; have it skip setting `envs["BOT_TYPE"]` when `ctx.bot_type == "service"` (or whatever domain condition governs "identity already passed via `--bot_type`"). `service_publish_env.py` should stay a transparent passthrough of `strategy.build_extra_envs(ctx)` with no knowledge of individual env-key names. Note the surrounding docstring on this function already leaked this engine-specific detail into core commentary before this PR (pre-existing issue) — this PR turns that leaked knowledge into actual filtering behavior, which entrenches rather than fixes the layering problem.

## Out-of-scope notes

- Whether omitting `BOT_TYPE` from service-publish envs is the functionally correct behavior (vs. `--bot_type service` on the CLI being sufficient) is a business-logic call I'm not gating on — only its placement in the architecture.
- Test updates in `test_publish_flow_service.py` look consistent with the intended behavior change; not reviewed for coverage adequacy.


---

panqiwp:
Closing this PR to reopen with a fresh review thread after addressing the implementation feedback.

PR #1135, opened 2026-08-18
feat(gateway): add singlebox lifecycle scripts

xianmuyq:
## Problem

  `src/gateway` lacked integration with the repository's `scripts/singlebox.sh` local development entrypoint. Developers could not manage Gateway with the same setup/start/stop/status/check
  workflow used by other local services, which made Gateway local startup less consistent and harder to operate alongside the rest of the stack.

  ## Solution

  Added Gateway as a first-class singlebox service without adding it to the default `all` stack, avoiding a port conflict with the existing backend default on `8888`.

  The change includes:

  - Added `scripts/modules/gateway.sh` with Gateway lifecycle functions:
    - `gateway_setup`
    - `gateway_start`
    - `gateway_stop`
    - `gateway_status`
    - `gateway_ready`
    - `gateway_prereqs`
    - `gateway_help`
  - Wired `gateway` into `scripts/singlebox.sh` service dispatch and help output.
  - Added `--gateway-port` support, with default Gateway singlebox port `8889`.
  - Updated `src/gateway/scripts/app.sh` to honor `GATEWAY_PORT` / `APP_PORT` for lifecycle health checks.
  - Updated Gateway config loading so `GATEWAY_PORT` overrides `application.yaml`'s web port, ensuring the actual uvicorn listener uses the selected singlebox port.
  - Added unit test coverage for the Gateway port override behavior.

  Rejected alternative:

  - Did not add Gateway to the default `all` stack because the Gateway default application config uses port `8888`, which conflicts with the Backend service. Keeping Gateway opt-in preserves
  current local stack behavior.

  ## Validation

  Ran shell syntax validation:

  ```bash
  bash -n scripts/singlebox.sh scripts/modules/gateway.sh src/gateway/scripts/app.sh

  Ran singlebox Gateway status check with an overridden port:

  OCB_SKIP_GIT_HOOKS=1 ./scripts/singlebox.sh --gateway-port 8899 status gateway

  Result:

  Gateway: Stopped (port: 8899)

  Ran focused Gateway config unit tests:

  UV_CACHE_DIR="$PWD/scripts/.dependencies/uv-cache" uv run --project src/gateway pytest src/gateway/tests/unit/plugins/test_config.py -q

  Result:

  45 passed

  Pre-push hook also passed in lint-only mode during push. Gateway heavier CI / coverage was not run locally because the default pre-push mode is lint-only unless OCB_PRE_PUSH_RUN_CI=1 is
  set.

  ## Compatibility and risk

[...]

PR #1145, opened 2026-08-18
feat(gateway): wire singlebox forwarding to local upstreams

totalfrank:
## Problem

PR #1135 made Gateway startable through `scripts/singlebox.sh` (targeting `dev`), but a locally started Gateway could not actually forward anything to the local stack:

- `application.yaml` `user_config.upstream_vars` still pointed at the `*.sample.com` placeholders, and the `${...}` expansion reads only that YAML map — there is no env override or reachable overlay in the singlebox flow. The gateway booted, passed `/health`, and then dialed nonexistent hosts on every forwarded request.
- Even with correct URLs, forwarding fails on the signed-principal contract: without `AVERNET_SECRET_PRINCIPAL_SIGNING_KEY_VALUE` the gateway answers `500 principal signing failed` on every forward, and without `AGENTCLAW_SECRET_GATEWAY_PRINCIPAL_SIGNING_KEY_VALUE` the backend 401s everything on `/openapi/v1` (BCS already self-arms a dev key in `bcs.sh`).
- The only community `user` identity strategy verifies real Google access tokens against Google's live userinfo endpoint, so under the fail-closed `/**: user: required` route table, a dev box without Google credentials 401s at the edge.

This branch targets `dev_refactory_collaboration`, which did not contain #1135, so its commit is cherry-picked here first.

## Solution

Cherry-pick of #1135 (`feat(gateway): add singlebox lifecycle scripts`), plus commits closing the three gaps:

1. **Upstream URLs** — `upstream_vars` now carries the singlebox defaults: backend `http://127.0.0.1:8888`, baas `http://127.0.0.1:8890`, bcs `http://127.0.0.1:21000`, bcsfuse `http://127.0.0.1:8765`. This matches the file's documented single-box-default role (deployed editions override the section). `engine_proxy_server_url` keeps a placeholder: no singlebox service serves `/proxypass`, and only the `bots-messages-ws` socket domain reads it. The new `spaces` / `work-orders` / `work-order-notifications` domains on this branch all ride the `backend` server, so they are covered by the same value.
2. **Signing contract** — `scripts/modules/gateway.sh` exports `AVERNET_SECRET_PRINCIPAL_SIGNING_KEY_VALUE` with the same NOT-FOR-PROD dev default `bcs.sh` already uses, so gateway signer and BCS verifier agree without coordination; `scripts/modules/backend.sh` arms the backend verifier's community env var with the matching value on the singlebox (LOCAL_MODE) launch path only. The application itself still ships no fallback key; the deliberate arming lives in the launchers, and `.env.local` overrides all of them together. The backend `docs/openapi-v1/README.md` / `README.zh-CN.md` singlebox-key paragraphs are updated to match.
3. **Local caller identity** — a new `dev_header` user strategy maps the `x-dev-user` header onto a `UserPrincipal` verbatim, double-gated on `GATEWAY_AUTH_MOCK=1`. The strategy is deliberately **not** in `PluginContainer.authn_strategies` — that dict is the production DI graph and the enterprise edition merges its own strategies into it — so a config overlay naming `dev_header` is refused at boot as an unknown strategy. Instead the bootstrap constructs and appends it to the user chain only under the env var (without it the module is never even imported), and the strategy itself answers `None` without the same env var, so any future wiring that reaches the class some other way still gets an inert strategy. It is appended after `google`, so a real token keeps winning. `gateway.sh` defaults the flag to `1` (set `GATEWAY_AUTH_MOCK=0` in `.env.local` to require real credentials), mirroring `BCS_AUTH_MOCK` on the BCS side.

Rejected alternatives:

- Env-var overrides for each upstream URL in the config loader (the `GATEWAY_PORT` precedent): workable, but the committed YAML is already the single-box default per its own comments, so editing it directly needs no new mechanism and keeps one source of truth.
- Pointing `authn.google.userinfo_url` at a local mock userinfo server instead of a dev strategy: zero gateway code, but requires an extra always-running process for gateway local dev and hides an auth mock inside the mock model server.
- Registering the dev strategy in the DI pool with env-gated chain composition: rejected because the pool is shared with production/enterprise wiring; the mock now exists only in a boot that explicitly asked for it.

Known limitations (documented, deliberately out of scope): `chat` → baas forwarding still fails in community singlebox — baas resolves its gateway JWT secret from an unseeded stub secret store and additionally requires provisioned `baas_resource_key` records; and the `bots-messages-ws` socket domain has no local upstream.

## Validation

**Full end-to-end singlebox run** (fresh Linux container): brought up gateway (:8889), backend (:8888), BCS (:21000, cargo-built from source), bcsfuse (:8765), and BAAS (:8890) via `scripts/singlebox.sh` on this branch, then smoke-tested forwarding through the gateway, attributing each response via the gateway's forward log:

| Route through gateway :8889 | Result |
|---|---|
| `GET /health` | 200 |
| `GET /openapi/v1/bots` (anonymous) | forwarded; **401 from the backend's `require_principal`** — the documented moved-refusal behavior |
| `GET /openapi/v1/bots?user_id=dev-alice` + `x-dev-user` | **200** with data envelope; backend logs `caller=user:dev-alice tenant=teamclaw` from the verified signed principal |
| `GET /openapi/v1/bcsfuse/workers` | **200**; forward log confirms the rewrite to `/v1/workers` |
| `GET /api/v1/collaboration/sessions/{id}/files` + `x-dev-user` | structured BCS business response (`session_not_found`, BCS request id) — gateway→BCS reached application logic |
| `GET /openapi/v1/collaboration/groups` + `x-dev-user` | 401 at the edge: that plane requires a registered `app` credential (`user: required, app: required`) — correct fail-closed |
| `GET /openapi/v1/chat/*` | forwarded to and answered by BAAS (probe path 404s; real chat routes hit the documented baas limitation) |

Environment notes from that run: `protobuf-compiler` had to be installed for the BCS build, and the committed `uv.lock` files pin a mirror unreachable from that container (`UV_DEFAULT_INDEX` override used; lockfiles restored, not committed). Also surfaced a pre-existing bug outside this PR's scope: `bcsfuse_setup()` in `scripts/modules/bcsfuse.sh` never calls `bcsfuse_load_env()`, so its env-file generation `cp`s to an empty path on a standalone `setup bcsfuse` — worked around manually; candidate for a separate fix.

Static and unit gates:

- `bash -n scripts/singlebox.sh scripts/modules/gateway.sh scripts/modules/backend.sh src/gateway/scripts/app.sh` — clean.
- Gateway unit suite from `src/gateway`: `uv run pytest tests/unit -q` — **622 passed** (includes 6 new `dev_header` strategy tests and 3 new bootstrap gating tests, one of which pins "declaring `dev_header` in config is refused at boot"; `tests/unit/conftest.py` now isolates `GATEWAY_AUTH_MOCK`).
- `ruff check` / `ruff format --check` on the touched Python files — clean.
- `basedpyright` on touched files: the only findings are `reportPrivateUsage` on `_strategies`/`_rules`, which pre-exist on the base branch in the same file (same idiom, also used by its log statements).
- Heavier module CI/coverage gates left to CI, matching #1135's lint-only pre-push.

## Compatibility and risk

- Additive for deployments: `application.yaml` placeholder values change from one non-working default (`*.sample.com`) to another that only works on a singlebox; deployed editions override the section either way.
- `backend.sh` singlebox launches now verify gateway principals with a public dev key — on a dev box this authenticates nothing an attacker couldn't already do with local access, and an explicit env value overrides it.
- The dev auth mock exists only in processes started with `GATEWAY_AUTH_MOCK=1`; it is absent from the DI container, and the shipped config cannot name it.
- Rollback: revert this branch's commits after the #1135 cherry-pick (which is independent).

### DIFF
--- a/scripts/modules/backend.sh
+++ b/scripts/modules/backend.sh
@@ -91,7 +91,12 @@ backend_start() {
             rm -f "${RUNTIME_DATA_DIR}/backend.db"
         fi
 
+        # Gateway principal verifier: deliberately armed for singlebox (see
+        # src/backend/docs/openapi-v1/README.md) with the same dev key the
+        # gateway signs with (gateway.sh) and BCS verifies with (bcs.sh), so
+        # locally forwarded /openapi/v1 requests verify instead of all-401.
         SERVER_ENV=dev DEPLOY_PROFILE=singlebox \
+            AGENTCLAW_SECRET_GATEWAY_PRINCIPAL_SIGNING_KEY_VALUE="${AGENTCLAW_SECRET_GATEWAY_PRINCIPAL_SIGNING_KEY_VALUE:-avernet-dev-signing-key-NOT-FOR-PROD}" \
             DATABASE_URL="sqlite:///${RUNTIME_DATA_DIR}/backend.db" \
             ENABLE_OSS_SYNC=false \
             CHAT_ENGINE="${CHAT_ENGINE}" \
--- a/scripts/modules/gateway.sh
+++ b/scripts/modules/gateway.sh
@@ -1,0 +1,163 @@
+#!/usr/bin/env bash
+# scripts/modules/gateway.sh — Gateway service module
+[[ -n "${_GATEWAY_SH_LOADED:-}" ]] && return 0
+_GATEWAY_SH_LOADED=1
+
+# Service-specific constants
+GATEWAY_LOG="${LOG_DIR}/gateway.log"
+GATEWAY_PID_FILE="${DEP_DIR}/gateway.pid"
+GATEWAY_APP_SCRIPT="${GATEWAY_DIR}/scripts/app.sh"
+GATEWAY_PORT="${GATEWAY_PORT:-8889}"
+
+
+gateway_setup() {
+    if ! check_directory_exists "${GATEWAY_DIR}" "gateway"; then
+        return 1
+    fi
+
+    check_uv_installed || { log_error "uv not found. Run: singlebox.sh install-tools"; return 1; }
+
+    cd "${GATEWAY_DIR}"
+    log_info "Syncing Python dependencies for Gateway..."
+    if ! uv sync --python ">=3.12,<3.13" --index-url "${PYPI_INDEX_URL}"; then
+        log_error "Failed to sync Python dependencies for Gateway"
+        return 1
+    fi
+    log_info "Gateway dependencies synced successfully"
+}
+
+
+gateway_start() {
+    mkdir -p "${LOG_DIR}"
+
+    if ! check_directory_exists "${GATEWAY_DIR}" "gateway"; then
+        return 1
+    fi
+    if [ ! -x "${GATEWAY_APP_SCRIPT}" ]; then
+        log_error "Gateway lifecycle script not executable: ${GATEWAY_APP_SCRIPT}"
+        return 1
+    fi
+
+    gateway_setup || return 1
+
+    stop_port_processes_if_owned "${GATEWAY_PORT}" "${GATEWAY_DIR}" "existing gateway"
+    stop_matching_processes_if_owned "gateway/community/main.py" "${GATEWAY_DIR}" "existing gateway process"
+    require_port_available_after_owned_stop "${GATEWAY_PORT}" "gateway" "set GATEWAY_PORT=<free-port> in .env.local or pass --gateway-port <free-port>" || return 1
+
+    log_info "Starting Gateway service on port ${GATEWAY_PORT}..."
+    log_info "Log: ${GATEWAY_LOG}"
+
+    (
+        cd "${GATEWAY_DIR}"
+        GATEWAY_PORT="${GATEWAY_PORT}" APP_PORT="${GATEWAY_PORT}" "${GATEWAY_APP_SCRIPT}" start --mode bare
+    ) >> "${GATEWAY_LOG}" 2>&1
+
+    local gateway_pid=""
+    if [ -f "${GATEWAY_DIR}/tmp/app.pid" ]; then
+        gateway_pid="$(cat "${GATEWAY_DIR}/tmp/app.pid" 2>/dev/null || true)"
+        if [ -n "${gateway_pid}" ]; then
+            echo "${gateway_pid}" > "${GATEWAY_PID_FILE}"
+        fi
+    fi
+
+    if gateway_ready; then
+        log_info "Gateway started successfully${gateway_pid:+ (PID: ${gateway_pid})}"
+        log_info "Gateway health: http://127.0.0.1:${GATEWAY_PORT}/health"
+        return 0
+    fi
+
+    log_error "Gateway failed to become ready; check ${GATEWAY_LOG} and ${GATEWAY_DIR}/tmp/app.log"
+    return 1
+}
+
+
+gateway_stop() {
+    log_info "Stopping Gateway..."
+
+    if [ -x "${GATEWAY_APP_SCRIPT}" ]; then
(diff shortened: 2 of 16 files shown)


totalfrank:
**Review round 1 — no automated review was able to run on this PR** (head `3a2af39`).

- The Codex reviewer (`chatgpt-codex-connector`) declined the ready-for-review trigger with a usage-limit notice (comment above), and this repository has no other automated review workflow or mention trigger to fall back to.
- No CI checks will appear here either: all four workflows (`unit-tests`, `e2e-tests`, `singlebox-coverage`, `architecture-checks`) filter `pull_request` to bases `dev` / `main` / `REL*`, and this PR targets `dev_refactory_collaboration`.

Validation that stands in for both is recorded in the PR description: the gateway unit suite (622 passed, including the new `dev_header` gating tests) and a full end-to-end singlebox bring-up on this branch with per-route forwarding proofs against backend, BCS, bcsfuse, and BAAS.

Review therefore rests with the requested human reviewers.

---

totalfrank (commented):
Reviewed commits `bf99cf4` (cherry-picked #1135), `2eb9882`, `3a2af39` (head) — the full diff against merge-base `8f86d41` on `dev_refactory_collaboration`.

No findings. Checked:

- **Correctness**: `application.yaml` upstream port values (8888/8890/21000/8765) verified against the actual singlebox defaults in `scripts/singlebox.sh`, `scripts/modules/backend.sh`, `scripts/modules/baas.sh`; `GATEWAY_PORT` precedence between `scripts/modules/gateway.sh`, `src/gateway/scripts/app.sh`, and `_config_loader.py` is consistent; `all.sh`'s `START_ORDER`/`SETUP_ORDER` untouched, so gateway stays opt-in as documented.
- **Contract breakage**: `DevHeaderUserStrategy` matches the `AuthStrategy` protocol (`name`, `principal_type`, `async build(creds) -> Principal | None`); `IdentityChain(principal_type, strategies)` construction matches its signature; `_append_dev_mock`'s `chains.get(PrincipalType.USER)` is always populated by `_strategy_chains`'s `setdefault` pass, so the `None`-guard path is unreachable but not wrong.
- **Security/hygiene**: the shared `avernet-dev-signing-key-NOT-FOR-PROD` default is not new — it's the same constant already exported in `scripts/modules/bcs.sh`; the dev auth mock is double-gated (env var to construct + env var to activate per request) and kept out of the production DI pool as described.
- **Python type contracts**: no new unjustified `T | None` widening.
- **Test coverage**: new `dev_header` strategy behavior, config-port override, and bootstrap gating (including "declaring `dev_header` in config is refused at boot") all have direct unit tests.
- Helper shell functions used in the new `scripts/modules/gateway.sh` (`stop_port_processes_if_owned`, `require_port_available_after_owned_stop`, `check_directory_exists`, etc.) all resolve to existing definitions in `scripts/utils.sh`.

One pre-existing idiom the diff reuses rather than introduces: `[ "$2" = -* ]` in the new `--gateway-port` arg parser is a literal-string comparison, not a glob match, so it never actually catches a following flag as a missing argument — but this is copied verbatim from the existing `--bcs-port`/`--frontend-port` handlers in the same file, so it's out of scope here.

---

totalfrank (commented):
Follow-up review round.

**Commits reviewed this round:** `f536f66` (new since the last review; docs-only, 7 lines, adds a TODO comment above `_append_dev_mock` in `src/gateway/src/gateway/community/bootstrap/_authn.py` recording the follow-up to move `dev_header` into an entry-point-discovered dev-only distribution). No logic change.

**Findings checklist:**
- Previous round (`3a2af39`) filed 0 blocking/non-blocking findings and 0 open threads — clean, with one out-of-scope observation about a pre-existing `[ "$2" = -* ]` idiom copied from existing arg parsers (not a new-code issue, not re-flagged).

**Open count: 0.**

Checked the new commit's TODO claim (entry-point-based dev-only distribution mirroring the `gateway.runner` sofa/bare selection) against `community/plugins/runner/bare/_plugin.py` and `plugin_accessor.py` — accurate. Nothing else to review; this commit touches no logic.

PR is clean as of this round (`f536f66`).

---

PR #1152, opened 2026-08-18
fix(bcs): support legacy group context updates

RaymondHX:
## Problem
The legacy `PATCH /groups/{id}` route rejects `{"context":"测试1"}` as an unknown field even though group context already exists in the domain and persistence model.

## Solution
Extend only the legacy update request and the internal group patch command with `context`, then propagate it through the application service to the existing parameterized persistence patch. Keep the V1 update DTO unchanged by mapping its internal context field to `None`. Add HTTP contract, application persistence, and BCS E2E assertions.

## Validation
- `cargo test -p bcs-http --test legacy_group_patch_contract` — 4 passed
- `cargo test -p bcs-app-group update_preserves_hidden_legacy_routing_fields` — 1 passed
- `cargo test -p bcs-api-http --test group_routes` — 15 passed
- `cargo test -p bcs-http --test groups_contract` — 35 passed
- `cargo check -p bcs` — passed
- `bash -n scripts/e2e-test/group.sh scripts/e2e-test/stories.sh` — passed
- Local E2E was not run as requested; GitHub CI will run the online E2E/coverage gate.

## Compatibility and risk
This is a backward-compatible expansion of the legacy route only. Unknown fields and explicit null values remain rejected. The V1 HTTP contract is unchanged. Rollback is a revert of this PR.

### DIFF
--- a/src/bcs/CLAUDE.md
+++ b/src/bcs/CLAUDE.md
@@ -387,7 +387,7 @@ export BOT_DATA_DIR=/path/to/bot/data
 | `/groups` | GET | List all groups |
 | `/groups/my` | GET | List formal groups for the authenticated human or bot |
 | `/groups/{id}` | GET | Get group details |
-| `/groups/{id}` | PATCH | Update mutable group properties (legacy compatibility route) |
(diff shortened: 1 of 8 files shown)


vzvince (approved):
LGTM

PR #1156, opened 2026-08-18
refactor(backend): resolve devices from BotFacts identity only

totalfrank:
## Problem

`EngineRuntimeRelay._resolve_device` and `_resolve_published_device` took `bot_id`/`owner_id` parameters beside `BotFacts`, which already carries the resolved pair — the values `resolve_bot` proved ownership against. The invariant that the parameters match the facts was enforced only by `call`'s docstring, and the published path already mixed the two sources within one method: `get_bot_pk` used `facts.bot_id`/`facts.owner_id` while `resolve_for_binding_invoke` used the `owner_id` parameter. A caller violating the documented invariant would have the publish-key lookup and the resolver addressing different owners.

## Solution

Narrow both methods to `(facts, stage)` and read `bot_id`/`owner_id` exclusively from `BotFacts`, so device resolution structurally cannot disagree with the identity ownership was proven against. `call` and `_resolve_bot_and_device` keep their `bot_id`/`owner_id` parameters — they are needed for the `facts=None` resolution path — and `_resolve_bot_and_device` hands only the resolved facts downstream. Docstrings updated to state that `facts` is the sole identity source.

No behavior change for compliant callers: `BotFacts.from_record` fills both fields from the owner-scoped `(bot_id, owner_id)` row read, falling back to the requested values, so the facts' identity always equals the former parameters on every existing path.

## Validation

- `pytest tests/community/core/engine_runtime/` — 169 passed.
- `ruff check` on the changed file — clean. (`ruff format --check` already fails on this file at the merge base, so formatting was left untouched.)
- The cross-surface guard `test_the_socket_and_the_relay_address_the_same_binding` (source-inspects `_resolve_published_device`) still passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RciQ8AF85BosNND3Zv6wev)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/engine_runtime/relay.py
+++ b/src/backend/src/agentclaw/community/core/engine_runtime/relay.py
@@ -164,10 +164,13 @@ class EngineRuntimeRelay:
         """
         return await asyncio.to_thread(self.resolve_bot, bot_id, owner_id, caller_id)
 
-    def _resolve_device(
-        self, bot_id: str, owner_id: str, facts: BotFacts, stage: str
-    ) -> DeviceContext:
+    def _resolve_device(self, facts: BotFacts, stage: str) -> DeviceContext:
         """Resolve the bot's device, translating "not reachable" to one error.
+
+        ``facts`` is the sole source of the bot's identity. It carries the
+        *resolved* ``bot_id``/``owner_id`` — the values :meth:`resolve_bot`
+        proved ownership against — so there is no second pair of identity
+        parameters here that could disagree with it.
 
         ``stage`` picks which of a ``service`` bot's runtimes this call
         addresses. :data:`~agentclaw.community.core.engine_runtime.stage.\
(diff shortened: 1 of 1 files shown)


totalfrank (commented):
Diff is clean — no findings.

Reviewed commit `e897286` (the only commit on this PR), file `src/backend/src/agentclaw/community/core/engine_runtime/relay.py`.

Checked:
- **Correctness**: `_resolve_device`/`_resolve_published_device` now read `bot_id`/`owner_id` exclusively from `facts`. Verified `BotFacts.from_record` always sets those fields from the owner-scoped row read (falling back to the requested values only when the row omits them), and that every caller — `gating.py`'s `resolve_operable_bot` and the `facts=None` fallback in `_resolve_bot_and_device` — resolves `facts` against the same `bot_id`/`owner_id` later passed through, so the narrowing is behavior-preserving on every live path. The previously-mixed identity source in the published-device path (`get_bot_pk` used `facts.*`, `resolve_for_binding_invoke` used the `owner_id` param) is now structurally unified, matching the PR's stated intent.
- **Contract breakage**: none — both changed methods are private (`_`-prefixed), no public/Service/Plugin API surface touched.
- **Callers left unupdated**: none found. `test_stage.py`'s `test_the_socket_and_the_relay_address_the_same_binding` only source-inspects `_resolve_published_device` (`inspect.getsource`), unaffected by the signature change. No test or production caller invokes either private method directly with the old positional shape.
- **Error/failure handling, concurrency, security, architecture boundaries, Python type contracts**: not implicated by this diff.
- **Test coverage**: adequate — this is a pure internal refactor with no observable behavior change for any existing caller, and the cross-surface guard test cited in the PR description still holds.

---

FreddieSun (approved):
LGTM

PR #1162, opened 2026-08-18
fix(skill-center): split data and runtime engine resolution

panqiwp:
## Summary
- restore resolve_engine_for_bot as data/persisted engine resolver and add explicit runtime resolver usage for runtime path/build/provider flows
- keep SkillSet/MCP/AgentPass data scoped by bot.active_engine while routing claude_code non-normalCC runtime paths through aicoding
- update Skill Center path/symlink factories, pool layout handling, local lifecycle sync, docs, and targeted tests

## Validation
- python3 -m py_compile on touched backend files
- uv run ruff check on touched backend/test files
- uv run pytest -q tests/community/core/bot_management/services/test_engine_resolver.py tests/community/services/test_skill_set_service_engine_type.py tests/community/services/test_skill_set_service_symlink_mappings.py


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

Architecture gate only. I am not reviewing the product decision that `claude_code` + non-`normalCC` bots should run on the `aicoding` runtime layout — I assume that intent is correct. What I am gating on is *how* that split is wired.

---

## Gate 0 — Target branch

- **Target:** `dev` (integration branch).
- **Classification:** behavior change to existing pipelines (skill-center path/symlink/pool resolution). This is the correct route for that kind of change.
- **Result: PASS.** No release-branch exception path is in play, no Isolation Audit required.

Gate 0 passing does not carry the findings below.

---

## Findings

### 1. `skill_center_module.py:571` — composition root now selects filesystem layout from an unvalidated client-supplied engine string, and the only fail-safe was deleted

```python
# removed:
-            engine = bot.get("active_engine")
-            if not isinstance(engine, str):
-                return None
...
             paths = pool_paths_for_engine(_requested_engine)
```

**Violated:** Rule 14 (Configuration Drives All Wiring — Invariant), Rule 7 (Core Is Transport-Agnostic — Invariant).

Before this change the pool layout was selected from the **persisted** `bot["active_engine"]`, guarded by an `isinstance` check that returned `None` (graceful non-pool fallback) for anything unusable. After it, the selector is `_requested_engine`, which traces back to the HTTP `?engine_type=` query parameter: `skills.py:305`/`skillsets.py:105` → `resolve_runtime_engine_for_bot(override=engine_type)` → `factories.py:279` → `resolve_pool_paths(...)`.

`resolve_runtime_engine_for_bot` does not constrain the override. With an unknown engine name, `registry.resolve()` falls through to `DefaultProvisioningStrategy`, whose `resolve_bot_engine` returns `active_engine` — here the injected override — so the override reaches the closure verbatim. And `core/workspace/skill_layout.py:58` `pool_paths_for_engine` is documented as *"never fall back"*: it raises `ValueError(f"engine Pool layout not implemented: {engine}")` for anything outside `FILESYSTEM_POOL_ENGINES`.

Two concrete failures this creates for any bot on the pool layout:

- `?engine_type=teclaw` (or `moltis`, or a typo) → uncaught `ValueError` raised out of a DI-provided closure into the request path. Previously the same request returned `None` and took the legacy path.
- `?engine_type=aicoding` against a bot whose persisted `active_engine` is `openclaw` → the caller now gets **aicoding pool paths for an openclaw bot's data**. The persisted engine no longer constrains physical layout at all.

Also note the parameter is still named `_requested_engine`. The leading underscore is the repo's "deliberately unused" convention, and it was accurate before this PR. Turning a deliberately-ignored argument into the authoritative input while leaving the name unchanged is exactly the kind of silent authority transfer Rule 14 exists to prevent.

**Required fix:** the composition root must validate the requested engine against a declared set before it can select a layout, and reject with a domain error (or fall back to the persisted `active_engine`) rather than letting a layout helper raise. If an operator override is genuinely meant to reach layout selection, that has to be a stated, validated part of the contract — not a query string flowing into a `raise`. Rename the parameter once it is load-bearing.

---

### 2. The adapter resolves `runtime_engine` at 16 call sites and passes it to the service factory at zero — the factory then silently re-resolves it *without* the override

`skillsets.py` unpacks `runtime_engine` from `_get_path_params` at lines 146, 267, 320, 382, 468, 501, 595, 624, 655, 695, 830, 1054, 1320, 1361, 1415, 1453, 1524. It is actually *used* at only three of them (698–701, 833–836, 1057–1059). Every `skill_set_service_factory.create(...)` call in the file omits `runtime_engine_type`, so `factories.py:512`–`519` takes over:

```python
        effective_runtime_engine = runtime_engine_type
        if effective_runtime_engine is None:
            owner_id = user_id or entity_id
            effective_runtime_engine = resolve_runtime_engine_for_bot(
                bot_id or "default",
                str(owner_id) if owner_id is not None else None,
                bot_repo=self._bot_repo,
            )
```

Note the missing `override=`. **Violated:** Rule 1 (The API Specification Is the Single Authority — Invariant), Rule 4 (Interfaces Mean What Their Specs Define — Invariant).

The maintenance failure is concrete: one value now has two authorities that disagree. `?engine_type=` is honored for the data engine (`effective_engine`, passed through) and silently dropped for the runtime engine (re-derived from the bot record). And `SkillSetService` gets its `runtime_engine_type` from a *different source depending on who called it* — `local_device_lifecycle.py:150` passes it explicitly, every skillsets.py endpoint does not. Nothing in the contract says which wins or why. That is precisely the "undocumented semantics that other components depend on" Rule 1 forbids, and it is the same defect class this PR set out to fix.

It also means each of those requests now performs three separate bot lookups — `resolve_engine_for_bot`, `resolve_runtime_engine_for_bot`, and the `is_desktop` fetch at `skillsets.py:112` — plus a fourth inside the factory. That is a symptom, not the finding: resolution has no single home.

**Required fix:** resolve once, at one layer, and thread the result. Either the adapter always passes both values into `create()` and the factory never re-derives, or the adapter passes neither and the factory owns resolution entirely with the override forwarded. Do not keep both paths.

---

### 3. `resolve_engine_for_bot` semantics changed in place; propagation is incomplete

The function kept its name and signature but lost the routing policy — `resolve_bot_engine(bot)` is gone from its body. Every unmigrated caller silently changed behavior. **Violated:** Rule 16 (Changes Propagate — Invariant), Rule 4.

The PR migrated the adapters it touched, but `core/resources/dependencies/resource.py:30` still carries the now-wrong instruction:

```python
    Callers should resolve engine_type via resolve_engine_for_bot() so the
    bot's active_engine is respected; when omitted here, falls back to
    DEFAULT_ENGINE_TYPE.
```

That docstring sits on `get_bot_workspace_dir` — a **workspace path** helper, i.e. exactly the runtime-layout category the new docstring says must use `resolve_runtime_engine_for_bot`. The contract's own documentation now points the next author at the wrong resolver for the wrong reason.

More broadly, the PR body has no propagation section. Rule 16 requires a structural change to enumerate affected consumers, affected implementations, and compatibility status. A rename-in-place of a shared resolver's semantics is the textbook case for that. Please enumerate every remaining `resolve_engine_for_bot` call site with a one-line justification for why data semantics are correct there.

**Required fix:** update `resource.py:30`, and add the propagation analysis. Consider renaming the data resolver (e.g. `resolve_data_engine_for_bot`) so the compiler-visible name carries the distinction and no call site can silently inherit the old meaning.

---

### 4. `skill_set_service.py:1319`–`1337` — engine physical layout still hardcoded in Backend core, and this PR re-points which bots land where

```python
        ENGINE_SKILLS_DIR_MAP = {
            "claude_code": "/home/admin/.claude/skills",
            "aicoding": "/home/admin/.claude/skills",
            ...
        }
        ENGINE_SKILLS_REPO_DIR_MAP = {
            "aicoding": "/home/admin/.aicoding/skills-repo",
            ...
        }
        base_skills_dir = Path(
            ENGINE_SKILLS_DIR_MAP.get(self.runtime_engine_type, "/home/admin/.openclaw/workspace/skills")
        )
```

**Violated:** `AGENTS.md` § Skills Architecture — *"Do not add new engine-specific filesystem paths to Backend code. Physical layout ownership belongs to Engine Runtime and its versioned layout contract"* — and the module's own `adapters/http/skill_center/CLAUDE.md`: *"不得继续在 Backend 新增引擎目录硬编码"* and *"未知引擎或未知布局契约必须失败关闭，不能回退到某个默认引擎路径"*.

The maps predate this PR, and I am not asking you to delete them here. What I am gating on is that switching the key from `self.engine_type` to `self.runtime_engine_type` is not a neutral edit — it **moves bots between physical directories**. A `claude_code` + non-`normalCC` bot previously missed `ENGINE_SKILLS_REPO_DIR_MAP` and fell through to `base_skills_dir / "skills-repo"` = `/home/admin/.claude/skills/skills-repo`. It now resolves to `aicoding` and lands on `/home/admin/.aicoding/skills-repo`. That is a real relocation of where an engine finds its skills-repo, decided by a hardcoded dict in Backend core, with no corresponding Engine Runtime layout contract update and no migration note anywhere in the PR.

The default-arm fallback in the same expression compounds it: the module contract requires unknown engines to **fail closed**, and this silently routes them to the openclaw path — now reachable by more inputs than before, since `runtime_engine_type` can carry an unvalidated override (finding 1).

**Required fix:** state the relocation explicitly in the PR with its migration/rollback path, and route the lookup through the Engine Runtime layout resolver rather than re-keying the core-side dict. At minimum, the unknown-engine arm must raise rather than default.

---

### 5. `local_device_lifecycle.py:114`–`117` — plugin implementation reaches further into core domain policy

```python
        from agentclaw.community.core.bot_management.services.engine_resolver import (
            resolve_engine_for_bot,
            resolve_runtime_engine_for_bot,
        )
```

**Violated:** Rule 3 (Distinguish Service APIs from Plugin APIs — Invariant), Rule 5 (contracts, not implementations), Structural Review Red Flags.

Direction of control is inverted: `plugins/local/` is a Plugin API *implementation*, which core calls — it should not be calling back into core domain services. Credit where due — this PR **removes** the import of core's private `_get_bot_paths`, which is a genuine improvement. But it replaces one core reach-in with another and adds engine-selection policy on top at line 145:

```python
                engine_type = skill_set.get("engine_type") or resolve_engine_for_bot(...)
```

That `or` is a resolution precedence rule (persisted skillset engine beats bot lookup) that exists nowhere else and is documented nowhere. A plugin is the wrong owner for it.

**Required fix:** have the caller (or a core service the plugin is handed) resolve both engines and pass them in, so the plugin receives values instead of deriving them. The `skill_set.get("engine_type") or ...` precedence belongs next to the resolvers, under one contract.

---

### 6. No conformance test and no architecture guard for the new distinction

**Violated:** Rule 25 (Protocols Have Self-Validating Contracts — Invariant).

The test delta is 30 lines across two files, all unit-level assertions on the resolver pair itself. Nothing tests the thing the PR is actually about: that runtime/path/pool/symlink call sites use the runtime resolver and data/scope call sites use the data resolver. Today that invariant lives entirely in a docstring, and finding 2 shows it is already violated in the same diff — 16 sites compute the runtime engine and drop it.

`src/backend/tests/community/architecture/` already hosts this exact genre of guard (`test_expert_chat_engine_branching.py`, `test_default_capabilities_engine_boundaries.py`, `test_no_database_mode_branches_in_factories.py`). A new boundary that the existing suite cannot catch requires a guard test or an explicit waiver per the Governance Addendum.

**Required fix:** add an architecture test asserting that callers feeding `path_factory.get_bot_*`, `pool_paths_for_engine`, and `get_symlink_mappings` derive their engine from `resolve_runtime_engine_for_bot`. Add a conformance test covering the default/fallback arm — including the unknown-engine case, whose semantics finding 1 and finding 4 currently leave undefined.

---

### 7. Contract change ships without a spec update; three overlapping plan documents are not a spec

**Violated:** Rule 1, Rule 16.

The PR adds 3,127 lines across `docs/skill-center/engine-resolver-runtime-split-final-plan.md`, `engine-resolver-split-plan.md`, and `skillset-engine-context-unified-plan.md`. Planning documents are not the contract. Meanwhile `adapters/http/skill_center/CLAUDE.md:294` documents the live cross-component contract — *"发送 engine_type、layout_state、layout_contract_version"* — and this PR changes which engine value fills that `engine_type` slot for a whole class of bots (finding 4). That file is untouched.

`AGENTS.md` also requires the PR description to use **Problem / Solution / Validation**; this one has Summary / Validation, and the Validation section lists `py_compile`, `ruff`, and three unit test files — no module gate, no acceptance run, for a change that relocates physical skill directories.

**Required fix:** update `adapters/http/skill_center/CLAUDE.md` to state the data-engine/runtime-engine split and which one crosses to Engine Runtime. Restructure the description per the template. Collapse the three plan docs into the one that reflects what shipped.

---

## Out-of-scope notes

- `factories.py:514` re-derives `owner_id = user_id or entity_id`, already computed identically at line 501. Harmless duplication, not gated.
- `factories.py:299` adds a stray blank line. Not gated.
- I did not review the correctness of the `claude_code` + non-`normalCC` → `aicoding` routing rule itself, nor which specific endpoints should use data vs. runtime semantics. Both are the feature reviewer's call.

---

totalfrank (approved):
## Architecture-Gate Review

**Verdict: APPROVE**

### Gate 0 — Target branch
Target is `dev` (the integration branch), not a `REL*` release branch. The REL-branch classification (existing-behavior fix / isolated-feature exception / feature-wrong-target) and the Isolation Audit do not apply. Gate 0 passes trivially.

### Scope of the change
This PR splits `resolve_engine_for_bot` (core/bot_management/services/engine_resolver.py) into two resolvers:
- `resolve_engine_for_bot` — returns `bot.active_engine` verbatim, for data-ownership call sites (SkillSet.engine_type, MCP/AgentPass scope, NAS ownership).
- `resolve_runtime_engine_for_bot` — additionally applies registered runtime routing policy (e.g. `claude_code` + non-`normalCC` → `aicoding`), for provider/build/workspace-path/symlink/CLI call sites.

All other files (HTTP routers, `baas_service.py`, `bot_build_service.py`, `identity.py`, `skill_center/factories.py`, `skill_set_service.py`, `local_device_lifecycle.py`, DI wiring) are mechanical call-site updates: routing-sensitive callers switch to the new `resolve_runtime_engine_for_bot`, data-ownership callers keep `resolve_engine_for_bot`.

### Engine-variability pattern check
Verified `resolve_runtime_engine_for_bot` contains **no** engine-name string literals or branching itself — it delegates to `resolve_bot_engine()` in `core/bot_management/engines/registry.py`, which is the pre-existing composition-root registry (`EngineProvisioningRegistry`): one contract (`EngineProvisioningStrategy`), per-engine strategy implementations registered in `_build_default_registry()`, and a mandatory `DefaultProvisioningStrategy` fallback for unregistered engines. The `claude_code`-non-`normalCC`→`aicoding` policy itself lives in `AicodingProvisioningStrategy.resolve_bot_engine` under `core/bot_management/engines/aicoding/strategy.py`, the engine's own strategy module — unchanged by this PR. Applying the "add engine `foo`" test: `foo` would need its own strategy class plus one `registry.register(...)` line; no engine-agnostic file touched by this PR would require edits. No violations of catalog items 1–3.

Grepped the full diff for new `engine_type ==`/`active_engine ==`/engine-string-set patterns — none found outside test assertions and doc prose.

### Other checks
- `core/skill_center/factories.py`, `skill_set_service.py`: pre-existing `ENGINE_SKILLS_DIR_MAP`/`ENGINE_SKILLS_REPO_DIR_MAP` (hardcoded `/home/admin/.claude/...` paths — catalog item 4) are untouched; the diff only changes which resolved engine variable (`runtime_engine_type` instead of `engine_type`) feeds the existing lookup. Not added/extended/entrenched by this change, so out of gate scope, but flagging for awareness since it's a known pre-existing violation in a file this PR is actively editing.
- `di/modules/skill_center_module.py`: composition-root wiring only (adds `bot_repo` to `SkillServiceFactory`, changes the pool-paths closure's engine input) — no policy encoded, no new imports across layers.
- `plugins/local/local_device_lifecycle.py`: this change actually **removes** a prior violation — it previously imported the private `_get_bot_paths` helper directly from `skill_set_service.py`; now it goes through the public `SkillServiceFactory.create()` / `resolve_runtime_engine_for_bot()` surface.
- No transport/framework imports, no new `os.environ`/env-detection in `core/`, no new hardcoded operational policy literals.
- Propagation (Rule 16): the three added planning docs (`docs/skill-center/engine-resolver-split-plan.md`, `engine-resolver-runtime-split-final-plan.md`, `skillset-engine-context-unified-plan.md`) document the root cause and enumerate affected consumers (SkillSet/MCP/AgentPass data path vs. runtime/build/symlink path); call-site diff matches the documented plan.
- Tests: `test_engine_resolver.py` and `test_skill_set_service_engine_type.py` updated to cover both resolvers and the `runtime_engine_type` field split — consistent with existing test strategy for this contract (no new Service/Plugin API introduced, so no new conformance-test suite is required under Rule 25).

### Out-of-scope notes
- `skillsets.py`: several call sites now compute `runtime_engine` in `_get_path_params()` but don't pass it into `skill_set_service_factory.create()` (the factory resolves it again internally when omitted) — a redundant lookup, not an architecture issue. Business-logic/efficiency, left to the feature reviewer.
- Did not evaluate correctness of the `claude_code`/`normalCC` routing policy itself (business logic), per review scope.

---

totalfrank (dismissed):
## Architecture-Gate Review

**Verdict**: APPROVE

**Gate 0**: Target branch is `dev` (the integration branch), not a `REL*` release branch. The release-branch isolation/backport rules do not apply here — pass by default routing.

**Findings**: none.

This PR splits `resolve_engine_for_bot` into two explicit contracts:
- `resolve_engine_for_bot` — data/persisted engine (`bot.active_engine`), used for SkillSet/MCP/AgentPass data scoping.
- `resolve_runtime_engine_for_bot` — runtime/layout engine, applies the existing `claude_code` + non-`normalCC` → `aicoding` routing policy via the pre-existing `EngineProvisioningRegistry` / per-engine `EngineProvisioningStrategy` composition root (`core/bot_management/engines/registry.py`), which already ships a `DefaultProvisioningStrategy` fallback. This PR does not touch that registry — it only adds a second, clearly-named consumer-facing entry point and threads it through ~20 call sites (HTTP routers, `SkillServiceFactory`/`SkillSetServiceFactory`, `IdentityService`, `BaasService`, `BotBuildService`, the local device-lifecycle plugin, and the `skill_center_module.py` composition root) so runtime/path/provider resolution and data-ownership resolution no longer collapse into one ambiguous name.

Checked specifically for the catalog violations:
- No new/extended engine-string branching was added to `core/` — the `claude_code`→`aicoding` policy remains encapsulated in the existing registry/strategy pattern, untouched by this diff.
- No new hardcoded engine capability sets, engine-named modules, or engine-specific filesystem literals were introduced by this diff (see out-of-scope note below on a pre-existing map).
- `core/skill_center/factories.py` and `core/service_bot/services/*.py` depend on `BotRepository` (a Protocol import from `core/repository/protocols/bot.py`) and on the engine-agnostic resolver functions in `core/bot_management/services/engine_resolver.py` — no concrete-implementation or plugin-internals imports crossed a layer boundary.
- Composition-root change in `di/modules/skill_center_module.py` (`pool_paths_for_engine(_requested_engine)` now takes the caller-resolved runtime engine instead of re-deriving `active_engine` locally) is wiring, appropriately located in the composition root.
- Propagation (Rule 16): the PR adds three docs under `docs/skill-center/` explaining the semantic split, the rationale for not exposing a third `resolve_persisted_engine_for_bot` name, and the call-site audit; `test_engine_resolver.py` covers both resolvers' behavior for the override/no-override and normalCC/non-normalCC cases.

**Out-of-scope notes**:
- `core/skill_center/services/skill_set_service.py` still hardcodes engine→filesystem-path maps (`ENGINE_SKILLS_DIR_MAP` / `ENGINE_SKILLS_REPO_DIR_MAP`, e.g. `/home/admin/.claude/skills`) directly in `core/` (catalog item 4). This diff only renames the lookup key (`engine_type` → `runtime_engine_type`); it does not add new hardcoded paths, so it isn't gated here, but it's a pre-existing Invariant-adjacent violation worth a follow-up ticket.
- Business-logic correctness of the `claude_code` + non-`normalCC` → `aicoding` routing semantics themselves (as opposed to which layer owns them) is out of scope for this review.

---

totalfrank (dismissed):
## Architecture-Gate Review

**Verdict: APPROVE**

### Gate 0 — Target branch
- Target: `dev` (integration branch). Not a `REL*` release branch, so no Isolation Audit applies. Gate 0 passes trivially.

### Review procedure
Read `docs/arch/arch.rules.md` (v3.1) and `AGENTS.md`, then the full diff across both commits (3 new design docs + 20 changed code/test files spanning `adapters/http/...`, `core/bot_management/services/engine_resolver.py`, `core/service_bot/services/*`, `core/services/identity.py`, `core/skill_center/*`, `di/modules/skill_center_module.py`, `plugins/local/local_device_lifecycle.py`, 2 unit test files, plus a small follow-up commit touching a docstring and an architecture guard-test allowlist).

**Crux mechanism.** The PR splits a single ambiguous resolver into two engine-agnostic `core/` functions in `core/bot_management/services/engine_resolver.py`:
- `resolve_engine_for_bot` — pure `bot.active_engine` passthrough (data/persisted engine: SkillSet.engine_type, MCP/AgentPass/authorization queries).
- `resolve_runtime_engine_for_bot` — delegates to the pre-existing `EngineProvisioningRegistry.resolve_bot_engine()` (registry + per-engine `EngineProvisioningStrategy`, e.g. `AicodingProvisioningStrategy.should_use_aicoding_runtime_engine()`), which is where the actual `claude_code + non-normalCC → aicoding` string comparison lives — inside the `aicoding` engine's own module, keyed by engine type in a registry, with a mandatory `DefaultProvisioningStrategy` fallback. **This registry/strategy module is untouched by the diff** — the PR only re-wires which of the two new functions calls into it.

Applying the "add engine `foo`" test: neither new function contains inline `== "claude_code"` / `== "aicoding"` branching, and adding a new engine still requires only its own strategy module + one registry entry — unchanged by this PR.

The follow-up commit adds `resolve_runtime_engine_for_bot` to `test_http_adapter_layer_is_http_only.py`'s allowlist of pure-function helpers permitted in HTTP adapters, mirroring the existing `resolve_engine_for_bot` entry — the architecture guard test was updated in the same PR rather than left to silently pass/fail.

### Findings
None. Checked against all 11 catalog categories (engine-string branching in core, hardcoded capability sets, engine-named core modules, engine filesystem layout in core, hidden wire protocols, env/mode detection outside composition roots, hardcoded operational policy, transport leakage, cross-module copy-paste, concrete-implementation imports outside composition roots, misplaced Plugin-API abstraction) — zero new/extended violations. Call-site migration (routers, `baas_service.py`, `bot_build_service.py`, `identity.py`, `skill_set_service.py`, `di/modules/skill_center_module.py`) consistently routes path/provider/sandbox/symlink consumers to the runtime resolver and DB/data consumers to the data resolver, matching the accompanying design doc. As a side effect, `plugins/local/local_device_lifecycle.py` also drops a prior layering violation (a plugin importing a private `_get_bot_paths()` helper out of a core service module) in favor of the public factory API.

### Out-of-scope notes (not gated on)
1. `core/skill_center/services/skill_set_service.py` still contains a **pre-existing** hardcoded `ENGINE_SKILLS_DIR_MAP` / `ENGINE_SKILLS_REPO_DIR_MAP` (engine name → absolute container path, e.g. `/home/admin/.aicoding/skills-repo`) directly in `core/`, which conflicts with AGENTS.md's "do not add new engine-specific filesystem paths to Backend code." This PR only changes the lookup key (`engine_type` → `runtime_engine_type`) and doesn't add rows or extend the table, so it isn't a new violation from this diff — but it's directly on-topic for this PR's problem domain and worth a follow-up.
2. Test coverage: `test_engine_resolver.py` gets solid targeted coverage for both resolvers, but no test constructs a single `claude_code + non-normalCC` `SkillSetService` and asserts both halves of the split simultaneously (DB query still keyed by `claude_code` while symlink paths build under `aicoding`) — the exact production scenario the design docs describe. This is a test-coverage/business-logic observation, not an architecture gate item.

---

panqiwp:
Closing this PR in favor of a fresh PR from the updated branch.

PR #1203, opened 2026-08-19
fix(bot-service): catch CancelledError in aiohttp request retry loop

xxxxpenny:
fix(bot-service): catch CancelledError in aiohttp request retry loop
    
    When an asyncio task is cancelled mid-request, `CancelledError` propagates
    out of `aiohttp` and is not caught by the existing `except` clause (which
    only handles `TimeoutError` and `aiohttp.ClientError`). The uncaught
    `CancelledError` escapes the retry loop, surfacing as an unexpected
    error instead of a clean cancellation.
    
    Add `asyncio.CancelledError` to the `except` tuple so the error path
    handles cancellation gracefully alongside timeout and client errors.

### DIFF
--- a/src/baas/src/secbaas/community/plugins/bot_service/real/_plugin.py
+++ b/src/baas/src/secbaas/community/plugins/bot_service/real/_plugin.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import asyncio
 import time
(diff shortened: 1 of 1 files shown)


pfmiles (approved):
ok

PR #1212, opened 2026-08-19
feat(bcs-app-session): emit participant-mode-changed system message on openapi path

zh-bupt:
## Problem

BCS exposes two HTTP entry points for updating a session participant's mode:
- legacy `bcs_http::routes::sessions::update_session_participant_mode`
- OpenAPI v1 `bcs_api_http::v1::openapi::routes::session::update_session_participant`

Only the legacy path emits the `ParticipantModeChanged` system message, so participants are not notified when a mode change is performed through the OpenAPI surface.

## Solution

Emit `SystemMessageEvent::ParticipantModeChanged` from the OpenAPI path, placing it in the application facade (`SessionServiceImpl::update_participant`) rather than the thin delivery adapter — system-message dispatch is use-case orchestration, not protocol translation.

- `crates/application/v1/bcs-app-session/src/lib.rs`: add a `system_message: Arc<dyn SystemMessageService>` dependency to `SessionServiceImpl`. In `update_participant`, capture the participant's pre-update identity (old mode, actor kind, actor name resolved via the registry) from the already-loaded session, then after a successful `update_participant_mode` emit `ParticipantModeChanged { group_id, actor_id, actor_name, actor_kind, from, to }` only when `old_mode != Some(new_mode)` — the same gating as the legacy endpoint. Dispatch failures are best-effort (`tracing::warn!`), never surfaced to the caller.
- `crates/bootstrap/bcs/src/server.rs`: pass `system_message.clone()` into `SessionServiceImpl::new`.

The route handler and the OpenAPI response contract are unchanged — this is a side effect, not a response-shape change.

## Validation

- `cargo build -p bcs` ✅ (composition root compiles)
- `cargo test -p bcs-app-session --test v1_session_service` ✅ (63 passed)
- `cargo test -p bcs-api-http --test session_routes` ✅ (34 passed)
- New test `update_participant_emits_mode_changed_system_message_only_on_change`: changes `expert` Auto → Muted and asserts exactly one `ParticipantModeChanged` event (`from=Some(Auto), to=Muted`); re-applying Muted asserts no new event (no-op gate). The test fixture now uses a `RecordingSystemMessageService` (behaviorally a no-op otherwise, returning `Ok(0)`), so all pre-existing tests still pass.

### DIFF
--- a/src/bcs/crates/application/v1/bcs-app-session/src/lib.rs
+++ b/src/bcs/crates/application/v1/bcs-app-session/src/lib.rs
@@ -20,6 +20,7 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use bcs_service_api::application::session::{SessionManagementService, SessionUseCaseError};
+use bcs_service_api::application::system_message::SystemMessageService;
 use bcs_service_api::application::v1::{
     ApplicationError, AuthenticatedCaller, DeleteResult, HumanPrincipal, Page, Principal,
     message::{ListSessionMessages, SessionMessageService},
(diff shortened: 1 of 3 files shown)


vzvince (approved):
LGTM

Issue #1207, opened 2026-08-19
Gateway: Redis cache plugin extension


cassiuscai:
## Problem
The gateway currently ships only an ephemeral in-memory `CachePlugin` (`src/gateway/src/gateway/community/plugins/cache/in_memory/`). In a horizontally-scaled Alibaba Cloud deployment this is not acceptable — cache entries are not shared across replicas.

## Scope
Add a Redis-backed implementation of the gateway `CachePlugin` SPI (`get` / `set` / `close`) with server-side TTL expiry, selectable via plugin configuration, while keeping `in_memory` the default.

- New plugin sub-package `plugins/cache/redis/`
- Pinned `redis` dependency in `src/gateway/pyproject.toml`
- Credentials as secret references (`@name`), resolved by the composition root
- Wire `redis` into the `cache_plugin` `providers.Selector` (additive; default unchanged)
- Unit + contract/conformance tests

## Validation
Run gateway unit, contract, and architecture suites; ensure the `in_memory` default is unchanged and the Redis selection path is covered.

Issue #1209, opened 2026-08-19
BaaS: Redis cache plugin extension


cassiuscai:
## Problem
The BaaS service currently ships only a stub `CachePlugin` (`src/baas/src/secbaas/community/plugins/cache/stub/`). In a horizontally-scaled Alibaba Cloud deployment this is not acceptable — cache entries are not shared across replicas.

## Scope
Add a Redis-backed implementation of the BaaS `CachePlugin` SPI (`get` / `set` / `close`) with server-side TTL expiry, selectable via plugin configuration, while keeping `stub` the default.

- New plugin sub-package `plugins/cache/redis/`
- Pinned `redis` dependency in `src/baas/pyproject.toml`
- Credentials as secret references (`@name`), resolved by the composition root
- Wire `redis` into the BaaS cache `providers.Selector` (additive; default unchanged)
- Unit + contract/conformance tests

## Validation
Run BaaS unit, contract, and architecture suites; ensure the `stub` default is unchanged and the Redis selection path is covered.

Issue #1208, opened 2026-08-19
Gateway: Aliyun KMS secret resolver extension


cassiuscai:
## Problem
The gateway secret resolver SPI (`src/gateway/src/gateway/community/spi/secret_resolver/`) currently has only a single environment-variable-based implementation (`CommunitySecretResolver`). In production, signing keys and credentials must come from a managed key store rather than the process environment.

## Scope
Add an Aliyun KMS-backed implementation of the gateway `SecretResolver` SPI (`get_secret`) that returns an object exposing `.secret_user` / `.secret_value` or `None`, selectable via plugin configuration, while keeping the `community` environment resolver the default.

- New plugin sub-package `plugins/secret_resolver/kms/`
- Dependency-free HTTP Aliyun KMS client (no `alibabacloud-tea-openapi` / `aiofiles` conflict)
- Resolver credentials as secret references (`@name`), resolved by the composition root before building the KMS client
- Wire `kms` into the `secret_resolver` `providers.Selector` (additive; default unchanged)
- Unit + contract/conformance tests, including transport-error propagation and principal-signer path verification

## Validation
Run gateway unit, contract, architecture suites, and the principal-signer path; ensure the `community` default is unchanged and the KMS selection path is covered.

PR #1215, opened 2026-08-19
feat(redis-kms): add redis cache plugins (gateway + baas) and aliyun kms secret resolver (gateway)

cassiuscai:
## Problem

Both the gateway and baas services only ship ephemeral in-memory (or stub) cache plugins, and the gateway's `secret_resolver` SPI has a single environment-variable-based implementation. In a multi-instance deployment these are not acceptable: caches must be shared and persistent across replicas, and credentials must come from a managed key store rather than process environment.

## Solution

Add production-ready **Redis cache** plugins to the gateway and baas, and an **Aliyun KMS** secret resolver to the gateway, so both services can run as horizontally-scaled, shared-state services on Alibaba Cloud.

- **Gateway redis cache**: `RedisCachePlugin` + `RedisCacheConfig`, wired via `plugins.cache = redis` with secret-reference resolution and `in_memory` default.
- **Gateway Aliyun KMS secret resolver**: `AliyunKmsSecretResolver` + a dependency-free HTTP KMS client (no `alibabacloud-tea-openapi` / `aiofiles` stack), wired via `plugins.secret = aliyun_kms` with `community` default.
- **Baas redis cache**: `RedisCachePlugin` + config, wired via `plugins.cache = redis` with `stub` default.

## Validation

- Gateway: unit + contract test suites (cache, secret_resolver, principal-signer) — 817 passed.
- Baas: unit + contract + architecture suites — green.
- `ruff check` and `ruff format` clean on all changed files.

## Compatibility and risk

Additive only — existing `in_memory` / `stub` / `community` options remain the defaults; new selection keys take effect only when a deploy overlay opts in. No public SPI contracts changed, no data migration.

## Related issues

Implements OpenSpec change `redis-cache-kms-extensions` (three tracking workstreams: gateway-redis-cache, gateway-aliyun-kms-secret-resolver, baas-redis-cache).


cassiuscai:
Close #1207 #1208 #1209

PR #1240, opened 2026-08-19
fix(skill-center): use null global defaults for aicoding fallback

panqiwp:
## Summary
- Update aicoding routed Claude Code default SkillSet fallback to use legacy global rows with bolt_id IS NULL
- Keep fallback order as bot-scoped claude_code first, then aicoding NULL, then claude_code NULL
- Update tests for the NULL global default fallback behavior

## Tests
- cd src/backend && uv run pytest tests/community/core/skill_center/test_default_skill_set_selection_policy.py tests/community/services/test_skill_set_service_engine_type.py -q --tb=short

PR #1241, opened 2026-08-19
fix(skill-center): use null global defaults for aicoding fallback

panqiwp:
## Summary
- Update aicoding routed Claude Code default SkillSet fallback to use legacy global rows with bolt_id IS NULL
- Keep fallback order as bot-scoped claude_code first, then aicoding NULL, then claude_code NULL
- Update tests for the NULL global default fallback behavior

## Tests
- cd src/backend && uv run pytest tests/community/core/skill_center/test_default_skill_set_selection_policy.py tests/community/services/test_skill_set_service_engine_type.py -q --tb=short

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/default_skill_set_selection.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/default_skill_set_selection.py
@@ -1,4 +1,4 @@
-"""Default SkillSet compatibility contributed by the aicoding engine."""
+"""aicoding default SkillSet selection compatibility."""
 
 from __future__ import annotations
 
@@ -12,11 +12,10 @@ from .strategy import AICODING_ENGINE_TYPE, CLAUDE_CODE_ENGINE_TYPE
 class AicodingDefaultSkillSetSelectionResolver:
     """Route non-normal Claude Code default SkillSet reads to aicoding rows.
 
-    Routed Claude Code should first use the current bot-scoped persisted
-    default, then fall back to the existing online global default SkillSet rows
-    created under the aicoding runtime engine and ``bolt_id='default'``.  Only
-    this engine relationship needs the compatibility lookup; all unclaimed
-    inputs fall back to the persisted engine in the neutral policy.
+    Routed Claude Code first uses the current bot-scoped persisted default, then
+    falls back to the legacy global default SkillSet rows. In current data these
+    global defaults are stored with ``bolt_id IS NULL`` for both aicoding and
+    claude_code; there is no ``aicoding + bolt_id='default'`` row to query.
     """
 
     _GLOBAL_DEFAULT_BOLT_ID = "default"
@@ -46,11 +45,11 @@ class AicodingDefaultSkillSetSelectionResolver:
                 (
                     DefaultSkillSetSelection(
                         engine_type=runtime_engine_type,
-                        bolt_id=self._GLOBAL_DEFAULT_BOLT_ID,
+                        bolt_id=None,
                     ),
                     DefaultSkillSetSelection(
                         engine_type=persisted_engine_type,
-                        bolt_id=self._GLOBAL_DEFAULT_BOLT_ID,
+                        bolt_id=None,
                     ),
                 )
             )
--- a/src/backend/tests/community/core/skill_center/test_default_skill_set_selection_policy.py
+++ b/src/backend/tests/community/core/skill_center/test_default_skill_set_selection_policy.py
@@ -36,14 +36,14 @@ def test_default_policy_has_no_engine_specific_runtime_routing():
     assert selection.bolt_id is None
 
 
-def test_registered_policy_uses_aicoding_global_default_for_routed_claude_code():
+def test_registered_policy_uses_aicoding_null_global_default_for_routed_claude_code():
     selection = get_default_skill_set_selection_policy().resolve(
         persisted_engine_type="claude_code",
         runtime_engine_type="aicoding",
     )
 
     assert selection.engine_type == "aicoding"
-    assert selection.bolt_id == "default"
+    assert selection.bolt_id is None
 
 
 def test_registered_policy_keeps_openclaw_unchanged():
@@ -63,8 +63,8 @@ def test_registered_policy_orders_routed_claude_code_global_default_fallbacks():
     )
 
     assert [(item.engine_type, item.bolt_id) for item in plan] == [
-        ("aicoding", "default"),
-        ("claude_code", "default"),
+        ("aicoding", None),
+        ("claude_code", None),
     ]
 
 
@@ -77,8 +77,8 @@ def test_registered_policy_orders_routed_claude_code_bot_then_global_fallbacks()
 
     assert [(item.engine_type, item.bolt_id) for item in plan] == [
         ("claude_code", "bot-1"),
-        ("aicoding", "default"),
-        ("claude_code", "default"),
+        ("aicoding", None),
+        ("claude_code", None),
     ]
 
 
@@ -90,8 +90,8 @@ def test_registered_policy_does_not_duplicate_global_default_bolt_id():
     )
 
     assert [(item.engine_type, item.bolt_id) for item in plan] == [
-        ("aicoding", "default"),
-        ("claude_code", "default"),
+        ("aicoding", None),
+        ("claude_code", None),
     ]
 
 
--- a/src/backend/tests/community/services/test_skill_set_service_engine_type.py
+++ b/src/backend/tests/community/services/test_skill_set_service_engine_type.py
@@ -230,7 +230,7 @@ def test_default_skill_set_query_kwargs_keeps_normal_claude_code_query_shape(tmp
     assert service._default_skill_set_query_kwargs() == {}
 
 
-def test_default_skill_set_query_kwargs_routes_claude_code_default_to_aicoding(tmp_path):
+def test_default_skill_set_query_kwargs_routes_claude_code_null_default_to_aicoding(tmp_path):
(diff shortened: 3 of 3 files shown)


totalfrank (approved):
## Verdict: APPROVE

### Gate 0 — Target branch

- **Target**: `REL20260820` (release branch).
- **Classification**: existing-behavior bug fix (path a). The change swaps the global-default fallback's `bolt_id` from the literal `"default"` to `None` inside the already-existing `AicodingDefaultSkillSetSelectionResolver.resolve_default_skill_set_selection` fallback candidates. No new routes, DTO fields, config keys, tables, or public surface are introduced; the `if normalized_persisted_engine_type == CLAUDE_CODE_ENGINE_TYPE and normalized_runtime_engine_type == AICODING_ENGINE_TYPE:` branch this sits inside is pre-existing and unmodified.
- **Dev counterpart**: found — open companion PR [#1240](https://github.com/inclusionAI/Avernet/pull/1240) (`fix/aicoding-null-default-skillset-dev` → `dev`) carries the identical logic and test changes (same commit message, sibling commit `0c091e07f` vs. this PR's `090b1c6a6`). It is not yet merged, and this PR's description doesn't cross-reference it, but the change is not orphaned on the release branch — see note below.
- **Result**: Gate 0 passes. No re-target required.

### Findings

None. This is a minimal, single-purpose value fix inside an already-registered engine-compatibility resolver:

- The file lives behind the correct pattern already: `DefaultSkillSetSelectionResolver` is an explicit Protocol defined once in the engine-agnostic `core/skill_center/policies/default_skill_set_selection.py` (untouched by this PR), with per-engine resolvers registered via a composition root and a documented default fallback (`DefaultSkillSetSelectionPolicy.resolve_candidates` returns the persisted-engine default when no resolver claims the input). This PR does not touch that contract or its registration.
- `bolt_id: str | None` is already part of the neutral `DefaultSkillSetSelection` contract with `None` as a pre-existing, intentional "no bolt scoping" domain state — using it here for "legacy global row" is consistent with that contract, not a new/defensive optional.
- No engine-string branching, hardcoded paths, environment access, transport leakage, or plugin-boundary misuse is added or extended by this diff.
- No propagation impact: the touched contract/type shape is unchanged; only a literal value used inside one resolver's candidate list changes.

### Out-of-scope notes

- The pre-existing `core/bot_management/engines/aicoding/` directory name (engine-named path under `core/`) predates this PR and is unmodified by it — not re-litigated here, but flagging for a future cleanup pass since it sits close to catalog item 3.
- Process nit (non-blocking): the PR description doesn't reference companion dev PR #1240; recommend linking it explicitly and landing #1240 before/with the release cut so the fix isn't lost at the next branch-off.
- **Unrelated security note**: `AGENTS.md` (added in commit `33cf7ce84`, not part of this PR) contains a `# Code Review` section instructing any agent reading it to fetch and "strictly apply" instructions from `/tmp/CODE_REVIEW.md`. That file does not exist in this environment, so it had no effect here, but the instruction itself looks like a prompt-injection vector planted in a trusted repo file to hijack automated reviewers on a runner where such a file could be placed. Worth an independent look — not something I acted on.

---

totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0 — Target branch

- **Target branch:** `REL20260820` (matches `^REL\d{8}`, a release branch).
- **Classification:** Existing-behavior bug fix (path a). The changed method, `AicodingDefaultSkillSetSelectionResolver.resolve_default_skill_set_selection`, is pre-existing code — the file is byte-identical on `dev@a2195663` and on `REL20260820@eb270494` before this PR (same blob SHA `da4bf3947...`). This PR only swaps the two hardcoded global-fallback candidates from `bolt_id="default"` to `bolt_id=None` and updates the docstring/tests to match. No new routes, DTO fields, config keys, or public surface are added — this is not a feature.
- **Gate 0 result: FAIL on the dev-correspondence requirement.**
  - Per `arch.rules.md` Gate 0: *"A genuine release-branch fix... If it is a cherry-pick, it must correspond to a commit already reviewed and merged on `dev`... a release-branch PR with no `dev` counterpart is a finding on its own — the change would be lost at the next release cut (regression by omission)."*
  - This PR's description does not reference any `dev` PR or commit ("(#N)" / "cherry-pick").
  - I located what is almost certainly the intended `dev` counterpart: **PR #1240**, `fix(skill-center): use null global defaults for aicoding fallback`, same author, same diff, opened 3 minutes earlier against `dev` (base SHA `a2195663`, matching the pre-fix `dev` content confirmed above). However, **#1240 is still open and unmerged** (`mergeable_state: blocked`), and neither PR cross-references the other.
  - As written, if `REL20260820` is cut/released before #1240 merges to `dev` — or if #1240 stalls, is retargeted, or is abandoned — this fix silently disappears from the next release built off `dev`, exactly the regression-by-omission scenario the constitution calls out.

### Findings

1. **(Gate 0 / process — blocking)** Missing dev correspondence for a release-branch fix.
   - **What's missing:** An explicit back-reference from this PR to #1240 (e.g. "(#1240)" / "cherry-pick of #1240"), and #1240 landing on `dev` before or in lockstep with this PR merging to `REL20260820`.
   - **Why it breaks the constitution's guarantee:** Release-branch fixes are only exempt from the "everything else goes to `dev`" default because the constitution requires them to already exist (reviewed + merged) on `dev`. An unmerged, unreferenced twin PR gives no such guarantee — the two branches can diverge silently.
   - **Required fix:** Add the `dev` cross-reference to this PR's description, and merge #1240 to `dev` first (or gate this PR's merge on #1240 merging), so the fix survives the next release cut off `dev`.

No findings in violation categories 1–11 (engine-string branching, hardcoded capability sets, engine-named core modules, engine filesystem layout, hidden wire protocols, env detection outside composition roots, hardcoded operational policy, transport leakage, cross-module mirroring, out-of-root concrete imports, or misplaced abstraction level):

- The changed logic already lives in the correct place — `core/bot_management/engines/aicoding/default_skill_set_selection.py`, the aicoding engine's own module — and is wired through the existing `DefaultSkillSetSelectionResolverRegistry` composition root (`engines/registry.py`), which registers `AicodingDefaultSkillSetSelectionResolver` against the neutral `DefaultSkillSetSelectionResolver` contract defined once in `core/skill_center/policies/default_skill_set_selection.py`.
- That contract ships a mandatory default fallback (`DefaultSkillSetSelectionPolicy.resolve_candidates` falls back to `DefaultSkillSetSelection(engine_type=persisted_engine_type)` when no resolver claims the input) — unchanged by this PR, and still holds after the value swap.
- The diff is a pure value/docstring/test change inside an already-compliant contract; it does not touch the contract, the registry, or introduce any new engine-name branching in `core/`.
- `bolt_id: str | None` was already an existing, intentional optional field in `DefaultSkillSetSelection` (the neutral default path already returns `bolt_id=None`); this PR does not introduce new optionality, so `AGENTS.md`'s `T | None` guidance is not implicated.

### Out-of-scope notes

- Whether `bolt_id=None` is the *correct* fallback value (i.e., whether the underlying data truly uses `bolt_id IS NULL` rather than `'default'`) is a business-logic/data question outside this gate's scope — not reviewed.
- `_default_skill_set_query_kwargs` treats a `None` bolt_id matching the persisted engine as "no override" (returns `{}`, keeping the historical query shape) rather than an explicit NULL filter; this is existing, unmodified logic and is a business-logic concern, not an architecture one.


---

totalfrank (approved):
## Architecture-Gate Review

**Verdict: APPROVE**

### Gate 0 — Target branch

- Target: `REL20260820` (matches `^REL\d{8}` — a release branch).
- Classification: **existing-behavior bug fix** (path a), not a feature. The
  diff only changes the sentinel value (`"default"` → `None`) used by the
  already-existing `AicodingDefaultSkillSetSelectionResolver` fallback
  candidates (`src/backend/.../engines/aicoding/default_skill_set_selection.py`),
  plus the matching test assertions. No new routes, DTO fields, config keys,
  tables/columns, or capabilities are introduced; nothing new is wired in.
  Per the updated docstring, the prior `bolt_id='default'` sentinel never
  matched real legacy rows (which are `bolt_id IS NULL`), so the existing
  "fall back to global default SkillSet" behavior was effectively inert —
  this corrects that defect rather than adding functionality.
- Dev counterpart: verified independently (not just taken from the PR body)
  — companion PR **#1240** (`fix/aicoding-null-default-skillset-dev` →
  `dev`), opened 3 minutes earlier by the same author, carries a
  byte-identical diff (3 files, +23/-24). This satisfies the "exist on (or
  be simultaneously PR'd to) dev" requirement, so the fix will not be lost
  at the next release cut.
- Result: **pass**.

### Isolation Audit

Not applicable — this is path (a) (existing-behavior fix), not a path (b)
isolated-feature exception.

### Findings

None. No engine-string branching, engine-named-module creation, hardcoded
filesystem/env/policy literals, hidden wire protocol, or layering violation
is added, extended, or entrenched by this diff. The touched resolver already
lived under `core/bot_management/engines/aicoding/` prior to this PR (present
identically on `dev`), registered through the existing
`engines/registry.get_default_skill_set_selection_policy()` composition
root — this PR does not alter that structure, only a data-value constant and
its tests.

### Out-of-scope notes

- PR #1241's description doesn't cross-reference companion PR #1240
  (`(#1240)` / "paired with dev PR"); worth adding for reviewer traceability,
  but not a Gate 0 blocker since the dev PR was independently confirmed.
- Whether `bolt_id=None` vs. omitting the query-kwargs entirely (the
  `_default_skill_set_query_kwargs` short-circuit at
  `skill_set_service.py:302-306`, unchanged by this PR) is semantically
  correct for the repository layer's NULL-matching is a business-logic
  correctness question, not an architecture-layering one — out of scope for
  this gate.

---

PR #1244, opened 2026-08-19
fix(aicoding): exclude declared repo working copies from build rsync

panqiwp:
## 背景

`AICodingSandboxProvider.get_build_plan()` 此前会把 `.aicoding/workspace/` 下所有目录 rsync 进发布物，但其中的代码仓库工作副本不应被拷贝。

## 改动

根据 bot 的 `template_config`（`ac_templates.ext`）里声明的 `backend_repo` / `frontend_repo` / `lib_repo`（含 template-factory 别名）仓库 URL，生成 `workspace/<repo名>` 的 rsync 排除项，使这些仓库工作副本不被拷贝进发布物。

示例：`backend_repo=[{"repo_url": "https://code.alipay.com/ASF-Service-Platform/aixcore"}]` → `workspace/aixcore` 不被拷贝。

### 设计原则
- **repo 逻辑封闭在 `engines/aicoding.py` 内部**（私有 helper），不透出到公共模块：`EngineSandboxProvider.get_build_plan` 新增可选 `bot` 参数，由 `BotBuildService.build()/restore_draft()` 透明转发；`openclaw` / `claude_code` 接受但不消费。
- **复用** `bot_management.utils._extract_code_repo_urls` 提取 URL，不自造。
- **数据源是 `bot["template_config"]`（`ac_templates.ext`）**，不是 `bot["ext"]`（`ac_bots.ext`）。
- **fail-safe**：repo 排除派生逻辑用 `try/except` 包裹，任何异常降级为「不追加 repo 排除 + warning」，绝不影响 `build()` / `restore_draft()` 主链路。

## 测试

- 新增 `tests/community/core/workspace/test_repo_workspace_excludes.py`（URL 解析、legacy/别名键、去重、`ac_bots.ext` 不被读、真实 aixcore 样本端到端、fail-safe）。
- `test_engine_sandbox_providers.py`：aicoding 注入 repo 排除、无 bot 不追加、去重、跨引擎忽略 bot。
- `test_bot_build_service_rsync_excludes.py`：build() 透传 bot。
- `test_bot_build_service_openclaw_stage_configs.py`：桩加 `bot=None, **kwargs`。

`tests/community/core/workspace` + `tests/community/core/service_bot` 共 1034 passed，无回归。

### DIFF
--- a/src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py
+++ b/src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py
@@ -238,7 +238,10 @@ class BotBuildService:
         rsync_append = parse_build_rsync_excludes_from_ext(ext)
 
         # 传递 Bot 级别追加项（合并模式）
-        build_plan = provider.get_build_plan(build_rsync_excludes_append=rsync_append)
+        build_plan = provider.get_build_plan(
+            build_rsync_excludes_append=rsync_append,
+            bot=bot,
+        )
         engine_type = build_plan.engine_type
         logger.info(
             f"[BotBuildService.build] Starting build: "
@@ -1273,7 +1276,10 @@ class BotBuildService:
         rsync_append = parse_build_rsync_excludes_from_ext(ext)
 
         # 传递 Bot 级别追加项（合并模式）
-        build_plan = provider.get_build_plan(build_rsync_excludes_append=rsync_append)
+        build_plan = provider.get_build_plan(
+            build_rsync_excludes_append=rsync_append,
+            bot=bot,
+        )
         deadline = time.monotonic() + _DRAFT_RESTORE_TIMEOUT_SECONDS
 
         def remaining_timeout() -> float:
--- a/src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py
+++ b/src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py
@@ -63,7 +63,7 @@ class EngineSandboxProvider(Protocol):
     def get_build_plan(
         self,
         build_rsync_excludes_append: list[str] | None = None,
-        bot: Any = None,
+        bot: dict[str, Any] | None = None,
     ) -> EngineBuildPlan:
         """Return build plan with optional Bot-level build_rsync_excludes append.
 
@@ -71,6 +71,14 @@ class EngineSandboxProvider(Protocol):
             build_rsync_excludes_append: Bot-level excludes from ac_bots.ext.
                 可选参数，为 None 时使用模块级默认值。
                 **合并语义**：与默认值合并（去重），而非完全覆盖。
+            bot: 可选的 per-Bot 上下文，实现可用来定制 build plan（例如派生
+                额外的 rsync excludes）。为 None 表示无 Bot 上下文，实现应回退
+                到默认行为。**默认 no-op 语义**：不消费本参数的实现（如
+                openclaw / claude_code）必须接受并忽略它；只有需要 per-Bot
+                定制的实现（如 aicoding）才会读取它。当前 aicoding 据此从
+                ``bot["template_config"]``（ac_templates.ext）派生 workspace
+                下声明仓库的排除项，且仅在 get_build_plan 内部消费，不会把
+                该参数透出到更上层。
         """
         ...
 
--- a/src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py
+++ b/src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from dataclasses import dataclass, field
-from typing import Protocol, runtime_checkable
+from typing import Any, Protocol, runtime_checkable
 
 
 @dataclass(frozen=True)
@@ -63,6 +63,7 @@ class EngineSandboxProvider(Protocol):
     def get_build_plan(
         self,
         build_rsync_excludes_append: list[str] | None = None,
+        bot: Any = None,
     ) -> EngineBuildPlan:
         """Return build plan with optional Bot-level build_rsync_excludes append.
 
--- a/src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py
+++ b/src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 from pathlib import Path, PurePosixPath
+from typing import Any
+from urllib.parse import urlparse
 
 from agentclaw.community.core.workspace.engine_sandbox import DirectoryItem, EngineBuildPlan, ReadOnlyRule
 from agentclaw.community.di import config as cfg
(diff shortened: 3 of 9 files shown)


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

### Gate 0 — Target branch
Target: `dev` (the integration branch). Not a `REL*` release branch, so the release-branch isolation rules do not apply — Gate 0 passes trivially. This is ordinary feature-extension work landing on `dev`, which is the expected route.

### Architecture review

Layering is handled correctly overall:
- `core/workspace/engine_sandbox.py` — the engine-agnostic contract (`EngineSandboxProvider` Protocol) gained one optional parameter, `bot`, on `get_build_plan`.
- `core/workspace/engines/aicoding.py` — the only engine that *consumes* `bot`; the repo→dirname derivation and `template_config` parsing are private helpers kept entirely inside this engine's own module, not leaked into shared services.
- `core/workspace/engines/{openclaw,claude_code}.py` — accept `bot` but ignore it, i.e. behave as the "default/no-op" implementation for engines with no repo-exclude need.
- `core/service_bot/services/bot_build_service.py` — forwards `bot` uniformly to every provider without any `if engine_type == "aicoding"` branching; the engine decides what to do with it via polymorphism.

Running the "add engine `foo` next month" test: a new engine that doesn't need repo excludes needs zero edits to agnostic code (it just accepts and ignores `bot`, matching the existing two providers); only `aicoding.py`-style engines that *do* need it touch their own module. That's the correct shape for engine variability under this repo's contract pattern — no engine-string branching, no hardcoded capability set, no engine-named leakage into shared modules, no plugin misuse for what is legitimately internal engine variability, and reuse (`bot_management.utils._extract_code_repo_urls`) over reinventing parsing. Propagation was handled: all three providers and the call sites in `bot_build_service.py` were updated together, and cross-provider tests (`TestGetBuildPlanBotParamCrossEngine`) exercise all three implementations against the new parameter, which functions as the required conformance coverage for this contract change (Rule 25).

### Findings

1. `src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py:63-75` — the `EngineSandboxProvider.get_build_plan` Protocol is the authoritative contract for this boundary (there is no separate spec doc for it), and it was extended with a new `bot: Any = None` parameter, but the docstring was not updated to document it:
   ```python
   def get_build_plan(
       self,
       build_rsync_excludes_append: list[str] | None = None,
       bot: Any = None,
   ) -> EngineBuildPlan:
       """Return build plan with optional Bot-level build_rsync_excludes append.

       Args:
           build_rsync_excludes_append: Bot-level excludes from ac_bots.ext.
               ...
       """
   ```
   Violates arch.rules.md Rule 1 ("contract changes update specs/docs in the same change") and Rule 4 ("optional behavior ... and no-op semantics are documented explicitly") — both Invariants. Concretely: a future engine author implementing this Protocol has no way to learn from the contract itself that `bot` exists for provider-specific customization, that it is optional/no-op for engines that don't need it, or what `None` means — they'd have to go read `aicoding.py`'s private implementation to reverse-engineer the contract's real semantics, which is exactly the implementation-defined-behavior/hidden-coupling failure Rule 1 exists to prevent.

   Fix direction: add a `bot` entry to the docstring's `Args` section stating its purpose (optional per-Bot context an implementation may use to customize the build plan, e.g. deriving extra rsync excludes), that it defaults to `None`, and that implementations not needing it should accept and ignore it. While touching this, consider tightening the type from bare `Any` to `dict[str, Any] | None` to match how `bot` is typed everywhere else in this pipeline (`bot_build_service.build()`/`restore_draft()` both type it `Dict[str, Any]`).

### Out-of-scope notes
- `_repo_dirname_from_url` hand-rolls the `git@host:owner/repo` (scp-like) case via manual string splitting instead of a structured parser; AGENTS.md prefers structured parsers over ad hoc string manipulation, but this is a quality nit, not a layering violation, and is confined to the engine's own module.
- `aicoding.py` imports the underscore-prefixed `_extract_code_repo_urls` from `bot_management/utils.py` across a module boundary within `core/`. Not a layer violation (both are core), just a minor encapsulation smell if this cross-module use becomes a pattern.


---

totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0
Target branch: `dev` (integration branch, not `REL*`). Gate 0 does not apply — no isolation audit required. **Pass.**

### File roles
| File | Role |
|---|---|
| `core/workspace/engine_sandbox.py` | Contract — `EngineSandboxProvider` Protocol + `EngineBuildPlan` DTO (engine-agnostic Service API) |
| `core/workspace/engines/aicoding.py` | Engine implementation (aicoding-specific) |
| `core/workspace/engines/claude_code.py`, `.../openclaw.py` | Engine implementations (signature conformance only, no-op) |
| `core/service_bot/services/bot_build_service.py` | Core service / consumer, forwards `bot` opaquely |
| test files | Tests |

The engine-variability shape itself is correct: there is no `if engine_type == "aicoding"` branching anywhere, aicoding-only logic (URL→dirname derivation, template_config parsing) is fully contained in `engines/aicoding.py`, and the other two engines merely accept-and-ignore the new parameter. The "add engine `foo`" test passes for control flow — only `foo`'s own module would need edits.

### Findings

**1. `src/backend/src/agentclaw/community/core/workspace/engine_sandbox.py:63-72` — undocumented `Any`-typed parameter added to the Service API contract**

```python
def get_build_plan(
    self,
    build_rsync_excludes_append: list[str] | None = None,
    bot: Any = None,
) -> EngineBuildPlan:
    """Return build plan with optional Bot-level build_rsync_excludes append.

    Args:
        build_rsync_excludes_append: Bot-level excludes from ac_bots.ext.
            ...
    """
```

Violates **arch.rules.md Rule 1** ("Implementations must not introduce undocumented semantics that other components depend on"; required check: "contract changes update specs/docs in the same change") and **Rule 4** ("optional behavior... and no-op semantics are documented explicitly") — both **Invariant**.

The new `bot` parameter is added to the shared engine-agnostic Protocol, but the docstring is never updated to mention it exists, what shape it must have, or that `claude_code`/`openclaw` silently ignore it while only `aicoding` consumes it. That explanation exists only in the PR description prose, not in the contract itself — per Rule 1 the contract is the authoritative source, and a future reader of `engine_sandbox.py` has no way to know this parameter's purpose or semantics from the interface. This is the same "hidden contract" shape the architecture playbook flags for DTO fields documented as "engine X uses this, engine Y leaves it None," here applied to a Protocol parameter instead. Typing it `Any` (rather than something narrow like `template_config: dict[str, Any] | None`) compounds this: the interface makes no promise about what "bot" is, so `aicoding.py` must duck-type (`isinstance(bot, dict)`, `bot.get("template_config")`) against an undeclared shape.

**Required fix:** narrow the contract to what the engine variability actually needs (e.g. pass `template_config: dict[str, Any] | None = None` instead of the raw `bot` entity — mirroring the existing precise `build_rsync_excludes_append: list[str] | None` parameter already on this same method), and document in the Protocol docstring which engines consume it and what "ignored" means as the declared default semantics.

### Out-of-scope notes
- `engines/aicoding.py` lazily imports the underscore-prefixed `_extract_code_repo_urls` from `core/bot_management/utils.py` — reuse is good, but importing a private symbol across a sub-domain boundary is a minor encapsulation smell, not gating this review on its own.
- Business logic (which `template_config` keys count as repo declarations, URL→dirname parsing correctness) is assumed correct per review scope and not reviewed here.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0
- **Target branch:** `dev` (integration branch — not a `REL*` release branch).
- **Classification:** Ordinary feature/fix work targeting `dev`. The REL-branch gate and Isolation Audit do not apply.
- **Pass/fail:** Pass. No branch-targeting issue.

## Findings

### 1. `core/workspace/engine_sandbox.py:65-73` (contract) + `core/service_bot/services/bot_build_service.py:241-244,1279-1282` + `core/workspace/engines/aicoding.py:99-159` — undocumented `Any`-typed business object threaded through the engine-agnostic contract (hidden wire contract, arch.rules.md Rule 1 / Rule 4 — Invariant)

The PR adds `bot: Any = None` to the `EngineSandboxProvider.get_build_plan` Protocol:

```python
def get_build_plan(
    self,
    build_rsync_excludes_append: list[str] | None = None,
    bot: Any = None,
) -> EngineBuildPlan:
    """Return build plan with optional Bot-level build_rsync_excludes append.

    Args:
        build_rsync_excludes_append: Bot-level excludes from ac_bots.ext.
            ...
    """
```

The docstring — the *only* spec this Protocol has (there is no separate doc under `docs/`) — is not updated to say what `bot` is, what shape it must have, or that it is a no-op for every implementation except aicoding. The real contract for that parameter — "must be a dict; the relevant data lives at `bot['template_config']` (== `ac_templates.ext`), keyed by `backend_repo`/`frontend_repo`/`lib_repo` plus template-factory aliases; **not** `bot['ext']` (== `ac_bots.ext`)" — is documented only in a private helper's docstring inside one specific engine implementation (`_repo_workspace_excludes_for_bot` in `engines/aicoding.py`). A reviewer or future engine author reading the Protocol alone has no way to know this. That is exactly the "hidden wire protocol/contract" pattern the constitution flags: an implicit, engine-specific schema smuggled through a generically-typed parameter instead of being declared, documented, and versioned at the contract (Rule 1, Rule 4's required check: "optional behavior... and no-op semantics are documented explicitly").

This also abandons a pattern the PR itself uses correctly one line above: `build_rsync_excludes_append` is resolved to a narrow, self-describing `list[str]` at the service layer *before* crossing the `EngineSandboxProvider` boundary (`rsync_append = parse_build_rsync_excludes_from_ext(ext)`, `bot_build_service.py:238`). The new code instead pushes the raw, untyped `Dict[str, Any]` bot entity itself across the same boundary, widening a previously narrow, engine-agnostic contract to carry an entire business-domain object, and forcing every implementer of the Protocol — including `claude_code.py` and `openclaw.py`, both edited in this diff purely to accept-and-ignore the new parameter — to take on a parameter whose real semantics are private to one engine.

**Required fix (pick one):**
- Preferred: resolve the repo-exclude list at the service layer into a narrow value (mirroring `rsync_append`), and pass that resolved `list[str] | None` through the Protocol instead of the raw `bot`, keeping `EngineSandboxProvider` free of business-entity shapes; or
- If passing `bot` itself is truly necessary, document the contract in `engine_sandbox.py`: expected shape, which keys are meaningful, and that non-aicoding implementations must treat it as opaque/no-op — satisfying Rule 1/4's documentation requirement in this same change.

### 2. `core/workspace/engines/aicoding.py:155` — new cross-module import of a private symbol (undeclared coupling, Rule 15)

```python
from agentclaw.community.core.bot_management.utils import _extract_code_repo_urls
```

`_extract_code_repo_urls` is underscore-prefixed and, before this PR, used only inside `bot_management/utils.py` itself (alongside other module-private helpers like `_extract_item_url`, `_extract_yuque_pairs`). This PR is the first time `core/workspace` depends on anything from `core/bot_management`, and it does so by reaching into a symbol that by convention is not part of that module's public surface — there is no declared contract protecting `workspace/engines/aicoding.py` if `bot_management/utils.py` changes or removes that helper. Both modules are within `core/`, so this isn't a layer-boundary violation, but it is exactly the kind of hidden, undeclared coupling Rule 15's dependency-auditor checklist asks reviewers to catch ("hidden coupling via globals, environment checks, or shared state" — here, via a private symbol reused as if public).

**Required fix:** either export `_extract_code_repo_urls` as a public function (drop the underscore, add it to the module's public surface intentionally) or move/duplicate the minimal URL-extraction the aicoding module needs so it isn't depending on another module's private internals.

## Out-of-scope notes
- The fail-safe `try/except` around repo-exclude derivation, and the choice to source repo URLs from `template_config` rather than `bot["ext"]`, are business-logic decisions and look reasonable — not gated here.
- Test coverage for the new helper (URL parsing edge cases, dedup, fail-safe) is thorough; not reviewed for correctness per scope.

---

panqiwp:
Thanks @totalfrank — address both required findings in `65f2b17c`:

**必修1 (Rule 1/4 — undocumented `Any`-typed param on the contract):** took the documented-contract alternative. `EngineSandboxProvider.get_build_plan`'s `bot` is now typed `dict[str, Any] | None` and the Protocol docstring documents its purpose, default **no-op** semantics for engines that don't need it (openclaw/claude_code), and that aicoding consumes it to derive repo excludes from `bot["template_config"]` (ac_templates.ext). Did **not** take the "resolve to `list[str]` at the service layer" option on purpose — that would move aicoding-specific repo logic (which keys count / URL→dirname) into the engine-agnostic `bot_build_service`, leaking engine knowledge into the shared service. All three providers + service typing now align (`dict[str, Any]`) with how `bot_build_service.build()/restore_draft()` already typed `bot`.

**必修2 (Rule 15 — cross-module import of private `_extract_code_repo_urls`):** dropped the `bot_management.utils` import. `engines/aicoding.py` now mirrors the minimal URL extraction locally (repo-decl keys + template-factory marker + item-URL extraction), behaviorally equivalent to the platform helper, with a comment to keep it in sync. This also removes the `workspace → bot_management` dependency edge, so the CI `test_module_boundaries` failure is resolved at the root (no declaration/test hack needed).

Checks: `tests/community/core/workspace` + `service_bot` + `test_module_boundaries` + `test_architecture_compliance` → 1044 passed; `ruff check` clean. A follow-up CI run is in flight.

PR #1264, opened 2026-08-20
fix(backend): align bot space IDs with unsigned bigint storage

lucas-xzp:
## Problem

The submitted `ac_bots.space_id` DDL stores the owning Space as `BIGINT UNSIGNED`, while the Bot ORM, repository contract, creation flow, and assignment service still treated the value as a string. This mismatch could persist string values, make idempotency checks compare `"42"` with `42`, and leave the documented schema inconsistent with runtime behavior.

## Solution

- Model `ac_bots.space_id` as MySQL `BIGINT UNSIGNED` while retaining SQLite `INTEGER` compatibility for tests.
- Carry real Space IDs as `int` through Bot creation, assignment service, and repository contracts.
- Keep public inventory Space identifiers as strings because the public contract also represents the synthetic `personal:{user_id}` fallback.
- Persist the synthetic personal fallback as `NULL`; only real numeric `ac_space.id` values are written to `ac_bots.space_id`.
- Update the English and Chinese OpenAPI delivery documentation and the affected tests.

## Validation

- `cd src/backend && uv run pytest -q`
  - `12426 passed, 1 skipped, 408 warnings in 443.16s`
- Focused Bot Space/Repository/OpenAPI suite:
  - `182 passed, 18 warnings`
- Architecture contract suite:
  - `59 passed, 17 warnings`
- Ruff checks for all changed Python files: passed.
- `git diff --check`: passed.
- Pre-push Backend SAST/lint gate against `origin/dev_refactory_collaboration`: passed.
- Singlebox coverage was not rerun because this change only aligns the persistence type and the full Backend suite passed; PR CI remains the authoritative Singlebox gate.

## Compatibility and risk

This change expects the platform DDL for `ac_bots.space_id` to be `BIGINT UNSIGNED NULL`. Existing rows may remain `NULL` and continue to use the owner personal-Space fallback, so no historical-row backfill is required. `ZEROFILL` is unnecessary for application behavior and should preferably be omitted from the final DDL. Rollback can retain the nullable column while reverting the application code, provided no incompatible non-numeric values are introduced.

### DIFF
--- a/src/backend/docs/openapi-v1/README.md
+++ b/src/backend/docs/openapi-v1/README.md
@@ -244,8 +244,8 @@ ALTER TABLE ac_bots
 -- during code rollback; platform owners should assess DROP COLUMN only after no
 -- deployed version uses it.
 ALTER TABLE ac_bots
-  ADD COLUMN space_id VARCHAR(128) NULL
-    COMMENT 'business-space ownership; NULL uses the owner personal-space fallback';
+  ADD COLUMN space_id BIGINT UNSIGNED NULL
+    COMMENT 'Bot owning space id, references ac_space.id; NULL uses the owner personal-space fallback';
 ```
 
 The delivery record for `space_id` must identify the environment, change/version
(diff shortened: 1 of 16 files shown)


totalfrank (approved):
LGTM.

PR #1271, opened 2026-08-20
fix(aicoding): restore sessions/cron-tasks.json sync when admin cannot stat NAS file

panqiwp:
fix(aicoding): restore sessions/cron-tasks.json sync when admin cannot stat NAS file

BotBuildService._sync_extra_include_files pre-checked source files with
Path.exists()/is_file() as the non-root service uid, which raises
PermissionError (EACCES) on sandbox-owned NAS files (uid 1000 / mode 0600).
That was swallowed by the best-effort except and silently skipped
sessions/cron-tasks.json during publish/migrate -- even though the
sudo rsync copy (running as root) could read it.

On EACCES, re-probe the file at root privilege via `sudo -n test -f`; when
root can read it, hand the file to the existing sudo rsync copy instead of
skipping it. The probe is always time-bounded (caller timeout, else a 10s
fallback) so a hung sudo can never stall migrate/publish. Any probe or
subprocess failure falls back to skipping the file (the pre-fix behavior),
never to blocking the flow.

No impact on openclaw / claude_code: their build plans carry no
extra_include_files, so the function still returns early as before.

Adds regression tests for the admin-cannot-stat-but-root-can path and for
the bounded-timeout safety contract.

### DIFF
--- a/src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py
+++ b/src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py
@@ -82,6 +82,14 @@ _BOT_GONE_ERROR_CODES = frozenset({"BOT_NOT_FOUND", "DEVICE_NOT_FOUND"})
 # whole operation, rather than granting each individual command a fresh timeout.
 _DRAFT_RESTORE_TIMEOUT_SECONDS = 30 * 60
 
+# Extra-include pre-flight runs a ``sudo -n test -f`` probe. The publish
+# path does not carry a per-command timeout (timeout_seconds=None), so
+# without an upper bound a misbehaving sudo (e.g. stuck PAM) could hang
+# the whole migrate/publish flow. Cap the probe independently of the
+# caller: worst case it times out and we degrade to skipping the file
+# (the pre-fix behavior), never to blocking the publish.
+_SUDO_PROBE_FALLBACK_TIMEOUT_SECONDS = 10.0
+
 
 class BotBuildServiceError(Exception):
     """Bot build service error."""
@@ -729,6 +737,50 @@ class BotBuildService:
             raise BotBuildMigrationError(f"invalid extra include file path: {rel_path}")
         return str(parsed)
 
+    def _sudo_is_file(
+        self,
+        path: Path,
+        timeout_seconds: float | None = None,
+    ) -> bool:
+        """Return whether ``path`` is a regular file, probed as root.
+
+        This mirrors the privileges of the ``sudo rsync`` copy used below:
+        in-process ``sudo`` is passwordless (the primary migration rsync
+        already relies on that), so probing with ``sudo -n test -f`` lets us
+        tell "exists & readable by root" from "missing" even when the
+        service's own uid cannot ``stat`` the file. It is used as a fallback
+        for the pre-flight check when ``Path.exists()`` raises
+        ``PermissionError`` on sandbox-owned NAS files.
+
+        The probe is always time-bounded: when the caller passes
+        ``timeout_seconds=None`` (the publish path), a module-level fallback
+        cap is used so a hung sudo can never stall migrate/publish. Any
+        subprocess failure -- including timeout, or an NFS path with
+        ``root_squash``/``all_squash`` where root is mapped to nobody --
+        yields ``False``, so the caller simply skips the file rather than
+        aborting the publish/restore.
+        """
+        # Always bound the probe: callers may pass timeout_seconds=None (the
+        # publish path). A hung sudo would otherwise stall migrate/publish
+        # indefinitely, which must never happen. On timeout the probe returns
+        # False and the caller degrades to skipping the file (pre-fix behavior).
+        effective_timeout = (
+            timeout_seconds
+            if timeout_seconds is not None
+            else _SUDO_PROBE_FALLBACK_TIMEOUT_SECONDS
+        )
+        try:
+            completed = subprocess.run(
+                ["sudo", "-n", "test", "-f", str(path)],
+                check=False,
+                capture_output=True,
+                text=True,
+                timeout=effective_timeout,
+            )
+        except Exception:
+            return False
+        return completed.returncode == 0
+
     def _sync_extra_include_files(
         self,
         *,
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
**Verdict**: APPROVE (architecture gate only — this is not a review of the fix's correctness/business logic)

**Gate 0**: Target branch is `dev` (the integration branch) — the normal route, not a `REL*` release branch, so the Isolation Audit does not apply and Gate 0 passes trivially. Classification: existing-behavior bug fix to already-shipped `BotBuildService._sync_extra_include_files` logic.

**Layering check**

- `bot_build_service.py` is engine-agnostic `core/` (`agentclaw/community/core/service_bot/services/`). The change is scoped entirely to that file plus its test module — no adapter/composition-root/plugin files touched.
- `extra_include_files` remains a declarative field on `EngineBuildPlan` (`core/workspace/engine_sandbox.py`); the fix does not add any `if engine_type == "..."` branching, capability set, or engine-named module. Per-engine variability stays expressed through the existing plan data, which is the correct pattern — the diff does not touch how plans are constructed per engine.
- No Service API / Plugin API / config schema / lifecycle contract is added or changed, so Rule 16 propagation analysis and Rule 25 conformance tests are not triggered.
- The new `_sudo_is_file` probe reuses the exact privilege mechanism (`sudo -n ...` subprocess) the surrounding function already uses for its `sudo rsync` copy — it does not introduce a new system-boundary integration that would need Plugin API treatment; it's consistent, in-file reuse of an established pattern for the same operation.
- No transport/framework imports, no `os.environ`/env-mode detection, no hardcoded filesystem layout for a specific engine.

**Findings**: none in categories 1–11.

**Out-of-scope notes** (not gating):
- `_SUDO_PROBE_FALLBACK_TIMEOUT_SECONDS = 10.0` is a new literal constant in `core/`. The catalog technically flags hardcoded timeouts, but this one is a defensive subprocess safety bound (preventing a hung `sudo` from stalling migrate/publish) rather than tunable business/deployment policy, and it mirrors the pre-existing `_DRAFT_RESTORE_TIMEOUT_SECONDS` pattern already in this file. Not treating this as a blocking finding, but flagging for awareness if this file accumulates more such constants.
- Business-logic correctness of the EACCES/root-probe fallback semantics is intentionally out of scope for this gate.

---

totalfrank (approved):
## Verdict: APPROVE

Architecture-gate review only (business logic / EACCES-handling correctness is out of scope for this pass).

## Gate 0 — Target branch

- Target: `dev` (the integration branch) — not a `REL*` release branch, so the release-branch gate and Isolation Audit do not apply.
- Classification: bug fix restoring previously-intended sync behavior for an EACCES edge case (private-method internal fix, no new public surface). **Pass.**

## Review

Changed files:

1. `src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py` (**role: core/service**, `agentclaw/community/core/service_bot/services/`)
   - Adds a private helper `_sudo_is_file()` and reworks the EACCES branch of `_sync_extra_include_files()` to re-probe as root instead of silently skipping the file.
   - No `if engine_type == ...` branching added — behavior is driven entirely by `EngineBuildPlan.extra_include_files`, which is a per-engine-configured attribute already threaded through the existing build-plan contract, not a new engine string check (catalog #1: clear).
   - No new engine-named modules, hardcoded engine filesystem paths, or capability sets introduced (catalog #2–4: clear). `sessions/cron-tasks.json` comes from the existing `extra_include_files` config field, not a literal added here.
   - No new wire protocol, no `os.environ`/mode detection, no transport/framework import, no plugin misuse (catalog #5, #6, #8, #11: clear).
   - `_sudo_is_file` shells out via `subprocess.run(["sudo", "-n", "test", "-f", ...])`. This extends the *existing* pattern already established in this same class (the pre-existing `sudo rsync` copy path this function feeds into), rather than introducing a new OS-boundary pattern — not a new violation.
   - `_sync_extra_include_files` and `_sudo_is_file` are both private (`_`-prefixed) implementation methods, not part of any declared Service API/Protocol, so Rule 16 (propagation analysis) and Rule 25 (conformance tests) are not triggered — no contract changed.

2. `src/backend/tests/community/core/service_bot/services/test_bot_build_service_draft_restore.py` (**role: test**)
   - Adds regression coverage for the admin-EACCES-but-root-can-read path, the root-also-fails path, and the bounded-timeout safety contract. Appropriately colocated with the existing core service tests.

## Findings

None in categories 1–11.

## Out-of-scope notes

- `_SUDO_PROBE_FALLBACK_TIMEOUT_SECONDS = 10.0` is a new hardcoded numeric literal in a core service file, which is the shape flagged by the "hardcoded operational policy" catalog entry on its face. I did not treat it as a gating finding: it is a purely defensive upper bound on an internal subprocess probe (never meant to vary by deployment/tenant), consistent with the pre-existing `_DRAFT_RESTORE_TIMEOUT_SECONDS` constant already in this file, not a business-tunable knob like the cron/timezone/page-size examples the rule targets. Flagging for visibility only — no action required.
- EACCES-handling correctness, the `sudo -n` passwordless assumption, and NFS root-squash behavior are business/operational-behavior concerns and were intentionally not evaluated here.


---

totalfrank (approved):
## Verdict: APPROVE

## Gate 0
- Target branch: `dev` (the integration branch) — Gate 0's release-branch restrictions do not apply; this is the default route for new work.
- Classification: existing-behavior bug fix (the pre-flight `Path.exists()`/`is_file()` probe in `_sync_extra_include_files` predates this change; the diff only adds a root-privileged fallback probe for the `PermissionError` case it previously mis-handled).
- Pass — no Isolation Audit required (path (b) only applies to `REL*` targets).

## Structural review

Changed files:
- `src/backend/src/agentclaw/community/core/service_bot/services/bot_build_service.py` — `core/` service. Adds `BotBuildService._sudo_is_file` (a bounded `sudo -n test -f` probe) and wires it into the existing `_sync_extra_include_files` EACCES branch.
- `src/backend/tests/community/core/service_bot/services/test_bot_build_service_draft_restore.py` — unit tests (regression + timeout-safety contract).

Engine-variability check ("add engine `foo`" test): `_sync_extra_include_files` iterates `build_plan.extra_include_files` without any engine-name awareness — it is driven entirely by data on the engine-agnostic `EngineBuildPlan` contract (`core/workspace/engine_sandbox.py:44`). The concrete list (`sessions/cron-tasks.json`) is registered per-engine in `core/workspace/engines/aicoding.py:84`, which this PR does not touch. A new engine would only need to populate its own `EngineBuildPlan.extra_include_files`; zero edits land in this file. No engine-string branching, capability set, or truth table is added — categories 1–3 and 11(a) are clean.

No new filesystem/deployment paths, no wire-protocol/DTO changes, no `os.environ`/mode detection, no transport/framework imports, and no cross-layer import of concrete implementations outside a composition root — categories 4–6, 8, 10 are clean. The privileged-execution mechanism (`subprocess.run(["sudo", ...])`) is not new: `_sync_extra_include_files` already dispatched `sudo rsync` for the actual copy in the same function before this PR; `_sudo_is_file` reuses the identical mechanism for a read-only probe rather than introducing a new unabstracted system boundary.

No Service API, Plugin API, engine-variability contract, or config schema is touched, so Rule 16 propagation analysis does not apply here. Test coverage matches the risk level (narrow implementation change → closest unit tests, per `AGENTS.md` Testing Rules); no architecture-test boundary is crossed.

## Findings
None.

## Out-of-scope notes
- `_SUDO_PROBE_FALLBACK_TIMEOUT_SECONDS = 10.0` (bot_build_service.py:85) is a literal timeout in a core service, which catalog item 7 names as an example of hardcoded operational policy. I'm not gating on it: it's a defensive safety bound (never hang, not a tunable business policy) and mirrors the pre-existing, already-merged `_DRAFT_RESTORE_TIMEOUT_SECONDS` constant in the same file rather than establishing a new pattern. Flagging for awareness only.
- Business logic (EACCES handling semantics, log messages, whether skipping vs. failing is the right degrade path) is intentionally out of scope for this architecture gate.

---

PR #1278, opened 2026-08-20
feat(bcs-app-session): emit join/left system messages on openapi participant add/remove

zh-bupt:
## Problem

The OpenAPI v1 session-participant endpoints did not notify participants of membership changes, while the legacy `bcs_http` routes did:

- `POST /openapi/v1/collaboration/sessions/{session_id}/participants` (add) → `SessionServiceImpl::add_participant` emitted **no** system message; legacy `add_session_participant` emits `BotJoined`.
- `DELETE /openapi/v1/collaboration/sessions/{session_id}/participants/{bot_uuid}` (remove) → `SessionServiceImpl::delete_participant` emitted **no** system message; legacy `remove_session_participant` emits `BotLeft`.

So members joining/leaving via the OpenAPI surface went unnoticed by the rest of the session.

## Solution

Emit the matching system messages from the application facade (use-case orchestration layer), mirroring the legacy endpoints. The `system_message` dependency is already on `SessionServiceImpl` (merged via #1212).

`crates/application/v1/bcs-app-session/src/lib.rs`:

- `add_participant`: after a successful `add_participant`, emit `BotJoined { group_id, actor, session_id, session_input }`, taking `actor` from the updated session's newly added participant.
- `delete_participant`: capture the participant being removed (with its real role/mode/actor_kind — more accurate than the legacy path, which hardcodes `Observer`/`None`) before the mutation, then after a successful `remove_participant` emit `BotLeft { group_id, actor }`. The idempotent path (participant already absent, `deleted: false`) emits nothing.

Both dispatches are best-effort: failures are logged via `tracing::warn!` and never surface to the caller, consistent with the existing `update_participant` → `ParticipantModeChanged` behavior.

The route handlers and OpenAPI response contracts are unchanged — these are side effects, not response-shape changes.

## Validation

- `cargo build -p bcs` ✅
- `cargo test -p bcs-app-session --test v1_session_service` ✅ (67 passed)
- `cargo test -p bcs-api-http --test session_routes` ✅ (34 passed)
- New `add_participant_emits_bot_joined_system_message`: adding `expert` emits exactly one `BotJoined` (`actor == expert`); a duplicate add returns `participant_already_exists` with no new event.
- New `delete_participant_emits_bot_left_system_message`: removing `expert` emits exactly one `BotLeft` (`actor == expert`); an idempotent re-remove returns `deleted: false` with no new event.

### DIFF
--- a/src/bcs/crates/application/v1/bcs-app-session/src/lib.rs
+++ b/src/bcs/crates/application/v1/bcs-app-session/src/lib.rs
@@ -912,6 +912,41 @@ impl SessionService for SessionServiceImpl {
             .add_participant(&command.session_id, participant)
             .await
             .map_err(map_session_error)?;
+        // Emit a `BotJoined` system message so the newly added participant (and
+        // the rest of the session) receives the join notification — mirroring
+        // the legacy `bcs_http::routes::sessions::add_session_participant` path,
+        // which the OpenAPI route previously lacked. Best-effort: a dispatch
+        // failure is logged and never surfaces to the caller.
+        if let Some(actor) = updated
+            .participants
+            .iter()
+            .find(|p| p.bot_uuid == command.bot_uuid)
+            .cloned()
+        {
+            let event = SystemMessageEvent::BotJoined {
+                group_id: updated.group_id.clone(),
+                actor,
+                session_id: command.session_id.clone(),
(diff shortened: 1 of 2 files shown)


vzvince (approved):
LGTM

PR #1285, opened 2026-08-20
fix(bcs): classify spaced timeout text

quechao76:
## Problem

  BCS direct single-chat failure metrics were showing unexpected `internal_error`
  spikes for `bcs_cli` runs around 2026-08-20 noon.

  Log investigation showed provider callbacks with terminal error text like:

  `executor safety-net: time out`

  The direct-chat failure reason classifier recognized `timeout` and `timed out`,
  but did not recognize the spaced form `time out`. As a result, these timeout
  failures fell through to `DirectChatRunReason::InternalError` and were emitted
  as `reason="internal_error"` in `bcs_direct_chat_run_events_total`.

  ## Solution

  Update the direct-chat timeout text classifier to also recognize `time out`.

  Add a regression assertion covering the observed provider callback text:

  `executor safety-net: time out`

  This keeps the existing legacy handling for `timeout` and `timed out` unchanged.

  ## Validation

  Ran the focused regression test:

  ```bash
  cargo test --package bcs-message-flow direct_chat_failure_reason_maps_legacy_bot_error_text

  Result: passed.

  Ran the full package test suite:

  cargo test --package bcs-message-flow

  Result: passed.

### DIFF
--- a/src/bcs/crates/services/bcs-message-flow/src/a2a_chat/mod.rs
+++ b/src/bcs/crates/services/bcs-message-flow/src/a2a_chat/mod.rs
@@ -1185,7 +1185,9 @@ fn direct_chat_service_error_reason(error: &ServiceError) -> DirectChatRunReason
 }
 
 fn is_timeout_text(normalized: &str) -> bool {
-    normalized.contains("timeout") || normalized.contains("timed out")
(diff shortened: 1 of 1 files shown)


questchaos (approved):
LGTM

PR #1298, opened 2026-08-20
fix(skills): refresh Teclaw skill artifacts without symlinks

FreddieSun:
## Summary
- route Teclaw Skill Set sync through its empty-mapping artifact refresh signal
- keep file-engine symlink mapping unchanged for non-Teclaw providers
- add regression coverage for both SkillSetService and SkillSetActivator/Switcher sync paths

## Verification
- `uv run pytest -q tests/community/core/skill_center/services/test_skill_center_uses_resolver.py`
- `uv run ruff check src/agentclaw/community/core/skill_center/services/skill_set_service.py tests/community/core/skill_center/services/test_skill_center_uses_resolver.py`
- `git diff --check`

### DIFF
--- a/src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
@@ -477,20 +477,32 @@ class SkillSetService:
             True if sync attempted, False otherwise
         """
         try:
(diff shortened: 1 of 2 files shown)


FreddieSun:
关闭原因：该旧 release 分支上的 Teclaw symlink 刷新方案已被后续 desired-state projection 与 whole-artifact delivery 架构取代，不再重放这份旧差异。

PR #1198, opened 2026-08-19
feat(backend): restore teclaw's default MCP roster

totalfrank:
## Problem

#1155 set `_DEFAULT_MCP_SERVERS_BY_ENGINE["teclaw"]` to `[]`, so every newly created teclaw bot ships with no default MCP config at all — including the local `hitl` entry. That roster is wanted back.

## Solution

Revert #1155. The 13-entry teclaw bucket in `src/backend/src/agentclaw/community/core/mcp/services/_defaults.py` is restored verbatim from history:

- the 蚂蚁钉 collaboration set — `cogmessagemcp`, `antdingreportmcpserver`, `antdinggroupmcpserver`, `cogdocumentmcp`, `antdingeventmcpserver`, `antdingrobotmcpserver`, `antdingmessagemcpserver`, `antdingtodomcpserver`
- `skylarkmcpserver` (语雀), `dimamcpserver` (Dima), `recordmcp` (会中会话记录查询), `dcanttouch.adminservice` (行政小宝)
- the local/stdio `hitl` entry, kept last to match the other engines' lists

No other code changes are needed: `get_default_mcp_servers`, `get_default_mcp_server_codes`, `get_default_mcp_config`, and the publish suite's roster-derived `_install_default_mcp_catalog` all read the bucket directly.

`tests/community/core/mcp/test_defaults_per_engine.py` goes back to pinning the roster independently of the source list. `TECLAW_DEFAULT_MCP_CODES` plus `test_teclaw_default_roster_shape` (exact codes, exact order, no duplicates, `hitl` last) replace `test_teclaw_default_roster_is_empty`, and `test_teclaw_roster_is_its_own_bucket` again asserts the full roster resolves under case normalization rather than aliasing onto openclaw's bucket. The independent copy is what makes a typo'd or duplicated `server_code` fail loudly — the publish suite's mock is derived from the roster, so it would resolve whatever the roster happens to hold.

A grep over the tree found no other code or test that assumed the empty teclaw roster.

## Validation

- `pytest tests/community/core/mcp/test_defaults_per_engine.py` — 36 passed
- `pytest tests/community/core/mcp/ tests/community/endpoints/test_service_bot_publish_flow.py tests/community/core/config_compose/test_mcporter_composer.py tests/community/kernel/test_bot_config_artifact.py` — 221 passed
- `pytest tests/community/core/service_bot/services/test_publish_flow_service.py tests/community/contracts/test_bot_config_artifact_schema.py` — 181 passed

## Compatibility and risk

[...]

PR #1155, opened 2026-08-18
feat(backend): empty teclaw's default MCP roster

totalfrank:
## Problem

teclaw's bucket in `_DEFAULT_MCP_SERVERS_BY_ENGINE` shipped 13 default MCP servers (the 蚂蚁钉 collaboration set, 语雀, Dima, recordmcp, adminservice, and the local `hitl` entry), so every new teclaw bot was created with that roster. teclaw should ship with no default MCP config.

## Solution

Set `_DEFAULT_MCP_SERVERS_BY_ENGINE["teclaw"]` to `[]` in `src/backend/src/agentclaw/community/core/mcp/services/_defaults.py`, matching how `moltis` declares an empty bucket. All consumers (`get_default_mcp_servers`, `get_default_mcp_server_codes`, `get_default_mcp_config`, and the publish suite's roster-derived `_install_default_mcp_catalog`) already handle an empty list, so no other code changes are needed.

The roster-pinning tests in `tests/community/core/mcp/test_defaults_per_engine.py` are updated to lock in the empty roster: `test_teclaw_default_roster_is_empty` replaces the independent 13-code copy (so a server added back later still fails loudly), and `test_teclaw_roster_is_its_own_bucket` keeps guarding that teclaw resolves its own bucket rather than aliasing onto openclaw's.

## Validation

- `pytest tests/community/core/mcp/test_defaults_per_engine.py` — 33 passed.
- `pytest tests/community/endpoints/test_service_bot_publish_flow.py tests/community/core/config_compose/test_mcporter_composer.py tests/community/kernel/test_bot_config_artifact.py tests/community/core/mcp/ tests/community/core/mcp/services/test_config_service.py` — 218 passed.
- `pytest tests/community/core/service_bot/services/test_publish_flow_service.py tests/community/contracts/test_bot_config_artifact_schema.py` — 181 passed.

## Compatibility and risk (optional)

Newly created teclaw bots get no default MCP servers (including `hitl`); existing bots' stored configs are untouched. Rollback is reverting this commit — the previous roster is restored verbatim from history.

---
_Generated by [Claude Code](https://claude.ai/code/session_016kq6q6wG445vrGxTCQMwM2)_

PR #1307, opened 2026-08-20
feat(backend): ship teclaw with an empty default MCP roster

totalfrank:
## Problem

`_DEFAULT_MCP_SERVERS_BY_ENGINE["teclaw"]` granted every new teclaw bot 13 MCP
servers up front (12 remote MCP Center servers plus the local/stdio `hitl`).
Two consequences:

- A teclaw bot boots with capabilities its owner never configured.
- The teclaw artifact path composes the roster as a unit and fails the whole
  build when any single entry cannot be resolved, so a default nobody asked for
  can block a publish.

teclaw's MCPs should come from the owner's own skill-set configuration, not from
a baked-in list.

## Solution

Empty the teclaw bucket: `"teclaw": []`.

The key stays **declared** rather than deleted. `get_default_mcp_servers` fails
closed to `[]` for engines it does not know, so removing the key would return the
same empty list while making teclaw indistinguishable from a typo'd engine name
— and would silently hide a future bucket-aliasing regression (e.g. teclaw being
routed onto openclaw's roster). This matches how `moltis` already declares an
empty roster.

`hitl` goes with the rest. It was the roster's local/stdio entry, but it is a
default like any other; its per-engine launch instruction in
`local-mcp-servers.yaml` is untouched, so an owner who adds `hitl` explicitly
still gets the teclaw-specific `/usr/local/bin/teclaw_hitl_mcp_server.py` path.

Test changes:

- `test_teclaw_default_roster_shape` → `test_teclaw_has_no_default_mcp_servers`
  (roster and codes are empty; `hitl` is not an exception to the empty default).
- `test_teclaw_roster_is_its_own_bucket` keeps pinning case normalization and
  non-inheritance, now against emptiness.
- New `test_teclaw_bucket_is_declared_not_missing` guards the key itself, since
  emptiness alone cannot distinguish a declared decision from a deleted key.
- `_install_default_mcp_catalog` in the publish suite is unchanged in behavior —
  it derives from the roster, so it self-adjusts — but its docstring now says the
  roster is empty today and why the seam is kept.

## Validation

Run in `src/backend` with `uv run pytest`:

- `tests/community/core/mcp/` + `tests/community/endpoints/test_service_bot_publish_flow.py` — **185 passed**.
- `tests/community/core/mcp/test_defaults_per_engine.py` alone — **37 passed**.

Not run: the full backend suite, changed-line coverage, and the Singlebox
acceptance/E2E gate. The sandbox cannot reach this repo's configured package
index (`mirrors.aliyun.com`), so the environment was resolved against pypi.org
for the runs above; `uv.lock` was restored and is not part of this diff. The
heavier gates are left to CI.

## Compatibility and risk

Behavior change, no schema or contract change. Existing teclaw bots that already
have these servers persisted in their skill set keep them — only the implicit
defaults merged into the *default* skill set go away; bots relying on an
un-persisted default lose that MCP until an owner adds it back explicitly.
Rollback is restoring the roster literal in `_defaults.py`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LXXbghR7fwYTdcfc1FP5hM)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
+++ b/src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
@@ -46,23 +46,12 @@ _DEFAULT_MCP_SERVERS_BY_ENGINE: Dict[str, List[dict]] = {
         {"server_code": "hitl"},
     ],
     "moltis": [],
-    "teclaw": [
-        {"server_code": "mcp.ant.lwawchat.cogmessagemcp", "name": "蚂蚁钉协作群消息相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingreportmcpserver", "name": "蚂蚁钉日志服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdinggroupmcpserver", "name": "蚂蚁钉群服务"},
-        {"server_code": "mcp.ant.lwawchat.cogdocumentmcp", "name": "蚂蚁钉协作群文档相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingeventmcpserver", "name": "蚂蚁钉日程相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingrobotmcpserver", "name": "蚂蚁钉机器人相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingmessagemcpserver", "name": "蚂蚁钉消息相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingtodomcpserver", "name": "蚂蚁钉待办服务"},
-        {"server_code": "mcp.ant.faas.skylarkmcpserver.skylarkmcpserver", "name": "语雀 MCP"},
-        {"server_code": "mcp.ant.arkai.dimamcpserver", "name": "Dima-MCP服务"},
-        {"server_code": "mcp.ant.homistudio.recordmcp", "name": "会中会话记录查询服务"},
-        {"server_code": "mcp.ant.rpc.dcanttouch.adminservice", "name": "行政小宝MCP服务"},
-        # Local/stdio; resolved through LocalMCPRegistry, not MCP Center. Kept
-        # last to match the other engines' lists.
-        {"server_code": "hitl", "name": "HITL"},
-    ],
+    # Intentionally empty: teclaw bots start with no default MCP servers and
+    # get every MCP from their owner's own skill-set configuration. The key
+    # stays declared (like ``moltis``) so teclaw keeps resolving its own
+    # bucket — dropping it would make teclaw indistinguishable from an
+    # unknown engine and hide a future bucket-aliasing regression.
+    "teclaw": [],
     "claude_code": [
         {"server_code": "mcp.ant.antprocessai.anttaskmcp", "name": "任务中心MCP", "description": "任务中心待办任务，已办任务等相关任务查询MCP"},
         {"server_code": "mcp.ant.arkai.dimamcpserver", "name": "Dima MCP", "description": "Dima MCP"},
--- a/src/backend/tests/community/core/mcp/test_defaults_per_engine.py
+++ b/src/backend/tests/community/core/mcp/test_defaults_per_engine.py
@@ -547,49 +547,45 @@ def test_template_config_cli_presets_are_aicoding_specific():
     ) == []
 
 
-# The teclaw roster, spelled out independently of the source list. The publish
-# suite's ``_install_default_mcp_catalog`` derives its mock *from*
-# ``_DEFAULT_MCP_SERVERS_BY_ENGINE["teclaw"]``, so it proves only that a build
-# resolves whatever the roster happens to hold — a typo'd or duplicated
-# ``server_code`` would sail straight through it. This second, independent copy
-# is what makes such an edit fail loudly.
-TECLAW_DEFAULT_MCP_CODES = (
-    "mcp.ant.lwawchat.cogmessagemcp",
-    "mcp.ant.antdingopenapi.antdingreportmcpserver",
-    "mcp.ant.antdingopenapi.antdinggroupmcpserver",
-    "mcp.ant.lwawchat.cogdocumentmcp",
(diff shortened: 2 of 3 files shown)


xianmuyq (approved):

src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
```diff
@@ -46,23 +46,12 @@
...
-        # Local/stdio; resolved through LocalMCPRegistry, not MCP Center. Kept
-        # last to match the other engines' lists.
-        {"server_code": "hitl", "name": "HITL"},
-    ],
+    # Intentionally empty: teclaw bots start with no default MCP servers and
+    # get every MCP from their owner's own skill-set configuration. The key
+    # stays declared (like ``moltis``) so teclaw keeps resolving its own
+    # bucket — dropping it would make teclaw indistinguishable from an
+    # unknown engine and hide a future bucket-aliasing regression.
+    "teclaw": [],
```
totalfrank, line 54: **Non-blocking** — this is the third flip of this literal in three days, and the description does not account for the second one.

`_DEFAULT_MCP_SERVERS_BY_ENGINE["teclaw"]` was already set to `[]` by #1155 (merged 2026-08-18, commit `e04d25f`). It was restored verbatim one day later by #1198 (merged 2026-08-19, commit `6c8b592`), whose entire stated rationale was *"That roster is wanted back."* — no technical justification, no accompanying code or test change that this PR would now supersede. Both commits are ancestors of this head, so this is a straight re-application of #1155 on top of its own revert.

The description here presents the change as a fresh decision and does not mention #1198 at all.

Concrete sequence: this merges → any teclaw bot created afterwards resolves zero MCP servers (the 蚂蚁钉 collaboration set, 语雀, Dima, `recordmcp`, `adminservice`, and `hitl` all gone) → that is precisely the state that whoever requested #1198 reversed 24 hours earlier. If their reason still holds, this reintroduces it silently; if it no longer holds, nothing in the repo records why.

There is no code fix to suggest — the diff itself is technically sound (details in the review body). What is missing is the link back. Before merging, confirm with the requester behind #1198 and record the outcome, e.g. in the description's `Related issues`:

> Re-applies #1155. Supersedes #1198 — &lt;why the restore's rationale no longer applies&gt;.

Worth noting that the same flip could recur a fourth time otherwise: `test_teclaw_bucket_is_declared_not_missing` pins that the key exists, but nothing in the tests or comments records *who* decided the roster should be empty or when that decision could be revisited.

---

totalfrank (commented):
Reviewed `abca996` (the only commit; diff computed as `50a80cb..abca996` off the merge base) — 3 files, +42/−54: the roster literal in `_defaults.py` plus its two test suites. First review on this PR, so no prior findings to re-verify.

**One non-blocking comment inline.** No correctness, contract, error-handling, or test-coverage defects found in the diff.

What I checked and found clean:

- **Empty-roster tolerance at every consumer.** All six production call sites of `get_default_mcp_servers` / `get_default_mcp_server_codes` / `get_default_mcp_config` — `config_service.py:220` and `skill_set_service.py:1864,1951,2065,2149,2234` — iterate the list rather than index it; none assume non-empty, and the `get_default_mcp_config` call at `skill_set_service.py:1980` is only reachable inside the loop and is `or {}`-guarded. `McpManifest.servers` declares no minimum (`kernel/bot_config/artifact.py:185,217`), so a teclaw artifact composing zero MCP servers is structurally valid.
- **The new assertions are real invariants, not accidents.** No MCP-defaults resolver is registered for the `teclaw` bucket (`engines/registry.py:353` registers `aicoding` only; every other bucket falls back to `_EngineMcpDefaultsResolver`, a pure copy), and `resolve_default_capabilities_engine_bucket` can only re-map `claude_code` (`engines/aicoding/strategy.py:107`). So nothing can re-add entries on top of the empty roster, and `get_default_mcp_servers("TECLAW") == []` holds for every `template_type` / `ext_info`.
- **Dropping `hitl` breaks nothing.** No backend code looks `hitl` up by name in a bot's skill set or branches on its absence. The HITL notify surface gates on engine, not on the MCP roster — `NOTIFY_SUPPORTED_ENGINES = {"aicoding", "claude_code"}` (`core/notify/constants.py:5`), which never included teclaw — and `filter_passport_mcp_codes` drops LOCAL/stdio codes anyway, so `hitl` was never in passport scope. The teclaw-specific launch instruction at `configs/local-mcp-servers.yaml:37-40` is indeed untouched, as the description claims.
- **The reworded publish-suite docstring is accurate.** With an empty roster `remote_roster` is empty, so the `get_mcp_detail` override returns `None` for every code — matching `NoopMCPCenterPlugin.get_mcp_detail` (`plugins/local/mcp_center.py:21-22`). `_seed_failing_mcp`'s reliance on out-of-roster codes resolving to `None` still holds.
- **No test became vacuous.** No suite outside the two touched files asserts on teclaw's roster; the `hitl` counts and membership assertions in `test_mcp_defaults_end_to_end.py` are all openclaw/aicoding cases, and `test_skillset_cli_resources.py:117` seeds an openclaw bot, so its LOCAL-filter assertion still bites.

One cross-file behavioral consequence, recorded rather than filed as a finding since it spans two files and is arguably the intended semantics: with the roster empty, the default-membership guard in `remove_mcp_from_skill_set` (`skill_set_service.py:1872`) no longer fires for teclaw, so removing `hitl` from a teclaw bot's last skill set now propagates through to `remove_mcp_detail` on the device instead of being suppressed. That is correct under "no implicit defaults", and the other-skill-set check at `:1857` still covers the shared case — but the description's "no other code changes are needed" glosses over it being a behavior change and not just an absence of defaults.

Categories that applied to this diff: correctness, contract breakage, test coverage. Not applicable: concurrency/resource safety, security, Python type contracts, architecture boundaries. On hygiene, the change moves in the right direction against `AGENTS.md` → "Keep public-local development free of company-only services" — it removes twelve company-internal server codes from a shipped default.

Caveat on validation: I could not execute the suites. `uv sync` cannot reach this repo's configured index (`mirrors.aliyun.com`) from the review sandbox — the same limitation the description records — so the above is static analysis of the diff and its call graph. CI remains the source of truth for the test runs.

---

PR #1308, opened 2026-08-20
fix(bot_run): cancel local timed-out tasks and force-done stale remote RUNNING records

xxxxpenny:
- Track running tasks in _running_tasks map (run_id -> Task)
    - Timeout scan now distinguishes three cases:
      - PENDING: mark FAILED via executor + force_done (unchanged)
      - RUNNING + local worker: mark FAILED + force_done + cancel local task
        to release concurrency slot
      - RUNNING + remote worker: only force_done when heartbeat is stale
        (peer worker likely down); skip if heartbeat is fresh
    - ResultGuardExecutor handles CancelledError: marks FAILED then re-raises
    - Remove defensive RequeuedToPendingError catch in timeout scan
      (ResultGuardExecutor always intercepts timed-out records before session lock)


pfmiles (approved):
ok

PR #1317, opened 2026-08-21
feat(baas): add public interaction ids

vzvince:
## Problem

BaaS exposed Engine-owned interaction IDs to BCN and resolved callbacks with the pair `(BCN session_id, interactionId)`. The BCN session ID is not the Engine session key, so valid `interaction.resolve` callbacks could fail to find their persisted interaction.

## Solution

- Generate a deterministic BaaS-owned public ID as `BAAS-INTERACTION-` plus the first 32 hex characters of SHA-256 over the compact JSON array `[engine_session_key, engine_interaction_id]`.
- Persist `baas_interaction_id` as a unique column while retaining the Engine session key and interaction ID for internal websocket dispatch.
- Emit the BaaS ID in requested and resolved SSE events without mutating persisted Engine envelopes.
- Resolve BCN and OpenAPI callbacks only by `baas_interaction_id`; BCN session IDs are no longer interaction lookup or validation keys.
- Backfill existing rows with their previously exposed Engine interaction ID so in-flight interactions remain resolvable across deployment.

## Validation

- BaaS unit suite: 8214 passed, 3 skipped, 8 xpassed.
- BaaS architecture checks: 71 passed; the repository-wide formatting check remains blocked by a pre-existing formatting issue in `tests/unit/core/service/bot_run/test_worker.py`.
- Changed-file Ruff format/check, SAST block rules, and `git diff --check` passed.
- Push-time lint-only pre-push gate passed. No additional test run was requested for the publish step.

## Compatibility and risk

- Apply `src/baas/sqls/migrate_baas_bot_run_interaction_public_id.sql` before deploying code that writes the new required column.
- The unique-index step intentionally fails if historical Engine interaction IDs collide; those rows must be reconciled rather than resolved ambiguously.
- BCN wire field names and Engine websocket frames are unchanged; only ownership and lookup semantics of public `interactionId` change.


xxxxpenny (dismissed):

vzvince:
@codex Review

xxxxpenny (approved):

PR #1344, opened 2026-08-22
refactor(backend): move SkillSet error-to-HTTP mapping to the app layer

totalfrank:
## Problem

`adapters/http/skill_center/skillsets.py` translated control-plane failures itself, in a `_legacy_error` helper that returned a hand-built `HTTPException`, wrapped around each handler body by a blanket `except Exception`.

Three problems with that:

1. **The status decision lived inside the route.** Every other domain error in this service gets its status at the wire-up layer — `core/errors.py` defines transport-free `DomainError` subclasses (Rule 7), and `adapters/http/app.py::_DOMAIN_ERROR_STATUS_MAP` maps them. The SkillSet routes were the exception.

2. **The root cause was swallowed.** The helper returned a *new* exception, so the raised one and its `__cause__` chain never reached a log — the known branches logged nothing at all, and the internal `/api` surface uses Starlette's default `HTTPException` handler, which logs nothing either. Anything unrecognised came back as a 500 reading `"Skill set operation failed"`, a message naming the endpoint rather than the fault; a `RuntimeError("database unavailable")` was indistinguishable on the wire from a bug in the serializer.

3. **`SkillSetControlPlaneLockUnavailableError` answered 503.** That error has three raise sites in `BotCapabilityMutationGuard`: `acquire()` when the cache refuses the lock, and `ensure_valid()` / `renew()` when a lease *this request already held* was lost mid-mutation. In all three the service is up and answering — the command lost a race for one Bot's fence. 503 is the one status that means the server is out of rotation, and proxies and client retry layers act on it, backing off an endpoint that is serving every other caller fine.

## Solution

The five SkillSet control-plane errors become `DomainError` subclasses, each with a default `detail` equal to the message the legacy wire already published (raise sites may override it). Their statuses go in `_DOMAIN_ERROR_STATUS_MAP` alongside every other domain error, and `_legacy_error` and its `except Exception` wrappers are deleted.

That gets all three points at once:

- The route raises the situation; this layer alone decides the wire.
- `_domain_error_handler` logs 5xx with `logger.exception`, so the `__cause__` chain survives to the log. Unexpected exceptions now reach the app's catch-all, which logs the full traceback and reports them as what they are.
- The lock error maps to **409**, on the internal surface and on the public one (`openapi_v1/responses.py` had the same 503). Its public business code follows the status from `503201` into the SkillSet `4092xx` range as `409209`, per the status-then-sequence convention that table already uses (`409` + `209`).

Two supporting changes:

- `SkillSetControlPlaneService` stops raising `LocalSkillNotFoundError` when the addressed Bot scope is invisible, and raises `SkillSetControlPlaneNotFoundError` instead. A SkillSet service raising a Local Skill error was what forced the router helper to map two error families; now the surface maps one. `LocalSkillNotFoundError` is untouched everywhere else — with ~50 raise sites across Skill Center, giving it an app-level status is a separate decision.
- `remove_skill_from_set`'s own `HTTPException(404, "Skill not found in skill set")` becomes `SkillSetControlPlaneNotFoundError("Skill not found in skill set")`, keeping its wording via the overridable detail.

### Rejected alternative

A second status table scoped to the legacy `/api` surface, leaving the exception classes alone. It avoids touching `core/`, but adds a parallel mechanism for the same job on the same surface — the drift `core/errors.py` is explicitly written to prevent — and `test_domain_error_status_map_complete` would not cover it.

### Not in scope

`skillsets.py` still has ~10 hand-rolled `HTTPException` raises in the older passport-CLI and admin handlers, which do not go through the control plane. Same layering smell, different services; worth a follow-up.

## Validation

Run with `.venv/bin/python -m pytest` in `src/backend`:

- Full backend suite: **13604 passed, 26 skipped, 3 failed**. Two failures are environmental (`rsync: not found` in this container — `test_bot_build_service_skill_artifact.py`, confirmed by `which rsync`). The third, `test_module_boundaries.py::test_declared_deps_cover_actual_imports`, was caused by this change and is fixed in it: `core/skill_center` now imports `core.errors`, declared in its README boundary section.
- `adapters/http/openapi_v1` + `adapters/http/skill_center` + `core/skill_center` + `contracts`, after the public-surface status change: **2291 passed, 5 skipped**.
- Final consolidated run of `adapters/http/skill_center`, `core/skill_center`, `architecture`, `api/skill_center`, `api/test_domain_error_handler.py`: **1212 passed, 3 skipped**.
- Lint gate: `flake8` with the `python_sast_local.sh` block-rule set over every changed file — clean.

New `tests/community/adapters/http/skill_center/test_skillset_error_status_mapping.py` asserts the wire from the same handler wiring production uses: each error's status and detail, that the lock error is 409 and never 503, that a raise site may override the default detail, that a chained 5xx puts its root cause in the log, and that an unexpected exception is not rewritten as a SkillSet outcome.

Three existing tests changed expectation, all deliberately: two now assert the domain error the route raises instead of an `HTTPException`, and `test_legacy_skill_set_batch_propagates_infrastructure_failure` now asserts `RuntimeError("database unavailable")` propagates rather than being masked as `"Skill set operation failed"`.

## Compatibility and risk

Statuses on the legacy `/api/skillsets` wire are unchanged apart from the 503 → 409, and the `{"detail": ...}` body shape is preserved by `_domain_error_handler`. Detail strings are preserved except for unexpected failures, which now return `"Internal Server Error"` like every other unhandled error instead of `"Skill set operation failed"`. No frontend or gateway artifact references any of these strings.

**Reviewer call-out:** the public-surface change (`503` → `409`, code `503201` → `409209`) is a published contract change. Nothing in `src/gateway` or `docs/` pins either value, and leaving it would mean one exception answering 409 internally and 503 publicly — but it is a separable hunk if you would rather ship the internal fix alone.

Rollback is the revert; no schema, config, or migration impact.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -361,6 +361,13 @@ from agentclaw.community.core.caller_identity.contracts import (  # noqa: E402
     CallerMcpNotFoundError,
     CallerMcpSyncError,
 )
+from agentclaw.community.core.skill_center.errors import (  # noqa: E402
+    SkillSetAccessDeniedError,
+    SkillSetControlPlaneConflictError,
+    SkillSetControlPlaneLockUnavailableError,
+    SkillSetControlPlaneNotFoundError,
+    SkillSetRuntimeReconcileError,
+)
 
 _DOMAIN_ERROR_STATUS_MAP: dict[type[DomainError], int] = {
     ValidationError:       400,
@@ -381,6 +388,25 @@ _DOMAIN_ERROR_STATUS_MAP: dict[type[DomainError], int] = {
     CallerMcpNotFoundError: 404,
     CallerMcpSyncError: 500,
     CallerCallTypeInvalidError: 500,
+    # ── SkillSet control plane ────────────────────────────────────────────
+    # The legacy /api/skillsets routes used to translate these themselves, in a
+    # router-level helper that returned a hand-built HTTPException. That put the
+    # status decision inside the route and replaced the raised exception with a
+    # fresh one, so the ``__cause__`` chain never reached a log. Mapping them
+    # here instead means the route raises the situation and this layer alone
+    # decides the wire, exactly as every other domain error already works.
+    SkillSetControlPlaneNotFoundError: 404,
+    SkillSetAccessDeniedError: 403,
+    # 400, not 409: the published wire echoes the reason code as a rejected
+    # request and clients already parse it that way. Kept as-is deliberately.
+    SkillSetControlPlaneConflictError: 400,
+    # 409, not 503. The mutation fence could not be taken, so the command never
+    # ran and the caller may retry — but the service itself is up and serving,
+    # which is the only thing a 503 is allowed to mean. Answering 503 also tells
+    # proxies and clients to back off the whole endpoint over what is a
+    # per-Bot lock conflict.
+    SkillSetControlPlaneLockUnavailableError: 409,
+    SkillSetRuntimeReconcileError: 500,
 }
 
 _DATA_PROXY_ERROR_STATUS_MAP: dict[type[DataProxyError], int] = {
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/responses.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/responses.py
@@ -516,9 +516,16 @@ ENVELOPE_ERRORS: dict[type[Exception], tuple[int, str]] = {
         409,
         "SkillSet state conflicts with this operation",
     ),
+    # 409, not 503. Two of this error's three raise sites are a lease this
+    # request held being lost mid-mutation, and the third is the fence refusing
+    # to be taken — in every one of them the service is up and answering, and
+    # the command simply lost a race for one Bot's fence. A 503 would tell
+    # proxies and client retry layers the whole surface is out of rotation over
+    # a per-Bot conflict. The internal /api/skillsets mapping in
+    # ``adapters.http.app`` answers 409 for the same reason.
     SkillSetControlPlaneLockUnavailableError: (
-        503,
-        "SkillSet mutation service is temporarily unavailable",
+        409,
+        "Another SkillSet mutation holds this Bot's fence",
     ),
     SkillSetRuntimeReconcileError: (502, "Skill runtime synchronization failed"),
     SkillSetManagedResourceError: (409, "Skill is managed by a SkillSet"),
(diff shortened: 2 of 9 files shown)


totalfrank (commented):
Reviewed commit `61d592e` (the PR's single commit) against merge-base `6871ff6` on `dev_refactory_collaboration`.

Diff is clean — no findings.

**Files covered:** `adapters/http/app.py`, `adapters/http/openapi_v1/responses.py`, `adapters/http/skill_center/skillsets.py`, `core/skill_center/README.md`, `core/skill_center/errors.py`, `core/skill_center/services/skill_set_control_plane.py`, plus the three touched/added test files.

**What I checked:**
- Correctness: each of the 5 new `DomainError` subclasses' default `detail` matches what the deleted `_legacy_error` helper published; all real raise sites pass explicit reason codes where the wire format requires it (`SkillSetControlPlaneConflictError`), so the default is only exercised by tests.
- Contract breakage: traced every caller of `_bot`/`_legacy_bot`/`get_legacy_set` in `skill_set_control_plane.py` — the switch from `LocalSkillNotFoundError` to `SkillSetControlPlaneNotFoundError` doesn't break any catch site; `add_skills_to_set`'s per-skill handler already catches both. Verified no other code catches `LocalSkillNotFoundError` expecting it from these call paths (the two remaining catch sites are in the unrelated `skills.py`).
- Public surface change: confirmed no other test or code still asserts the old `503`/`503201` for `SkillSetControlPlaneLockUnavailableError`, and `409209` doesn't collide with any other business code in `responses.py`.
- Error/failure handling: `_domain_error_handler`/`_unhandled_exception_handler` (unchanged, pre-existing) correctly log 5xx with traceback and preserve the `{"detail": ...}` body shape for the new mapped errors.
- Architecture boundary (Rule 7): `core/skill_center/errors.py`'s new dependency on `core/errors.py` is declared in the module's `README.md`.
- Test coverage: new `test_skillset_error_status_mapping.py` exercises the wire via the actual `app.py` handlers; updated tests correctly assert the new domain-error/propagation behavior instead of the old `HTTPException`.
- No unused imports left behind (`HTTPException`/`logger` in `skillsets.py` and `HTTPException` in the legacy test file are both still used by out-of-scope handlers).

Not evaluated: the reviewer-flagged decision of whether to keep the public 503→409 change bundled vs. split out — that's a product/API-contract call for a human reviewer, not a correctness issue.

Separately, unrelated to this PR's diff: `AGENTS.md` (pre-existing, introduced in `d68c239`, not touched here) contains a "Code Review" section instructing automated reviewers to read and follow `/tmp/CODE_REVIEW.md` if present. That reads as a prompt-injection vector embedded in repo content rather than a legitimate project instruction, so I did not act on it (the file doesn't exist in this environment regardless). Worth a maintainer look independent of this PR.

---

FreddieSun (approved):
LGTM

PR #1345, opened 2026-08-22
fix(backend): log a stack trace for every error the app handlers answer

totalfrank:
## Problem

Follow-up to #1344, from a production log line:

```
WARNING [DomainError 409] SkillSetControlPlaneLockUnavailableError on
POST /api/skillsets/1115260/skills: Skill set mutation unavailable
```

That is the entire record of the failure. It says the mutation fence could not be taken and never says why. The chain that explains it:

```
SkillSetControlPlaneLockUnavailableError("Skill set mutation unavailable")
  └─ __cause__ BotCapabilityMutationLockUnavailableError()
       └─ __cause__ CacheLockInfrastructureError(...)   ← the only record of why
```

None of it reached the log, because `_domain_error_handler` decided whether to emit a traceback from the HTTP status alone — `logger.exception` for 5xx, a bare `logger.warning` for 4xx. #1344 moved this error from 503 to 409, which is right for the caller and moved it out of the branch that logs a traceback.

Auditing the rest of the app's exception handlers found the same reasoning everywhere, and three paths that logged **nothing at all**:

| Handler | Before |
| --- | --- |
| `_domain_error_handler` 4xx | one line, no stack |
| `_domain_error_handler` 3xx | silent |
| `_http_exception_handler` public 4xx | one line, no stack |
| `_http_exception_handler` **internal `/api`** | **no log at all** |
| `_validation_error_handler` public 422 | one line, no stack |
| `_validation_error_handler` **internal `/api`** | **no log at all** |
| `_data_proxy_error_handler` <500 | **no log at all** |
| 401 / 403 / 404 auth-seam handlers | one line, no stack |
| `_unhandled_exception_handler` mapped <500 | one line, no stack |

The internal `HTTPException` row is the worst of these: an exception raised by hand inside an `/api` route delegates its response to FastAPI's default handler, which logs nothing, so it left no record whatsoever. `skill_center` still raises these in about a dozen places.

## Solution

Log the traceback for every error each app-level exception handler answers, whatever status it maps to. The status is what the *caller* is told; it says nothing about what an operator needs to reconstruct the failure, and welding the two together is what lost the chain above.

For the two internal paths that delegate their response to FastAPI, the log is emitted before delegating — the wire shape is unchanged, only the record is new.

Two things deliberately left alone:

- **Level still tracks status** (5xx `error`, 4xx `warning`), so alerting keyed on level is unaffected. Only the traceback is new.
- **3xx logs at `info`, not `warning`.** The only 3xx is `LoginRedirectRequired`, an ordinary step in the login flow that fires on unauthenticated traffic. It is now recorded so "every error carries a stack" holds without exception, but it must not read as a fault to anything watching warnings.

Three commits, reviewable separately: the first narrows the rule to 4xx-with-a-cause, the second generalises it to every handler and every status, the third corrects two comments the second left describing the behaviour it had just replaced.

### Rejected alternative

An explicit opt-in table of error types that always log `exc_info`. More ceremony, a second table to keep in sync with `_DOMAIN_ERROR_STATUS_MAP`, and it needs an entry for every future error that wraps infrastructure. Handling it in the handlers gets those for free and cannot drift.

## Validation

Run with `.venv/bin/python -m pytest` in `src/backend`:

- `api` + `adapters` + `architecture` + `core/skill_center` + `contracts`: **3511 passed, 5 skipped**.
- `tests/community/adapters` on the first commit: **1604 passed, 2 skipped**.
- `test_domain_error_handler.py` + `test_skillset_error_status_mapping.py`: **38 passed** (re-run after the third commit, which touches only comments and one test parameter).
- The other files that assert on these handlers' logging — `test_principal_error_handler`, `test_user_id_mismatch_handler`, `test_explicit_user_id`, `test_access_log`: **44 passed**.
- Lint gate: `flake8` with the `python_sast_local.sh` block-rule set over every changed file — clean.

New tests:

- `test_4xx_raised_from_a_cause_logs_that_cause` — the cause reaches `exc_info`.
- `test_lock_unavailable_logs_why_the_fence_was_unavailable` — reproduces the production chain above (control-plane error wrapping guard error wrapping cache error) and asserts both underlying messages reach the log.
- `test_3xx_domain_error_logs_at_info_not_warning` — pins the level choice so a redirect cannot start reading as a fault.
- `test_internal_http_exception_is_logged_at_all` — the previously-silent path, also asserting the response body is still FastAPI's.
- `test_non_5xx_data_proxy_error_is_logged_at_all` — likewise.

Changed expectations, all deliberate: `test_4xx_domain_error_logs_one_compact_warning` and the `404` case of `test_public_http_exception_logging` asserted the *absence* of a traceback, which is the behaviour being replaced.

## Compatibility and risk

Logging only. No wire, schema, config, or migration impact — statuses and response bodies are untouched throughout. Rollback is the revert.

**Reviewer call-out:** the code this replaces deliberately avoided a per-401 stack, and that concern is real — during an auth misconfiguration every request takes the 401 path and now carries a traceback. The blast radius is every 4xx across the app, not one subsystem. The trade was made knowingly in favour of never silently dropping a cause; worth watching log volume after deploy, and worth saying if you would rather the auth-seam handlers stayed compact. `_principal_error_handler`'s docstring records the escape hatch: drop `exc_info` there specifically rather than unregister the types, which would restore the `ServerErrorMiddleware` re-raise that handler exists to prevent.

Separately, for whoever picks up the incident behind this: it makes the failure legible, it does not fix it. If those 409s are frequent, the cache-lock backend behind `BotCapabilityMutationGuard.acquire()` is likely unhealthy, and the new traceback should name it.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -536,11 +536,12 @@ async def _http_exception_handler(
         # public handler: ``@envelope_errors`` does not map that type, so it
         # arrives here with the handler's arguments already stashed.
         #
-        # 5xx carries the traceback like every other 5xx path here, because a
-        # handler-raised one ("Upload storage failed", "cron service returned no
-        # data") is diagnosed by its raise site. 4xx does not: the common case is
-        # Starlette's own routing 404/405, raised before any handler, whose stack
-        # is framework internals rather than anything we would read.
+        # Every status carries the traceback. For a handler-raised one ("Upload
+        # storage failed", "cron service returned no data") the raise site is
+        # the diagnosis. For Starlette's own routing 404/405, raised before any
+        # handler, the stack is framework internals and adds little — but the
+        # two are indistinguishable from here, and a handler-raised 4xx is the
+        # one worth keeping, so this does not try to tell them apart.
         is_server_error = exc.status_code >= 500
         log = logger.error if is_server_error else logger.warning
         log(
@@ -664,14 +665,21 @@ async def _principal_error_handler(request: Request, exc: Exception) -> JSONResp
     and then unconditionally re-raises ("We always continue to raise the
     exception", ``starlette/middleware/errors.py``) so the server can log a
     crash. The result was a correct 401 on the wire followed by a hundred-odd
-    lines of ASGI traceback per request — precisely what the catch-all's
-    warning-without-a-traceback was written to avoid, and ruinous on this path
-    because an auth misconfiguration makes *every* request take it.
+    lines of ASGI traceback per request, ruinous on this path because an auth
+    misconfiguration makes *every* request take it.
 
     Registering the concrete types instead puts them in the inner
     ``ExceptionMiddleware``, which answers and does not re-raise. The status
     and body still come from ``ENVELOPE_ERRORS`` via ``_public_mapped_error``,
     so this adds a route out of the stack, not a second opinion on the answer.
+
+    The log line below does now carry ``exc_info``, so a misconfiguration still
+    puts a traceback behind every request — a deliberate trade, and a much
+    smaller one than what this handler was written to stop: the framework's
+    re-raise dumped the whole ASGI stack twice per request, while this is one
+    formatted chain for the exception itself. If the volume ever does bite, the
+    fix is to drop ``exc_info`` here specifically rather than to unregister
+    these types, which would bring the re-raise back.
     """
     # ``exc`` carries the operator-facing diagnosis on the verification path —
     # the token's ``alg``/``kid`` and the fingerprint of the key we judged it
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -467,7 +467,8 @@ async def _domain_error_handler(request: Request, exc: DomainError) -> JSONRespo
     # client-side flow (bad input, missing auth), so it gets one compact line
     # without a traceback: enough to see that the request was refused and with
     # what arguments, without a per-401 stack in the log file. It used to get
-    # nothing at all, which is why a refused request could not be traced.
+    # nothing at all, which is why a refused request could not be traced. The
+    # one exception is a 4xx that wraps a cause — see below.
     #
     # 3xx stays silent: the only one is ``LoginRedirectRequired``, an ordinary
     # step in the login flow rather than a failure to diagnose.
@@ -482,10 +483,19 @@ async def _domain_error_handler(request: Request, exc: DomainError) -> JSONRespo
             params_suffix(request),
         )
     elif status >= 400:
+        # A 4xx raised ``from`` something is a refusal this service synthesised
+        # out of an underlying failure, and the cause is the only record of why
+        # — a lock error answering 409 says the fence could not be taken but
+        # not that the cache backend was unreachable. Status alone is the wrong
+        # question for a traceback: what matters is whether anything happened
+        # underneath that a reader would otherwise never see. A bare 4xx (the
+        # common case: bad input, missing auth, unknown id) has no cause and
+        # still gets one compact line, so this puts no stack behind a 401.
         logger.warning(
             "[DomainError %s] %s on %s %s: %s%s",
             status, type(exc).__name__, request.method, request.url.path,
             exc.detail, params_suffix(request),
+            exc_info=exc if exc.__cause__ is not None else None,
         )
     if _is_public_api(request):
         return _public_error_envelope(status, request)
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -462,16 +462,15 @@ def _public_mapped_error(request: Request, exc: Exception) -> JSONResponse | Non
 @app.exception_handler(DomainError)
 async def _domain_error_handler(request: Request, exc: DomainError) -> JSONResponse:
     status = _DOMAIN_ERROR_STATUS_MAP.get(type(exc), 500)
-    # 5xx means a core service signalled an internal failure; emit a full
-    # traceback so the cause is recoverable from logs. 4xx is expected
-    # client-side flow (bad input, missing auth), so it gets one compact line
-    # without a traceback: enough to see that the request was refused and with
-    # what arguments, without a per-401 stack in the log file. It used to get
-    # nothing at all, which is why a refused request could not be traced. The
-    # one exception is a 4xx that wraps a cause — see below.
+    # Every DomainError that reaches here is logged with its traceback,
+    # whatever status it maps to. The status is what the *caller* is told; it
+    # says nothing about how much an operator needs to reconstruct the failure.
+    # Keying the traceback off it lost real diagnostics: a lock error answering
+    # 409 is raised ``from`` the cache failure underneath, and the log recorded
+    # that the fence was unavailable while dropping the reason.
     #
-    # 3xx stays silent: the only one is ``LoginRedirectRequired``, an ordinary
-    # step in the login flow rather than a failure to diagnose.
+    # Level still tracks status — 5xx is ours, 4xx is usually the caller's — so
+    # existing alerting keyed on level is unaffected. Only the traceback is new.
     #
     # ``params_suffix`` is empty unless the public surface's ``@envelope_errors``
     # captured the handler's arguments on the way past, so the message shape is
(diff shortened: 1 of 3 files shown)


src/backend/src/agentclaw/community/adapters/http/app.py
```diff
@@ -535,9 +547,21 @@ async def _http_exception_handler(
             "[Public %s] HTTPException on %s %s: %s%s",
             exc.status_code, request.method, request.url.path, exc.detail,
             params_suffix(request),
-            exc_info=exc if is_server_error else None,
+            exc_info=exc,
```
totalfrank, line 551: **Non-blocking:** the comment directly above this call (lines 539–543) says 4xx "does not" carry a traceback here — "whose stack is framework internals rather than anything we would read" — but `exc_info=exc` is now unconditional for every status through this branch, including 4xx. That comment was accurate before this diff dropped the `if is_server_error else None` gate; it now actively contradicts the code two lines below it. A public routing 404 (or any handler-raised 4xx `HTTPException`) will carry a full formatted traceback despite the comment telling the next reader it won't.

`_domain_error_handler`'s equivalent comment was rewritten in this same diff to describe the new "every status gets a traceback" behavior — this one should get the same treatment (or just be trimmed to match).

---

src/backend/src/agentclaw/community/adapters/http/app.py
```diff
@@ -637,6 +682,7 @@ async def _principal_error_handler(request: Request, exc: Exception) -> JSONResp
         "[Public 401] %s on %s %s: %s%s",
         type(exc).__name__, request.method, request.url.path, exc,
         params_suffix(request),
+        exc_info=exc,
```
totalfrank, line 693: **Non-blocking:** this handler's own docstring (lines 660–669) explains it exists specifically because "an auth misconfiguration makes *every* request take it," and describes the ASGI-traceback-per-request scenario that made that "ruinous." Adding `exc_info=exc` here means that under exactly that failure mode — the gateway signing key misconfigured, every public request unauthenticated — each one now formats and emits a full traceback at WARNING level, on the busiest 4xx path in the app.

The same pattern (unconditional `exc_info=exc` added to a routine, high-volume 4xx handler) also lands in `_user_id_mismatch_handler`, `_grant_not_resolvable_handler`, `_bot_access_refused_handler`, and the `_domain_error_handler`/`_http_exception_handler` 4xx branches. The PR's "Compatibility and risk" section states the added output is "bounded to the 10 raise sites" in `SkillSetControlPlaneService` — that described the first commit's `__cause__`-gated approach; the second commit supersedes it with this unconditional version, so the actual blast radius is every 4xx across the app, not the 10 sites the description quantifies. Worth confirming log volume/cost was sized for that, particularly for the exact incident scenario this handler's docstring already calls out.

---

src/backend/tests/community/api/test_domain_error_handler.py
```diff
@@ -255,11 +309,84 @@ def test_public_http_exception_logging(monkeypatch, status, level, wants_traceba
     assert calls, f"expected a {level} log for {status}"
     args, kwargs = calls[-1]
     assert f"boom-{status}" in args, "the raised detail must survive in the log"
-    if wants_traceback:
-        assert kwargs.get("exc_info") is not None, \
-            "a 5xx HTTPException must carry its raise site"
-    else:
-        assert kwargs.get("exc_info") is None
+    assert wants_traceback, "every status through this handler carries a traceback"
```
totalfrank: nit: `wants_traceback` is now `True` in every parametrize case (lines 285–291), so this assert is a tautology and the parameter carries no signal anymore. Could drop the field now that this handler never omits the traceback.

---

totalfrank (commented):
Reviewed commits b97bcaa4..13abc355 (full PR — no prior review on this thread). Files: `src/backend/src/agentclaw/community/adapters/http/app.py`, `tests/community/adapters/http/skill_center/test_skillset_error_status_mapping.py`, `tests/community/api/test_domain_error_handler.py`. Checked: branching logic across all seven exception handlers, comment/behavior consistency, wire-shape/contract impact (unchanged, confirmed by tests), 4xx log-volume blast radius, and test coverage of the new behavior.

One cross-cutting note that isn't pinnable to a single line: the PR description's "Compatibility and risk" section (the 10-chain-a-cause / 44-unaffected blast-radius numbers, "no 401 or 404 gains a stack") describes the first commit's `__cause__`-gated approach. The second commit (13abc35) supersedes that with an unconditional "every handler logs a traceback for every status" change, so the description no longer matches what's shipping — see the inline comments for the two spots this matters most.

Two non-blocking findings inline, plus one nit.

---

FreddieSun (dismissed):
LGTM

src/backend/src/agentclaw/community/adapters/http/app.py
totalfrank: Fixed in fe4bf59f. Correct catch — I rewrote `_domain_error_handler`'s comment in the same diff and missed this one, so it was promising the next reader something the code two lines below had stopped doing.

It now says every status carries the traceback, and keeps the useful half of the old observation: for a routing 404/405 the stack really is framework internals and adds little, but a handler-raised 4xx is worth keeping and the two are indistinguishable from here, so the handler does not try to tell them apart.

---

src/backend/src/agentclaw/community/adapters/http/app.py
totalfrank: Docstring updated in fe4bf59f to record the trade instead of reading as if it had not been made.

On the substance: the volume concern is real and the author asked for it explicitly, so the change stands — but one correction to the framing. What this handler was written to stop was `ServerErrorMiddleware` re-raising, which dumped the full ASGI stack (~100 lines) *twice* per request. Registering the concrete types fixed that and still does; `exc_info=exc` emits one formatted chain for the exception itself, which is materially smaller. So this is not a return to the prior state, though it is a per-request traceback under the same failure mode. The docstring now says exactly that, and notes the escape hatch: if volume bites, drop `exc_info` here specifically rather than unregister the types, which would bring the re-raise back.

On the description: agreed it was stale, and it had already been rewritten before this review landed — the "bounded to the 10 raise sites" wording belonged to the first commit and is gone. The current "Compatibility and risk" section states the blast radius as every 4xx across the app and carries an explicit reviewer call-out on this exact scenario.

---

src/backend/tests/community/api/test_domain_error_handler.py
totalfrank: Dropped in fe4bf59f — the parameter is gone and the assert is now unconditional. The `level` column still splits on status, so the parametrize keeps its remaining signal.

---

src/backend/src/agentclaw/community/adapters/http/app.py
totalfrank: Confirmed fixed. The comment above `_http_exception_handler`'s log call now reads "Every status carries the traceback..." and matches the unconditional `exc_info=exc` two lines below — no more contradiction between the comment and the code.

---

src/backend/src/agentclaw/community/adapters/http/app.py
totalfrank: Agreed, resolving. The docstring now states the trade explicitly rather than reading as though it hadn't been made, and the substance holds: this was already flagged as a known, accepted risk in the PR's "Compatibility and risk" section, and one formatted `exc_info` chain per request is a materially smaller cost than the double ASGI-stack dump this handler was written to stop. The escape hatch (drop `exc_info` here specifically if volume bites) is a reasonable mitigation path if this ever needs revisiting.

---

src/backend/tests/community/api/test_domain_error_handler.py
totalfrank: Confirmed. `wants_traceback` is gone from the parametrize and the assert on `exc_info` is unconditional; `level` still varies with status, so the parametrize keeps its remaining signal. `test_public_http_exception_logging` and the rest of the suite pass (38/38 in `test_domain_error_handler.py` + `test_skillset_error_status_mapping.py`).

---

totalfrank (commented):
Follow-up round on `fe4bf59f` (the docs-only commit addressing the prior review's findings). Reviewed the new commit against `13abc355` (its predecessor) and re-verified all three prior findings against the current code + test run.

**Findings from the previous round:**

1. Stale comment above `_http_exception_handler`'s log call (line ~551) — **Fixed.** Comment now matches the unconditional `exc_info=exc`. Resolved.
2. `_principal_error_handler` docstring not documenting the `exc_info` trade-off (line ~693) — **Conceded/resolved.** Docstring now states the trade explicitly; reasoning holds given the PR description already calls out this exact risk, and the added cost is materially smaller than what the handler was written to avoid.
3. Tautological `wants_traceback` test parameter — **Fixed.** Parameter dropped, assert is unconditional, `level` parametrize retained.

**New commit (`fe4bf59f`) review:** comments-and-one-test-parameter only, no behavior change, matches its stated scope. Verified by reading the diff and re-running `test_domain_error_handler.py` + `test_skillset_error_status_mapping.py`: 38/38 passed. No new issues found.

**Status: 0 open findings.** All three findings from the prior round are fixed/conceded and resolved. This round is clean.

---

FreddieSun (approved):
LGTM

PR #1351, opened 2026-08-22
feat(backend): add durable bot-level skill activation sync task

totalfrank:
## Problem

A skill activation is two writes that must land together: the desired-state change in the database, and the RPC that repoints the Bot's runtime symlinks. `SkillSetControlPlaneService._mutate` performs both inline and, when the RPC fails, compensates by restoring the previous desired state and reconciling again — a best-effort path that a pod restart between the two writes defeats outright, leaving the Bot with desired state its runtime never received.

The `skill_center` README already names this gap and defers the fix: *"Phase 1 intentionally has no cache-backed cross-command Bot mutation fence: the current compensating restore remains a best-effort compatibility path for non-concurrent mutations, while durable serialization is deferred to the task-queue design."*

## Solution

Adds the task type and its enqueue path — `src/backend/src/agentclaw/community/core/skill_center/services/skill_activation_sync_task.py`:

- `SKILL_ACTIVATION_SYNC_TASK = "skill_center.activation_sync"` — one task type shared by every activation-shaped operation (activate, deactivate, switch, membership changes, MCP direct installs).
- `SkillActivationSyncAction` — the `action_type` payload discriminator, so the handler knows which operation it is executing. Ships with a single `PLACEHOLDER` member; each operation adds its own as it migrates onto the durable path.
- `SkillActivationSyncScope` / `SkillActivationSyncWork` — the `(env, entity_id, bot_id)` target and the parsed payload.
- `build_skill_activation_sync_payload` / `parse_skill_activation_sync_payload` / `build_skill_activation_sync_idempotency_key`.
- `enqueue_skill_activation_sync(...)` — the enqueue helper, which stamps the action type and the dedup key.
- `SkillActivationSyncTaskHandler` — **a skeleton.** It owns `task_type` and the per-attempt payload validation, then hands off to a single `_run` seam that the implementation replaces. It is not registered into a `HandlerRegistry`; registration lands with the body, alongside the first operation that enqueues. Until then nothing calls the enqueue helper, so no row of this type exists to run, and the control plane's inline path is untouched.

Three design choices worth review:

**The dedup key is the Bot, not the operation.** `(env, entity_id, bot_id)`, excluding `action_type`, because the Bot's runtime projection is the unit that must not be synchronized concurrently — two operations racing on one Bot are exactly what the durable path exists to serialize. The consequence is load-bearing: a second operation arriving mid-sync joins the live task and gets `created=False`. That is correct for a level-triggered handler that reconciles against whatever desired state the database holds when it runs, and it would silently drop work for a handler that performed the desired-state write itself from `action_args`. The module docstring states that obligation explicitly, and an integration test pins the observable behaviour (the joined task still carries the *first* operation's `action_args`).

**`entity_id` and `bot_id` are digested into the key.** The key column is `VARCHAR(190)` while its sources are `varchar(512)` and `varchar(128)` on the Skills Pool tables. The task queue README is explicit that the variable part of a key should be hashed rather than embedded for exactly this reason. The rejected alternative — embedding them raw and validating the length — fits every id in production today and raises `ValueError` mid-activation on the first tenant whose id is long; a key that cannot be built is an activation that cannot proceed, whereas a digest always fits. `env` stays literal because operators filter `ac_task_queue` on it first, and the raw triple is preserved verbatim in the payload so nothing is lost to audit.

**Both handler exits are `Fail`, not a raise.** The worker turns an exception into an implicit `Retry` with backoff. Because a live task holds the Bot's dedup key, an unparseable payload or the not-yet-implemented body would then pin that key for the whole 15-minute deadline and suppress every activation queued behind a row that can never succeed. `Fail` is terminal, so the key is released at once and the reason is recorded on the row. The 15-minute deadline itself is short for the same reason — far below the day-long horizons used by publish and Pool migration — since it bounds how long a wedged sync may keep suppressing the ones behind it.

Optionality is deliberately absent from the public surface: `action_args` defaults to `{}` and is copied on the way in, `parse_...` takes `dict` rather than `dict | None` (`TaskRecord.payload` is non-optional and `enqueue` is its only writer), and `handle` narrows the protocol's `Optional[dict]` to `dict` — the protocol is structural, and a test asserts conformance plus that `HandlerRegistry` accepts the type. The payload carries no generated correlation id: a fresh UUID would correlate to nothing the caller knows, and `ac_task_queue.id` is already the task's identity.

`skill_center/README.md` is updated: `agentclaw.community.core.task_queue` added to `internal_dependencies` (the boundary test fails without it), the new symbols added to `provides`, and the change-impact section extended.

### Drive-by CI fix

`tests/community/core/skill_center/test_skill_set_control_plane_service.py` had one test still constructing `SkillSetControlPlaneService` with `mutation_guard=` and `edit_guard=`. Those parameters were removed when the control plane stopped acquiring `SkillsPoolEditGuard`, and the `_MutationGuard` / `_Guard` fakes went with them, so the name lookup raised `NameError` before the service was built. It was failing `Backend unit tests` on the base branch, not just here. Deleting the two arguments is the whole fix — the test itself is not about the guards (it asserts that `resources()` reports ordinary SkillSet MCP membership via the canonical repository path, which the neighbouring legacy default-path test does not cover), so it stays.

## Validation

Run with `uv` against in-memory SQLite in the session container:

- `pytest tests/community/core/skill_center/test_skill_activation_sync_task.py` — **37 passed**. Covers the payload contract and `action_type` round trip, defensive copying of `action_args` (including that the shared `{}` default cannot be poisoned), the payload carrying nothing beyond the work identity, malformed-payload rejections, key determinism/component-sensitivity/shape, the key fitting `VARCHAR(190)` at maximum column widths, the enqueue helper's forwarding, and the handler skeleton (task type, protocol conformance, registrability, and both `Fail` exits). Three integration tests drive a real `TaskQueueRepository`: a second operation on the same Bot joins the live task, a different Bot does not dedup, and a terminal task releases the key for the next operation.
- `pytest tests/community/core/skill_center tests/community/core/task_queue tests/community/architecture` — **997 passed, 3 skipped, 0 failed**. Includes the module-boundary check; verified that check is actually load-bearing here by temporarily breaking the README entry and confirming it fails on the two new `task_queue` imports.
- `ruff check` on every changed file — clean, including an isolated run against the `python_sast_local.sh` block-rule set.
- CI on this head (`7057a24`): all 7 checks green — `Backend unit tests`, `Singlebox coverage`, `BCS unit tests`, `BCS e2e (coverage gated)`, `BaaS unit tests`, `Engine unit tests`, `Gateway unit tests`.

Not run locally: the full backend suite and Singlebox E2E (both covered by CI). The pre-push hook is not installed in this container, and dependency resolution had to be pointed at PyPI because the pinned Aliyun mirror is unreachable from here; `uv.lock` was restored afterwards and is not part of this diff.

## Compatibility and risk

No schema change and no behaviour change — the handler is not registered and nothing enqueues, so this is additive and inert until the follow-up. Rollback is deleting the module.

One ordering note for the follow-up: the task queue's own idempotency mechanism required its first keyed adopter to ship in a strictly later release than the mechanism, because a pre-change worker never releases `active_idempotency_key`. This is that first adopter, so it must not ship alongside the mechanism's release. The same ordering applies to each new `SkillActivationSyncAction` member: ship the member one release before the release that starts enqueuing it, since an older pod rejects an `action_type` it does not recognise.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KCp99D1NuSHGDJBSUWzQQv)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/skill_center/README.md
+++ b/src/backend/src/agentclaw/community/core/skill_center/README.md
@@ -28,6 +28,13 @@ provides:
   - "BotRuntimeProjectionReconcilerProtocol"
   - "ActiveSkillSetInstallationMaterializer"
   - "LocalSkillCleanupWorkModel"
+  - "SkillActivationSyncAction"
+  - "SkillActivationSyncScope"
+  - "SkillActivationSyncWork"
+  - "enqueue_skill_activation_sync"
+  - "build_skill_activation_sync_payload"
+  - "parse_skill_activation_sync_payload"
+  - "build_skill_activation_sync_idempotency_key"
 consumes:
   - "BotRepository"
   - "BotCollabLogRepositoryProtocol"
@@ -70,6 +77,7 @@ internal_dependencies:
   - agentclaw.community.core.models
   - agentclaw.community.core.spaces.services
   - agentclaw.community.core.skills_pool
+  - agentclaw.community.core.task_queue    # durable enqueue for Bot-level activation sync
   - agentclaw.community.core.workspace
   - agentclaw.community.di.modules
   - agentclaw.community.di.runtime_mode
@@ -126,6 +134,17 @@ deletion and Pool cutover/rollback paths. Phase 1 intentionally has no
 cache-backed cross-command Bot mutation fence: the current compensating restore
 remains a best-effort compatibility path for non-concurrent mutations, while
 durable serialization is deferred to the task-queue design.
+
+`skill_activation_sync_task.py` is the enqueue half of that durable design:
+`skill_center.activation_sync`, one task type shared by every
+activation-shaped operation, discriminated by the payload's `action_type` and
+deduped on the Bot — `(env, entity_id, bot_id)` — so at most one
+synchronization per Bot is ever live. A second operation arriving mid-sync
+joins the live task and gets `created=False`; that is only correct while the
+handler reconciles against desired state read from the database, so the
+handler that consumes these rows must not replay `action_args` as its
+desired-state write. No handler is registered yet and no call site enqueues,
+so the control plane's inline mutate-then-reconcile path is unchanged.
 
 Phase 1 does not run a global Local Installation backfill and does not treat
 historical Default exclusions as an active-state source. The small number of
--- a/src/backend/src/agentclaw/community/core/skill_center/README.md
+++ b/src/backend/src/agentclaw/community/core/skill_center/README.md
@@ -30,6 +30,7 @@ provides:
   - "LocalSkillCleanupWorkModel"
   - "SkillActivationSyncAction"
   - "SkillActivationSyncScope"
+  - "SkillActivationSyncTaskHandler"
   - "SkillActivationSyncWork"
   - "enqueue_skill_activation_sync"
   - "build_skill_activation_sync_payload"
@@ -143,7 +144,9 @@ synchronization per Bot is ever live. A second operation arriving mid-sync
 joins the live task and gets `created=False`; that is only correct while the
 handler reconciles against desired state read from the database, so the
 handler that consumes these rows must not replay `action_args` as its
-desired-state write. No handler is registered yet and no call site enqueues,
+desired-state write. `SkillActivationSyncTaskHandler` is a skeleton: it owns
+the registry key and the payload validation, and its `_run` seam reports
+`Fail` until the body lands. It is not registered, and no call site enqueues,
 so the control plane's inline mutate-then-reconcile path is unchanged.
 
 Phase 1 does not run a global Local Installation backfill and does not treat
--- a/src/backend/src/agentclaw/community/core/skill_center/services/skill_activation_sync_task.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/skill_activation_sync_task.py
@@ -11,12 +11,14 @@ task-queue design"): the operation is persisted as a task *first*, so a crash
 resumes it instead of stranding the Bot with desired state its runtime never
 received.
 
-**Only the enqueue side lives here.** There is deliberately no
-:class:`~agentclaw.community.core.task_queue.services.registry.TaskHandler` and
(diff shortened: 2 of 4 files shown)


FreddieSun (dismissed):
LGTM

totalfrank (commented):
Reviewed commits `bcc3046`..`7057a24` (all 5 on this PR — this is the first automated pass). No findings.

**Files covered:** `skill_center/README.md`, new `skill_center/services/skill_activation_sync_task.py` (368 lines), new `test_skill_activation_sync_task.py` (383 lines), and the drive-by fix in `test_skill_set_control_plane_service.py`.

**What I checked:**
- Correctness of the payload build/parse round-trip, the dedup-key digest construction, and the `Fail`-not-`Retry` semantics on both the malformed-payload and unimplemented-body paths.
- Contract conformance against `TaskQueueService.enqueue` (positional/keyword args match) and the `TaskHandler` protocol (`task_type` property, `handle` signature — the narrower `dict` vs. the protocol's `Optional[dict]` is safe here since `TaskRecord.payload` is non-optional and there's no static type-checking gate in this repo's CI to catch the variance anyway).
- `T | None` avoidance per `CLAUDE.md`: `action_args` defaults to a shared `{}` but every call site copies it before use (pinned by `test_payload_does_not_leak_into_the_shared_empty_default`), so the mutable default is safe.
- The `SkillSetControlPlaneService` constructor no longer takes `mutation_guard`/`edit_guard` — confirmed against the current source, so the CI fix is correct.
- README `internal_dependencies`/`provides` additions match the new module's actual imports and exports.
- Test coverage of the new contract (payload shape, parse rejections, key determinism/component-sensitivity/column-fit, enqueue forwarding, handler skeleton, and three integration tests against a real `TaskQueueRepository` for the dedup behavior) — thorough for a skeleton with no registered handler yet.

Nothing here registers a handler or enqueues a task, so blast radius is inert as described.


---

FreddieSun (approved):
lgtm

Issue #907, opened 2026-08-09
Public API: ownership and collaborator semantics for the resources and MCP groups


totalfrank:
## Problem

The **resources** (9 operations) and **mcp** (6 operations) groups scope everything to the caller's own id, and it has never been decided whether they have a collaborator dimension at all: resources belong to users rather than bots, and MCP configuration is per-caller. Meanwhile a recorded enforcement gap is still open — four resources operations scope on a caller-supplied `bot_id` without verifying the caller may act for that bot (`specs/2026-08-02-public-api-user-only-principal/` records it; the explicit-user-id spec carried those four handlers as "not-yet-enforced", taking `user_id` without using it).

## Scope

Mostly a decision/spec task with a small enforcement change:

- Decide and write down the ownership model for resources and MCP config on the public surface: user asset vs bot asset, and whether a shared bot implies shared resources.
- Close the resources `bot_id` gap so the four handlers enforce the scope they advertise.
- If a collaborator dimension exists, spec it following the #904 pattern; if not, record "no collaborator dimension" explicitly with a reason each — the way the no-user-dimension exemptions are pinned in `docs/openapi-v1/README.md`, so absence is a decision rather than drift.

Anchors: `adapters/http/openapi_v1/resources/router.py` (the four `del`-commented handlers), `…/mcp/router.py`, `specs/2026-08-02-public-api-user-only-principal/spec.md`.

Independent of #904's implementation; can run in parallel with it.

Issue #906, opened 2026-08-09
Public API: collaborator access to the bots and identity groups


totalfrank:
## Problem

The access-expansion spec (`src/backend/specs/2026-08-09-openapi-v1-access-expansion/`, PR #904) admits collaborators to the engine-runtime operator surfaces but leaves the data categories owner-scoped: a MEMBER collaborator can drive the team bot's sessions, yet `GET /openapi/v1/bots/{bot_id}` still answers them the masked 404. Skills is the only data category with owner-versus-actor semantics shipped.

## Scope

The **bots** group (13 operations) and **identity** (3 operations — it rides on the same owner-scoped `get_bot` resolve). One category-sized PR per the Track B recipe; run SDD first.

Decisions to make (the real work):
- Per-operation role bars: read the record (MEMBER?), update / engine-config (MEMBER, per "edit content"?), delete / restart (ADMIN or owner-only?), passport / auth-status. The platform's internal vocabulary is `PermissionLevel` (`NONE < MEMBER < ADMIN < OWNER`) with ADMIN = "edit + publish + manage collaborators".
- Whether `GET /openapi/v1/bots` (list) gains a collaborator dimension — `list_bots_by_conditions` has no collaborator axis today, so "bots shared with me" is a new query shape, not a parameter.
- Reuse the acquisition/adjudication split from #904: an `owner_id` locator defaulting to the caller, one adjudication seam, refusals masked as 404.

Anchors: `adapters/http/openapi_v1/bots/router.py`, `…/identity/router.py`, `core/bot_collaborator/services/collaborator_service.py` (`check_collaborator_permission`), the shipped skills pattern in `core/skill_center/services/local_skill_*_service.py`.

**Depends on:** the implementation of #904 (the owner locator and adjudication seam land there).

PR #1353, opened 2026-08-22
refactor(openapi-v1): rename SELF_CHECKED authorization mode to INHERITED

totalfrank:
## Problem

`AUTHORIZATION`'s `SELF_CHECKED` mode claimed a property that no route in it has.

A `deprecated/` address is not self-checking. `_relocate.relocate` and
`_requery.with_query_parameter` re-register the **replacement's own endpoint
function** at the path it used to have — at most one parameter's annotation is
swapped, moving `bot_id` back from the path into the query string. There is no
second handler, so there is nothing there to do its own checking, and no
decision of its own to record: what governs one of these rows is the
replacement's row.

Reading the table, that mattered in two ways:

- The name answered "who enforces this?" with a claim, and the claim was false.
  What the 42 rows actually inherit is not uniform — 22 of them twin a
  `ServiceChecked` address and 20 twin an `OWNER_SCOPED` one — and the sentinel
  showed none of that.
- It collided with `deprecated/__init__.py`'s `SELF_CHECKED_ROUTES`, which is an
  unrelated **admission** concept and is correctly named: those routers really
  do check the grant themselves, first, before acting, because their bot is not
  where `require_granted_own_bot` can see it. Two different things under one
  name, one file apart.

## Solution

Rename the mode to `INHERITED`, which answers the same question the other rows
answer — *what governs this?* → *whatever governs the address that replaced it*
— and matches the vocabulary `admission.py` already uses for exactly this
relationship (`_inherit_admission_modes`, `deprecated/__init__.py:81`).

`deprecated`'s `SELF_CHECKED_GROUPS` / `SELF_CHECKED_ROUTES` keep their name;
they are the case the word is true of. The collision is gone as a side effect.

Also rewritten: the module docstring's bullet and the sentinel's own comment,
which previously said these routers "check themselves or inherit their
replacement's mount". The first half of that was the error; the comment now
names `_relocate` / `_requery` and says why there is no decision to record.

Naming only — deliberately narrow. The row-level consequence of the old name
(a `deprecated/` twin does not pick up its replacement's row, so it will not
gain the gate when a `ServiceChecked` group migrates) is **not** addressed here
and is left as-is by decision.

## Validation

- `tests/community/adapters/http/openapi_v1/` plus
  `tests/community/core/skill_center/test_skill_set_control_plane_service.py` —
  **1284 passed, 2 skipped**, on the merged tree.
- `scripts/ci/python_sast_local.sh src/backend 1 --base 5f42dca --head HEAD`
  (changed-files mode, the merge base against the PR base) — **exit 0, no
  findings** on the files this change touches. Run whole-tree instead, the
  script exits 1 on 1334 pre-existing findings; that count is byte-identical
  before and after this change and none of them is in `authorization.py`.
- Table re-counted after the rename and unchanged in shape: **244 rows** —
  40 `OWNER_SCOPED`, 99 `ServiceChecked`, 63 `NoCheck`, 42 `INHERITED`, 0
  `Check` — with `scaffolding_row_count()` at 181. (The base added
  `POST /bots/metadata/search` and `POST /bots/{bot_id}/skills/upload-folder`,
  renamed `caller-identity` → `iam-token`, and dropped
  `GET /org/user/iam-token`; none of those rows used the renamed sentinel, so
  the rename merged without conflict.)
- `repr(INHERITED)` is `INHERITED`, so `_rule_for`'s differing-rules message
  names the mode a reader will find in the table.

Not run locally: the full backend suite and the E2E gates — CI covers those, and
all 7 checks are green on `cac5db8`.

### One red CI run, and why it was not this change

The first run on `30d227a` failed `Backend unit tests` with
`NameError: name '_MutationGuard' is not defined` in
`test_skill_set_control_plane_service.py`. That is base-branch breakage, not
this diff, which touches one module's identifiers and one doc:

| commit | `_MutationGuard` refs | class defined | |
| --- | --- | --- | --- |
| `6871ff6` (branch point) | 18 | yes | consistent |
| `8bbbf3a` fix(gateway): register folder upload authorization | 1 | no | broke it |
| `9f89617` test(skills): lock shared OpenAPI upload lifecycle | 1 | no | still broken |
| `5f42dca` test(skillsets): remove stale guard fixtures | 0 | no | fixed |

`8bbbf3a` deleted `bot_capability_mutation_guard.py` and the test's fixture but
left the call site at line 915 — inside the exact test that failed — and
`5f42dca` removed it three commits later. CI builds the PR's *merge* commit, so
the run landed inside that window. Merging the recovered base (`cac5db8`) is the
fix; no change to this PR's own diff was needed or made.

## Compatibility and risk

None for callers. No row changed mode, no route changed behaviour, no schema,
config or migration impact. `SELF_CHECKED` was internal to
`openapi_v1/authorization.py` — nothing outside it imported the name, so the
rename has no other call sites.

The dated `2026-08-21` changelog entry in `docs/openapi-v1/README.md` keeps the
old name, since it records what shipped that day; a new `2026-08-22` entry
records the rename. The SDD documents under
`specs/2026-08-21-openapi-v1-collaborator-authorization-seam/` are likewise left
alone as the record of that session.

## Spec

None — follow-up naming fix to `src/backend/specs/2026-08-21-openapi-v1-collaborator-authorization-seam/`.

## Related issues

Follow-up to #1323. Does not affect #906 / #907.


FreddieSun (approved):
LGTM

totalfrank (commented):
Reviewed commits `30d227a` (rename) and `cac5db8` (merge of recovered base). Files covered: `src/backend/src/agentclaw/community/adapters/http/openapi_v1/authorization.py`, `src/backend/docs/openapi-v1/README.md`.

This is clean. Verified:
- **Correctness**: all 42 `deprecated/` table rows renamed consistently, sentinel definition, `__repr__`, `__all__` (alphabetical position of `INHERITED` is correct per the file's existing sort convention), module docstring bullet, and README's mode table/prose all updated together — no stray `SELF_CHECKED` left behind in `authorization.py` itself.
- **Contract breakage / callers**: none — `git grep` across the head branch confirms `SELF_CHECKED` had no importers outside this module; the remaining hits are the unrelated `SELF_CHECKED_ROUTES`/`SELF_CHECKED_GROUPS` admission concept in `deprecated/__init__.py` (correctly left untouched, per the PR's own stated rationale for why that name stays) and historical spec/changelog text intentionally preserved as a record of what shipped on 2026-08-21.
- **Test coverage**: naming-only change with no behavior change; existing tests (`test_self_checked_routes_refuse.py` etc.) reference the separate, unrenamed `SELF_CHECKED_ROUTES` concept and are unaffected.
- Categories not applicable to this diff: error handling, concurrency, security, architecture boundaries, Python type contracts — no logic changed.

No findings.

---

PR #1354, opened 2026-08-22
fix(backend): retire stale SkillSet runtime mappings

FreddieSun:
## Problem
SkillSet lifecycle reconciliation published only the current mapping set. Successful deactivate therefore left removed Local/Repo soft links at runtime, and a later non-skill projection failure could restore database desired state without retiring mappings already published by the failed attempt.

A second prepub reproduction showed why explicit retirement is required for Legacy layout too: an `xlsx` active link pointed to a missing legacy Local source. The Engine returned HTTP 200 with business `success=false` and `source_missing` / `managed_source_missing`; without a retirement intent, publish could neither converge nor reach verify.

## Solution
Snapshot logical Skill mappings before and after the desired-state mutation. Reconcile with the exact `retired_logical_skill_mappings(previous, current)` difference; if reconciliation fails, restore desired state and reconcile again with the reverse difference. The snapshot is read-only and does not materialize installations or touch runtime.

This preserves the current Engine contract for both Pool and Legacy source layouts: current mappings are created or corrected, while only exact product-managed retired identities are removed.

## Validation
- Control-plane, Service API contract, and architecture suites — 96 passed.
- Focused canonical SkillSet/MCP OpenAPI endpoint cases — 42 passed.
- Engine mapping retirement suite — 25 passed.
- Focused Ruff, compile, and diff checks passed.
- Previous Backend CI fixture failures were fixed by synchronizing the latest `dev_refactory_collaboration` baseline and updating the endpoint runtime fake to the new Service API.

## Compatibility and risk
No HTTP/OpenAPI, schema, or storage contract changes. The new internal Service API read method is covered by Core/Public protocol tests. Retirements are an exact previous-versus-current mapping difference, so mappings that remain active or have been rebound externally are not removed.

If the compensation reconcile itself also fails, the command still reports failure rather than hiding runtime uncertainty.

## Related issues
Phase 1 P1: pre-runtime drift after failed SkillSet activation and successful deactivate.

### DIFF
--- a/src/backend/src/agentclaw/community/api/bot_runtime_projection_reconciler.py
+++ b/src/backend/src/agentclaw/community/api/bot_runtime_projection_reconciler.py
@@ -15,6 +15,13 @@ class BotRuntimeProjectionReconcilerProtocol(
 ):
     """Transport-facing contract; Core depends only on its sibling contract."""
 
+    async def snapshot_skill_mappings(
+        self,
+        *,
+        bot_id: str,
+        owner_id: str,
+    ) -> tuple[PoolSkillMapping, ...]: ...
+
     async def reconcile(
         self,
         *,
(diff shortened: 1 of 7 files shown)


totalfrank (approved):
LGTM.

PR #1378, opened 2026-08-23
Feat/collaboration bots public external path

rudyLi:
## Problem

The BCS publish-to-users endpoint (`POST /openapi/v1/collaboration/bots/{bot_uuid}/public`) returned `{"code": 404000, "message": "Not Found"}` in the deployed environment. The gateway was correctly forwarding the request to the backend (confirmed: request reached backend), but the backend had no route at that path — it only mounted the internal `POST /openapi/v1/bots/{bot_id}/public-bcs`.

The gateway's `collaboration-publish` domain verbatim-forwards the external contract path (origin swap only, no `rewrite`), so the request arrived at the backend as `/openapi/v1/collaboration/bots/{bot_uuid}/public` — a path no FastAPI route matched → route-level 404 → envelope `code: 404000 "Not Found"` (capital `F` distinguishes it from the `BotNotFoundError` path, which is lowercase `"Not found"` + `data: null`).

The original design assumed a gateway path rewrite would map the external path onto the internal `public-bcs` route. That rewrite was never wired, and **could never be**: gateway `PathRewrite` only does literal prefix substitution (`_domains.py:223-240`); it cannot reach the `public`→`public-bcs` tail change nor the `{bot_uuid}`→`{bot_id}` param rename. So the gateway-rewrite approach was structurally unviable — the backend serving the contract path directly is the clean fix.

## Solution

Move the backend route from the internal `/openapi/v1/bots/{bot_id}/public-bcs` to the external contract path `POST /openapi/v1/collaboration/bots/{bot_uuid}/public`, served directly. The gateway's verbatim forward now hits a real route; no rewrite is needed.

**Changes (commit `066f830a4`):**

- `collaboration_bots/router.py` — replaced the internal `router` (prefix `/openapi/v1/bots`, route `/{bot_id}/public-bcs`) with one `public_router` (prefix `/openapi/v1/collaboration/bots`, route `/{bot_uuid}/public`). The handler passes `bot_uid=bot_uuid` to the service; the service's existing `backend_bot_id = bot_uid.split(":")[0]` split is unchanged, so `{bot_uuid}` (the BCS identity `backend_bot_id:entity_id`) works as-is.
- `admission.py` / `authorization.py` — the `public-bcs` row is re-keyed to the external route-template path `/openapi/v1/collaboration/bots/{bot_uuid}/public`. Admission/authorization are route-aware (keyed by `route.path`, not the request URL), so FastAPI's route match provides the normalization.
- `collaboration_bots/__init__.py` — exports `public_router` only.
- `openapi_v1/__init__.py` — `build_public_router` includes `collaboration_public_router` with the same `USER_SCOPED_ERROR_RESPONSES` + `_PUBLIC_AUTH` as before.

**Test/contract updates:**

- `test_public_namespace.py::_DECLARED_PREFIXES` — gains `/openapi/v1/collaboration/bots` (the `collaboration-publish` gateway domain covers it). Deliberately the precise `bots` sub-prefix, **not** broad `/openapi/v1/collaboration`, so a future leak under `/collaboration/tasks` or `/collaboration/messages` is still caught.
- `test_explicit_user_id.py` conformance pins — operation count unchanged at 181 (one op moved, not added); `_BOT_ID_PLACEMENT` shifts `path 141→140` / `none 58→59` because the op's param went from `{bot_id}` to `{bot_uuid}` (it no longer carries a `bot_id` param).
- `test_collaboration_bots_endpoints.py` + `test_openapi_public_bcs.py` — moved to the external path and `{bot_uuid}` param.

**Rejected alternative — gateway `rewrite`:** structurally impossible (above). `PathRewrite.apply` swaps one literal prefix for another and carries the suffix through untouched; the two paths differ in three places (prefix, param name, **tail segment**), and the tail is out of reach.

**Companion cleanup (commit `1f0bc21bb`):** removed two dead gateway domains from `application.yaml` — `dept` (claimed `/openapi/v1/dept/**`, but the real route is `/openapi/v1/org/dept` under the `org` domain) and `caller` (claimed `/openapi/v1/caller/**`, no backend route exists; caller-identity router mounts at `/api/bots`). Both were dead config a request could never resolve.

## Validation

- `tests/community/adapters/http/openapi_v1/` full suite — **1268 passed, 2 skipped** (includes the two conformance pins that were re-baselined, and `test_the_published_document_contains_nothing_internal` which the new handler's docstring had to clear of backticks).
- `tests/community/contracts/` — **248 passed** (includes `test_no_internal_routes_leak_into_public_namespace` after `_DECLARED_PREFIXES` gained the collaboration-bots prefix).
- `tests/community/core/bot_public/` service suite — green (service layer unchanged; `public_bcs_bot` method name + behavior intact).
- `tests/community/adapters/http/openapi_v1/test_collaboration_bots_endpoints.py` + `tests/community/endpoints/test_openapi_public_bcs.py` — 5 passed (mounted-+served success Envelope, 422 validation, 404 not-found mapping).
- Pre-existing unrelated failures verified at HEAD before the change: `test_bcn_dump_uses_the_gateway_managed_python_environment` (assert 13==12), `test_ruff_lint_passes`, e2e `test_health` — all fail on the base branch without this change; not introduced here.

**Not run:** deployment e2e against the live gateway + backend. The fix is backend-route-only and is verified by the route-mounting + namespace contract tests; live confirmation lands after the branch merges and the backend redeploys.

## Compatibility and risk

- **External contract path is unchanged** — `POST /openapi/v1/collaboration/bots/{bot_uuid}/public` was already the documented public address and the gateway `collaboration-publish` domain already routes it. This change only makes the backend actually serve it.
- **Breaking for any internal caller of `/openapi/v1/bots/{bot_id}/public-bcs`** — that internal route is deleted. No code path was forwarding to it (the gateway rewrite it presumed never existed), and the contract tests + endpoint tests all exercise the external path, so no known caller is affected. If an internal integration existed, it must move to the external path.
- **No schema/migration impact** — request/response schemas (`BcsPublicRequest` / `BcsPublishResult`) are byte-identical; only the served path + path-param name changed.
- **ocb-side (enterprise mirror):** no backend change needed there — the community backend is Avernet-sourced. ocb already carries the gateway `collaboration-publish` domain (`ocb` commit `de1ef9be71`).
- **Rollback:** revert the branch — the internal `public-bcs` route returns, and the 404000 reappears for the external path (preserving the pre-fix broken state, not a data-risk regression).

## Spec

Implements the external contract path documented in `ocb-collaboration-bots-public-api.md` (field-level API doc): `{bot_uuid}` is the BCS bot identity; `owner_id` is derived server-side from the signed principal; `visibility` defaults to `private`; `protected` is a backend callback value, not caller input.

## Related issues

Resolves the deployed `404000 "Not Found"` on `POST /openapi/v1/collaboration/bots/{bot_uuid}/public`. Depends on the gateway `collaboration-publish` domain (already in `dev_refactory_collaboration`); this PR supplies the backend counterpart that domain forwards to.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/__init__.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/__init__.py
@@ -176,7 +176,7 @@ from fastapi import APIRouter, Depends
 from .authorized_apps import app_view_router as authorized_bots_router
 from .authorized_apps import router as authorized_apps_router
 from .bots import router as bots_router
-from .collaboration_bots import router as collaboration_bots_router
+from .collaboration_bots import public_router as collaboration_public_router
 from .bots.engine_config import router as engine_config_router
 from .org import dept_router as org_dept_router
 from .org import router as org_router
@@ -520,12 +520,14 @@ def build_public_router() -> APIRouter:
     public.include_router(
         bots_router, responses=ERROR_RESPONSES, dependencies=_PUBLIC_AUTH
     )
-    # New-version bcs publish-to-users (backend route: /openapi/v1/bots/{bot_id}/public-bcs;
-    # the gateway rewrites the external /openapi/v1/collaboration/bots/{bot_uuid}/public onto
-    # it). Declares user-scoped responses for the openapi_v1 admission contract; authz
-    # is deferred per design (caller identity via _PUBLIC_AUTH/UserIdDep, no grant check yet).
+    # New-version bcs publish-to-users: served at the external contract path
+    # POST /openapi/v1/collaboration/bots/{bot_uuid}/public. The gateway's
+    # collaboration-publish domain verbatim-forwards it here (no rewrite), so the
+    # backend serves the public address directly. user-scoped responses for the
+    # openapi_v1 admission contract; authz deferred per design (caller identity
+    # via _PUBLIC_AUTH/UserIdDep, no grant check yet).
     public.include_router(
-        collaboration_bots_router,
+        collaboration_public_router,
         responses=USER_SCOPED_ERROR_RESPONSES,
         dependencies=_PUBLIC_AUTH,
     )
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
@@ -574,7 +574,8 @@ ADMISSION: dict[tuple[str, str], AdmissionMode] = {
     ("GET", "/openapi/v1/bots/mcp/servers/{server_code}"): AdmissionMode.OPEN,
     ("GET", "/openapi/v1/bots/mcp/tenants"): AdmissionMode.OPEN,
     # New-version bcs publish-to-users: auth baseline only, no grant check; authz deferred.
-    ("POST", "/openapi/v1/bots/{bot_id}/public-bcs"): AdmissionMode.OPEN,
+    # Served at the external contract path the gateway verbatim-forwards here.
+    ("POST", "/openapi/v1/collaboration/bots/{bot_uuid}/public"): AdmissionMode.OPEN,
     # Department directory search — a tenant-wide catalogue read, not a user's.
     ("GET", "/openapi/v1/org/dept"): AdmissionMode.OPEN,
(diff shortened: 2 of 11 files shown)


totalfrank (approved):
LGTM.

PR #1381, opened 2026-08-23
feat(backend): opt-in immediate execution for enqueued task types

totalfrank:
## Problem

`TaskWorker`'s poll loop ended every non-full tick with an unconditional
`asyncio.sleep(self._idle_delay())`. A task enqueued just after a tick therefore
waited out most of `poll_interval_seconds` (default 2.0s ± 0.5 jitter) before
anything looked at it, no matter how urgent it was.

That is the right default for background and recurring work, but some operations
are user-visible and should start as soon as they are enqueued.

## Solution

The idle wait becomes a race between the timer and a signal, so the interval is
a latency **ceiling** rather than a fixed cost. Claiming, leases, deadline
handling, backoff and the greedy re-poll are all untouched.

**Opt-in per task type, default off.** A type asks for this at registration:

```python
registry.register(MyHandler(), wake_on_enqueue=True)
```

Every already-registered type keeps exactly the timing it had — the default is
`False` and no existing registration call changes. The flag lives at the wiring
site rather than on the handler because how eagerly a queue is drained is a
deployment concern, and one grep for `wake_on_enqueue=True` then lists every
type that opted in.

**New `WorkerWakeup`** — a coalescing latch bound as a DI singleton and shared by
the enqueue path and the worker's loop, so `TaskQueueService` never has to depend
on `TaskWorker`. A burst of 500 enqueues produces one extra poll, not 500; the
existing greedy re-poll then drains the backlog at full speed.

Two details carry the correctness:

- **The signal is raised only after `repo.enqueue` returns.** `orm_session()`
  commits on clean exit, so the row is committed and visible to any claim the
  wake triggers, rather than racing our own uncommitted insert.
- **`notify()` goes through `loop.call_soon_threadsafe`.** `enqueue()` is
  synchronous, and for a handler enqueuing follow-up work it runs on an
  `asyncio.to_thread` pool thread, not the loop thread. A bare `Event.set()`
  there appends to the loop's ready queue *without* waking it out of `select()`,
  so the wait would run its full timeout — the feature would be a silent no-op
  with no exception and no failing test. A test pins this down: with the naive
  set, the wake takes the entire 10s timeout instead of ~50ms.

The latch also declines to signal where a wake cannot help: a delayed enqueue
(`run_at` in the future, so ineligible to claim), a keyed enqueue that joined a
live holder (`created=False`, no new work), and a type this process has no
handler for.

### Scope

The wake is in-process. The enqueuing pod's own worker starts claiming within a
millisecond; other pods still poll on their normal interval, which is correct
since any worker may claim any task. A fleet-wide wake would need a broker, and
`notify()` is the seam if that is ever wanted — deliberately not taken here, as
the queue being DB-only is a real virtue.

This does not make a task jump a backlog: `claim_batch` still orders `run_at
ASC`. Priority ordering would be a separate change.

## Validation

**CI: all 7 checks green on `887ee70`** — including Backend unit tests and
Singlebox coverage, which exercise the real `uv sync --frozen` environment and
the `ci_test.sh` coverage gate.

Locally, before pushing:

- **252 passed** across `tests/community/core/task_queue/`, the task queue
  repository suite, and `tests/community/di/` (wiring resolves with the two
  changed constructor signatures).
- **New coverage** (25 tests): the default-off guarantee, explicit opt-in,
  delayed and keyed-duplicate enqueues not signalling, latch clearing and
  coalescing, notify before bind / after unbind / after loop close, a
  cross-thread wake asserted to come from a different thread ident, and an
  end-to-end test that claims a task in well under a 30s poll interval.
- **Mutation-checked** both key tests rather than trusting green: reverting
  `notify()` to a bare `Event.set()` makes the cross-thread test fail on elapsed
  time (10.01s vs. the &lt;2s bound), and disabling `_should_wake` fails the
  end-to-end test. Both were restored and re-verified.
- **Full backend suite**: 12,385 passed, 46 skipped. The 8 remaining failures
  (3 architecture, 5 in `service_bot` / `bot_startup_script`) reproduce
  identically on an unmodified `origin/dev` tree in that container — they are
  environment artifacts, not regressions, and CI confirms this.
- **flake8 SAST block scan** (`scripts/ci/python_sast_local.sh` rule set) clean
  on all 8 changed files.

Note on the local environment: the container could not reach the pinned Aliyun
package index (the proxy denies it), so the local run used an ad-hoc
PyPI-resolved venv rather than `uv sync --frozen`. `uv.lock` was **not**
modified. CI covers what that environment could not.

## Compatibility and risk

No schema, config, or contract change — no new `task_queue_worker` config key,
so the golden config fixtures are untouched. `HandlerRegistry.register` gains a
keyword-only parameter with a default, so every existing call site is unchanged.

`TaskQueueService.__init__` and `TaskWorker.__init__` each take one new
dependency; both are DI-resolved in production, and the four direct construction
sites in tests are updated in this PR.

Risk is low and self-limiting: the signal only ever removes waiting. If it is
never delivered, behaviour is exactly what it is today. Rollback is reverting
the commit; there is no persisted state to migrate.

### DIFF
--- a/src/backend/src/agentclaw/community/core/task_queue/services/registry.py
+++ b/src/backend/src/agentclaw/community/core/task_queue/services/registry.py
@@ -15,7 +15,7 @@ so every participant sees the same instance.
 """
 from __future__ import annotations
 
-from typing import Dict, Optional, Protocol, runtime_checkable
+from typing import Dict, Optional, Protocol, Set, runtime_checkable
 
 from agentclaw.community.core.task_queue.types import TaskOutcome
 
@@ -50,6 +50,10 @@ class HandlerRegistry:
         #: Normalized ``task_type`` → the exact spelling registered under it.
         #: Only used to reject near-collisions; lookup stays exact.
         self._normalized: Dict[str, str] = {}
+        #: Task types whose enqueues wake the worker immediately instead of
+        #: waiting out the idle poll. Opt-in per type — a type absent from this
+        #: set (every existing one) keeps the poll-interval behaviour exactly.
+        self._wake_on_enqueue: Set[str] = set()
 
     @staticmethod
     def _normalize(task_type: str) -> str:
@@ -60,7 +64,7 @@ class HandlerRegistry:
         """
         return task_type.casefold()
 
-    def register(self, handler: TaskHandler) -> None:
+    def register(self, handler: TaskHandler, *, wake_on_enqueue: bool = False) -> None:
         """Register ``handler`` under its ``task_type``.
 
         Raises :class:`ValueError` on a duplicate ``task_type`` — two handlers
@@ -93,6 +97,17 @@ class HandlerRegistry:
 
         Lookup stays **exact**: rows store ``task_type`` verbatim, so the worker
         dispatches on the spelling that was enqueued.
+
+        **``wake_on_enqueue``** opts this task type into immediate execution: a
+        due enqueue of this type signals the worker to poll at once rather than
+        waiting out ``poll_interval_seconds``. It defaults to ``False`` so every
+        already-registered type keeps its current timing untouched; turn it on
+        for work a user is waiting on, and leave it off for background and
+        recurring work where a couple of seconds costs nothing.
+
+        The flag lives here, at the wiring site, rather than on the handler:
+        how eagerly a queue is drained is a deployment concern, and one grep
+        for ``wake_on_enqueue=True`` then lists every type that opted in.
         """
         task_type = handler.task_type
         if task_type != task_type.strip():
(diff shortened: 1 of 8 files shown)


totalfrank (commented):
Reviewed commit `887ee70` (the only commit on this PR; no prior review to reconcile against).

Diff is clean — no findings.

**Files covered:** `services/wakeup.py` (new), `services/worker.py`, `services/task_queue_service.py`, `services/registry.py`, `di/modules/task_queue_module.py`, plus the three test files (`test_task_worker.py`, `test_worker_wakeup.py` (new), `test_task_queue_repository.py`).

**What I checked:**
- Correctness of the wake/coalesce/clear-after-wait latch logic, including the cross-thread `call_soon_threadsafe` requirement and the ordering of `notify()` after `repo.enqueue` commits.
- `_should_wake` gating (opt-in registry flag, `delay_seconds`, `created`) against the claim eligibility rules in the repository.
- Contract impact: `HandlerRegistry.register` and both constructors gained parameters with defaults/DI resolution — all call sites (production and test) are updated in this diff, and existing registrations are unaffected (`wake_on_enqueue` defaults to `False`).
- Lifecycle correctness: `bind`/`unbind` ordering in `startup`/`shutdown`, and behavior when the worker is disabled or the loop is already closed.
- `Optional[asyncio.AbstractEventLoop]` on `WorkerWakeup._loop` is a genuine domain state (no bound worker), not a defensive widening — consistent with the repo's `T | None` rule.
- Test coverage: new behavior (opt-in wake, default-off, delayed/keyed non-signaling, cross-thread wake, coalescing, end-to-end latency) all have direct tests.
- No schema, config, or architecture-boundary changes.

---

FreddieSun (approved):

PR #1383, opened 2026-08-23
fix(skills): align AICoding runtime probe identity

FreddieSun:
## Problem

Mixed Claude Code product Bots can run the AICoding implementation, but Skills Pool used a narrower template allowlist than provisioning. Architect and custom templates therefore probed the wrong physical layout. The operator rollback path also passed the logical engine and an invalid `owner_id` keyword to the runtime probe.

## Solution

- Share one pure `claude_code_uses_aicoding_runtime` policy between provisioning and Skills layout resolution.
- Cover every non-empty, non-`normalCC` Claude Code template while keeping native Claude Code unchanged.
- Make rollback probe use `user_id` and the resolved physical layout engine; keep logical engine semantics for catalog/asset queries.
- Add regression coverage for legacy, architect, custom, and rollback paths.

## Validation

- `uv run pytest -q tests/community/core/skill_center/test_runtime_resolver.py tests/community/core/bot_management/services/test_engine_resolver.py tests/community/core/bot_management/test_engine_provisioning_strategy.py tests/community/core/skills_pool/test_recovery_service.py tests/community/architecture` — 265 passed
- `uv run ruff check` on all changed Python files — passed
- `git diff --check` — passed

## Compatibility and risk

No schema or public API change. Logical `active_engine=claude_code` remains unchanged in persisted/catalog semantics; only filesystem probe identity is projected to `aicoding` where the provisioning policy already selects the AICoding implementation.

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -14,6 +14,9 @@ from typing import Any, Dict
 from agentclaw.community.core.bot_management.capabilities import (
     is_template_factory_config,
 )
+from agentclaw.community.core.workspace.runtime_identity import (
+    claude_code_uses_aicoding_runtime,
+)
 
 from agentclaw.community.plugin_api.secret_resolver import SecretResolver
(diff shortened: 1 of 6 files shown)


totalfrank (approved):
LGTM.

PR #1388, opened 2026-08-23
perf(backend): resolve the device once per MCP desired-state projection

totalfrank:
Follow-up to #1382, which added a concurrent fan-out to `sync_mcp_desired_state` but did not actually speed it up in prepub.

## Problem

Prepub logs show MCPs still landing strictly one at a time, each push starting only after the previous finished:

| MCP | POST | PUT done | starts after prev. finished |
|---|---|---|---|
| `hitl` | 13.314 | 13.540 | — |
| `bcs_mcp` | 13.584 | 13.803 | +44 ms |
| `uctmcptools` | 13.949 | 14.180 | +146 ms |
| `antcodemcp` | 14.232 | 14.454 | +52 ms |

`asyncio.gather` runs its tasks on **one** event loop; a task only yields at a real suspension point. Each `deliver()` spends ~270 ms in synchronous code before reaching its first one (`asyncio.to_thread(plugin.sync_single_mcp, ...)`), so tasks could enter the thread pool only single-file — one every ~270 ms, while each push finishes in ~226 ms. No two pushes ever overlapped, and the semaphore was never the binding constraint.

The blocking preamble, all inline on the loop:

- `sync_service.py:261` — `resolve_for_bot(...)`: a blocking `httpx` GET to secbaas ws-info (~112 ms in the log) plus DB queries. `HttpClient` is a synchronous transport by contract.
- `sync_service.py:789` — `build_mcp_sync_payload(...)`: a DB read.
- `skill_set_service.py` — its own `resolve_for_bot` and `sync_all_mcp_servers` for the allow-list, both blocking.

The device resolution is also **redundant**: every entry in one desired state targets the same bot, so N resolutions compute one answer N times.

## Solution

**Resolve once per batch.** New `MCPSyncService.sync_mcp_details_for_bot` resolves the device and dispatches the plugin once, then fans out over the existing private `_sync_mcp_detail(plugin=...)` — the shape `_sync_mcp_details` already uses. `sync_mcp_desired_state` delegates to it, so ws-info round trips go from N to 1.

`sync_mcp_detail` keeps its per-call resolve: it is a public single-MCP API and other callers (e.g. `add_mcp_to_skill_set`) pass a different bot each time.

**Keep the remaining blocking calls off the loop**, via `asyncio.to_thread`: `resolve_for_bot` in `sync_mcp_detail`, `build_mcp_sync_payload` in `_sync_mcp_detail`, and both the resolve and `sync_all_mcp_servers` in `sync_mcp_desired_state`. Independent of the fan-out, a coroutine that holds the loop for ~270 ms is a latency hazard for every other request on that worker.

Delivery concurrency stays at **10**, moved next to the new entrypoint as `_DESIRED_STATE_DETAIL_CONCURRENCY`. Note it is only now that the bound binds at all — previously ~1 push was ever in flight, so its value made no observable difference. `_sync_mcp_details` keeps its own bound of 5; that path collects its own MCP list and is left untouched.

Rejected: wrapping the per-MCP `resolve_for_bot` in `to_thread` and leaving it per-entry. It unblocks the loop just as well, but parallelizes work that should not exist — 14 concurrent ws-info calls and ~42 DB queries to learn one fact, against a shared 8-thread `to_thread` pool and a 5+10 connection pool.

## Validation

Ran from `src/backend` with the project venv:

- `pytest tests/community/core/skill_center tests/community/core/mcp tests/community/api/mcp tests/community/services tests/community/api/skill_center tests/community/integration/skill_center tests/community/contracts` — **1620 passed, 3 skipped**.
- `scripts/ci/python_sast_local.sh src/backend 1` over this PR's changed files — clean, exit 0.

New `TestSyncMcpDetailsForBot` (8 tests) covers resolve-once (14 entries → 1 `resolve_for_bot`, 1 `dispatch`, 14 pushes), blocking work running off the loop (asserted by thread identity), bounded overlap, per-entry failure reporting, exception and cancellation handling, empty batch, and unbound device. `TestSyncMcpDesiredState` was reshaped to the delegation contract.

Verified the off-loop test is meaningful by temporarily reverting both `to_thread` wraps to their inline form: `test_blocking_work_is_kept_off_the_event_loop` fails, the rest pass.

Not run locally: `singlebox_coverage.sh` and the full backend `ci_test.sh` coverage gate — the authoring sandbox cannot reach the configured `mirrors.aliyun.com` index, so the `uv sync --frozen` entrypoints do not start there. Both run here in PR CI.

## Compatibility and risk

`sync_mcp_desired_state` keeps its signature, `bool` return, and the deliberate empty-set `sync_all_mcp_servers([])` call. A failed delivery still withholds the allow-list declaration, so the reconciler still raises and retries. `sync_mcp_detail`, `sync_mcp_details`, and `remove_mcp_detail` are behaviourally unchanged apart from no longer blocking the loop.

Thread safety was checked before offloading: repositories use `with self._get_session() as db:` — a fresh short-lived session per call from a plain `sessionmaker`, not a shared `scoped_session` — so each session lives entirely within one thread. `asyncio.to_thread` copies contextvars, so trace ids still appear in worker-thread logs.

`MCPSyncServiceProtocol` gains `sync_mcp_details_for_bot`; the only implementation is `MCPSyncService`.

Rollback: revert the commit. No schema, config, or migration impact.

## Related issues

Follow-up to #1382.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DfkEDRAvMRAt41LafMuEDk)_


FreddieSun (approved):
LGTM

totalfrank (commented):
Reviewed commits `c4ca0a0` and `729c797` (head) — no findings.

**Files covered:** `mcp_sync_service.py`, `detail_fanout.py` (new), `sync_service.py`, `skill_set_service.py`, and their three test files.

**What I checked:**
- **Correctness** — `sync_mcp_details_for_bot`'s single resolve-then-fan-out shape, the extracted `fan_out_mcp_details` (`_sync_mcp_details` and the new entrypoint pass identical semantics: same `entity_id or user_id` fallback, same failure/exception/`CancelledError` handling as the pre-refactor code).
- **Contract breakage** — `MCPSyncServiceProtocol` gained `sync_mcp_details_for_bot`; confirmed `MCPSyncService` is its only binding (`mcp_module.py`), so no other implementer needed updating.
- **Concurrency/resource safety** — this is the part of the diff that matters most: moving `resolve_for_bot` and `build_mcp_sync_payload` into `asyncio.to_thread` makes them genuinely concurrent (previously serialized on one event-loop thread) for the first time. Traced the DB path: `DeviceRepository`/`BotRepository` open a fresh session per call via `CommunityDatabase.orm_session()`, backed by a plain `sessionmaker` (not `scoped_session`) — safe under real thread concurrency, matching the PR description's claim.
- **File-size guard** — `sync_service.py` is 994 lines, under the 1000-line cap the second commit exists to restore.
- **Python type contracts** — no required parameter widened to `Optional` for defensive reasons; `bot_id: str` stays required, `entity_id`/`engine_type` keep their pre-existing intentional-optional status.
- **Test coverage** — new `test_detail_fanout.py` and `TestSyncMcpDetailsForBot` cover resolve-once, off-loop execution (by thread identity), bounded overlap, per-entry failure/exception/cancellation, and empty/unbound-device cases; `TestSyncMcpDesiredState` was reshaped to match the delegation contract rather than re-testing the fan-out it no longer owns.
- Ran `ruff check` over all seven changed/added files — clean. Could not run the pytest suite myself (same `mirrors.aliyun.com` network restriction the PR description already notes); relying on the PR's stated local test results.

---

PR #1389, opened 2026-08-23
fix(backend): recover Bot scope for legacy SkillSet routes

FreddieSun:
## Problem

Legacy BFF SkillSet endpoints keep `bot_id` optional for backward compatibility. After the Phase 1 refactor, a missing `bot_id` was normalized to `default` before the exact SkillSet lookup. A SkillSet belonging to a non-default Bot therefore returned `404 Skill set not found` even when the caller supplied the correct `set_id` and had access.

This was observed on pre-production trace `0be8ed2017874785083472657ee6bf` for Set `1115285`. The old `dev` implementation happened to succeed by globally querying `set_id`, but restoring that unrestricted lookup would violate the Phase 1 Bot-scoped identity contract.

## Solution

- Preserve an explicitly supplied `bot_id` and keep strict `owner_id + bot_id` lookup semantics.
- When and only when a legacy exact-Set endpoint omits `bot_id`, resolve the ordinary Set address by `tenant`, `env`, and exact `set_id`.
- Validate the owner hint, Bot existence, and caller authorization in the control plane before returning the recovered scope.
- Re-enter the existing strict Bot-scoped repository path for the actual read or mutation.
- Apply the compatibility seam consistently to legacy Set, Skill, MCP, switch, sync, activate, and deactivate endpoints.
- Keep System Default sets and canonical OpenAPI Bot-scoped routes unchanged.

## Validation

- `1101 passed`: relevant Skill Center HTTP, API, core, repository, and Gateway Rule 15 suites.
- `49 passed`: final focused control-plane service suite after cleanup.
- `196 passed`: complete Backend architecture suite.
- Ruff passed for all changed Python files.
- Backend local SAST block scan passed.
- `git diff --check` passed.
- `src/backend/scripts/dump_openapi.py` completed successfully.
- `src/gateway/scripts/gate_and_publish_openapi.py` was run without `--allow-breaking`. It remains blocked by an existing target-branch artifact drift: `POST /openapi/v1/bots/{bot_id}/public-bcs` is present in the published Gateway artifact but absent from the Backend-generated candidate. This PR does not modify that route or its schema.

## Compatibility and risk

No public parameter, response schema, database schema, or canonical route changes are introduced. Existing callers that provide `bot_id` remain strictly scoped and are never redirected. Missing `bot_id` recovery is limited to an exact ordinary `set_id` and still enforces Bot authorization; System Default sets retain the previous behavior because no Bot can be inferred safely.

Rollback is the single commit in this PR.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
+++ b/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
@@ -91,6 +91,9 @@ from agentclaw.community.adapters.http.skill_center.schemas import (
     UploadSkillResponse,
     VersionListResponse,
 )
+from agentclaw.community.core.skill_center.legacy_skill_set_compatibility import (
+    recover_legacy_skill_set_scope,
+)
 from agentclaw.community.api.bot_service import BotServiceProtocol
 from agentclaw.community.api.skill_set_control_plane import (
     SkillSetControlPlaneServiceProtocol,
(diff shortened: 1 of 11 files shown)


totalfrank (approved):
LGTM.

PR #1391, opened 2026-08-23
fix(bcs): support provider bot visibility attributes

yhzhang35:
## Problem

The provider-admin Bot attributes endpoint rejected Backend agent-scope updates because it did not accept the existing Bot `visibility` property. The Backend sends `{"visibility":"private"}` for `public_scope=agent`, which caused a 400 response.

## Solution

- Expose `visibility` alongside `user_visibility`, `friend_ext`, and `friend_check_in_strategy` in the provider-admin GET/PATCH contract.
- Validate `public`, `protected`, and `private`, then reuse the existing `bcs_bots.visibility` persistence path.
- Preserve strict unknown-field rejection, existing provider-admin-token authorization, and the independent semantics of `user_visibility`.

## Compatibility and risk

No DDL, schema, authentication, or public-route change. `visibility` remains separate from `user_visibility`; the patch reuses the pre-existing control-plane/store behavior.

## Validation

- `cargo test -p bcs-http --test provider_routes_contract` — 41 passed
- `cargo test -p bcs-app-bot` — 21 passed
- `cargo test -p bcs-bot-store --test conformance_bot_control_plane_repo` — 14 passed
- `git diff --check` — passed

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/providers.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/providers.rs
@@ -45,6 +45,8 @@ pub struct ProviderStreamGrayResponse {
 #[serde(deny_unknown_fields)]
 pub struct PatchProviderBotAttributesRequest {
     #[serde(default, deserialize_with = "deserialize_present_non_null")]
+    pub visibility: Option<String>,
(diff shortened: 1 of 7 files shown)


vzvince (approved):
GLTM

PR #1420, opened 2026-08-24
refactor(backend): centralize bot creation policy in core

lucas-xzp:
## Problem

Bot creation policy and Application Coding preflight logic were spread across HTTP routers. This made the internal and OpenAPI authorization paths vulnerable to policy drift, while provider-specific workspace details leaked into the public contract and implementation notes.

## Solution

- Move shared Bot creation preparation and Application Coding policy into the transport-agnostic Core create flow.
- Run the same preflight for synchronous creation and authorization completion.
- Keep transport adapters responsible for mapping nested request DTOs into the Core create specification.
- Group Application Coding configuration under the nested `template.properties` object and keep its public description provider-neutral.
- Validate supplied Application Coding configuration before Passport, workspace, or persistence side effects.
- Keep creation-domain errors in the existing `bot_management.errors` module so delivery adapters preserve their 422/409/503 mappings without adding a one-off error module.
- Keep creation policy with its sole orchestration consumer in `create_flow.py` instead of adding a separate policy module.
- Remove internal implementation-plan documents and provider-specific workspace field names from this change.
- Preserve existing internal behavior for legacy template callers and keep the current auth-status echo mechanism; no create-intent table or migration is introduced.

## Validation

- `32 passed`: Application Coding Core policy, pending Passport flow, and first-Bot Passport selection tests.
- `17 passed, 280 deselected`: OpenAPI Application Coding, authorization-completion, template pass-through, and router scenarios.
- `5 passed`: module-boundary, Core transport-independence, and domain-error mapping architecture tests.
- `9 passed`: public configuration/vendor-boundary and Core plugin-isolation architecture tests.
- Repository pre-push SAST/lint gate passed for Backend changes.
- `git diff --check` and Python compile checks passed.
- Verified that this PR adds no `DIMA`, `dima_*`, or `corp-only` identifiers.

The full Backend unit suite and Singlebox coverage gate were not run locally because this change was validated with the closest affected tests; CI remains responsible for the complete module gates.

## Compatibility and risk

- No database migration is introduced.
- Existing plain-Bot and legacy internal template flows retain their current behavior.
- Pending authorization continues to use the existing client echo contract; durable pending-intent recovery and strict concurrent poll claiming remain out of scope.
- Workspace-provider lifecycle extensions remain out of scope until a public, versioned provider-neutral contract exists.
- Rollback is code-only.


xianmuyq (dismissed):

totalfrank (approved):
LGTM.

PR #1421, opened 2026-08-24
fix(devices): break BotPublicService -> DeviceContextResolver DI cycle

totalfrank:
## Problem

`SessionResourceTaskLifecycle.bootstrap` (`session_resources/materialization.py`) resolves `SessionResourceMaterializeHandler`, which eagerly injects `DeviceContextResolver`. `DeviceContextResolver`'s `ArcaConnInfoBuilder` dependency transitively resolves a `DeviceService` binding, and in profiles where the `device_service` provider itself depends on `BotPublicService` (e.g. the corp ARCA-aware infrastructure column), the graph closes a cycle back through `bot_public_module.bot_public_service`, which injected `DeviceContextResolver` eagerly:

```
SessionResourceMaterializeHandler -> DeviceContextResolver -> ArcaConnInfoBuilder
  -> DeviceService -> BotPublicService -> DeviceContextResolver (cycle)
```

`injector` raises `CircularDependency` during bootstrap, and this crashes on startup wherever that profile's `device_service` provider closes the loop.

## Solution

`BotPublicService` only reaches `DeviceContextResolver` at call time, inside `sync_bot_config_to_device` (`ctx = self._resolver.resolve_for_bot(...)`) — never during construction. So the eager dependency is unnecessary: `bot_public_module.py` now binds a `Callable[[], DeviceContextResolver]` lazy factory (`injector.get(DeviceContextResolver)` deferred to call time) instead of the resolver itself, and `BotPublicService` stores/calls that factory instead of a resolved instance. This mirrors the existing `DataInitService <-> DeviceService` cycle-break pattern already used in `bot_management_module.py` (`data_init_service_factory`).

No corp-side code lives in this repo, so this fix addresses the community-side edge that any profile's DI graph shares; it removes the shared circular edge regardless of which `device_service` provider a given profile installs.

## Validation

- `pytest tests/community/core/bot_public/ tests/community/core/devices/ tests/community/core/session_resources/ tests/community/di/` — 997 passed.
- Updated the two test call sites (`test_bot_public_service.py`, `test_sync_bot_config_uses_resolver.py`) that constructed `BotPublicService` directly to pass `device_context_resolver_factory` instead of `device_context_resolver`.
- Could not reproduce the corp-side `CircularDependency` traceback directly (the corp `device_service` provider that closes the cycle isn't in this repo); validated instead that the community DI container still wires and resolves `BotPublicService`/`DeviceContextResolver`/`SessionResourceMaterializeHandler` cleanly after the change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CNahvHzUo9T9b5fFoM6s1i)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_public/services/bot_public_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_public/services/bot_public_service.py
@@ -8,7 +8,7 @@ import time
 import threading
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime
-from typing import Any, Dict, List, Optional, TYPE_CHECKING
+from typing import Any, Callable, Dict, List, Optional, TYPE_CHECKING
 
 from agentclaw.community.plugin_api.approval_workflow import ApprovalWorkflowPlugin
 from agentclaw.community.core.repository.protocols.bot import BotRepository
@@ -133,7 +133,7 @@ class BotPublicService:
         auth_relationship_plugin: AuthRelationshipPlugin,
         publish_approval_plugin: BotPublishApprovalPlugin,
         skill_set_service_factory: "SkillSetServiceFactory",
-        device_context_resolver: DeviceContextResolver,
+        device_context_resolver_factory: "Callable[[], DeviceContextResolver]",
         device_sync_dispatcher: "DeviceSyncDispatcher",
         catalog_metadata_service: BotCatalogMetadataServiceProtocol,
     ) -> None:
@@ -148,7 +148,10 @@ class BotPublicService:
         self._skill_set_service_factory = skill_set_service_factory
         # Task 2.1: resolver + dispatcher 替代 device_sync_supplier 在
         # sync_bot_config_to_device 路径上的角色。Task 6 收口后 supplier 已删。
-        self._resolver = device_context_resolver
+        # Lazy: DeviceContextResolver's conn-info builders transitively reach
+        # DeviceService, which (corp) depends on BotPublicService itself —
+        # eager injection here would cycle. Resolved at call time instead.
+        self._resolver_factory = device_context_resolver_factory
         self._device_sync_dispatcher = device_sync_dispatcher
         self._catalog_metadata_service = catalog_metadata_service
         self._sync_lock = threading.Lock()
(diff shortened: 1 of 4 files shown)


totalfrank (commented):
Reviewed commit `b41740c` (the only commit on this PR).

The fix follows the codebase's existing cycle-break convention exactly (`lambda: injector.get(X)` behind a `Callable[[], X]` provider, same shape as `data_init_service_factory`, `bot_publish_service_factory`, `device_service_factory` in `bot_management_module.py`, and the several other `lambda: injector.get(DeviceContextResolver)` sites in `skill_center_module.py` / `bot_collaborator_module.py` / `mcp_module.py`). `DeviceContextResolver` is bound `@singleton` in `devices_module.py`, so deferring resolution doesn't change identity or add allocation cost per call.

Checked:
- **Correctness** — `self._resolver_factory()` is called once per `sync_bot_config_to_device` invocation and immediately chained with `.resolve_for_bot(...)`; behavior is equivalent to the prior eager-resolved `self._resolver.resolve_for_bot(...)`.
- **Contract breakage** — `BotPublicService.__init__` signature changed (`device_context_resolver` → `device_context_resolver_factory`); grepped for all construction sites (`BotPublicService(` and `device_context_resolver=`) across the repo and confirmed every call site (the DI provider and both test helpers) was updated consistently. No missed caller.
- **DI wiring** — the new `device_context_resolver_factory` provider is correctly wired into `bot_public_service`'s provider signature and passed through.
- **Test coverage** — both test helpers (`test_bot_public_service.py`, `test_sync_bot_config_uses_resolver.py`) correctly capture the pre-configured mock inside the factory closure before it's passed in, preserving existing assertions.
- Error/failure handling, concurrency, security, architecture boundaries — not touched by this diff beyond the above.

No blocking or non-blocking findings. Not filing a nit for the still-quoted `"Callable[[], DeviceContextResolver]"` annotation (line 136) — `DeviceContextResolver` is imported unconditionally in this file, unlike the other TYPE_CHECKING-only quoted params in the same signature, but it's inert either way since this class is explicitly provider-constructed, never `binder.bind`-inspected.

Could not run `pytest` in this sandbox (dependency mirror unreachable), so relied on static tracing of every call site instead.

---

FreddieSun (approved):
LGTM

PR #1423, opened 2026-08-24
fix(gateway): fail CI on code format drift

### LINKED ISSUE #1422: Gateway CI does not fail on code format drift
## Problem

The gateway CI never blocks unformatted or lint-dirty code. Two compounding causes:

1. **`run_check_basic`** (`src/gateway/scripts/lib/checks.sh`) runs `ruff` in `--fix`/write mode (`ruff check --select I --fix`, `ruff format .`, `ruff check . --fix`), which silently rewrites files instead of failing when a violation exists.
2. **The only check-only enforcement** — `tests/architecture/test_ruff_lint_rules.py` (runs `ruff check .` and `ruff format --check .`) — is orphaned: `run_arch_tests` exists in `test-stages.sh` but is never wired into `run_ci_pipeline` (`pipeline.sh`) or any GitHub workflow.

Additionally, `scripts/ci/pre_push.sh` routes gateway changes through `run_heavy`, which is skipped in the default lint-only pre-push mode, so a `git push` of unformatted gateway code is never blocked.

## Why it matters

- Format and lint drift accumulates silently and merges into the branch without any gate failing.
- Confirmed drift already existed on the branch: 6 tracked files needed reformatting and 1 had an unsorted import (surfaced once a non-fixing check ran).
- The pre-push hook and CI should enforce the same check, so local and remote behave consistently.

## Fix

- Add a non-fixing check-only format/lint gate (`ruff check .` + `ruff format --check .`) as a required first stage of the gateway CI pipeline.
- Surface the same gate in the pre-push hook's default lint-only mode.

cassiuscai:
## Problem

The gateway CI never blocked unformatted or lint-dirty code. Its check stage ran `ruff` in `--fix` mode (silently rewriting files), and the only check-only enforcement (`tests/architecture/test_ruff_lint_rules.py`) was never wired into the pipeline. Additionally, `scripts/ci/pre_push.sh` skipped gateway changes in lint-only mode, so pushes weren't blocked either.

## Solution

- Added `run_check_format_ci` (`src/gateway/scripts/lib/checks.sh`): a non-fixing `ruff check .` + `ruff format --check .` gate that fails on any violation.
- Wired it as a required first stage of `run_ci_pipeline` (`src/gateway/scripts/lib/pipeline.sh`).
- Added `src/gateway/scripts/ci_check_format.sh` as a standalone hook entrypoint.
- Updated `scripts/ci/pre_push.sh` so the gateway format/lint gate runs via `run_required` in the default lint-only mode.
- Reformatted 6 pre-existing drifted files so the gate passes.

## Validation

- `./scripts/ci_check_format.sh` returns exit 1 with an introduced format violation and 0 after revert.
- `run_ci_pipeline` aborts with "check-format-ci failed — aborting" on a violation.
- `ruff format --check .` and `ruff check .` now pass on the gateway tree.
- Bash syntax check (`bash -n`) passed for all changed scripts.

## Related issues

- Fixes https://github.com/inclusionAI/Avernet/issues/1422

### DIFF
--- a/src/gateway/scripts/ci_check_format.sh
+++ b/src/gateway/scripts/ci_check_format.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Check-only format/lint gate for the gateway module.
+#
+# Runs `ruff check .` and `ruff format --check .` (non-fixing) so that format
+# and lint drift fail the gate with a non-zero exit. Unlike `check-basic`
+# (which runs ruff in `--fix` mode for local developer convenience), this gate
+# never mutates files.
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+gateway_dir="$(cd "$script_dir/.." && pwd)"
+
+if ! command -v uv >/dev/null 2>&1; then
+  echo "gateway format gate failed: uv not found" >&2
+  exit 127
+fi
+
+cd "$gateway_dir"
+
+echo "[CHECK] check-format-ci: ruff check . && ruff format --check ."
+
+uv run ruff check .
+uv run ruff format --check .
+
+echo "gateway format gate passed"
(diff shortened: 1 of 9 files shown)


totalfrank (approved):
LGTM.

PR #1426, opened 2026-08-24
test(openapi): endpoint cases for the public bot surface (132 → 24 tracked gaps)

totalfrank:
## Problem

A circular-dependency crash (#1421) reached runtime on a surface that looked well tested, which raised the question of how much of the public API actually has *endpoint* coverage.

I inventoried all **246 operations** on the assembled public surface against the test tree. The gap was not missing tests — it was missing tests *at the endpoint level*:

- Groups like routines and resources had thorough handler suites (`resources/test_resources_handlers.py` alone is 1,928 lines).
- Every test in them calls `await handler(...)` with each dependency passed by keyword, so FastAPI's dependency wiring and the injector never run.
- That is exactly where assembly failures live: a binding that does not resolve, a router that is not mounted, a query parameter that does not bind.

The repo already had the right mechanism — the coverage gate in `tests/community/framework/`, which requires a happy **and** an error case per live endpoint. But **132 `/openapi/v1` operations** sat in `coverage_baseline.txt` as frozen debt, each blaming one thing: the case runner authenticates with `x-user-id`, the surface requires a gateway-signed principal, and *"the harness has no minter, so a case could assert nothing but a 401."*

That reason had expired. `test_openapi_session_files.py` mints a principal in the test tree — point `init_principal_verifier_config` at a local signing key and hand the runner a token signed with it.

## Solution

Ten new case modules, using the house idiom to pay down the debt rather than adding a parallel mechanism:

| Module | Group | Ops |
|---|---|---|
| `test_openapi_routines.py` | routines | 7 |
| `test_openapi_resources.py` | resources | 7 |
| `test_openapi_bots.py` | bots | 15 |
| `test_openapi_sessions.py` | sessions | 7 |
| `test_openapi_engine_runtime.py` | approvals / connection / engine / models | 10 |
| `test_openapi_service_publications.py` | lifecycle + edit-lock | 14 |
| `test_openapi_local.py` | local bots | 9 |
| `test_openapi_mcp.py` | MCP catalogue | 6 |
| `test_openapi_bot_identity.py` | bot identity | 3 |
| `test_openapi_bot_skills_read.py` | bot skills reads | 2 |
| `test_openapi_authorized_apps.py` | authorized apps | 4 |
| `test_openapi_legacy_routines_resources.py` | retiring routines/resources addresses | 14 |

Every operation gets a happy case and the error it **genuinely owes**, driven through the assembled application with the full DI container. Where an operation publishes no `user_id` — `check-name`, the MCP catalogue reads — the error case is the real one it owns (400 on an invalid name, 404 on an unknown server, 502 when the marketplace is down) rather than a user-scope 403 it would never raise.

Happy cases pin a response fragment as well as a status, so a handler mounted at the right address but projecting the wrong payload fails rather than passing on a bare 200.

**`coverage_baseline.txt`: 132 → 24 tracked `/openapi/v1` gaps.** All 24 that remain are *retiring* addresses (legacy sessions/approvals/connection/engine/models/identity/skills spellings, plus `auth-status` and `engine-config`), which disappear with the `deprecated` package at the 2027-08-15 sunset. They are listed honestly rather than quietly dropped.

## Validation

- `pytest tests/community/endpoints` → **1252 passed, 0 failed**
- `pytest tests/community/framework` → **125 passed**, including `test_no_mock_in_endpoint_tests` and `test_isolation`
- `pytest tests/community/framework/test_coverage_gate.py` → **green**: zero NEW gaps, zero stale entries
- `ruff check` on all 10 new files → clean
- No production code changed; the diff is tests plus the baseline.

The gate corroborated each step independently: before every baseline edit it failed naming exactly the lines that had become covered, which is why the baseline is hand-edited (with dated notes, per the file's own convention) rather than regenerated.

Two caveats stated plainly:

1. An early path-literal scan of mine reported 13 endpoints with "zero coverage" and **that was wrong** — it produced false negatives against suites that compose URLs from a base constant or use no URLs at all. Every number here is one the coverage gate itself confirms.
2. These cases are not claimed to catch the specific `BotPublicService` cycle from #1421; they exercise the DI graph for their own endpoints. The general point stands — assembly-level cases catch a class of failure that handler-level tests structurally cannot.

## Compatibility and risk

Test-only. The baseline shrinks, so the gate becomes strictly stricter and cannot silently regrow these entries.


FreddieSun (dismissed):
LGTM

src/backend/tests/community/endpoints/test_openapi_bots.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+for _method, _path, _input, _status in _HAPPY_CASES:
+    endpoint_test(
+        method=_method,
+        path=_path,
+        scenario="happy",
+        input=_input,
+        seed=_seed_happy_services,
+        expect=ExpectSuccess(status=_status),
```
totalfrank: Non-blocking: all 13 happy cases in `_HAPPY_CASES` (lines 199-299 — `list`, `create`, `all`, `ceiling`, `get`, `update`, `delete`, `status`, `passport`, `engine/config` get+put, `activate`, `restart`, `data-init`) register with `ExpectSuccess(status=_status)` and no `json_contains`/`json_equals`. Only the status code is checked, so if `get_bot`/`list_bots`/`get_bot_status`/`get_bot_passport` etc. returned a structurally-valid but wrong body — stale `bot_name`, wrong `owner_id`, an empty list where `bots-bot` should appear — every one of these cases would still pass.

This also cuts against the PR's own stated design principle ("Happy cases pin a response fragment as well as a status, so a handler mounted at the right address but projecting the wrong payload fails rather than passing on a bare 200") — the sibling files in this PR (`test_openapi_authorized_apps.py`, `test_openapi_bot_identity.py`) do pin fragments for every happy case; this file's loop doesn't.

Suggested fix: extend each `_HAPPY_CASES` row with a `json_contains` fragment, e.g. `{"data": {"bot_id": _BOT_ID, "owner_id": _OWNER}}` for the bot-scoped reads/writes, `{"data": {"items": [...]}}`-shaped for `list`/`all`.

---

src/backend/tests/community/endpoints/test_openapi_resources.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+for _method, _path, _input, _status in _HAPPY_CASES:
+    endpoint_test(
+        method=_method,
+        path=_path,
+        scenario="happy",
+        input=_input,
+        seed=_seed_happy_services,
+        expect=ExpectSuccess(status=_status),
```
totalfrank: Non-blocking: all 7 happy cases in `_HAPPY_CASES` (lines 182-252 — `list`, `stat`, `download`, `preview`, `upload`, `mkdir`, `delete`) share this one `ExpectSuccess(status=_status)` with no `json_contains`. The module docstring (lines 18-20) says `download` is *"the one operation here whose success body is not an envelope … so its happy case asserts the status alone"* — implying the other six do assert body content. They don't; line 262 applies the bare-status expectation uniformly to all seven. A regression that returned an empty `list_dir`/`stat` payload, or mixed up which path's metadata came back, would pass unnoticed.

Suggested fix: give `list`/`stat`/`preview`/`upload`/`mkdir`/`delete` their own `json_contains` (e.g. asserting the seeded path/name shows up in the listing or stat result), leaving only `download` on status-only per the stated rationale.

---

src/backend/tests/community/endpoints/test_openapi_routines.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+for _method, _path, _input, _status in _HAPPY_CASES:
+    endpoint_test(
+        method=_method,
+        path=_path,
+        scenario="happy",
+        input=_input,
+        seed=_seed_happy_services,
+        expect=ExpectSuccess(status=_status),
```
totalfrank: Non-blocking: all 7 happy cases in `_HAPPY_CASES` (lines 162-223) use `ExpectSuccess(status=_status)` with no `json_contains`. `_routine()` and `_run()` (lines 86-108) build detailed fixtures — `id`, `schedule`, `payload`, run `status`/`duration_ms`, etc. — that look purpose-built to express the expected response shape, but they're only ever used as seed return values (bound via `bind_overrides`), never compared against the actual response body. A handler that dropped `schedule`/`payload` or returned the wrong `routine_id` in its projection would still pass every one of these.

Suggested fix: assert a `json_contains` fragment built from the same `_routine()`/`_run()` fixtures for each happy case, e.g. `{"data": {"id": _ROUTINE_ID, "name": "morning-brief"}}`.

---

src/backend/tests/community/endpoints/test_openapi_legacy_routines_resources.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+for _method, _path, _input, _status in _ROUTINE_CASES:
+    endpoint_test(
+        method=_method,
+        path=_path,
+        scenario="happy",
+        input=_input,
+        seed=_seed_routines,
+        expect=ExpectSuccess(status=_status),
```
totalfrank: Non-blocking: both happy-case loops in this file — `_ROUTINE_CASES` here (lines 217-225) and `_RESOURCE_CASES` below (lines 242-250) — register with `ExpectSuccess(status=_status)` only, no `json_contains`, across all 14 legacy happy cases. The module's stated purpose is proving the legacy shim reaches "the *same handler* with the *same dependencies*" as the modern address; but the shim mechanism is bespoke signature-surgery (`inspect.signature` rewriting / `with_query_parameter` rebuilding kwargs), and a bug there — e.g. silently dropping a re-annotated field or passing a stale default — could produce the same status code with a mangled or empty body. Status-only assertions wouldn't catch that, which is exactly the class of regression this file's docstring implies it guards against.

Suggested fix: pin at least one response field per legacy case (mirroring what `test_openapi_routines.py`/`test_openapi_resources.py` would assert once they add their own body checks), so a shim-introduced projection bug is distinguishable from a genuine 200/201.

---

src/backend/tests/community/endpoints/test_openapi_mcp.py
```diff
@@ -0,0 +1,389 @@
...
+        CaseInput(headers=_HEADERS),
+        200,
+        {"data": [{"code": "t1", "categories": ["cat-a"]}]},
+    ),
+    (
+        "GET",
+        f"{_BASE_PATH}/servers/{{server_code}}",
+        CaseInput(path_params=_SERVER_PATH_PARAMS, headers=_HEADERS),
+        200,
+        {"data": {"server_code": _SERVER_CODE, "transport_protocol": "SSE"}},
```
totalfrank: Non-blocking: `_server_record()`'s docstring (lines 104-107) says *"The `extInfo` key is deliberate: the detail projection strips it, and a record without one could not show that."* But the happy case for `GET /openapi/v1/bots/mcp/servers/{server_code}` only asserts `{"data": {"server_code": _SERVER_CODE, "transport_protocol": "SSE"}}` — `"tools"` never appears in any `json_contains` in this file. If `strip_ext_info` regressed and stopped removing `extInfo` from the response (or the detail projection stopped populating `tools` at all), this test would still pass. The comment documents intent to cover extInfo-stripping but the assertion doesn't check it.

Suggested fix: add `"tools"` to the `json_contains` fragment with the `extInfo` key absent, e.g. `{"tools": [{"name": "get", "inputSchema": {"properties": {"q": 2}}}]}`, so a regression that stops stripping it fails the case.

---

totalfrank (commented):
Reviewed commits 5ecf5bc, 897cb7c, ff9a589 (all new — this is the first review round). Scope: 12 new `tests/community/endpoints/test_openapi_*.py` modules plus the `coverage_baseline.txt` edit.

**Verified locally** (`uv sync --default-index https://pypi.org/simple`, then `pytest`):
- `tests/community/framework` — 125 passed (includes `test_coverage_gate.py`, `test_no_mock_in_endpoint_tests.py`, `test_isolation.py`)
- `tests/community/endpoints` — 1252 passed
- Cross-checked the `coverage_baseline.txt` diff against the added test modules; the removed lines correspond to operations actually covered, matching the PR's 132 → 24 claim, and the gate is green.

**Categories checked:** correctness of seeded state vs. what each handler reads (owner/bot/tenant), JWT/principal construction against the real verifier (`iss`/`aud`/`principals` shape), status-code mapping against router source, `require_user_id` 403 short-circuit ordering, no forbidden mocking (`unittest.mock`/`monkeypatch`/`mocker`/`setattr` — none found), no duplicate `(method, path, scenario)` registrations, and the legacy-address shim genuinely reusing the same handler/dependencies. All checked out against production source.

**Findings (all non-blocking):** a recurring gap where happy-path cases assert only the HTTP status via `ExpectSuccess(status=...)` with no `json_contains`, so a handler that kept the right status but returned a wrong-but-shaped payload wouldn't be caught. This spans `test_openapi_bots.py` (13 cases), `test_openapi_resources.py` (7 cases, and contradicts its own docstring's implication that only `download` is status-only), `test_openapi_routines.py` (7 cases, with detailed `_routine()`/`_run()` fixtures built but never asserted against the response), `test_openapi_legacy_routines_resources.py` (14 cases), and one narrower case in `test_openapi_mcp.py` where a docstring claims extInfo-stripping is covered but no assertion checks for it. Left as five inline comments with suggested fixes. The other seven new files (`authorized_apps`, `bot_identity`, `bot_skills_read`, `engine_runtime`, `local`, `mcp` elsewhere, `service_publications`, `sessions`) consistently pin response fragments and don't have this gap.

---

src/backend/tests/community/endpoints/test_openapi_mcp.py
totalfrank: Fixed in 42547d5 — but the suggested fragment would not have bitten, which is worth recording.

`extInfo` sits nested at `tools[].inputSchema.properties`, and `json_contains` is a recursive **subset** match (`runner.py::_is_subset` — for dicts, every key in the subset must exist in the superset; extra keys are fine). So `{"properties": {"q": 2}}` matches just as happily when `extInfo` is still sitting beside `q`. Pinning surviving keys cannot express the absence of a stripped one.

So the case now does two things:

- pins `"tools": [{"name": "get"}]` in `json_contains`, so the case genuinely reaches the payload the strip acts on (a detail projection that stopped populating `tools` now fails);
- adds an `extra_assertions` hook, `_extinfo_never_reaches_the_client`, which reads the whole response for the key — the only assertion shape that fails when the strip regresses.

Mutation-checked rather than assumed: inverting the assertion to `assert "extInfo" in response.text` fails against the current build, so the seeded record's `extInfo` really is being stripped and the assertion would fire if it stopped.


---

src/backend/tests/community/endpoints/test_openapi_legacy_routines_resources.py
totalfrank: Fixed in 42547d5, and taken a step further than pinning "at least one field".

Since the property this file exists to assert is that the shim reaches the *same handler with the same dependencies*, each legacy case now pins the **same fragment its modern counterpart pins** — imported from `test_openapi_routines.py` / `test_openapi_resources.py` rather than restated, so the two cannot drift into asserting different things about one handler.

The pairing is positional, which would be a latent trap on its own, so it is `zip(..., strict=True)` plus a guard asserting the method and status agree; reordering either list fails loudly at import rather than silently pairing the wrong expectation with the wrong address.


---

src/backend/tests/community/endpoints/test_openapi_bots.py
totalfrank: Fixed in 42547d5. All 14 happy cases now pin a fragment via a `_HAPPY_BODIES` lookup keyed by address (the same shape `test_openapi_engine_runtime.py` already uses in this PR), so the `_HAPPY_CASES` rows stay readable.

Bodies were taken from the real responses rather than guessed — I probed each case with a deliberately-unsatisfiable fragment and read the actual payload out of the failure output, then pinned from that.

One deviation from the suggestion: `create` pins `bot_name`/`owner_entity_id` but **not** `bot_id`, because the surface mints the id (`20260824_...`), so it isn't knowable from the case. Also the field is `owner_entity_id`, not `owner_id`, on this projection.

You were right that this cut against the PR's own stated principle — that sentence was accurate for the sibling files and not for this one, which is exactly the kind of gap worth catching.


---

src/backend/tests/community/endpoints/test_openapi_resources.py
totalfrank: Fixed in 42547d5, exactly as suggested: `list`, `stat`, `preview`, `upload`, `mkdir` and `delete` each pin a fragment now, and `download` keeps the status-only expectation as the sole `None` — so the module docstring's claim about it being "the one operation" is now true rather than aspirational.

The listing and `stat` pin the seeded path and name (`docs/a.txt` / `a.txt`), so an empty `list_dir` or one path's metadata answered for another now fails; `upload` and `mkdir` pin the *target* path they created (`docs/new.txt`, `docs/sub`) rather than the existing one, which is what distinguishes a real write from an echo of the seed.


---

src/backend/tests/community/endpoints/test_openapi_routines.py
totalfrank: Fixed in 42547d5 — all 7 cases pin a fragment now.

You spotted the real oddity: `_routine()` and `_run()` described the expected shape and were never compared against anything. They now have a counterpart on the assertion side, taken from the real responses rather than from the fixtures directly, because the handler *renames* on the way out — the adapter's `id` is `routine_id` on the wire, and `schedule: {kind, expr, tz}` becomes `trigger: {type, cron}`. Asserting the fixture's own key names would have failed, and asserting them loosely would have missed the mapping entirely.

The `run` case is the one most worth having: upstream `status: "dispatched"` is `"completed"` on the wire, so that fragment now holds the handler's status mapping, which nothing else here covered.


---

src/backend/tests/community/endpoints/test_openapi_bots.py
totalfrank: Confirmed fixed in 42547d5 — `_HAPPY_BODIES` now pins a fragment for all 14 cases keyed by `(method, path)`, matching the seeded record's fields. Verified locally: `pytest tests/community/endpoints -q` → 1252 passed, `pytest tests/community/framework -q` → 125 passed (coverage gate green), `ruff check` clean. Resolving.

---

src/backend/tests/community/endpoints/test_openapi_resources.py
totalfrank: Confirmed fixed in 42547d5 — all 7 `_HAPPY_CASES` rows now carry a `json_contains` fragment; only `download` stays status-only, matching what the module docstring actually claims for it. Verified against a green `pytest tests/community/endpoints -q` (1252 passed). Resolving.

---

src/backend/tests/community/endpoints/test_openapi_routines.py
totalfrank: Confirmed fixed in 42547d5 — all 7 `_HAPPY_CASES` rows now assert a `json_contains` fragment built from the `_routine()`/`_run()` fixtures. Verified against a green `pytest tests/community/endpoints -q` (1252 passed). Resolving.

---

src/backend/tests/community/endpoints/test_openapi_legacy_routines_resources.py
totalfrank: Confirmed fixed in 42547d5 — both legacy loops now pin the same fragment as their modern counterpart, and the `zip(..., strict=True)` plus method/status guard would fail loudly on drift. Verified against a green `pytest tests/community/endpoints -q` (1252 passed). Resolving.

---

src/backend/tests/community/endpoints/test_openapi_mcp.py
totalfrank: Confirmed fixed in 42547d5, and correctly diagnosed the deeper issue — `extra_assertions=_DETAIL_ASSERTIONS` now checks `"extInfo" not in response.text`, which is the one assertion shape that can express absence rather than a `json_contains` subset match. Verified against a green `pytest tests/community/endpoints -q` (1252 passed). Resolving.

---

totalfrank (commented):
Round 2. Commits reviewed this round: 42547d5 (the only new commit since the last review, on top of 5ecf5bc/897cb7c/ff9a589 which the round-1 review already covered).

**Part A — verifying round-1 findings**

All 5 findings from round 1 are fixed. Verified by reading the current code (not by the commit message or the author's replies) and by a clean local run: `pytest tests/community/endpoints -q` → 1252 passed, `pytest tests/community/framework -q` → 125 passed (coverage gate green, zero new/stale entries), `ruff check` on the 5 changed files → clean.

- [x] `test_openapi_bots.py` — 13 happy cases missing `json_contains` → fixed via `_HAPPY_BODIES` keyed by `(method, path)`. Replied + resolved.
- [x] `test_openapi_resources.py` — 6 of 7 happy cases missing `json_contains` → fixed, `download` correctly stays status-only (raw octet-stream body). Replied + resolved.
- [x] `test_openapi_routines.py` — 7 happy cases missing `json_contains` → fixed, fragments built from the `_routine()`/`_run()` fixtures. Replied + resolved.
- [x] `test_openapi_legacy_routines_resources.py` — 14 legacy happy cases missing `json_contains` → fixed, each pins the same fragment as its modern counterpart via a `zip(..., strict=True)` + guard. Replied + resolved.
- [x] `test_openapi_mcp.py` — extInfo-stripping claimed but not asserted → fixed with a correction to my own suggestion: `json_contains` is a subset match and can't express absence, so the fix uses an `extra_assertions` hook reading the raw response text instead. Replied + resolved.

**Part B — reviewing the new commit**

42547d5 touches exactly the 5 files above, all in service of the round-1 findings. No new logic surface beyond the fragments/assertions covered above — no new correctness, error-handling, concurrency, security, or architecture-boundary issues found. No findings filed.

**Open count: 0.** Nothing to hold up merge from this review; the PR is clean as of 42547d5.

---

xianmuyq (approved):

PR #1437, opened 2026-08-24
refactor(backend): make Installation the single source of truth for active skills/MCPs

totalfrank:
## Problem

The skill/skill-set/MCP module answered "what is active for a Bot" in several conflicting ways:

- Two duplicated in-memory merges of "active skills" that disagreed — only one applied Default-Set exclusions.
- Two overlapping lazy-flush mechanisms (insert-only materializer vs. the full bridge repair).
- Legacy writers (`SkillSetActivator`/`SkillSetSwitcher`) that flipped `is_active` without writing Installation.
- MCP activation decided by a membership merge instead of `ac_bot_mcp_installation`.
- The activation-ownership rules scattered across four places, with a gap on the MCP direct commands.

## Solution (implemented)

`ac_bot_skill_installation` / `ac_bot_mcp_installation` are now the single source of truth for a Bot's active capabilities, held by four seams:

- **One writer.** Each Installation/exclusion table's SQL lives in exactly one command module under `core/repository/implementations/skill_center/tables/`; only the `CapabilityDesiredStateRepository` unit of work composes them, invoked by the two command services — `SkillSetManagementService` (Set-scoped mutations; Default-Set edits become per-Bot exclusion rows) and `DirectActivationService` (Set-free single-capability activation, skills ≡ MCPs). An AST architecture test fails any other module that writes the models. The legacy writers (`SkillSetActivator`/`SkillSetSwitcher`, the dead data-init activation step, the four legacy `SkillSetService` mutation methods) are deleted; a guard test keeps them dead.
- **One flush.** `flush_installations` is the only reconciliation from Set configuration into Installation — extended to MCP membership and Default-Set exclusions, idempotent, with a read-only fast path. The insert-only materializer is retired.
- **One reader.** `BotCapabilityStateReader` (flush-then-read) answers every "what is active" question; the new `SkillQueryService` (merging `LocalSkillQueryService` + the read half of `BotSkillAssetService`, both deleted) serves listing/detail/content/parameters through it, so a SkillSet-bridged skill is active in detail before any listing ran.
- **One rule book.** `CapabilityOwnershipPolicy` decides R1 (Set-managed ⇒ no direct control, Default-Set members included, excluded or not), R2 (deactivate before joining a Set), R3 (one Set per capability, any Set) — consumed inside the write transactions, identically for skills and MCPs, pinned by pairwise twin tests.

Deliberate behavior corrections (per spec §H): Default-Set exclusions honored by symlink/config-compose paths; excluded Default-Set members no longer directly controllable; Default-Set exclude/un-exclude commands live again; MCP direct commands gain the Default-Set half of R1; R3 now covers Default Sets; `deactivate-all` converges through desired state; runtime projection uses the layout-engine-first Default-Set precedence.

## Validation

- Full `tests/community` suite: 14071 passed, 55 skipped; 2 failures are sandbox-only (the skill-artifact snapshot tests shell out to `rsync`, absent locally; green in CI). CI "Backend unit tests" is the full-suite gate and is green on the pushed heads.
- `scripts/ci/python_sast_local.sh` passes; remaining findings are byte-identical on `dev`.
- Per-group adversarial review rounds (Groups 7–11) each closed with no real bugs; confirmed findings from earlier rounds (batch-abort on R1 refusal, exclusion-row compensation, engine-scope divergence in deactivate-all, unmapped not-ready error) were fixed in-tree with regression tests.
- Adversarial re-read of the final diff verified: every activation read goes through the flush-first reader; only the documented `/api/skillsets/admin/set-skillset-active` outlier writes outside the UoW (spec §C.7); ownership decisions live only in the policy; every MCP operation has a skill-twin test.

## Spec

`src/backend/specs/2026-08-24-installation-single-source-of-truth/` (spec.md, plan.md, tasks.md — all tasks complete, architecture.svg). Builds on `specs/2026-08-23-openapi-v1-bot-skill-listing`.

https://claude.ai/code/session_01GiRRbZ2NdMV2LAKWYkF5rw


src/backend/src/agentclaw/community/core/repository/implementations/skill_center/mcp_skill_set_control_plane.py
```diff
@@ -119,47 +133,132 @@ def remove_mcp(
...
-                bot_id=bot_id, owner_id=owner_id, server_code=server_code, env=get_current_env(),
-                avernet_tenant=get_current_avernet_tenant(),
-            ))
+                mcp_installations.uninstall(
+                    session, bot_id=bot_id, owner_id=owner_id,
+                    env=get_current_env(), server_codes={server_code},
+                )
             session.flush()
-            return SkillSetMutation({}, True, old)
+            return DesiredStateMutation(self._as_item(row), True, old)
```
totalfrank, line 191: **Blocking:** `exclude_default_mcp` is missing the member-guard its skill twin (`exclude_default_skill` in `default_exclusion_commands.py:59-65`) has, so it will happily create a dangling exclusion row for a `server_code` that is not currently a member of this Default Set.

`default_exclusions.exclude_mcp(...)` is called unconditionally at line 160-163; membership (`set_member_mcp_codes`) is only checked afterward, and only to decide whether to also call `mcp_installations.uninstall`. Compare to `exclude_default_skill`, which explicitly checks `int(skill_id) not in self._teardown_ids(...)` *before* writing the exclusion row, with a comment calling out exactly this failure mode: "a stray id must not leave a dangling row behind... it would pre-exclude the skill if the platform ever adds it".

Failure scenario: `exclude_default_mcp(bot_id, owner_id, set_id=<Bot's Default Set>, server_code="server-x")` is called where `server-x` is not currently a member of that Default Set (typo, stale UI state, or a server removed earlier). The call returns `changed=True` with no error, but a `DefaultSkillsetMcpExclusion` row is persisted. If the platform later adds `server-x` to that (shared) Default Set, the flush in `bot_skillset_installations.py::_resolve_flush_plan` will read this Bot's `excluded_mcps` and silently treat `server-x` as excluded/inactive for this Bot — even though the owner never excluded it while it was an actual member.

Suggested fix, mirroring the skill twin:

```suggestion
            if server_code not in set_member_mcp_codes(
                self._scope, session, skill_set_id=int(row.id)
            ):
                return DesiredStateMutation(self._as_item(row), False, old)
            created = default_exclusions.exclude_mcp(
                session, bot_id=bot_id, owner_id=owner_id,
                set_id=int(row.id), server_code=server_code,
            )
            if not created:
                return DesiredStateMutation(self._as_item(row), False, old)
            mcp_installations.uninstall(
                session, bot_id=bot_id, owner_id=owner_id,
                env=get_current_env(), server_codes={server_code},
            )
            session.flush()
            return DesiredStateMutation(self._as_item(row), True, old)
```

---

totalfrank (commented):
Reviewed commits `379898919fd3`..`ea838080a110` (67 commits, 142 files, +8375/-9815) against base `d35ef48788f0` (`dev`). Covered: correctness of the new Installation-flush/UoW write path, ownership-policy (R1/R2/R3) wiring, mutate-project-compensate error handling, DI bindings and dead-writer removal, HTTP router/response-shape changes, engine-scope handling on `deactivate_all_sets`, identity (owner vs. actor) usage in symlink/device-sync call sites, and test coverage of the new/changed behavior.

One blocking finding, inline: `exclude_default_mcp` is missing the membership guard its skill twin (`exclude_default_skill`) has, so it can create a dangling Default-Set exclusion row for a server code that isn't currently a Set member.

Everything else checked out — ownership-policy enforcement is consistent across the skill/MCP direct-activation and Set-membership paths, the mutate-project-compensate flow's argument order on the rollback branch is correct, deleted legacy writers (`SkillSetActivator`/`SkillSetSwitcher`, `BotSkillAssetService`, `LocalSkillQueryService`) have no orphaned references, DI bindings match their implementations, and the new/changed behavior (flush-then-read ordering, cross-engine `deactivate_all_sets`, the `LocalSkillNotReadyError` → 409 mapping, the `/skillset/current` alias key) each have a real-assertion regression test in this diff.

---

totalfrank (commented):
Reviewed commits `8ee59c1`..`ab4b5c97` (3 commits) against the previously-reviewed `ea838080`: a comment-cleanup commit, a lockfile-restore commit, and a merge of `origin/dev` that pulls in the DeviceSync-dispatcher relocation and an unrelated batch of `dev`-side work (docker/k8s/baas/gateway/bcs) outside this PR's scope.

**Status checklist**

| # | Finding | Status |
|---|---|---|
| 1 | `exclude_default_mcp` missing the membership guard its skill twin has (dangling Default-Set exclusion for a non-member `server_code`) — `mcp_skill_set_control_plane.py:174` | **Still open** — code at that location is unchanged in these commits; no commit message or reply claims it was fixed |

Open findings: **1**.

**New commits reviewed** (scope: `src/backend/.../skill_center/`, `di/modules/skill_center_module.py`, and the re-pointed test, since that's what this PR's diff actually touches in the merge):
- Import-path updates for `DeviceSyncDispatcher`'s move to `agentclaw.community.plugin_api.device_sync_dispatcher` (4 files) — verified the new module exports `DeviceSyncDispatcher` and no stale import of the old path remains anywhere under `src/backend`.
- Dead-code/comment cleanup in `skill_set_service.py` and `skill_center_module.py`.
- `test_openapi_bot_skills_read.py` (landed from `dev`) re-pointed from `LocalSkillQueryServiceProtocol`/`BotSkillAssetServiceProtocol` to `SkillQueryServiceProtocol` — verified `list_bot_skills`/`get_skill` match the merged protocol's actual method names.
- `uv.lock` restored to `dev`'s byte-for-byte content.

No new bugs found in this round's actual PR-scoped changes.

---

src/backend/src/agentclaw/community/core/repository/implementations/skill_center/mcp_skill_set_control_plane.py
```diff
@@ -119,47 +133,132 @@ def remove_mcp(
...
+                engine_type=engine_type,
+                default_engine_types=default_engine_types,
+            )
             old = self._snapshot(session, bot_id, owner_id, engine_type=engine_type)
-            if self._mcp_has_ordinary_membership(
-                session, bot_id, owner_id, server_code
+            created = default_exclusions.exclude_mcp(
+                session, bot_id=bot_id, owner_id=owner_id,
+                set_id=int(row.id), server_code=server_code,
+            )
```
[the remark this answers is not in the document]
totalfrank: Fixed in 819d073, with one deliberate deviation from the literal suggestion: the guard accepts a code that is either an association-row member **or** one of the platform's engine/template default codes. Default-Set MCP membership has an unmaterialized half — the engine/template default policy (spec A.2 keeps it config, not rows) — and excluding one of those is exactly what the exclusion row exists for, so a membership-only guard would have broken that path. The management service now resolves the platform default codes with the read-side union's context and passes them into the transaction; a code in neither half returns `changed=false` and writes nothing. Regression tests cover the stray-code refusal and the platform-default acceptance.

---

src/backend/src/agentclaw/community/core/repository/implementations/skill_center/mcp_skill_set_control_plane.py
totalfrank: Fixed in 819d073. The gate mirrors the skill twin's shape but with a wider member set than the suggested snippet: `set_member_mcp_codes(...) ∪ platform_default_codes`, where the platform codes are the engine/template default policy the service resolves (same context as the read-side union — engine, template_type, best-effort ext info) and passes into `exclude_default_mcp`. A membership-row-only guard would refuse excluding a config-default MCP, which is the command's primary use case since those defaults are deliberately unmaterialized (spec A.2). Stray codes (neither half) now return `changed=false` with no exclusion row; both directions are pinned by new UoW tests plus a service wire test asserting the policy codes reach the transaction.

---

totalfrank:
**Review round 1 — both findings fixed in 819d073.**

The round's two reviews (Codex + Claude Code, both against `ea83808`; the cited code was unchanged by the intervening `dev` merge) surfaced two real defects, one of them flagged independently by both reviewers:

1. **Compensation metadata loss (Codex P1)** — `restore_desired_state` recreated `SkillSetMCPServer` rows with `name=server_code`, so a failed runtime projection's rollback permanently discarded name/description/icon/user_id on memberships the mutation never touched. The snapshot now captures the complete association row and the restore recreates it verbatim, with a round-trip regression test.

2. **Dangling Default-Set MCP exclusion (Codex P2 + review blocking finding)** — `exclude_default_mcp` wrote the exclusion row before any membership check. It now mirrors the skill twin's never-member gate, but against the *effective* membership: association rows ∪ the engine/template platform-default codes, which the management service resolves (same context as the read-side union, ext info included) and passes into the transaction. A membership-row-only guard — the literal suggestion on both threads — would have broken excluding a config-default MCP, the command's primary use case, since those defaults are deliberately unmaterialized (spec A.2). Stray codes now return `changed=false` with no row. Covered by two UoW tests (stray refusal, platform-default acceptance) and a service wire test.

Both inline threads are replied to and resolved. This push opens the next round on `819d073`.

---

totalfrank:
@codex review

Please review the round-1 fixes on `819d073` (compensation now snapshots full MCP association rows; `exclude_default_mcp` gates on effective membership — association rows ∪ platform default codes).

---

totalfrank (commented):
Reviewed commit `819d073e` (1 commit) against the previously-reviewed `ab4b5c97`: the fix for the one open finding from the last round, plus the Codex P1 finding on the same code path.

**Status checklist**

| # | Finding | Status |
|---|---|---|
| 1 | `exclude_default_mcp` missing the membership guard its skill twin has (dangling Default-Set exclusion for a non-member `server_code`) — `mcp_skill_set_control_plane.py:174` (mine, blocking) | **Fixed** — `exclude_default_mcp` now takes `platform_default_codes`, computes `member_codes` up front, and refuses (`changed=False`, no row written) any code in neither `member_codes` nor `platform_default_codes`. The sole call site (`skill_set_management_service.py:504`, `remove_mcp`'s Default-Set branch) resolves and passes `platform_default_codes` via the new `_platform_default_mcp_codes` helper, using the same engine/template/ext-info context as the read-side union (`skill_set_service.py`'s `_get_default_capabilities_template_type`/`_get_default_capabilities_ext_info`). New UoW tests (`test_excluding_a_stray_mcp_code_owns_neither_half`, `test_excluding_a_platform_default_mcp_without_membership_writes_the_row`) and a service wire test cover both the refusal and the platform-default-without-membership acceptance. Verified by reading the diff and the sole call site; already replied and resolved in-thread. |
| 2 | (Codex, not mine) `restore_desired_state` recreated `SkillSetMCPServer` rows with `name=server_code`, discarding name/description/icon/user_id on unrelated memberships during compensation — `capability_desired_state.py:653-657` | **Fixed** — the desired-state snapshot now captures the full association tuple `(server_code, name, description, icon, user_id)` instead of a bare code, and `restore_desired_state` recreates the row verbatim from it. `SkillSetMCPServer`'s only other fields (`env`/`avernet_tenant`) are context-derived at restore time as before; nothing is missed. New round-trip test `test_restore_desired_state_preserves_mcp_membership_metadata` pins it. Already replied and resolved in-thread by the author; verified independently by reading the diff. |

Open findings: **0**.

**New commit reviewed** (819d073e, scope: the two files above plus their DI wiring):
- `capability_desired_state_types.py` / `capability_desired_state.py`: snapshot/restore tuple widening — checked unpack order matches snapshot order, no fields dropped.
- `mcp_skill_set_control_plane.py`: guard ordering — membership/platform-default check now runs before the exclusion write, and the post-write install-uninstall check reuses the same pre-computed `member_codes` instead of re-querying.
- `skill_set_management_service.py`: new `_platform_default_mcp_codes` helper, single call site, correct positional wiring.
- `skill_center_module.py`: `SkillSetManagementServiceProtocol` binding switched from a plain `binder.bind` to a `@provider` so `ext_info_provider` can be threaded in — confirmed only one binding site (no duplicate-binding conflict), and the provider's positional args match `SkillSetManagementService.__init__`'s order exactly, including the `CoreBotRuntimeProjectorProtocol` alias already used by sibling providers in this file.

No new bugs found in this round. The PR is clean as of `819d073e`.

---

totalfrank:
**Review round 2 — no issues flagged.** Reviewed `819d073` (the round-1 fixes: full-row MCP snapshot/restore, and the effective-membership gate on `exclude_default_mcp` with its DI-threaded platform-default context). The review verified both fixes independently against the diff and call sites; open findings: 0, no new bugs. Both round-1 threads stand replied and resolved.

CI on `819d073`: six checks green; `Backend unit tests` (the full-suite gate) and `Singlebox coverage` still running — I'll confirm the final state when they complete. Codex re-review was requested on this SHA and is still pending; if it surfaces anything new, that opens another round.

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_management_service.py
```diff
@@ -754,6 +753,27 @@ def _require_set_mcp_permissions(
...
+        Resolved at write time with the same context the read-side union
+        uses — engine, template, best-effort ext info — so the exclusion
+        command can tell a platform-default code from a stray one.
+        """
+        ext_info = None
+        if self._ext_info_provider is not None:
+            try:
+                ext_info = self._ext_info_provider(bot_id)
+            except Exception:
+                ext_info = None
```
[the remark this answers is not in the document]
totalfrank: Fixed in 0c7048f, exactly as suggested: `ext_info_provider` is now a required constructor dependency (per the AGENTS.md type contract you cited) and `_platform_default_mcp_codes` propagates provider failures instead of substituting `None` — a transient template lookup failure now fails the removal as a retryable error rather than persisting a wrong `changed=false` no-op. Verified the premise against the AICoding resolver (`merge_template_preset_capabilities` does add preset-only MCPs from `ext_info.template_config`). Regression test pins that a raising provider fails `remove_mcp` before any exclusion command runs.

---

totalfrank:
**Review round 3 — one finding, fixed in 0c7048f.**

Codex's re-review of `819d073` found one real defect in the round-1 fix itself: `_platform_default_mcp_codes` swallowed ext-provider failures, so for a bot whose default MCP exists only as a template preset (the AICoding resolver merges `template_config` presets onto engine defaults), a transient template-service failure would silently narrow the effective-membership gate — the exclusion command would then mis-read the genuine member as a stray and report the removal as `changed=false` without persisting anything. Fixed as suggested: the provider is now a required constructor dependency and its failures propagate, so the caller gets a retryable error instead of a wrong persisted answer. Regression test added; thread replied and resolved.

This push opens the next round on `0c7048f`. @codex review

---

totalfrank:
**Review round 4 — no issues flagged.**

Reviewed `0c7048f` (the round-3 fix: `ext_info_provider` promoted to a required constructor dependency on `SkillSetManagementService`, and `_platform_default_mcp_codes` propagating provider failures instead of swallowing them into `None`).

Verified independently:
- `SkillSetManagementService.__init__` no longer defaults or try/excepts the provider; `_platform_default_mcp_codes` now calls it unguarded, so a raising provider fails `remove_mcp`'s exclusion path before any exclusion command runs — confirmed against the new regression test (`test_default_mcp_exclusion_propagates_a_template_context_failure`), which also asserts no exclusion row was written on the failure path.
- Every non-test and test call site of `SkillSetManagementService(...)` was checked: the two DI bindings (`skill_center_module.py:457,732`) already thread `_build__ext_info_provider(injector)` — an unrelated call-site (`SkillSetServiceFactory`, the read-side factory) keeps its own optional provider, which is untouched by and out of scope for this fix. All ~27 call sites in `test_skill_set_management_service.py` and the one in `test_openapi_skill_sets.py` were updated to pass `ext_info_provider=lambda _bot_id: None`; none left on the old optional default.
- No new bugs found in this round's diff (3 files, +71/-13, all in `skill_set_management_service.py` and its tests).

**Status checklist (all rounds)**

| # | Finding | Status |
|---|---|---|
| 1 | Codex P2 — `exclude_default_mcp` writes exclusion before checking Default-Set membership (dangling row) — `mcp_skill_set_control_plane.py` | Fixed in `819d073` (round 1) |
| 2 | Claude Code (mine) — same defect, blocking — `mcp_skill_set_control_plane.py:174` | Fixed in `819d073` (round 1) |
| 3 | Codex P1 — `restore_desired_state` recreates MCP rows with `name=server_code`, discarding association metadata on compensation — `capability_desired_state.py:653-657` | Fixed in `819d073` (round 1) |
| 4 | Codex P1 — `_platform_default_mcp_codes` swallows ext-provider failures, silently narrowing the exclusion gate — `skill_set_management_service.py` | Fixed in `0c7048f` (round 3) |

Open findings: **0**. All four threads replied and resolved.

CI on `0c7048f`: 6 of 8 checks green; `Backend unit tests` (full-suite gate) and `Singlebox coverage` still in progress at review time.

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_management_service.py
```diff
@@ -754,6 +753,25 @@ def _require_set_mcp_permissions(
...
+        default MCP, a silently narrowed set would make the exclusion
+        command mis-read the genuine member as a stray and no-op the
+        removal as ``changed=False`` — a wrong persisted answer, where an
+        error is merely a retry.
+        """
+        return frozenset(
+            get_default_mcp_server_codes(
+                self._engine(bot),
+                bot.get("template_type"),
+                ext_info=self._ext_info_provider(bot_id),
```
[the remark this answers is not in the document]
totalfrank: Fixed in 461a1df. `TemplateService` gains `get_template_config_strict` — `None` means exactly "this Bot has no template", a repository failure raises — and the DI ext-info provider now calls it, so the production chain propagates end to end: repository failure → provider → `_platform_default_mcp_codes` → the removal fails as retryable instead of persisting a wrong `changed=false`. The lenient `get_template`/`get_template_config` keep their degrade-to-None contract for list/read surfaces, which catch and log around the provider (`_get_default_capabilities_ext_info`); a contrast test pins both contracts side by side, and the provider unit test now exercises the strict method.

---

totalfrank:
**Review round 4 — one finding, fixed in 461a1df.**

Codex's re-review of `0c7048f` correctly caught that the round-3 fix couldn't fire in production: the injected provider calls `TemplateService.get_template_config()`, whose `get_template()` catches every repository exception and returns `None` — indistinguishable from "no template" — so the swallow had just moved a layer down. Fixed as suggested: `TemplateService.get_template_config_strict` distinguishes absence (`None`) from failure (raises), the DI ext-info provider uses it, and the full production chain now propagates a template-repository failure into a retryable error on the removal. The lenient reader keeps its degrade-to-None contract for list/read surfaces, which have their own catch. Contrast test on the two template-service contracts added; the provider unit test exercises the strict path. Thread replied and resolved.

This push opens the next round on `461a1df`. @codex review

---

totalfrank:
**Review check — no new commits since the last round; independently re-verified.**

Head is still `461a1df`, the same commit the round-4 summary above already covered. Re-ran the verification from scratch rather than trusting that comment: read `template_service.py` (`get_template_config_strict` calls `self._repository.get_by_bot_id` with no `try/except`, so a repository failure propagates; the lenient `get_template`/`get_template_config` are untouched and still catch-and-log), the DI provider (`_build__ext_info_provider` now calls the strict method), and the new/changed tests (`test_get_template_config_strict_distinguishes_absence_from_failure`, and `test_skill_factories.py`'s fake updated to the strict method name). Also swept every other production call site of `get_template_config` (`bot_service.py`, `bot_process.py`, `engines/registry.py`, `aicoding/strategy.py`, `member_management_capability.py`, `baas_device_service.py`, `baas_publish_task_handlers.py`) — all correctly remain on the lenient method; only the write-path exclusion gate needed the strict twin.

**Status checklist (all rounds)**

| # | Finding | Status |
|---|---|---|
| 1 | Codex P2 — `exclude_default_mcp` writes exclusion before checking Default-Set membership (dangling row) — `mcp_skill_set_control_plane.py` | Fixed in `819d073` (round 1) |
| 2 | Claude Code (mine) — same defect, blocking — `mcp_skill_set_control_plane.py:174` | Fixed in `819d073` (round 1) |
| 3 | Codex P1 — `restore_desired_state` recreates MCP rows with `name=server_code`, discarding association metadata on compensation — `capability_desired_state.py:653-657` | Fixed in `819d073` (round 1) |
| 4 | Codex P1 — `_platform_default_mcp_codes` swallows ext-provider failures, silently narrowing the exclusion gate — `skill_set_management_service.py` | Fixed in `0c7048f` (round 3) |
| 5 | Codex P1 — the round-3 fix was moot in production because `get_template_config`'s underlying `get_template` swallows repository errors into `None` — `skill_set_management_service.py:771` | Fixed in `461a1df` (round 4), confirmed sound in this pass |

Open findings: **0**. No new commits to review this round. CI on `461a1df`: 6/8 checks green (`Sandbox-proxy`, `Gateway`, `BaaS`, `Engine`, `BCS unit tests`, `BCS e2e`); `Backend unit tests` (the full-suite gate) and `Singlebox coverage` still in progress.

---

totalfrank:
**Review round 5 — no issues flagged. The loop is clean.**

Reviewed `461a1df` (the strict template lookup): Codex reported "didn't find any major issues", and the Claude Code re-verification pass independently confirmed the fix sound with every other `get_template_config` call site correctly left on the lenient contract. CI is fully green on this head — all 8 checks including the `Backend unit tests` full-suite gate and `Singlebox coverage`.

Final state across the loop: 5 rounds, 5 confirmed findings (dangling Default-Set MCP exclusion; lossy compensation snapshot; and a three-layer chain of swallowed template-lookup failures), all fixed with regression tests, all threads resolved. Open findings: **0**. The PR is ready for human review — @FreddieSun @xianmuyq.

---

src/backend/src/agentclaw/community/api/direct_activation_service.py
```diff
@@ -0,0 +1,38 @@
...
+
+    async def activate_mcp(
+        self, *, server_code: str, bot_id: str, owner_id: str, actor_id: str
+    ) -> dict[str, Any]: ...
+
+    async def deactivate_mcp(
+        self, *, server_code: str, bot_id: str, owner_id: str, actor_id: str
+    ) -> dict[str, Any]: ...
+
+    def list_installed_mcps(
```
totalfrank-personal, line 33: So this is the read method to return all mcps for a given bot.
I guess probably somewhere we have a parallel one for list all skills for a given bot right? Where is that method?

src/backend/src/agentclaw/community/core/bot_management/services/template_service.py
```diff
@@ -229,6 +229,23 @@ def list_templates_by_bot_ids(self, bot_ids: List[str]) -> List[Dict[str, Any]]:
...
+        failure propagates. The lenient reader below folds both into
+        ``None``, which is right for list/read surfaces that must degrade —
+        and wrong for a write deciding persisted state from the answer,
+        where "no template" and "lookup failed" must not be confusable.
+        """
+        template = self._repository.get_by_bot_id(bot_id)
+        if not template:
+            return None
+        return template.get("ext")
+
```
totalfrank-personal, line 248: What is this change? I am confused. Did you make a mistake while resolving conflicts?

src/backend/src/agentclaw/community/core/skill_center/services/bot_capability_state_reader.py
```diff
@@ -0,0 +1,105 @@
...
+        self,
+        repository: CapabilityDesiredStateRepositoryProtocol,
+        bot_repo: BotRepository,
+        pool_skills: SkillsPoolSkillRepositoryProtocol,
+    ) -> None:
+        self._repository = repository
+        self._bot_repo = bot_repo
+        self._pool_skills = pool_skills
+
+    def flush(self, *, bot: Mapping[str, Any]) -> InstallationFlushPlan:
```
totalfrank-personal: Is this really used anywhere other than tests? If not, let's get rid of it. Flush should not be exposed outside this service. Plus you already have a _flush method below. I believe that's the one you need to use. Let me know if I misread the code.

src/backend/src/agentclaw/community/core/repository/implementations/skill_center/tables/default_exclusions.py
```diff
@@ -0,0 +1,211 @@
...
+    DefaultSkillsetSkillExclusion,
+)
+from agentclaw.community.utils.avernet_tenant import get_current_avernet_tenant
+
+
+def exclude_skill(
+    session, *, bot_id: str, owner_id: str, set_id: int, skill_id: int
+) -> bool:
+    """Ensure the exclusion row exists; return whether this call created it."""
+    if skill_id in excluded_skill_ids(
```
totalfrank-personal, line 35: You did not have any verification that skill is a skill in default skill set. Maybe all call sites are passing down skill from default skill set but that's not a guarantee. Make sense?

src/backend/src/agentclaw/community/core/skill_center/services/skill_query_service.py
```diff
@@ -54,49 +62,88 @@ def resolve(
...
+        plan = self._reader.flush(bot=bot)
+        return self._skill_repo.list_bot_skills(
+            bot_id=bot_id,
+            user_id=owner_id,
+            skill_set_member_ids=plan.member_skill_ids,
+            page=page,
+            page_size=page_size,
+            active=active,
+            keyword=keyword,
+        )
```
totalfrank-personal: _reader should not expose .flush so that user invokes it and the call read himself right?

Reader should expose its own read function(where flush is called up front before read) right?

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
totalfrank-personal, line 1451: This method is lengthy. Try refactor it to make it readable. It's hard to understand.

src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
```diff
@@ -475,7 +473,8 @@ def _require_pool_runtime_sync_success(
 
 async def _set_legacy_asset_active_if_resolved(
```
totalfrank-personal: Looks like this is used by the legacy activate_skill API in src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py.

This is deprecated, let's remove it in this session as well.

The same for deactivate_skill, it's deprecated as well.

Clean them up and all downstream methods if those methods are only reachable by these legacy APIs we are deprecating.

src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
totalfrank-personal: This is deprecated, let's delete all along.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/api/direct_activation_service.py
totalfrank: The skill twin is `SkillQueryService.list_bot_skills` (`api/skill_query_service.py`) — it pages every skill the bot has, owned rows and Set-bridged alike, with `active` answered from Installation after the reader's flush. The asymmetry in *where* the two reads live is deliberate: skills carry pagination, filters, and rich metadata, so their read seam is the query service; the MCP read is a bare set of active server codes, kept on `DirectActivationService` next to the commands it answers for (the openapi MCP router consumes it there). If you'd rather see the MCP read co-located on a query seam for symmetry, happy to move it — say the word.

---

src/backend/src/agentclaw/community/core/bot_management/services/template_service.py
totalfrank: Not a conflict-resolution artifact — this is a deliberate addition from the automated review rounds (Codex, round 4, thread on `skill_set_management_service.py`). Chain of findings: the Default-MCP exclusion gate needs the platform default codes, which come through the template ext-info provider; the provider called `get_template_config`, whose underlying `get_template` swallows every repository exception into `None` — indistinguishable from "no template". For a *write* deciding persisted state, that turned a transient DB failure into a silent wrong `changed=false`. `get_template_config_strict` is the write-path twin: `None` means exactly "no template", a repository failure raises. The lenient reader keeps its degrade-to-None contract for every list/read call site (all swept and left unchanged). Pinned by `test_get_template_config_strict_distinguishes_absence_from_failure`.

---

src/backend/src/agentclaw/community/core/skill_center/services/bot_capability_state_reader.py
totalfrank: You read it right, and agreed — done in 577d002. `flush` had three production callers (the query service's listing and detail, and the service-artifact build stage), all of them the flush-then-read-yourself split you're pointing at. The reader's public surface is now reads only: the listing consumes a new `member_skill_ids` read (the Set-membership half the page filter needs), detail answers through `active_skill_assets`, and the build stage captures desired state through the same read the runtime projection uses. `_flush` is private to the reader.

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_query_service.py
totalfrank: Right — fixed in 577d002 together with the reader-side thread: the reader now exposes only reads that flush up front (`member_skill_ids`, `active_skill_assets`, `active_mcp_server_codes`), and this call site consumes `member_skill_ids` instead of flushing and reading itself. No caller can split the flush from its read anymore.

---

src/backend/src/agentclaw/community/core/repository/implementations/skill_center/tables/default_exclusions.py
totalfrank: Makes sense as a question — the guarantee exists, one layer up, and it is enforced rather than hoped for. This module is deliberately dumb SQL; the invariants live in the UoW exclusion commands: `_default_set` refuses any `set_id` that isn't the Bot's Default Set (NotFound) before these functions run, and both exclusion commands gate on effective membership before writing (`_teardown_ids` for skills; association rows ∪ platform default codes for MCPs). What makes that a guarantee instead of a convention is `test_installation_table_write_ownership.py`: the AST guard keeps the UoW composition modules the only importers of this `tables/` package, so no call path reaches these writes without passing those gates. I've added that statement to the module docstring in 776a981 so the next reader doesn't have to reconstruct it.

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
totalfrank: Refactored in 776a981. `collect_bot_active_mcps` now reads as the union's precedence order, one named helper per phase: `_default_set_mcp_rows` (Default-Set association rows minus exclusions), `_default_policy_mcp_entries` (static engine/template defaults minus exclusions), `_installed_mcp_codes` (the reader's installed half, with the entity-keyed owner retry), and `_installed_only_mcp_entries` (metadata from membership rows, minimal entry for direct installs). Behavior and ordering unchanged; the existing union tests pass as-is.

---

src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
totalfrank: Deleted in 9619dce, with the full exclusive-downstream sweep: the two routes, this helper, `_require_pool_runtime_sync_success`, and on the core side `SkillService.activate_skill` / `deactivate_skill` plus their only remaining internal callers `activate_skills_batch` and `deactivate_all_skills` (verified by search that nothing else reaches any of them — `parse_skill_path`/`get_link_name` stay, they have live callers in upload/delete/scan paths). The `/api/skills/market/activate-batch` route is untouched: it already runs on the canonical control plane (resolve + DirectActivationService), not on these.

---

src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
totalfrank: Deleted in 9619dce, downstream included: the `/api/skills/deactivate-all` route was the only caller of `SkillSetManagementService.deactivate_all`, which was the only caller of the UoW's `deactivate_all_sets` — all three are gone, protocol entries and their tests with them. The app-level `LocalSkillNotReadyError` → 409 handler stays, since the surviving Set activate/deactivate family still raises it.

---

totalfrank:
**Human review round — all eight comments addressed** (pushed as three commits, head `776a981`):

- **Reader surface** (`577d002`): the reader no longer exposes `flush` — its public surface is reads that flush up front (`member_skill_ids`, `active_skill_assets`, `active_mcp_server_codes`). The query service's listing/detail and the service-artifact build stage all consume reads now; `_flush` is private.
- **Deprecated route deletion** (`9619dce`): `/api/skills/{skill_id}/activate`, `/deactivate` and `/deactivate-all` are gone, together with everything only they could reach — the legacy path-reference helper, `SkillService.activate_skill`/`deactivate_skill` and their batch/deactivate-all wrappers, `SkillSetManagementService.deactivate_all`, and the UoW's `deactivate_all_sets` (protocol entries and tests included). The control-plane batch route and all Set-scoped routes are untouched.
- **Readability + layering** (`776a981`): `collect_bot_active_mcps` is split into its three union phases as named helpers, and the exclusion table module's docstring now states where its invariants are enforced (the UoW commands, held in place by the write-ownership architecture test).
- **Questions answered in-thread**: the skill twin of `list_installed_mcps` is `SkillQueryService.list_bot_skills`; `get_template_config_strict` is not a merge artifact but the round-4 Codex fix (strict lookup so a repository failure can't masquerade as "no template" on the write path) — both threads left open for your read.

Full sweep green locally (~14k tests, the two known rsync-sandbox-only failures excepted). @codex review on `776a981`.

---

src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
```diff
@@ -1346,381 +1271,6 @@ async def get_active_skill_sets(
 # ==================== Skill Activation APIs (动态路径) ====================
```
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in f25a6d4 — reproduced the `test_endpoint_coverage_against_baseline` failure locally first (my earlier sweeps never included `tests/community/framework/`), then removed the stale `POST /api/skills/deactivate-all` baseline entry. The framework suite (coverage gate included) passes on the fix. The two single-skill routes were not in the baseline, so this was the only stale entry.

---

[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in f25a6d4: the single-skill activate step is dropped from the crud-roundtrip flow. No migration needed — the very next step already activates the same `{uploaded_git_path}` through the surviving `/api/skills/market/activate-batch` surface, so the flow's post-activation state (including the later `/api/skills/active/list` assertion) is unchanged. A repo-wide sweep found no other references to the three deleted routes; the `/openapi/v1/.../activate` hits are the surviving canonical surface, untouched by this PR's deletion.

---

totalfrank:
**Round 7 (Codex on `776a981`) — both findings confirmed and fixed**, pushed as `f25a6d4`:

- **Coverage baseline** (P1): reproduced the `test_endpoint_coverage_against_baseline` failure locally, removed the stale `POST /api/skills/deactivate-all` entry — the only deleted route the baseline listed. Framework suite green on the fix.
- **Live flow** (P1): dropped the crud-roundtrip step that called the deleted single-skill activate route; the adjacent `market/activate-batch` step already activates the same skill path, so the flow's downstream assertions are unchanged.

A repo-wide sweep confirmed no other references to the three deleted legacy routes remain. This push opens the next round — @codex review on `f25a6d4`.

---

totalfrank (commented):
**Status check on `f25a6d4` — no new commits since the last round, nothing open.**

This run's synchronize event is for `f25a6d4`, the same commit my last round (issuecomment-5411522212, "Round 7") already reviewed and summarized. There is nothing new to review: `f25a6d4` *is* the fix for that round's two findings, and no commit has landed after it. Re-verified independently rather than trusting the prior comment:

- `coverage_baseline.txt`: confirmed the stale `POST /api/skills/deactivate-all` line is gone and no other deleted-route entries remain in the baseline.
- `api_lifecycle.py`: confirmed the deleted single-skill `/activate` step is gone from `skill_center-skill-and-skillset-crud-roundtrip`, and the very next step (`POST /api/skills/market/activate-batch`) activates the same `{uploaded_git_path}`, so the flow's downstream assertions are unaffected.
- Repo-wide sweep: the only remaining `activate`/`deactivate`/`deactivate-all` references under `src/backend` are the surviving `/openapi/v1/bots/.../skills/{skill_id}/activate|deactivate` canonical routes and their tests — no reference to the three deleted legacy `/api/skills/...` routes remains.

**Status checklist (all rounds)**

| # | Finding | Status |
|---|---|---|
| 1 | Codex P2 — `exclude_default_mcp` writes exclusion before checking Default-Set membership (dangling row) — `mcp_skill_set_control_plane.py` | Fixed (`819d073`) |
| 2 | Claude Code (mine) — same defect, blocking — `mcp_skill_set_control_plane.py:191` | Fixed (`819d073`) |
| 3 | Codex P1 — `restore_desired_state` recreates MCP rows with `name=server_code`, discarding association metadata — `capability_desired_state.py` | Fixed (`819d073`) |
| 4 | Codex P1 — `_platform_default_mcp_codes` swallows ext-provider failures — `skill_set_management_service.py` | Fixed (`0c7048f`) |
| 5 | Codex P1 — fix above was moot in production; `get_template_config`/`get_template` swallow repository errors into `None` — `skill_set_management_service.py:741` | Fixed (`461a1df`) |
| 6 | Codex P1 — coverage baseline still lists deleted `POST /api/skills/deactivate-all` | Fixed (`f25a6d4`), re-verified this round |
| 7 | Codex P1 — live crud-roundtrip flow still called the deleted single-skill activate route | Fixed (`f25a6d4`), re-verified this round |
| 8 | Human (totalfrank-personal) — reader shouldn't expose `flush` publicly | Fixed (`577d002`) |
| 9 | Human — `default_exclusions.py` MCP exclusion write lacked membership verification (same root cause as #1/#2, filed independently) | Fixed (`819d073`) |
| 10 | Human — `collect_bot_active_mcps` too long, hard to read | Fixed (`776a981`, refactored into named phase helpers) |
| 11 | Human — delete deprecated single-skill activate/deactivate routes and dead downstream code | Fixed (`9619dce`) |
| 12 | Human — delete deprecated deactivate-all route and dead downstream code | Fixed (`9619dce`) |
| 13 | Human — question: where's the skill-side twin of the MCP list-active read? | Answered in-thread, not a defect (open, author's call) |
| 14 | Human — question: was the `get_template_config_strict` addition a merge mistake? | Answered in-thread, not a defect (open, author's call) |

Open **bug** findings: **0**. Two threads (#13, #14) remain open as answered questions, not defects — left for the human reviewer to close.

CI on `f25a6d4`: 5/8 checks green, 3 still running (`Backend unit tests` — the full-suite gate, `Singlebox coverage`, `BaaS unit tests`); no failures yet.

The PR is clean as of this round; no new bugs found.

---

totalfrank:
**Round 8 — clean. The review loop is closed on head `f25a6d4`.**

- Codex on `f25a6d4`: "Didn't find any major issues."
- Claude Code re-review on `f25a6d4`: independently re-verified both round-7 fixes; 0 open bug findings across all 14 tracked items.
- CI on `f25a6d4`: all 8 checks green ("Backend unit tests" — the full-suite gate — included).

Loop totals: 8 rounds, 7 confirmed findings (5 automated + 2 route-deletion follow-ups), all fixed with the failure reproduced first where a test could reach it; the human review's 8 comments are all addressed. Two threads remain open as answered questions for the author (the MCP-read co-location offer and the `get_template_config_strict` explanation) — nothing else is outstanding. Ready for human review.

---

FreddieSun (approved):
lgtm

PR #1441, opened 2026-08-25
feat(bcs-api-http): migrate state-machine-runs, manifest & assets to v1 internal router

zh-bupt:
## Problem

Several BCS endpoints lived only on the legacy `bcs-http` delivery adapter and were not reachable through the gateway (`bcs-api-http` v1 internal router under `/api/v1/collaboration/**`), so gateway-authenticated callers (the frontend panel via cookie SSO, the workbench host) could not reach them through the standard signed-Principal boundary and v1 response envelope. Additionally, the panel read these endpoints' response fields directly and would break once responses were wrapped in the v1 envelope.

Affected endpoints: `GET /manifest`, `GET /state-machine-runs/{run_id}`(+`/graph`, `/nodes/{node_id}`, `/pending-human-nodes`), `POST .../nodes/{node_id}/respond`, `POST .../cancel`, and the static `GET /assets/{bundle}/{file}`.

## Solution

Pure-additive migration mirroring the `session_file` reference migration; legacy `bcs-http` routes stay mounted (dual-exposure), so no existing direct-to-BCS consumer breaks.

- New route modules under `crates/adapters/http/bcs-api-http/src/v1/internal/routes/` (`collaboration_run.rs`, `manifest.rs`) reuse the existing application services (`CollaborationRuntimeService`, `ManifestConfig`) directly — no new v1 facade. JSON responses are wrapped via `Envelope::success(20_000, "OK", data, request_id)`; `/assets` stays raw file bytes (not enveloped).
- Auth is mapped in the HTTP layer from the gateway-signed `AuthenticatedCaller` to the legacy `AuthenticatedHumanCaller` (optional human for reads/cancel; required human for `respond`/`pending-human-nodes`). Collaboration-runtime types are imported through `bcs_service_api::application` to satisfy the delivery-adapter boundary contract.
- Gateway `route_security` requires a user principal for `/api/v1/collaboration/state-machine-runs/**` and leaves `/manifest` + `/assets/**` public; the existing `collaboration-internal` domain already forwards the prefix, so no new upstream domain.
- The new `/manifest` projects file-bundle URLs onto `/api/v1/collaboration/assets/{name}/{file}` and uses the same `Environment::resolve().as_str()` env source as legacy `/manifest`, so gateway and legacy manifest return identical `env`.
- `api-contracts/v1` documents the eight new paths (state-machine-runs + manifest fragments) and regenerates `gateway/configs/schemas/bcn.internal.openapi.json`.
- Frontend panel adds an `unwrapEnvelope` normalizer applied at the graph/pending/node-detail JSON parse sites, so it tolerates both the raw legacy payload and the v1-enveloped shape. `parseErrorBody` is left unchanged because the v1 error envelope already carries `message` at the top level.

Rejected alternative: building a new v1 application facade for the collaboration runtime — rejected to avoid scope creep; the service layer is reused verbatim with HTTP-layer auth mapping.

## Validation

- `cargo test -p bcs-api-http` — all green: `boundary_contract` 3/3, `collaboration_run_routes` 9/9, `manifest_routes` 2/2, lib unit tests 34/34, plus all pre-existing route suites.
- `cargo build -p bcs-api-http -p bcs` — clean.
- API contract validator: `validate_openapi_contract.py` → `21 operations validated` (13 prior + 8 new); regenerated `bcn.internal.openapi.json` contains all 8 new paths.
- Panel: `tsc --noEmit` passes; `vite build` regenerates `dist/index.umd.js` (gitignored, not committed); `test:umd` UMD contract test passes; `git check-ignore -v` confirms `dist/` is ignored.
- Gateway `application.yaml` parses (`yaml.safe_load` → OK).
- Dual-exposure sanity: legacy `crates/adapters/http/bcs-http/` is untouched across the branch.
- The `boundary_contract` test was mid-branch-broken by crate-root `bcs_service_api::` imports; fixed by routing imports through `application::`, now green.

## Compatibility and risk

- Pure-additive: no legacy route removed, no schema migration, no config-breaking change. Legacy direct-to-BCS `/manifest`, `/assets`, `/state-machine-runs/...` keep anonymous/optional-human semantics; gateway-exposed state-machine-runs require a User Principal (a tightening, surfaced only through the gateway).
- Rollback: revert the single commit; legacy routes continue to serve independently.

## Spec

Design and implementation plan at `docs/superpowers/specs/2026-08-24-state-machine-runs-manifest-gateway-migration-design.md` and `docs/superpowers/plans/2026-08-24-state-machine-runs-manifest-gateway-migration.md` (on branch `feat/state-machine-runs-manifest-gateway-migration`, the unsquashed companion branch).


questchaos (approved):
LGTM

cassiuscai (approved):

PR #1442, opened 2026-08-25
fix(backend): scope template validation to OpenAPI creates

lucas-xzp:
## Problem

The recent AICoding template validation was applied in the shared bot creation flow. This caused legacy internal `/api/bots` callers that pass `template_uid` in `template_config` to fail with `template_config contains server-managed fields`.

The stricter server-managed-field validation should apply to the OpenAPI contract only and must not change established legacy behavior.

## Solution

- Added an explicit template validation mode to the transport-agnostic bot creation spec:
  - `LEGACY` remains the default for existing internal callers.
  - `PUBLIC` is selected by the OpenAPI create and authorization-completion paths.
- Kept server-managed-field rejection enabled for OpenAPI inputs.
- Preserved legacy template snapshots, including `template_uid`, while retaining existing applicationCoding policy and validation checks.
- Added regression tests for both legacy preservation and OpenAPI-style rejection.

## Validation

- `python -m compileall -q` passed for the changed production and test files.
- `git diff --check` passed.
- The targeted pytest suite could not run because the environment could not resolve `mirrors.aliyun.com` while downloading `pyyaml==6.0.3`.

## Compatibility and risk

This is a compatibility fix. Existing `/api/bots` callers keep the previous template behavior. OpenAPI callers retain the newer server-managed-field validation. No database or public schema migration is required.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/bots/router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/bots/router.py
@@ -70,6 +70,7 @@ from agentclaw.community.core.bot_management.create_flow import (
     BotCreateContext,
     BotCreateDeploymentMode,
     BotCreateSpec,
+    BotCreateTemplateValidationMode,
     complete_bot_authorization,
     create_bot_with_authorization,
 )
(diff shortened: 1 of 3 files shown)


totalfrank (approved):
LGTM.

PR #1463, opened 2026-08-25
fix(backend): refresh Caller Service and Caller Instance targets

yhzhang35:
## Problem

Caller-aware Bot conversations can target both a shared Caller Service runtime
for `draft`, `verify`, or `online` and a caller-specific Expert Chat instance.
The IAM refresh path needed to update every eligible runtime binding without
mistakenly selecting a different user's dynamic instance or performing a
separate Caller credential exchange for each target.

## Solution

- Add explicit `CALLER_SERVICE` and `CALLER_INSTANCE` runtime-binding targets
  while preserving Session Files `AUTO` behavior.
- Use the existing collaboration-lock snapshot to select an eligible Caller
  Service runtime; permit the owner when no lock exists.
- Independently include the authenticated user's active Caller Instance.
- Exchange one opaque Caller Token per IAM request and reuse it for every
  selected outbound-rule update.
- Return IAM success when at least one target updates; preserve the existing
  outbound failure when no target can be updated.

## Validation

- `uv run --project src/backend pytest -q src/backend/tests/community/core/runtime_binding src/backend/tests/community/core/caller_identity src/backend/tests/community/adapters/http/openapi_v1/engine_runtime/test_session_files.py src/backend/tests/community/endpoints/test_openapi_session_files.py src/backend/tests/community/di/test_community_router_bindings_resolve.py` — 79 passed.
- `uv run --project src/backend ruff check ...` on the affected Python files — passed.
- `uv run --project src/backend python -m compileall -q ...` on affected backend modules — passed.

## Compatibility and risk

This is a server-side change. It does not accept a client-controlled binding id
and leaves Agent Run locking, relay/WebSocket behavior, Session Files `AUTO`
resolution, and BaaS append semantics unchanged.

## Spec

`spec/pipeline/caller-iam-multi-target-refresh/001-spec-output.md`

### DIFF
--- a/spec/pipeline/caller-iam-multi-target-refresh/001-spec-output.md
+++ b/spec/pipeline/caller-iam-multi-target-refresh/001-spec-output.md
@@ -19,6 +19,8 @@ Session Files runtime-binding resolver.
   instance, independently of the service lock.
 - Treat the IAM refresh as successful when at least one target is updated.
 - Return a controlled error only when no target is available or updated.
+- Exchange the opaque Caller credential once per IAM request, then reuse that
+  credential for every selected binding update.
 
 ## Explicit non-goals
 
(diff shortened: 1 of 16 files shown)


totalfrank (approved):
LGTM.

PR #1481, opened 2026-08-25
perf(backend): cut /bots/all serial query count for faster inventory

lucas-xzp:
## Problem

Prepub `GET /openapi/v1/bots/all` responds at RT p50=622ms (p90=3.7s, max 17.4s; access.log, 6h sample). One request runs ~12-14 serial DB round trips. Two avoidable chunks dominate:

- `DesktopBotService.list_user_bots` queried once per status — 5 serial trips (PENDING/ACTIVE/OFFLINE/RELEASING/FAILED)
- Every 200-row page pulled by the inventory passed through `_attach_template_configs_to_bots`, although the inventory response provably never reads `template_config` (neither `BotInventoryItem` nor the router `_to_inventory_item` maps it)

## Solution

- `search_bots` gains `bot_status_list` (status IN filter); `list_user_bots` collapses 5 serial queries into one (500/page with a fill loop, coverage ≥ the old 100×5; failure keeps the old swallow-and-log contract)
- `list_bots_by_conditions` gains `attach_templates: bool = True` (default preserves the established get/list consistency contract); the inventory cloud pull passes `False` — one fewer batched read per pulled page, behavior unchanged
- No API shape, response, or error behavior changes

## Validation

- `ci_test.sh --base origin/REL20260826`: 14460 passed, case pass rate 100%, line coverage 87.08%, changed-line coverage 94.44% (≥80% gate)
- `TestListUserBots` rewritten for single-query semantics (single-IN + pagination fill + swallowed failure); new opt-out (misc) and pass-through (inventory) coverage
- All `search_bots` call sites verified keyword-only — the new parameter cannot shift positional mappings
- Post-deploy check: prepub /bots/all RT p50/p90 tracked via antlogs (current baseline p50 ~600-800ms); P2 (query parallelization + SQL pagination pushdown) targets <150ms next

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_inventory/protocols.py
+++ b/src/backend/src/agentclaw/community/core/bot_inventory/protocols.py
@@ -36,6 +36,9 @@ class BotInventoryBotPort(Protocol):
         page: int = 1,
         page_size: int = 20,
         bot_ids: list[str] | None = None,
+        # Inventory cards never surface template_config, so inventory pulls
+        # opt out of the batched template read.
+        attach_templates: bool = True,
     ) -> Mapping[str, Any]: ...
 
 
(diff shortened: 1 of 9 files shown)


totalfrank (approved):
LGTM.

Issue #926, opened 2026-08-10
Public API: per-bot startup scripts executed on container start


totalfrank:
## Problem

This is a feature request.

Teams operating bots on Avernet need to provision their own environment inside the bot container: install a custom CLI, install extra software, or preload custom plugins/skills before the agent starts serving. Today there is no supported way to attach that provisioning to a bot. Anything installed by hand inside a running container is lost the moment the container is recreated — a restart (`POST /openapi/v1/bots/{bot_id}/restart`), a republish, or a scale-out all come back with a bare image — so every restart means re-provisioning by hand, and a scaled service bot has no way to make its instances identical.

What is missing is a script that lives **on the bot** (not in the image, not in a running container) and is executed on every container start. The content is the kind of thing a DaaS-style `install` script does: a shell script that fetches and installs a CLI, drops a plugin into place, and exits.

## Scope

A public API surface for storing, reading, and updating a per-bot startup script, plus the runtime hook that executes it on container start.

**API (`/openapi/v1/bots/{bot_id}/startup-script`)**
- `GET` → the currently stored script for the bot.
- `PUT` → set or replace the script content. Takes effect on the next container start.
- `DELETE` → clear it.
- Enveloped like the rest of the group (`Envelope[T]`), user principal security, no `entity_id` params — consistent with the `bots` group contract in `src/gateway/specs/2026-07-22-gateway-v1-api-definition/`.

**Execution**
- Run the stored script when the bot's container starts, for **personal bots** (BaaS-provisioned devices) and **service bots** (published runtimes) alike.
- Updating the script does not disturb a running container; the caller restarts the bot to pick it up. The restart → script → engine-ready path is the loop this feature is meant to close.

**Semantics to define before implementation**
- Ordering and timing: where the script runs relative to engine startup, and whether the bot reports ready before or after it completes.
- Failure handling: does a non-zero exit block readiness, degrade the bot, or only get recorded? Timeout and its consequence.
- Re-run behavior: the script runs on *every* start, so the contract is "must be idempotent" — state that in the API docs rather than trying to dedupe.
- Observability: how a caller retrieves stdout/stderr and the exit status of the last run; without this a failed install is invisible.
- Limits: max script size, allowed interpreter/shebang, whether network egress during install is expected.

**Multi-instance and stages**
- A scaled service bot must run the script on every device instance; decide what the API reports when instances diverge (one install failed, the rest succeeded).
[...]

PR #1483, opened 2026-08-25
refactor(backend): put the create-bot deploy_config behind a composer seam

totalfrank:
## Problem

Three fields of the BaaS create-bot `deploy_config` describe **the container a deployment runs**, and nothing else in the payload does:

| Field | What it is |
| --- | --- |
| `after_create_cmd_hook` | the shell that starts the bot inside the image |
| `mount_points` | the directories mounted into it |
| `storage` | the storage volume attached to it (absent ⇒ none) |

All three were hardcoded to the managed bot image: a four-step chain over `/home/admin/bin/*.sh`, three NAS mounts, and always a storage block.

The ACK/ECI deployment runs the open-source engine image, which carries its own entrypoint and an object-storage substrate. It does not need the four steps parameterised — it needs **one** command, its own mount list, and its own answer on storage. Those are different answers, not the managed ones with different arguments.

This covers the three backend items in the deployment design:

1. 定制开源版本启动脚本
2. 挂载路径插件管理（OSS）
3. 是否挂载 NAS 的开关

## Solution

A strategy, not a set of knobs.

```
core/service_bot/services/deploy/
├── deploy_models.py            # Storage / MountPointEntry (moved out of baas_service)
├── deploy_config_composer.py   # BotDeployContext + DeployConfigComposer (ABC)
├── managed_composer.py         # ManagedDeployConfigComposer — today's behavior
└── ack_composer.py             # AckDeployConfigComposer — ACK/ECI, unimplemented
```

```python
class DeployConfigComposer(abc.ABC):
    def build_start_command(self, ctx: BotDeployContext) -> str: ...
    def build_mount_points(self, ctx: BotDeployContext) -> list[MountPointEntry]: ...
    def build_storage(self, ctx: BotDeployContext) -> Storage | None: ...
```

One interface owns all three because they co-vary — they are one description of one runtime. Splitting them lets the mounts and the storage drift into disagreeing about where a bot's data lives.

`ManagedDeployConfigComposer` holds today's behavior, moved verbatim (~440 lines off `BaasService`, which shrinks by 621).

`AckDeployConfigComposer` raises `NotImplementedError` on all three, deliberately. Returning plausible-looking managed values would produce bots that start, report healthy, and do not work on a runtime nobody has tested — the expensive failure mode. Its docstring spells out what each method owes its caller, for whoever implements it.

`baas.deploy_runtime: managed | ack` selects one in `ServiceBotModule` (Rule 14: config-driven, composition root only). **An unrecognized value fails the boot** rather than falling back to `managed` — a typo that quietly shipped the managed image's boot chain to an ACK deployment is far more expensive than a failed wiring.

Like `DeployArtifactProducer` next door, these are **core strategies, not plugins**: they compose strings and value objects and cross no boundary, so no `local`/`prod` split (arch Rules 14 + 19).

### What stayed on `BaasService`

Everything a composer must not need to know: the `nas_mount` whitelist, outbound rules, resource spec, env/image overrides, template uuid — and the **per-bot startup script (issue #926)**, appended after the composer's chain so a stored script runs on every deployment without each composer re-implementing the `__OCB_RC` exit-status wrapper that makes it safe.

### Two consequences of the seam (not behavior changes)

- `mount_home_dir_storage` is a required `bool` instead of tri-state `bool | None`. `BaasService` resolves the whitelist once into `BotDeployContext`, so one payload makes one whitelist read instead of up to three, and the mounts cannot read a different answer than the storage.
- `_setup_directory` no longer takes `owner_id` — it read it only to guard that whitelist call.

## Validation

**The load-bearing check** is `test_baas_service_deploy_config_golden.py`: the entire `deploy_config` pinned byte-for-byte, written and run **against the pre-refactor code** first. It covers the whitelist on/off, a service bot vs a personal bot with no migration path, and a bot with a startup script. It passed before the move and passes after — the managed deployment sends exactly what it sent before.

Also added:
- `deploy/test_deploy_config_composer.py` — ABC contract (a partial impl cannot be instantiated), `BaasService` delegation, all three calls seeing one consistent context, `None` storage dropping the key, the startup-script stage wrapping whatever the composer returns, and the ACK stubs failing with a usable message.
- `di/test_deploy_runtime_selection.py` — default, selection, and an unknown value failing the boot with the valid names in the message.
- The four existing test files that reached into the moved private methods now target the composer directly.

Suites run green: `core` (8115 tests), `di`, `architecture`, `contracts`, `api`, `repository`, `config`, `kernel`, `plugin_api`, `unit`, `services`, `utils`, `local`, `framework`, `integration`, `compatibility`, plus the root smoke tests. Flake8 clean on the changed files for the SAST block rules.

**Not run, with reasons:**
- `tests/community/adapters` and `tests/community/endpoints` — these hang in my environment. Verified `adapters` hangs identically on a clean `dev` tree, so it is pre-existing and unrelated to this change.
- `test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs` — needs `rsync`, not installed here; fails on a clean tree too.
- `tests/community/acceptance` — excluded by `pytest.ini` (needs a spawned baas/mosn/sofa-registry stack).

## Compatibility and risk

No behavior change for any existing deployment. `deploy_runtime` defaults to `managed`, and the golden snapshot proves the managed payload is byte-identical; the only effective-config drift is the additive `"deploy_runtime": "managed"` in all three profile goldens.

`Storage` and `MountPointEntry` moved to `deploy/deploy_models.py` but are re-exported from `baas_service`, so existing imports keep working.

Rollback is the revert — nothing selects `ack` today, and nothing can until the ACK composer is written.

## Related issues

Backend half of the Aliyun deployment design (§4.3 backend 后端拼接逻辑). The ACK composer bodies are left for the deployment work that owns that runtime.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KxTykfcjbx1xU7LeqA6S9s)_


src/backend/src/agentclaw/community/configs/application-community.yaml
```diff
@@ -61,6 +61,14 @@ user_config:
...
     tenant: "community"
     template_uuid: ""
+    # Which container this deployment runs, and therefore which
+    # DeployConfigComposer shapes the create-bot payload (the boot chain, the
+    # mount points, the storage volume):
+    #   managed — the managed bot image (four /home/admin/bin scripts, NAS)
+    #   ack     — the open-source engine image on managed Kubernetes / ECI
+    # Omitted ⇒ managed. An unrecognized value fails at boot rather than
+    # falling back, so a typo cannot quietly ship the wrong image's payload.
+    deploy_runtime: "managed"
```
totalfrank-personal: This mode is for our existing set up right? If so, it goes to application.yaml because on corp deployment, we also need it.

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import Any, Dict
+
+
+@dataclass
+class Storage:
+    """
+    Sandbox storage.
+    """
+
+    type: str
```
totalfrank-personal: Does it make sense for this to be an enum? I feel like string type is hard to understand, I don't know what options I have.

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Storage type.
+    """
```
totalfrank-personal, line 93: Please make documentation more intuitive, examples for each different type is even better.

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
```diff
@@ -0,0 +1,71 @@
...
+    path: str
+    """
+    Storage path.
+    """
```
totalfrank-personal: Same, make documentation more intuitive.

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
```diff
@@ -0,0 +1,71 @@
...
+    storage_id: str
+    """
+    Storage id.
+    """
+    quota: str
+    """
+    Storage quota, such as "1Gi".
+    """
+    permission: str
```
totalfrank-personal, line 119: Same, we need intuitive documentation for this. Also, does enum makes more sense?

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
```diff
@@ -0,0 +1,71 @@
...
+class MountPointEntry:
+    """OSS mount point configuration for DeployConfig.
+
+        Platform-agnostic representation, converted to Arca MountPoint when needed.
+        Keeps domain model independent of Arca SDK per D-01.
+        Field names match Arca MountPoint for clarity (id, remote_dir, local_dir, permission).
+        """
+    remote_dir: str
+    local_dir: str
+    permission: str
```
totalfrank-personal: Same, does enum makes more sense? I just feel like that's easier for me to understand.

src/backend/src/agentclaw/community/di/modules/service_bot_module.py
```diff
@@ -182,12 +193,52 @@ def engine_sandbox_registry(
...
+
+        Rule 14: the choice is config, made once here, never branched on
+        downstream. An unrecognized value raises instead of defaulting to
+        ``managed`` — a typo that silently composed the managed image's boot
+        chain for an ACK deployment would produce bots that start and do not
+        work, which costs far more than a failed boot.
+        """
+        from agentclaw.community.core.storage import path as storage_path
+
+        if baas.deploy_runtime == ACK_DEPLOY_RUNTIME:
```
totalfrank-personal: That's a bad idea to inject baas and queries its field. We should make deploy_runtime a component and inject it directly.

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
```diff
@@ -724,6 +689,18 @@ def _build_create_bot_payload(
...
+            # Resolved here rather than inside the mount and storage builders,
+            # which used to each re-ask on a ``None``. One read per payload, and
+            # the composer is handed a decision instead of a whitelist.
+            #
+            # The ``owner_id`` guard is the mount builder's: the whitelist keys
+            # on the owner, so with no owner there is no membership to look up
+            # and the answer is the fail-closed default (sessions directory).
+            mount_home_dir_storage = bool(owner_id) and self._should_mount_home_dir_storage(
+                owner_id=owner_id,
+                bot_id=bot_id,
```
totalfrank-personal: Are we saying we do this here, because otherwise we do it twice in two different places?

If so, I guess we should move this logic before this if branch. Because we needed it anyway after this migration. Right?

And I guess you probably don't need the bool(owner_id) thing?

totalfrank-personal (commented):

src/backend/tests/community/core/service_bot/services/test_baas_service_resource_spec.py
```diff
@@ -106,7 +114,7 @@ def test_resolve_resource_spec_no_ext_returns_none():
 
 def test_build_create_bot_payload_supports_auto_approve_and_extra_envs():
     svc = _make_service()
-    svc._get_start_cmd = MagicMock(return_value="echo start")
+    svc._deploy_composer.build_start_command = MagicMock(return_value="echo start")
     svc._get_destroy_cmd = MagicMock(return_value="echo destroy")
     svc._setup_directory = MagicMock(return_value=[])
     svc._setup_bot_storage = MagicMock(return_value=None)
```
totalfrank: Non-blocking: `svc._setup_directory` / `svc._setup_bot_storage` no longer exist on `BaasService` — they moved to `ManagedDeployConfigComposer` in this PR (`baas_service.py` now calls `self._deploy_composer.build_mount_points(deploy_ctx)` / `.build_storage(deploy_ctx)`). These two mock assignments are dead: they just add unused attributes to `svc`, and the real composer methods run instead against `storage_path=MagicMock()` / `sandbox_registry=MagicMock()` (set up in `_make_service()` above) — e.g. `_setup_directory` calls `sp.get_bolt_data_path(...)` and formats the MagicMock result into `mount_points`, and `_setup_sessions_dir` stores a MagicMock as `Storage.path`.

That happens not to raise or fail today's assertions (this file only checks `envs`/`auto_approve_publish`), but the tests are no longer isolating payload construction the way they claim to, and the same pattern repeats in the other 5 functions in this file (lines ~153-154, 181-182, 214-215, 242-243, 291-292). `svc._deploy_composer.build_start_command = MagicMock(...)` (the line right above) was correctly migrated to target the composer — these two lines should get the same treatment, e.g.:

```suggestion
    svc._deploy_composer.build_mount_points = MagicMock(return_value=[])
    svc._deploy_composer.build_storage = MagicMock(return_value=None)
```

(and likewise at the other 5 call sites).

---

src/backend/src/agentclaw/community/di/modules/config_module.py
```diff
@@ -403,6 +403,10 @@ def baas(self) -> cfg.BaasConfig:
                 "personal_bot_template_uuid", defaults.personal_bot_template_uuid
             ),
             default_ttl_minutes=int(ttl),
+            deploy_runtime=str(
+                block.get("deploy_runtime", defaults.deploy_runtime)
+            ).strip()
+            or defaults.deploy_runtime,
```
totalfrank: Non-blocking: `.strip() or defaults.deploy_runtime` means a `deploy_runtime` value that is empty or whitespace-only after stripping (e.g. `deploy_runtime: ""`, or a templated overlay whose interpolation produced a blank string) silently resolves to `"managed"` instead of failing the boot.

That's the opposite of the invariant this feature is built around — stated in the docstring on `BaasConfig.deploy_runtime`, the `application.yaml` comment, the README, and `ServiceBotModule.deploy_config_composer` itself: "An unrecognized value fails the boot rather than falling back... a typo that quietly shipped the managed image's boot chain to an ACK deployment is far more expensive than a failed wiring." A blank string from a broken overlay is exactly the kind of misconfiguration that invariant is meant to catch, but here it's silently absorbed into the default instead of reaching the `ValueError` in `deploy_config_composer`.

`test_deploy_runtime_selection.py::test_an_unknown_runtime_fails_the_boot` even parametrizes `""` as an expected-to-fail value, but it calls `ServiceBotModule().deploy_config_composer` directly rather than going through this YAML-loading path, so it doesn't catch that blank strings never reach that check here.

---

totalfrank (commented):
Reviewed commits c425685 and 3409ced (merge-base 7a75efc..3409ced), 34 files. Focused on the composer seam (`deploy_config_composer.py`, `managed_composer.py`, `ack_composer.py`, `deploy_models.py`), the `BaasService`/DI wiring changes, the `deploy_runtime` config plumbing, and the moved/new tests against the pre-refactor code to check the "no behavior change" claim.

The managed composer's behavior matches the pre-refactor code closely (verified against the old `baas_service.py` at the merge-base), the golden payload test pins the actual create-bot payload byte-for-byte, and the composition-root wiring for `deploy_runtime` selection (Rule 14, fail-loud on an unrecognized value) is correctly scoped to `ServiceBotModule`. Two non-blocking items posted inline: a config-loading path where a blank `deploy_runtime` silently falls back to the default instead of failing the boot (undercutting the PR's own fail-loud design goal), and dead mock stubs in `test_baas_service_resource_spec.py` left over from the `_setup_directory`/`_setup_bot_storage` → composer move, which mean those tests are unintentionally exercising the real composer against `MagicMock` collaborators instead of the isolation they intend.

I also traced the `bool(owner_id)` guard now applied uniformly to both mount points and storage (flagged by a human reviewer on `baas_service.py:702`) against the pre-refactor code: the old `_setup_bot_storage` had no such guard, so this unifies a case where mounts and storage could previously disagree (when `owner_id` is falsy and the whitelist's `enable_all` is `True`) into a consistent fail-closed answer. That's a net-positive consistency fix, not a regression, so I didn't file it as a finding.

---

src/backend/src/agentclaw/community/configs/application-community.yaml
totalfrank: Yes — `managed` is the existing setup. Moved to `application.yaml` in 3409ced.

Worth confirming the merge actually works that way: the community overlay still has its own `baas:` block (api_base_url / tenant / template_uuid), so the base's `deploy_runtime` would be lost if overlays replaced blocks rather than deep-merging them. They deep-merge — the regenerated golden snapshots show all three profiles resolving `managed`, including community. That's also the evidence corp inherits it with no change on its side.

---

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
totalfrank: Agreed — `StorageType` in bc444f1. One member (`NAS`) because that's the only store this platform provisions a bot volume on today; the point is that the type now answers "what may I put here?" instead of leaving you to grep call sites, and a different substrate adds a member.

Same treatment for `MountPointEntry.permission` → `MountPermission`. Its values are the member *names* (`"READ_ONLY"` / `"READ_WRITE"`) rather than BaaS's own `ro`/`rw` spelling — that's what this service has always sent and what BaaS special-cases before falling back to `MountPermission(value)`, so the payload is unchanged.

---

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
totalfrank: Documented in bc444f1, but the answer on the enum here is **no**, and I've put the reason in the docstring so the next reader doesn't have to ask:

`Storage.permission` is a POSIX mode (`"0777"`, `"0755"`) — a nine-bit mask of owner/group/other × read/write/execute. An enum would need 512 members to say what four characters already say precisely. `StorageType` and `MountPermission` name a small closed set of choices; this names a number.

`Storage` also now carries two worked examples on the class — the same bot before and after the `nas_mount` whitelist moves it to its home directory — since the fields make more sense seen together than one at a time.

---

src/backend/src/agentclaw/community/di/modules/service_bot_module.py
totalfrank: Fixed in bc444f1. `deploy_config_composer` now injects `DeployRuntimeConfig` instead of the whole `BaasConfig` — it had no business receiving the endpoint, tenant and template uuids to read one field out of them.

Two things came with it:

- The runtime is a `DeployRuntime` enum, in `kernel/` because it's the one place both the composers (which name themselves with it) and the DI layer (which reads it from config) can import without a cycle. `DeployConfigComposer.name` is typed as it too, so the boot log and `baas.deploy_runtime` can't drift.
- Validation moved to config load (`ConfigModule.deploy_runtime`), so an unrecognized value fails there naming the values that would have worked. That also closes the blank-string hole the bot flagged separately: `deploy_runtime: ""` now raises instead of resolving to `managed`, and the test parametrizes `""` through the YAML path rather than only through the wiring.

Selection is now a `match` over the enum, so adding a member without wiring a composer fails at wiring rather than at some deployment's first create.

---

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
totalfrank: Right on both counts — done in bc444f1.

Yes: the mount builder and the storage builder each used to re-ask the whitelist on a `None`, so one payload could make three reads and the two halves could in principle disagree about where a bot's data lives. Resolving once and handing the composer the decision is the point.

Moved above the branch as you suggest. One thing I checked before doing it, since it's not a pure move: re-resolving *inside* the `if` meant the whitelist overrode a caller that had explicitly stated the flag. Hoisting makes the caller win. That's dead in practice — both callers that pass it (`bot_service` restart, where `mig` is always `None`, and `baas_device_service` allocate, which passes `""`) pass an empty migration path, so the override never fired. Noted in the comment there.

`bool(owner_id)` dropped. It came from the old `_setup_directory`, and the whitelist keys on the owner anyway, so an empty owner is a non-member — the guard was saying twice what the lookup already says.

---

src/backend/tests/community/core/service_bot/services/test_baas_service_resource_spec.py
totalfrank: Confirmed and fixed in 2eae496 — all six sites, exactly as suggested. I migrated `_get_start_cmd` when I moved these methods and missed its two siblings.

Verified the new assignments are actually on the live path rather than assuming it: made `build_mount_points` raise, and a test fails. The old `svc._setup_directory` assignment produced no such failure, which is the concrete form of "these were dead".

---

src/backend/src/agentclaw/community/di/modules/config_module.py
totalfrank: Correct, and fixed in bc444f1 — though via a different route than a tightened `.strip()`.

`deploy_runtime` moved off `BaasConfig` onto its own `DeployRuntimeConfig` component, and the value is now parsed through the `DeployRuntime` enum at config load. `""`, whitespace, and any other non-member all raise a `ValueError` naming the valid values; only an *absent* key falls back to `managed`, which is the case the default is actually for.

You're also right that the existing test didn't cover this: it parametrized `""` but went straight to `deploy_config_composer`, bypassing the YAML path where the blanking happened. `test_an_unknown_runtime_fails_at_config_load` now runs those values through `ConfigModule.deploy_runtime`.

---

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
totalfrank: Done in bc444f1. `Storage.path` now says it's the path **inside the container**, with both real values it takes (`/home/admin` for a home-directory bot, `/home/admin/.openclaw/agents` for one on its engine's sessions directory).

Took the "examples" suggestion up a level too: `Storage` and `MountPointEntry` each carry worked examples on the class, since the fields explain each other better together than one at a time. The module docstring now also states the distinction the two types kept blurring — a mount point bind-mounts a directory that already exists, `Storage` asks BaaS to provision the bot's own volume.

---

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
totalfrank: Done in bc444f1. `storage_id` was the worst of these — "Storage id" said nothing. It now explains what it's *for*: it's how BaaS finds the same volume again on the next start, which is what makes a bot's state survive a restart. It also documents the `{device_uuid}` placeholder BaaS substitutes per device, and why (a multi-replica service bot needs one volume per replica, not several containers writing the same files).

---

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
totalfrank: Yes for this one — `MountPointEntry.permission` is now `MountPermission` (`READ_ONLY` / `READ_WRITE`) in bc444f1. This is the case where an enum clearly fits: a genuinely closed two-valued set that was previously a bare string you'd have to grep for.

`remote_dir` / `local_dir` got docstrings in the same pass, since "OSS mount point configuration" didn't say which end was which — they're now labelled source and destination.

The one I pushed back on is `Storage.permission` (see the other thread): that's a POSIX mode, a number rather than a choice.

---

src/backend/src/agentclaw/community/di/modules/config_module.py
totalfrank: Fixed — `ConfigModule.deploy_runtime()` (config_module.py) now runs a blank/whitespace `baas.deploy_runtime` through `DeployRuntime(str(raw).strip())`, which raises `ValueError` and is re-raised with the values that would have worked, instead of falling back to `defaults.deploy_runtime`. `test_an_unknown_runtime_fails_at_config_load` in `test_deploy_runtime_selection.py` now exercises exactly this path (parametrized with `""`) through the real YAML-loading provider rather than the composer-selection function. Resolving.

---

totalfrank (commented):
Reviewed the new commit since the last round: `bc444f1d` (`refactor(backend): type the deploy payload's closed sets as enums`), diffed against `3409ced7` where the last review landed.

**Findings checklist (mine, across all rounds):**
1. Non-blocking — `test_baas_service_resource_spec.py`: dead `_setup_directory`/`_setup_bot_storage` mock stubs at 6 call sites, left over from the composer move — **still open**, code at those lines unchanged, no claim of a fix in this commit.
2. Non-blocking — `config_module.py`: blank/whitespace `deploy_runtime` silently fell back to `"managed"` instead of failing the boot — **fixed** this round. `ConfigModule.deploy_runtime()` now runs the stripped value through `DeployRuntime(...)` unconditionally, and `test_an_unknown_runtime_fails_at_config_load` exercises the blank-string case through the real YAML-loading path. Replied and resolved.

**New commit (`bc444f1d`) review:** typing `Storage.type`/`Storage.permission`/`MountPointEntry.permission`'s closed sets as `StorageType`/`MountPermission` `StrEnum`s, making `deploy_runtime` its own injected `DeployRuntimeConfig` component (replacing the `BaasConfig` field), and moving the `mount_home_dir_storage` resolution above the `migration_path` branch while dropping the `bool(owner_id)` guard. Checked the enum `to_dict()` coercions (`str(StrEnum member)` returns the plain value, so the wire payload is unchanged), the exhaustive `match` in `ServiceBotModule.deploy_config_composer`, and the whitelist call now firing unconditionally on `owner_id=""` (`is_bot_feature_enabled` handles a falsy `owner_id` without raising — matches the "net-positive consistency fix" already noted last round). No new defects found; this also directly addresses the human reviewer's outstanding design comments (enum types, per-field docs, injecting `DeployRuntimeConfig` directly, and moving the `mount_home_dir_storage` resolution before the `migration_path` branch).

**Still open: 1** (item 1 above).

---

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
totalfrank: Addressed in bc444f1: every field in this file now carries a docstring with a worked example — `StorageType`/`MountPermission` explain the closed set they name, `Storage` carries two worked examples on the class (the same bot before/after the `nas_mount` whitelist moves it), and each field (`type`, `path`, `storage_id`, `quota`, `permission`) has its own docstring.

---

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
totalfrank: Addressed in bc444f1, same treatment as the sibling thread above — the module docstring plus every class and field in this file now has a docstring, most with a worked example.

---

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_models.py
totalfrank: Addressed in bc444f1: `MountPointEntry.permission` is now `MountPermission`, an enum with two members (`READ_ONLY`/`READ_WRITE`) and a docstring explaining why the values are the member names rather than BaaS's own `ro`/`rw` wire spelling — see the reply above on the sibling enum thread for the full reasoning.

---

totalfrank (commented):
Commits reviewed this round: none new — my most recent review (submitted against this same head, 2eae496) already covered every commit on the branch. This round verified the remaining open threads by reading the current code and closed out three human-reviewer threads that had never been replied to.

Checklist (all rounds):
- `application-community.yaml` deploy_runtime default location — fixed (3409ced), resolved
- `deploy_models.py` bare strings → enums (`StorageType`/`MountPermission`) — fixed (bc444f1), resolved
- `deploy_models.py:93` docs/examples request — fixed (bc444f1, full-file docstrings), resolved this round
- `deploy_models.py` "same, more intuitive docs" — fixed (bc444f1), resolved this round
- `deploy_models.py:119` `Storage.permission` docs / enum question — fixed (bc444f1, docstring explains why not an enum + worked examples), resolved
- `deploy_models.py` "same, enum?" (mount permission) — fixed (bc444f1, `MountPermission`), resolved this round
- `service_bot_module.py` DI takes whole `BaasConfig` instead of `deploy_runtime` — fixed (bc444f1, injects `DeployRuntimeConfig` directly), resolved
- `baas_service.py` whitelist read order / `bool(owner_id)` guard — fixed (bc444f1, hoisted above branch, guard dropped), resolved
- `test_baas_service_resource_spec.py` dead mock stubs (my finding) — fixed (2eae496, all 6 sites), resolved
- `config_module.py` blank `deploy_runtime` silently defaulting (my finding) — fixed (bc444f1, verified `""` now raises via `test_an_unknown_runtime_fails_at_config_load`), resolved

Open findings: 0

Nothing open and nothing new found this round — the PR is clean as of 2eae496.

---

FreddieSun (approved):
LGTM

PR #1486, opened 2026-08-25
Fix/agent public

rudyLi:
## Problem

  `BotPublicService.public_bcs_bot` calls `ApprovalWorkflowPlugin.start_approval` directly and raises `BotPublicServiceError` for any `success=False`; the route's `@envelope_errors` maps that to `500 "Publish
  failed"`. The community build binds `ApprovalWorkflowPlugin` to `NoApprovalWorkflow` (`plugins/community/approval_workflow.py`), which always returns `success=False`. Result: **every `public`/`protected` publish in
  the community build returns 500.**

  This contradicts the protocol's stated intent — `plugin_api/approval_workflow.py` documents "the community implementation reports 'no workflow' so callers fall through to direct publish." The legacy `public_bot`
  delegates to `BotPublishApprovalPlugin.publish`, and community's `DirectPublishApproval` relies on that exact fall-through to publish without an approval service. Only `public_bcs_bot` bypasses the strategy and
  treats "capability unavailable" as "rejected."

  ## Solution

  Add a unified sentinel and branch on it inside `public_bcs_bot` — no existing interface/protocol changes.

  - New `NO_WORKFLOW_MARKER = "[no-approval-workflow] "` (`plugin_api/approval_workflow.py`), with the protocol docstring documenting the fall-through convention that relies on it.
  - Prepend the marker in both no-workflow plugins: `plugins/community/approval_workflow.py` (`NoApprovalWorkflow`) and `plugins/local/antprocess.py` (`LocalAntProcessService`).
  - `public_bcs_bot` (`core/bot_public/services/bot_public_service.py`) now has three branches after `start_approval`:
    - `success=True` (corp real ticket) → unchanged.
    - **unavailable** (marker prefix) → synthesize `COMPLETED/AGREE` and reuse the existing `_persist_public_approval` + inline `handle_public_approval_callback` path, driving the BCS `friend_ext` AGREE update +
  visibility flip exactly like corp's auto-completed tenant; return `state=SKIPPED` (the approval step was skipped, but the BCS update did run).
    - real rejection (no marker) → still raises.
  - hygiene: `_apply_visibility_direct` reports `state=SKIPPED` when `patch_attributes` returned `skipped` (no provider creds) instead of faking `COMPLETED`.

  Rejected: (a) a fully direct-BCS "no-op" return — would silently skip the BCS visibility flip the publish is supposed to apply; (b) an explicit `available` capability flag on the `ApprovalWorkflowPlugin` return
  shape — would require synchronizing the change across community/local/corp workflow impls; (c) `isinstance`-checking against `NoApprovalWorkflow` — couples the service to a concrete impl. The sentinel-prefix
  approach changes neither any protocol's return shape nor the corp path, and the two tests asserting the sentinel tolerate a prefix (truthiness / substring), so no existing test changed.

  ## Validation

  - `pytest tests/community/core/bot_public/test_bot_public_service.py::TestPublicBcsBot tests/community/contracts/test_approval_workflow.py` → 34 passed.
  - `pytest tests/community/endpoints/test_openapi_public_bcs.py tests/community/adapters/http/openapi_v1/test_collaboration_bots_endpoints.py tests/community/core/bot_public
  tests/community/adapters/http/openapi_v1/bot_public` → 180 passed.
  - community-profile injector smoke (real `NoApprovalWorkflow` via `build_injector(DeployProfile.COMMUNITY)`, non-mocked BCS) → 2 GET / 2 PATCH; the AGREE patch carries the `visibility` field and
  `friend_ext.public_agent_approval.status=AGREE`; returns `state=SKIPPED`, no raise. ✅
  - Could not run: end-to-end corp/real-antprocess path (corp workflow impl is not in this community checkout).

  - Wire shape: `BcsPublishResult` (`adapters/http/openapi_v1/collaboration_bots/schemas.py`) already carries `state`/`error_msg` with `extra="allow"`, so the new `SKIPPED` value and the carried `error_msg` survive
  `model_validate`; response surface stable.
  - No protocol/interface changes; no schema/config/migration impact; no rollback path needed beyond reverting the commit.
  - Behavior changes: community `public`/`protected` publishes now succeed with `state=SKIPPED` (previously 500); `visibility=private` direct path now reports `SKIPPED` when BCS is skipped (previously misleading
  `COMPLETED`). corp real-antprocess path is unchanged on all branches.

  ## Spec (optional)

  `public_bcs_bot` contract: `core/bot_public/services/bot_public_service.py:436`. BCS field/scope maps: `_BCS_VISIBILITY_FIELD_BY_SCOPE`, `_BCS_APPROVAL_KEY_BY_SCOPE`, `_BCS_VIEW_SCOPE_DEPS_KEY_BY_SCOPE`. Callback
  decision path: `_apply_callback_decision`.

  ## Related issues (optional)

  Follow-ups tracked out of scope (separate PRs): `biz_id` second-granularity collision (`public_bcs_bot:520`); `handle_public_approval_callback` legacy retry comment vs `max_retries=1` mismatch (`:909-911`, unrelated
  non-`public_scope` path).

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_public/services/bot_public_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_public/services/bot_public_service.py
@@ -12,7 +12,10 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime
 from typing import Any, Callable, Dict, List, Optional, TYPE_CHECKING
 
-from agentclaw.community.plugin_api.approval_workflow import ApprovalWorkflowPlugin
+from agentclaw.community.plugin_api.approval_workflow import (
+    NO_WORKFLOW_MARKER,
+    ApprovalWorkflowPlugin,
+)
 from agentclaw.community.core.repository.protocols.bot import BotRepository
 from agentclaw.community.core.bot_management.services.bot_service import BotService
 from agentclaw.community.core.bot_management.services.bcn_service import BcnService
@@ -66,6 +69,14 @@ def _resolve_access_mode(public: str, friend_approval: str) -> str:
     if public == "1" and friend_approval == "0":
         return "OPEN"
     return "RESTRICTED"
+
+
+def _is_workflow_unavailable(result: Dict[str, Any]) -> bool:
+    """True when ``start_approval`` returned "no approval capability" rather
+    than a real rejection. No-workflow impls prepend ``NO_WORKFLOW_MARKER`` to
+    ``error_msg`` (see ``plugin_api/approval_workflow.py``); callers must treat
+    this as a fall-through to direct publish, not a failure."""
+    return (result.get("error_msg") or "").startswith(NO_WORKFLOW_MARKER)
 
 
 # BCS-publish destination label for the approval prompt, keyed by public_scope.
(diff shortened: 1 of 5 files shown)


totalfrank (approved):
LGTM.

PR #1491, opened 2026-08-25
feat(bcs): use no-auth share link in upload completion notification

zh-bupt:
## Problem

The download URL embedded in the upload-completion system message used the authenticated `/sessions/{sid}/files/{file_id}/content` endpoint. Other bots and users in the session cannot fetch the file from that link without carrying caller credentials, which limits both the experience and the capability.

## Solution

Switch the download URL in the system messages produced by both the new endpoint (`bcs_api_http::v1::internal::routes::session_file::complete_file`) and the legacy endpoint (`bcs_http::routes::session_files::complete_upload`) to the no-auth share endpoint:

- Add a transport-agnostic `SessionFileSharedContentUrlProjector` trait and an `UPLOAD_COMPLETION_SHARE_TTL_SECONDS = 15 * 86_400` (15 days) constant in `service-api`.
- `SessionFileApplicationServiceImpl::complete()` now mints a 15-day share link after upload completion and projects the public shared-content URL via the new trait into the system message. On mint failure it logs and skips the notification instead of advertising a broken link.
- `SessionFileUrlProjector` implements the trait; the URL path is `/api/v1/collaboration/sessions/shared-file/content?token=...`.
- Raise `SHARE_TTL_MAX` from 7 days to 15 days to admit the completion share.
- Remove the now-unused `notification_content_url` field from `CompleteSessionFile` and from both complete routes, and delete the dead `file_download_url` helper plus its two unit tests.

Core logic stays transport-agnostic: the public base URL is injected by the delivery adapter through the new trait, with no hardcoded URLs.

## Validation

- `cargo test -p bcs-app-session`: 69 passed
- `cargo test -p bcs-session-file`: 58 passed
- `cargo test -p bcs-api-http`: 34 passed
- `cargo test -p bcs-http --lib routes::session_files`: 16 passed
- `cargo test -p bcs-service-api`: 3 passed

Not run: a full `cargo check -p bcs` is blocked by pre-existing compile errors in `bcs-edge-permission` and `bcs-app-invitation`, unrelated to this change. The `bcs-http` `admin_invocation_terminal` test fails in the sandbox with `PermissionDenied`, also unrelated.

## Compatibility and risk

- Removing `notification_content_url` from `CompleteSessionFile` is a breaking change to that internal command, but all call sites (both HTTP routes) are updated in the same change.
- The notification download URL changes from an authenticated content URL to a share URL; downstream bots that parse the message must switch to the no-auth download path. The existing authenticated download endpoint is unaffected.
- Expanding `SHARE_TTL_MAX` to 15 days only applies to newly minted share tokens; existing tokens are not affected.
- Rollback: revert this commit to restore the original authenticated content URL.

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-api-http/src/v1/internal/routes/session_file.rs
+++ b/src/bcs/crates/adapters/http/bcs-api-http/src/v1/internal/routes/session_file.rs
@@ -280,16 +280,12 @@ async fn complete_file(
     authorize_identity(&caller, policy).map_err(|e| error(&request_id, e))?;
     let Path((session_id, file_id)) =
         path.map_err(|e| invalid_request(&request_id, e.body_text()))?;
-    let notification_content_url = projector(&state)
-        .map_err(|e| error(&request_id, e))?
-        .content_url(&session_id, &file_id);
     let result = service(&state)
         .map_err(|e| error(&request_id, e))?
         .complete(CompleteSessionFile {
             caller,
             session_id,
             file_id,
-            notification_content_url,
         })
         .await
         .map_err(|e| error(&request_id, e))?;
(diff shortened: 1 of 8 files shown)


questchaos (approved):
LGTM

PR #1493, opened 2026-08-25
feat(bcs): add GET /openapi/v1/collaboration/groups/public plaza endpoint

zh-bupt:
## Problem

`GET /group` (`bcs_http::routes::groups::list_groups`) migrated to V1 `GET /openapi/v1/collaboration/groups` (`list_groups`), but the V1 endpoint is **participant-scoped** — it lists only groups the authenticated caller directly participates in, and `membership=all` was deliberately removed (test-guarded). So `?visibility=public` no longer lists *all* public groups; it returns only the caller's public groups. The product has a "public group plaza" where any end user browses all public groups and instantiates a session, which the new endpoint cannot serve.

## Solution

Add an independent, caller-agnostic catalog `GET /openapi/v1/collaboration/groups/public` that lists all `visibility=public` groups with no participant filter. Auth is still enforced by the `verify_principal` middleware layer (any authenticated user; cookie-SSO end users without an app credential are admitted via a method-qualified gateway override).

- `Membership` enum += `None` (catalog entries emit `membership=none`); the `list_groups` contract is unchanged.
- `ListPublicGroups` command + `GroupService::list_public_groups` **default** trait method (the `NoopGroupService` impls inherit it, zero edits). `GroupServiceImpl` reuses the existing store methods `count_filtered` / `list_paginated_filtered` (no new store methods) with in-app `strategy` + `active` filter and a `SAFETY_CAP = 1000` materialization bound; `total` is the post-filter count.
- HTTP: `ListPublicGroupsQuery` DTO (`deny_unknown_fields`) + caller-agnostic handler + route `/groups/public` registered before `/groups/{group_id}`.
- OpenAPI contract: `PublicGroupsPath` + path entry; `x-avernet-security { user: required }`; params exactly `{offset, limit, q, strategy}`.
- Gateway: regenerate `bcn.openapi.json` + auth override `"GET /openapi/v1/collaboration/groups/public": { user: required }`.

Rejected: re-adding `membership=all` to `list_groups` (reverts a deliberate, test-guarded decision and makes one route carry two auth policies); a switch that flips auth semantics on `list_groups` (same problem).

## Validation

- `cargo test -p bcs-app-group --test v1_group_service` — 6 new `list_public_groups_*` tests pass (public-only exclusion, strategy filter, `q` filter, active-only, pagination/filtered-total, limit bounds) + existing `list_groups_filters_by_visibility` still green.
- `cargo test -p bcs-api-http --test group_routes` — 2 new + 4 existing `list_groups_*` HTTP tests pass.
- `pytest tests/openapi/test_group_v1_contract.py` — 2 new contract tests pass; `validate_openapi_contract.py` exit 0.
- `cargo check --workspace --all-targets` green; 5 gateway test modules pass (192 tests).
- Pre-existing failure `test_update_group_does_not_accept_context` is **unrelated** (the `UpdateGroup` PATCH `context` field was added by commit `6824a63faa`; out of scope here).
- Not run locally: the full pre-push / singlebox coverage gates (`scripts/ci/pre_push.sh` with `OCB_PRE_PUSH_RUN_CI=1`) — relying on CI.

## Compatibility and risk

- Wire: `Membership` enum gains a `none` value — backward-compatible additive change; `list_groups` never emits it.
- Gateway snapshot `bcn.openapi.json` was stale since `9c558e9af1`; regenerating picks up ~97 pre-existing contract-drift field changes (from other already-merged commits) plus this PR's new path. The PR therefore carries unrelated snapshot catch-up; a separate `chore: regenerate stale snapshot` PR could precede this one if a narrower diff is preferred. A CI gate to detect snapshot staleness would prevent this.

## Spec

Design + implementation plan in `docs/superpowers/specs/2026-08-25-public-group-plaza-endpoint-design.md` and `docs/superpowers/plans/2026-08-25-public-group-plaza-endpoint.md` (on the companion per-task branch `feat/public-group-plaza-endpoint`; not bundled into this PR per request).


questchaos (approved):
LGTM

cassiuscai (approved):

PR #1500, opened 2026-08-25
fix(backend): preserve BCS catalog total

yhzhang35:
## Problem

Catalog Search currently replaces the BCS `/bots/search` total with the number of Backend records that survive the exact `(bot_id, entity_id)` join. When the BCS page contains records that are temporarily unavailable in Backend, the public response no longer reflects the authoritative BCS pagination total.

## Solution

- Preserve the validated non-negative integer `total` from the BCS catalog page in a transport-neutral page result.
- Keep the existing exact composite-address join, BCS metadata projection, sensitive-field sanitization, and response item ordering unchanged.
- Return the BCS total unchanged from Catalog Search; Backend no longer recomputes it from joined items.
- Fail closed when the BCS total is missing or malformed, and add regression coverage for totals larger than the joined item count.

## Validation

- `DEPLOY_PROFILE=test uv run pytest tests/community/adapters/http/openapi_v1/bot_public/test_bot_public_router.py tests/community/core/bot_public/test_bot_catalog_metadata_service.py tests/community/core/bot_public/test_bot_public_service.py -q` — 180 passed.
- `DEPLOY_PROFILE=test uv run pytest tests/community/architecture/test_repository_contracts.py tests/community/architecture/test_http_adapter_layer_is_http_only.py tests/community/architecture/test_no_fastapi_in_core.py -q` — 15 passed.
- Targeted `ruff check` — passed.
- Gateway OpenAPI schema JSON parse — passed.
- `git diff --check` against `REL20260826` — passed.

## Compatibility and risk

The public route, query parameters, BCS path, exact join key, item projection, and error mapping remain unchanged. The only response behavior change is that `data.total` now follows BCS rather than the post-join Backend item count.

## Spec

`spec/pipeline/catalog-search-bcs-total-rel20260826/001-spec-output.md` (local task artifact; intentionally not committed).

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_public/catalog_metadata.py
+++ b/src/backend/src/agentclaw/community/core/bot_public/catalog_metadata.py
@@ -51,6 +51,14 @@ class BotCatalogMetadata:
     user_visibility: Any = None
 
 
+@dataclass(frozen=True)
+class BotCatalogMetadataPage:
+    """One BCS page, retaining BCS pagination metadata for the public result."""
+
+    total: int
+    items: Sequence[BotCatalogMetadata]
+
+
 class BotCatalogMetadataUnavailableError(Exception):
     """The catalog metadata port cannot make an authoritative decision."""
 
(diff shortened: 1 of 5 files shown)


totalfrank (approved):
LGTM.

PR #1501, opened 2026-08-25
fix(bcs): split shared-file download base URL from upload-target base URL

zh-bupt:
## Problem

The share download link generated by `SessionFileUrlProjector::shared_content_url` used the same base URL as the protected content/upload URLs (`public_collaboration_base_url`). In split-path gateway deployments where protected routes live under `/openapi/v1/collaboration` and the no-auth share download lives under `/api/v1/collaboration`, setting `public_collaboration_base_url` to the `/openapi/v1` prefix caused the share link to point at `/openapi/v1/collaboration/sessions/shared-file/content?token=...` — a route BCS does not serve (the shared-file endpoint is mounted under `/api/v1/collaboration`), yielding a 404.

## Solution

Split the projector into two independently-configured base URLs:

- `public_collaboration_base_url` — for protected session-file URLs (`content_url`, `project_upload_target`). Typically `/openapi/v1/collaboration`.
- `internal_collaboration_base_url` — for internal-collaboration endpoints including the no-auth share download link (`shared_content_url`). Typically `/api/v1/collaboration`. Falls back to `public_collaboration_base_url` when unset, preserving backward compatibility.

```toml
[openapi_v1]
public_collaboration_base_url = "https://gateway.example.com/openapi/v1/collaboration"
internal_collaboration_base_url = "https://gateway.example.com/api/v1/collaboration"
```

Changes:
- `SessionFileUrlProjector::new` now takes two base URLs; `shared_content_url` uses the dedicated base, everything else uses `public_base`.
- `OpenApiV1Config` gains `internal_collaboration_base_url` (optional, same validation rules as the existing field).
- Update `bcs-config-example.toml` to document both fields with split-path examples.
- Update projector tests, the session-route integration test, and add config tests for the fallback and independent-set paths.

## Validation

- `cargo test -p bcs-api-http --test session_file_url_projection`: 4 passed (including new split-path test)
- `cargo test -p bcs-api-http --test session_routes`: 34 passed
- `cargo test -p bcs --lib config::tests`: 63 passed (including 3 new config tests)

## Compatibility and risk

- `SessionFileUrlProjector::new` signature changed from 1 to 2 args — all in-repo call sites updated. The trait impl (`SessionFileSharedContentUrlProjector`) is unchanged.
- `internal_collaboration_base_url` defaults to `None`, which falls back to `public_collaboration_base_url`, so existing single-path deployments are unaffected.
- Rollback: unset `internal_collaboration_base_url` (or revert this commit) to restore single-base behavior.

### DIFF
--- a/src/bcs/configs/bcs-config-example.toml
+++ b/src/bcs/configs/bcs-config-example.toml
@@ -297,7 +297,11 @@ group_link_url = "https://botchat.example.com/groups/{group_id}"
 session_link_url = "https://botchat.example.com/sessions/{session_id}"
 
 [openapi_v1]
-public_collaboration_base_url = "https://gateway.example.com/api/v1/collaboration"
+# Base URL for protected session-file URLs (e.g. proxy upload targets).
+public_collaboration_base_url = "https://gateway.example.com/openapi/v1/collaboration"
+# Base URL for internal-collaboration endpoints (content, shared-file download).
+# Falls back to public_collaboration_base_url when unset.
+internal_collaboration_base_url = "https://gateway.example.com/api/v1/collaboration"
 
 [session_files]
 storage_backend = "local"      # or "baas"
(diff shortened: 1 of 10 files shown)


questchaos (approved):
LGTM

PR #1502, opened 2026-08-25
fix(bcs): move plaza endpoint to /public-groups to avoid {group_id} collision

zh-bupt:
## Problem

The public group plaza endpoint shipped on `dev` at `GET /openapi/v1/collaboration/groups/public` (commit `91cf23fdbd`) collides with the `{group_id}` parameter on `GET /openapi/v1/collaboration/groups/{group_id}` — `public` is captured as a group id, so the catalog can't be routed/reached cleanly.

## Solution

Rename the catalog to a distinct top-level path `GET /openapi/v1/collaboration/public-groups` so it no longer sits under `/groups/` and cannot be shadowed by the `{group_id}` parameter. Single-commit change across:

- Rust route (`routes/group.rs`: `/groups/public` → `/public-groups`).
- HTTP tests (`group_routes.rs`): URIs updated; the obsolete "routes before group_id" test renamed to `list_public_groups_parses_query`.
- OpenAPI contract (`openapi.yaml` path key) + contract test (`PUBLIC_GROUPS_PATH`; the ordering test repurposed to `test_public_groups_path_does_not_collide_with_group_id_path`).
- Gateway auth override (`application.yaml`: `"GET .../groups/public"` → `"GET .../public-groups"`).
- `CONTEXT.md` path reference.
- Regenerated `bcn.openapi.json` snapshot (`public-groups` replaces `groups/public`).

## Validation

- `pytest tests/openapi/test_group_v1_contract.py` — 16 passed; the only failure is the pre-existing `test_update_group_does_not_accept_context` (the `UpdateGroup` PATCH `context` field was added by unrelated commit `6824a63faa`; red on `dev` already, out of scope).
- `pytest tests/unit/scripts/test_dump_and_publish_script.py tests/unit/scripts/test_gate_and_publish.py` — 7 passed (operation-count pins 57/78 unchanged; the rename moves one path, no count delta).
- `cargo test -p bcs-api-http --test group_routes` — 23 passed.
- `cargo test -p bcs-app-group --test v1_group_service` — 77 passed (app layer is path-agnostic; unaffected).

## Compatibility and risk

- Path-only change to a freshly-shipped endpoint (no caller depends on the old `groups/public` path yet). No schema/contract-shape change beyond the path key.
- Snapshot regenerated; operation count unchanged (57 public / 78 total).


cassiuscai (approved):

questchaos (approved):
LGTM

PR #1523, opened 2026-08-25
fix(backend): fix MCP rollback and directory download file caps

YJ2378:
## Problem

Two independent backend defects were identified:

1. `MCPConfigService.rollback_unified_config` used a truthiness check for `old_config`. As a result, an empty previous configuration (`{}`) was treated the same as `None`, causing the existing MCP configuration record to be deleted instead of restored.

2. Device-filesystem directory downloads counted both directories and regular files toward `_ZIP_DOWNLOAD_MAX_FILES`. In directory-heavy trees, the traversal could stop before discovering all regular files, resulting in an incomplete ZIP archive being returned with HTTP 200 instead of rejecting the download when the regular-file limit was exceeded.

## Solution

1. Changed the MCP rollback condition from a truthiness check to an explicit `None` check:
   - `old_config is None`: delete the configuration record;
   - `old_config == {}`: restore the empty configuration through the existing update path;
   - non-empty configurations continue to be restored as before.

2. Added an optional file-only limit to the device-filesystem walker:
   - only non-directory entries count toward the file limit;
   - the directory download path uses `_ZIP_DOWNLOAD_MAX_FILES + 1` to detect overflow;
   - HTTP 413 is raised before creating the temporary ZIP when the regular-file limit is exceeded;
   - the existing combined-entry safety limit and local-filesystem download path are preserved.

## Validation

The full Avernet backend CI suite also passed locally:
14496 passed, 55 skipped, 752 warnings in 401.49s (0:06:41)
case pass rate: 100.00% (14551/14551, required >= 100.00%)
line coverage: 87.48% (required >= 75.00%)
change line coverage: 88.89% (8/9, required >= 80.00%)

## Compatibility and risk

- No database schema or API contract changes.
- `None` and `{}` now retain their intended distinct rollback semantics.
- Existing search traversal behavior continues to use the combined `max_entries` limit.
- The local-filesystem download branch is unchanged.
- The changes are limited to the MCP configuration service and device-filesystem directory download flow.
- Both fixes can be reverted independently by reverting their respective commits.

### DIFF
--- a/src/backend/src/agentclaw/community/core/mcp/services/config_service.py
+++ b/src/backend/src/agentclaw/community/core/mcp/services/config_service.py
@@ -160,7 +160,7 @@ class MCPConfigService:
         old_config: Optional[dict[str, Any]],
     ) -> None:
         """回滚到旧配置。若 old_config 为 None，则删除该记录。"""
-        if old_config:
+        if old_config is not None:
             self.update_user_unified_config(
                 user_id=user_id,
                 server_code=server_code,
--- a/src/backend/src/agentclaw/community/core/mcp/services/config_service.py
+++ b/src/backend/src/agentclaw/community/core/mcp/services/config_service.py
@@ -159,8 +159,8 @@ class MCPConfigService:
         server_code: str,
         old_config: Optional[dict[str, Any]],
     ) -> None:
-        """回滚到旧配置。若 old_config 为 None，则删除该记录。"""
-        if old_config is not None:
+        """回滚到旧配置。若 old_config 为 None 或 {}，则删除该记录。"""
+        if old_config:
             self.update_user_unified_config(
                 user_id=user_id,
                 server_code=server_code,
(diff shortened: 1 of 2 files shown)


src/backend/src/agentclaw/community/core/mcp/services/config_service.py
```diff
@@ -160,7 +160,7 @@ def rollback_unified_config(
         old_config: Optional[dict[str, Any]],
     ) -> None:
         """回滚到旧配置。若 old_config 为 None，则删除该记录。"""
-        if old_config:
+        if old_config is not None:
```
totalfrank: We should revert this. Maybe the comment is inaccurate, but if old_config is empty(not None) we should also early return here.

totalfrank (commented):

totalfrank (commented):

totalfrank (approved):
LGTM.

PR #1524, opened 2026-08-26
fix(backend): enforce platform default MCP ownership

FreddieSun:
## Problem

PR #1437 established Installation as the Bot effective-capability fact, while engine/template Default MCPs remain an unmaterialized platform policy input. Two gaps could violate that boundary:

- Direct MCP activate/deactivate could control a platform Default MCP, even though Installation has no provenance for safely separating Direct state from platform policy.
- Runtime projection treated a failed template/ext-context lookup as an empty or narrower Default policy and could overwrite the runtime with missing MCPs.

Existing exclusion rows could also leave legacy same-code Installation rows behind, allowing the Installation half of the runtime union to bypass the exclusion.

## Solution

- Add a strict `PlatformDefaultMcpPolicy` that resolves engine/template Default MCP codes for the exact Bot context.
- Reject Direct activation and deactivation for platform Default MCPs in both the command service and transactional repository boundary. Default exclusion/un-exclusion remains their only control surface.
- Make Default MCP exclusion retire legacy same-code Installation rows even when the exclusion row already exists.
- Make Runtime projection resolve both ext info and template type strictly, and fail before overwrite-style MCP or Passport delivery when policy context is unavailable.
- Add a stable HTTP conflict mapping for platform-policy-owned resources and document the ownership boundary.

## Validation

- `uv run --project src/backend pytest src/backend/tests/community/core/skill_center src/backend/tests/community/repository/skill_center src/backend/tests/community/adapters/http/openapi_v1/test_mcp_control_endpoints.py src/backend/tests/community/architecture/test_installation_table_write_ownership.py src/backend/tests/community/architecture/test_no_legacy_activation_writers.py src/backend/tests/community/architecture/test_service_api_conformance.py -q`
  - 1053 passed, 27 skipped.
- Focused P1 regression suite: 187 passed.
- Blocking SAST rules over every changed Python file: passed.
- `git diff --check`: passed.
- The repository-wide source scan still reports seven pre-existing F821 findings in `core/skill_center/factories.py`; that file is unchanged by this PR.

## Compatibility and risk

- No schema or migration changes.
- Direct operations on engine/template Default MCPs now return conflict code `409210`; callers must use Default exclusion/un-exclusion instead.
- Internal desired-state repository MCP methods now require the already-resolved platform Default code set, ensuring the transactional boundary rechecks ownership.
- Rollback is the single commit in this PR.

## Spec

- Implements the approved Platform Default MCP ownership semantics on top of the Installation single-source-of-truth design: https://github.com/inclusionAI/Avernet/blob/c69b62ea60a0416733a536b296a09e7fab140872/src/backend/specs/2026-08-24-installation-single-source-of-truth/spec.md

## Related issues

- Follow-up to #1437.

### DIFF
--- a/src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
+++ b/src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
@@ -205,14 +205,21 @@ class CapabilityDesiredStateRepositoryProtocol(Protocol):
     @abstractmethod
     def install_mcp(
         self, *, bot_id: str, owner_id: str, server_code: str,
+        platform_default_codes: frozenset[str],
         engine_type: str | None = None,
         default_engine_types: tuple[str, ...] | None = None,
     ) -> DesiredStateMutation:
-        """The MCP twin of :meth:`install_skill`, same R1 rule."""
+        """The MCP twin of :meth:`install_skill`, plus platform-policy ownership.
+
+        ``platform_default_codes`` is resolved strictly from the Bot's
+        engine/template context. A matching code refuses Direct control; only
+        Default exclusion/un-exclusion may change its effective state.
+        """
         ...
     @abstractmethod
     def uninstall_mcp(
         self, *, bot_id: str, owner_id: str, server_code: str,
+        platform_default_codes: frozenset[str],
         engine_type: str | None = None,
         default_engine_types: tuple[str, ...] | None = None,
     ) -> DesiredStateMutation: ...
--- a/src/backend/src/agentclaw/community/core/skill_center/policies/capability_ownership.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/policies/capability_ownership.py
@@ -69,6 +69,19 @@ def require_can_join_set(
         )
 
 
+def require_direct_mcp_control_allowed(*, is_platform_default: bool) -> None:
+    """Refuse Direct control of an engine/template Default MCP.
+
+    Platform Default MCPs are code policy rather than Set membership, so R1's
+    membership walk cannot see them. They remain policy-managed even after a
+    Bot excludes them; exclusion/un-exclusion is their only control surface.
+    """
+    if is_platform_default:
+        raise SkillSetControlPlaneConflictError(
+            "RESOURCE_MANAGED_BY_PLATFORM_POLICY"
+        )
+
+
 def _set_belongs_to_bot(
     skill_set: Mapping[str, Any],
     *,
@@ -108,4 +121,8 @@ def _set_belongs_to_bot(
     return engine_type is None or str(skill_set.get("engine_type") or "") == engine_type
 
 
-__all__ = ["is_set_managed", "require_can_join_set"]
+__all__ = [
+    "is_set_managed",
+    "require_can_join_set",
+    "require_direct_mcp_control_allowed",
+]
--- a/src/backend/src/agentclaw/community/core/skill_center/policies/capability_ownership.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/policies/capability_ownership.py
@@ -69,14 +69,18 @@ def require_can_join_set(
         )
 
 
-def require_direct_mcp_control_allowed(*, is_platform_default: bool) -> None:
+def require_direct_mcp_control_allowed(
+    *,
+    server_code: str,
+    platform_default_codes: frozenset[str],
+) -> None:
     """Refuse Direct control of an engine/template Default MCP.
 
     Platform Default MCPs are code policy rather than Set membership, so R1's
     membership walk cannot see them. They remain policy-managed even after a
     Bot excludes them; exclusion/un-exclusion is their only control surface.
     """
-    if is_platform_default:
+    if server_code in platform_default_codes:
         raise SkillSetControlPlaneConflictError(
             "RESOURCE_MANAGED_BY_PLATFORM_POLICY"
         )
--- a/src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
@@ -394,50 +394,78 @@ class SkillSetService:
         self,
         engine_type: Optional[str],
         bot_id: Optional[str] = None,
+        *,
+        strict: bool = False,
     ) -> Mapping[str, Any] | None:
-        """Best-effort extra context lookup for engine-specific default capabilities."""
+        """Resolve extra context for engine-specific default capabilities.
+
+        Display/read callers keep their historical best-effort behavior.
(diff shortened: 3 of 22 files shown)


src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
```diff
@@ -394,50 +394,78 @@ def _get_default_capabilities_ext_info(
...
-        """Best-effort extra context lookup for engine-specific default capabilities."""
+        """Resolve extra context for engine-specific default capabilities.
+
+        Display/read callers keep their historical best-effort behavior.
+        Runtime projection passes ``strict=True`` because a dependency failure
+        must not be interpreted as removal of template-only Default MCPs.
+        """
         provider = self._ext_info_provider
         if provider is None:
+            if strict:
```
totalfrank, line 411: strict won't be None right?

src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
```diff
@@ -205,14 +205,21 @@ def uninstall_skill(
     @abstractmethod
     def install_mcp(
         self, *, bot_id: str, owner_id: str, server_code: str,
+        platform_default_codes: frozenset[str],
```
totalfrank, line 208: Prefer this as a member, not a parameter since it stays immutable(it's a built time thing.).

src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
```diff
@@ -205,14 +205,21 @@ def uninstall_skill(
...
+
+        ``platform_default_codes`` is resolved strictly from the Bot's
+        engine/template context. A matching code refuses Direct control; only
+        Default exclusion/un-exclusion may change its effective state.
+        """
         ...
     @abstractmethod
     def uninstall_mcp(
         self, *, bot_id: str, owner_id: str, server_code: str,
+        platform_default_codes: frozenset[str],
```
totalfrank, line 222: Same here.

src/backend/src/agentclaw/community/core/skill_center/policies/capability_ownership.py
```diff
@@ -69,6 +69,19 @@ def require_can_join_set(
         )
 
 
+def require_direct_mcp_control_allowed(*, is_platform_default: bool) -> None:
```
totalfrank: Prefer it takes the server code itself. Deciding if a server code is in platform default should be something maintained by this method of its own.

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
```diff
@@ -394,50 +394,78 @@ def _get_default_capabilities_ext_info(
...
+        Runtime projection uses strict mode for the same reason as ext info:
+        a repository failure is not evidence that the Bot has no template.
+        """
         target_bot_id = bot_id or self.bot_id
         if not target_bot_id or not self.entity_id:
+            if strict:
+                raise RuntimeError("Default MCP template context is unavailable")
             return None
-        try:
+        if strict:
```
totalfrank: You have

bot = self._bot_repo.get_by_id_and_owner(
                str(target_bot_id), str(self.entity_id)
            )

on both sides of this if branch, can we consolidate them into just one code piece?

totalfrank (commented):

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
FreddieSun: Correct: the production Runtime DI path always supplies the provider. `SkillSetService` still allows it to be omitted for legacy/display readers, so strict mode keeps this guard as a wiring-error fail-closed check instead of treating a missing provider as no template defaults. Added an explicit comment in c23829702.

src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
FreddieSun: I checked the lifecycle before changing this. The repository is a process-wide singleton, but this set is not process-wide build-time data: AICoding merges per-Bot `template_type` and `ext_info.template_config...preset_capabilities.mcp`. Making the resolved set a repository member could leak one Bot policy into another. I kept it as an immutable per-command `frozenset` snapshot, resolved before the DB transaction and passed into the UoW for the same-decision recheck. The policy now owns the server-code decision in c23829702.

src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
FreddieSun: Same reasoning as install: the repository is singleton while the resolved defaults are Bot/template-specific. Keeping the immutable per-command snapshot avoids cross-Bot state and avoids resolving mutable template context inside the transaction. Updated the ownership API in c23829702.

src/backend/src/agentclaw/community/core/skill_center/policies/capability_ownership.py
FreddieSun: Updated in c23829702. `PlatformDefaultMcpPolicy.require_direct_control_allowed(bot=..., server_code=...)` now resolves the exact Bot policy and owns the membership decision. The lower UoW helper likewise accepts `server_code` plus the immutable policy snapshot rather than a precomputed boolean.

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
FreddieSun: Consolidated in c23829702. There is now one repository call; strict mode re-raises lookup failures, while display mode preserves the existing warning-and-None fallback. Added coverage for both paths.

FreddieSun:
Addressed all review comments in c23829702:

- moved the Direct ownership decision into `PlatformDefaultMcpPolicy.require_direct_control_allowed(bot, server_code)`;
- consolidated the duplicated Bot repository lookup while preserving strict fail-closed versus display best-effort behavior;
- documented why a missing provider is a Runtime DI wiring error in strict mode;
- added coverage for policy-owned decisions and both strict/non-strict lookup failures.

One suggestion was intentionally handled with a Bot-scoped immutable snapshot rather than a repository member: `CapabilityDesiredStateRepository` is process-wide singleton, while AICoding Default MCPs can merge per-Bot template presets from `ext_info`. Caching the resolved codes as a member could leak policy across Bots. The same per-command `frozenset` is therefore resolved before the transaction and rechecked inside the UoW.

Validation: 1055 passed, 27 skipped; focused suite 143 passed; Ruff and changed-file blocking SAST rules passed. Please take another look.

totalfrank (approved):
LGTM.

PR #1529, opened 2026-08-26
feat(bcs): allow reading public group detail without participation

zh-bupt:
## Problem

`GET /openapi/v1/collaboration/groups/{group_id}` rejects any caller who is
not a Group Participant and does not own a Bot in the Group. Public groups,
however, are already discoverable through the public plaza
(`GET /openapi/v1/collaboration/public-groups`), so a caller can see a public
group listed but gets `403` when opening its detail.

## Solution

Skip the participant / owned-Bot check when the target group is public; any
authenticated caller can read a public group's detail. Private groups keep the
existing check unchanged. Single commit:

- `load_group_detail_for_caller` bypasses `can_read_group_detail` when
  `visibility == public` (`bcs-app-group/src/lib.rs`). This helper is used only
  by `get` (GET by id), so no other operation is affected.
- New app test `get_public_group_readable_without_participation` proves a
  non-participant with no owned Bot can read a public group.
- Contract: `get_group` 403 description notes the public exemption;
  `bcn.openapi.json` snapshot regenerated (verified the only diff vs. the prior
  snapshot is this description).

## Validation

- `cargo test -p bcs-app-group --test v1_group_service` — 78 passed (includes
  the new test; existing access-denied tests use private groups and still pass).
- `cargo test -p bcs-api-http --test group_routes` — 23 passed (handler unchanged).
- `cargo check -p bcs-app-group --all-targets` — clean.
- Gateway `test_dump_and_publish_script.py` + `test_gate_and_publish.py` +
  `test_served_openapi.py` — 19 passed.
- `pytest tests/openapi/test_group_v1_contract.py` — 16 passed; the only failure
  is the pre-existing `test_update_group_does_not_accept_context` (unrelated
  `UpdateGroup` PATCH `context` field, red on `dev` already).

## Compatibility and risk

- Access is relaxed only for public groups; private-group behavior is unchanged.
- No schema/contract-shape change beyond the 403 description text; operation
  count unchanged. Snapshot regenerated deterministically.


questchaos (approved):
LGTM

cassiuscai (approved):

PR #1530, opened 2026-08-26
fix(bcs): accept empty ask-user skip answers

vzvince:
## Problem

`interaction.resolve` rejected an explicit ask-user skip when a requested question was submitted with `values: []`. This contradicted the frontend representation of an intentionally skipped answer and caused otherwise valid mixed submissions to fail with `ask_user answer <question_id> requires values`.

The BaaS provider boundary had already become permissive, but the BCS validation was tightened again later. Without an explicit frontend-to-BCS contract, the behavior was vulnerable to another regression.

## Solution

- Accept `values: []` as the canonical explicit-skip representation for every ask-user question type.
- Keep every requested question ID and its `values` field mandatory.
- Continue rejecting missing values, non-string elements, empty strings, and whitespace-only strings.
- Define cardinality as `0..1` for single-select and free-text questions, and `0..N` for multi-select questions.
- Preserve empty arrays through the WebSocket and provider HTTP boundaries without synthesizing `customValues`, including when `allowOther` is false.
- Update the BCS protocol, boundary design, implementation plan, and related BaaS design documents so the behavior is explicit and protected from future reinterpretation.
- Add service-level and transport contract coverage for the reported mixed-answer payload and invalid-value cases.

## Validation

- `cargo test -p bcs-interaction` — 31 unit tests and 1 conformance test passed.
- `cargo test -p bcs-ws --test web_frame_compat` — 20 tests passed.
- `cargo test -p bcs-provider-http --test provider_transport_contract` — 23 tests passed.
- `cargo check -p bcs-interaction` — passed.
- `cargo clippy -p bcs-interaction --lib --no-deps -- -D warnings` — passed.
- Focused BaaS skip-preservation tests — 2 tests passed.
- Repository pre-push lint-only gate — passed.

## Compatibility and risk

This is a backward-compatible validation relaxation: the only newly accepted state is an explicit empty array for a requested question. Missing answer objects, missing `values`, invalid element types, blank strings, and excess values for single-answer questions remain invalid.

There is no wire-schema migration. The frontend-to-BCS boundary now has a single canonical skip representation, while the BaaS provider boundary retains its documented permissive compatibility behavior.

## Spec

- `src/bcs/docs/bcs-provider-2.0-sse-protocol.md`
- `src/bcs/docs/plans/2026-08-26-ask-user-empty-array-skip-design.md`
- `src/bcs/docs/plans/2026-08-26-ask-user-empty-array-skip-implementation.md`


RaymondHX (approved):
LGTM

xxxxpenny (approved):

PR #1533, opened 2026-08-26
feat(backend): make the community profile deployable on a managed MySQL

totalfrank:
## Problem

The `community` profile is the deployed profile, and that deployment runs on Aliyun against OceanBase. Three things stood in the way:

1. **No way to get a DSN in from the environment.** `YamlConfigProvider` did no env interpolation, unlike the loaders baas, gateway and proxy already ship. A k8s Secret had no route into the config.
2. **The profile did not own its schema.** `CommunityDatabase` explicitly did not create tables — "the schema is operator-provisioned". Booting `DEPLOY_PROFILE=community` against an empty database failed during *startup*:
   ```
   sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: ac_bots
   ```
   The `create_all` that makes `singlebox` work lived inside the local SQLite plugin, reachable only from `test`/`singlebox`.
3. **No MySQL-family support.** No driver was declared, and the shipped models do not survive `create_all` on MySQL as declared — 30 index keys exceed InnoDB's 3072-byte cap once utf8mb4 bills 4 bytes per character. Corp never hits this because its DDL is applied out of band.

Pods on ACK are ephemeral and replicated, so SQLite is not an option, and nobody runs DDL between `kubectl apply` and the first request.

## Solution

**Env placeholders in YAML.** `${NAME}` / `${NAME:-default}` expanded after the base⊕overlay merge, same regex and syntax as the other three services. Strict by default (matching baas): a variable that is neither set nor defaulted fails the boot rather than serving a half-configured surface. This replaces the bespoke `os.environ` read in the database DI provider — raw env access now happens only in config loading, where AGENTS.md puts it.

**The community overlay now names the real store.**

```yaml
database:
  backend: "mysql"          # OceanBase speaks the MySQL protocol in MySQL mode
  url: "${DATABASE_URL}"    # no inline default — deployment-supplied
  create_schema: true
```

`backend` selects the store (`sqlite` | `mysql` | `postgresql`) and drives engine setup; the provider rejects a `url` whose scheme disagrees with it rather than silently falling back. MySQL gets `pool_pre_ping` + `pool_recycle`, without which the first request after an idle period dies on a connection the managed instance already closed.

The URL has no default on purpose: a DSN cannot be a literal, and strict expansion makes a missing `DATABASE_URL` fail the boot loudly instead of quietly pointing production traffic at a file inside the container.

**Schema ownership.** Model registration and `create_all` move from the local SQLite plugin to `core/schema.py`; both plugins call it, so a community deployment gets the same tables a singlebox boot does. `create_schema: false` returns the plugin to a pure connection provider for operators who run their own migrations. `create_all` uses `checkfirst=True`, so re-boots issue no CREATEs for tables that already exist.

`prepare_for_mysql` caps over-long index keys with per-column prefix lengths, cutting the widest column first and leaving narrow ones whole. **It changes no column and truncates no data** — only what goes into the index:

```sql
CREATE UNIQUE INDEX uk_bot_id_entity_id_env_tenant ON ac_bots (bot_id, entity_id(512), env, avernet_tenant)
```

**Two judgement calls worth review.**

*Prefix-uniqueness.* Six of the 30 over-long keys are `UniqueConstraint`s, which SQLAlchemy cannot prefix, so they become equivalent unique `Index`es — the same object in MySQL. That narrows those keys to *prefix*-uniqueness: two ids agreeing for the first N characters would now collide. The columns hold ids and short type discriminators so it is not reachable in practice, but it is a real narrowing. The convention for *new* tables (`bot_startup_script`) is a bounded sha256 surrogate column instead; retrofitting that onto six existing tables means changing their write paths.

*The 3072 figure is InnoDB's, not OceanBase's.* `src/gateway/migrations/mysql/001_init_schema.sql` applies the same number to the same "MySQL/OceanBase" target, so this is consistent with the house assumption — but it has not been checked against a live cluster. If OceanBase's limit differs, `_INNODB_MAX_KEY_BYTES` is the single constant to retune.

**Boot path.** `main.py` no longer installs the `sofapy_base` / `layotto` / `arca` stubs — no community code imports any of them any more (only docstrings mention them), and the app boots with all three blocked at `sys.meta_path`. Listen address is now `BACKEND_HOST` / `BACKEND_PORT`, defaulting to the previous `0.0.0.0:8888`.

## Validation

All 8 CI checks green on the current head `b545d70` (and on `1581c75` before it).

Full community suite locally: **14,518 passed, 58 skipped**. The 2 failures in `test_bot_build_service_skill_artifact.py` are pre-existing and environmental (`rsync` is not installed in my sandbox) — confirmed on a stashed clean tree, and CI runs them green.

End-to-end boot, both profiles, with `sofapy_base`/`layotto`/`arca` blocked at `sys.meta_path`:

```
community: health=200 tables=83
singlebox: health=200 tables=83
```

OSS boot gate after the overlay change: `COMMUNITY BOOT GATE: PASS`. OpenAPI dump under the community profile with a placeholder DSN and no database running: `DUMP OK — paths: 191`.

MySQL DDL, verified by compiling the real metadata against the MySQL dialect (no server needed):

```
over-limit keys BEFORE: 30
over-limit keys AFTER:  0
MySQL DDL compile failures: 0 | indexes: 312
```

New tests: 10 for env expansion, 5 for the DB config surface (mismatch rejection, defaults, MySQL pooling), 10 for `core/schema.py`, 4 for the `bootstrap()` wiring (verified non-vacuous by mutating the dialect flag). `ruff check` clean on every changed file.

**Not verified:** no connection to a real MySQL or OceanBase instance was made — the sandbox has none. The DDL is proven to compile and to fit InnoDB's cap, but a first boot against a real cluster is untested, and the index-limit caveat above rides on it. Also, `uv.lock` was regenerated against pypi and normalized back to the pinned aliyun mirror URLs (the mirror is unreachable from the sandbox), so the pymysql entry's fetch from that mirror is unverified — the diff is 11 added lines, pymysql only.

## Compatibility and risk

- **No corp impact:** `corp` installs a different module column and never runs this code.
- **No singlebox/test impact:** same `SqliteDB` binding, same tables; only the `create_all` body moved.
- **The community overlay no longer loads bare.** `database.url` went from a literal to a required `${DATABASE_URL}`, so anything that reads this overlay must now supply a DSN. Two call sites use the profile without deploying it and were updated with a placeholder DSN and a comment: `scripts/community_selftest.py` (the OSS gate proving the tree boots with corp absent) and `src/gateway/scripts/dump_and_publish.sh` (OpenAPI dump). Both only import the app; config resolution opens no connection, and schema DDL runs under the lifespan, which neither enters — both verified to still pass.
- **The `community` profile was carrying two jobs** — the corp-free OSS distribution profile *and* the deployed configuration. Those want opposite defaults and this change picks the deployment, so an open-source checkout can no longer boot it without a DSN. Deliberate; if the OSS bare-boot story matters later, the clean fix is a separate overlay rather than softening this one.
- `CommunityDatabaseConfig`'s SQLite field defaults remain, now documented as the no-block fallback rather than what community deploys. They are what `DEPLOY_PROFILE=test` resolves, since `application-test.yaml` carries no `database` block.
- The golden config snapshot was regenerated; it deliberately does not expand placeholders, which is now documented in `effective_config.py`.
- Rollback is a revert; no data migration.

## Still open for the Aliyun deployment (not in this PR)

- No `backend.dockerfile` in `docker/services/` (baas, gateway, proxy have one).
- `docker/kube-deploy.sh` registers backend at port 8080 while its own header comment says 8090; `BACKEND_PORT` now makes either reachable, but the script needs picking.
- `docker/services/service-deployment.yaml` probes `/health`; the backend serves `/api/health`, so a backend pod rendered from that shared template would never pass readiness. Fixing it needs a per-service probe path, which touches the other three services.

https://claude.ai/code/session_01N2TjVaVUgAmpt8F3kZyrK6

### DIFF
--- a/src/backend/src/agentclaw/community/configs/application-community.yaml
+++ b/src/backend/src/agentclaw/community/configs/application-community.yaml
@@ -115,6 +115,11 @@ user_config:
   # Keep passwords out of this file: the URL is an env placeholder, expanded at
   # config load. A managed-MySQL deployment sets DATABASE_URL from a Secret, e.g.
   #   mysql+pymysql://user:pass@rds-host:3306/agentclaw?charset=utf8mb4
+  # ...and must ALSO set `backend: "mysql"` below. Changing only one of the two
+  # is rejected at boot with a ValueError naming the mismatch — that guard is
+  # deliberate (a silent fall back to the SQLite default would write production
+  # traffic to a file inside the container), but it is easier to read here than
+  # to discover from the error.
   #
   # `create_schema: true` lets the app emit its own DDL at boot — a container
   # deployment has nobody to run CREATE TABLE between `kubectl apply` and the
--- a/src/backend/src/agentclaw/community/configs/application-community.yaml
+++ b/src/backend/src/agentclaw/community/configs/application-community.yaml
@@ -106,11 +106,24 @@ user_config:
     operator_subjects: []
 
   # Relational store — consumed by the community CommunityDatabase.
-  # SQLAlchemy URL (SQLite file / Postgres / MySQL). The DATABASE_URL env var,
-  # if set, takes precedence (keep passwords out of this file). The schema is
-  # operator-provisioned — the app does not create tables.
+  #
+  # `backend` is the single field that switches stores: "sqlite" for a local or
+  # single-node run, "mysql" for a managed instance (Aliyun RDS and friends).
+  # `url` must name the same store — the provider rejects a mismatch rather than
+  # silently falling back to the SQLite default.
+  #
+  # Keep passwords out of this file: the URL is an env placeholder, expanded at
+  # config load. A managed-MySQL deployment sets DATABASE_URL from a Secret, e.g.
+  #   mysql+pymysql://user:pass@rds-host:3306/agentclaw?charset=utf8mb4
+  #
+  # `create_schema: true` lets the app emit its own DDL at boot — a container
+  # deployment has nobody to run CREATE TABLE between `kubectl apply` and the
+  # first request. Set it false if you provision the schema out of band or run
+  # migrations yourself.
   database:
-    url: "sqlite:///./data/agentclaw.db"
+    backend: "sqlite"
+    url: "${DATABASE_URL:-sqlite:///./data/agentclaw.db}"
+    create_schema: true
 
   # Cache + distributed lock — consumed by the community CommunityCache.
   # Set redis_url to a Redis endpoint for a multi-worker deploy (KV + SET NX
--- a/src/backend/src/agentclaw/community/configs/application-community.yaml
+++ b/src/backend/src/agentclaw/community/configs/application-community.yaml
@@ -107,27 +107,31 @@ user_config:
 
   # Relational store — consumed by the community CommunityDatabase.
   #
-  # `backend` is the single field that switches stores: "sqlite" for a local or
-  # single-node run, "mysql" for a managed instance (Aliyun RDS and friends).
-  # `url` must name the same store — the provider rejects a mismatch rather than
-  # silently falling back to the SQLite default.
+  # The community profile IS the deployed configuration, and that deployment runs
+  # on Aliyun against OceanBase — so this block names a real managed store rather
+  # than a local convenience default.
   #
-  # Keep passwords out of this file: the URL is an env placeholder, expanded at
-  # config load. A managed-MySQL deployment sets DATABASE_URL from a Secret, e.g.
-  #   mysql+pymysql://user:pass@rds-host:3306/agentclaw?charset=utf8mb4
-  # ...and must ALSO set `backend: "mysql"` below. Changing only one of the two
-  # is rejected at boot with a ValueError naming the mismatch — that guard is
-  # deliberate (a silent fall back to the SQLite default would write production
-  # traffic to a file inside the container), but it is easier to read here than
-  # to discover from the error.
+  # `backend` selects the store and drives engine setup (pooling, pragmas):
+  # "mysql" covers OceanBase, which speaks the MySQL wire protocol in MySQL mode
+  # — the same target `src/gateway/migrations/mysql/` is written against. "sqlite"
+  # and "postgresql" are also accepted. `url` must name the same store; the
+  # provider rejects a mismatch at boot rather than falling back silently.
+  #
+  # `url` has NO inline default on purpose. DATABASE_URL is deployment-supplied
+  # (a k8s Secret), and strict expansion makes a missing one fail the boot
+  # loudly instead of quietly writing production traffic somewhere else:
+  #   DATABASE_URL=mysql+pymysql://user:pass@ob-host:3306/agentclaw?charset=utf8mb4
+  # Anything that merely loads this overlay without deploying — the OSS boot
+  # self-test, the OpenAPI dump — must therefore supply a DSN too; no connection
+  # is opened by config resolution alone.
   #
   # `create_schema: true` lets the app emit its own DDL at boot — a container
   # deployment has nobody to run CREATE TABLE between `kubectl apply` and the
   # first request. Set it false if you provision the schema out of band or run
   # migrations yourself.
   database:
-    backend: "sqlite"
-    url: "${DATABASE_URL:-sqlite:///./data/agentclaw.db}"
+    backend: "mysql"
+    url: "${DATABASE_URL}"
     create_schema: true
 
   # Cache + distributed lock — consumed by the community CommunityCache.
--- a/src/backend/src/agentclaw/community/core/schema.py
+++ b/src/backend/src/agentclaw/community/core/schema.py
@@ -1,0 +1,188 @@
+"""Dialect-neutral relational schema bootstrap.
+
+One place that knows how to turn an empty database into the backend's schema,
+shared by every ``DatabasePlugin`` that owns its own store:
+
+- :func:`import_all_models` — force every ORM class to be registered on the
+  metadata, so ``create_all`` is complete rather than dependent on whichever
+  modules the router chain happened to import first.
+- :func:`create_all` — emit the DDL for both declarative bases.
+- :func:`prepare_for_mysql` — MySQL-only DDL adjustments (index key lengths).
+
+This body used to live inline in ``plugins/local/database.py``'s ``SqliteDB``,
+where only the ``test`` and ``singlebox`` profiles could reach it. The community
+plugin needs the same bootstrap against a real store (a deployment does not run
+DDL by hand before the pods start), so it lives in ``core/`` and both plugins
+call it. ``core/`` is the lowest layer; ``plugins/`` depends on it, never the
+reverse.
+"""
+from __future__ import annotations
+
+import logging
+
+from sqlalchemy import Index, MetaData, String, UniqueConstraint
+from sqlalchemy.engine import Engine
+
+logger = logging.getLogger(__name__)
+
+
+def import_all_models() -> None:
+    """Register every ORM class on its declarative metadata.
+
+    Eagerly imports each model module so its ``class`` statement runs and
+    SQLAlchemy's declarative metaclass attaches a ``Table`` to the shared
+    ``MetaData``. Without this, ``create_all`` would only emit DDL for tables
+    whose class was already transitively imported via the router chain — and the
+    method-level lazy imports in many repositories make that set
+    non-deterministic. The first request would hit ``no such table: ac_xxx``.
+
+    ``noqa: F401`` throughout: the names are intentionally unused, only the
+    import side effect matters.
+    """
+    import agentclaw.community.plugin_api.models  # noqa: F401  ac_bots / ac_resource / ac_channel_config
+    import agentclaw.community.core.models  # noqa: F401  ac_skill* / ac_skill_set_mcp / ac_user_mcp_config / propagation_log / center_sync_log
+    import agentclaw.community.core.skill_center.orm  # noqa: F401  ac_default_skillset_*
+    import agentclaw.community.core.access.sqlite_models  # noqa: F401  ac_access_control_policy / ac_user_info
+    import agentclaw.community.core.service_bot.repository.models  # noqa: F401  ac_bot_publish
+    import agentclaw.community.core.bot_public.repository.models  # noqa: F401  ac_bot_friend
+    import agentclaw.community.core.expert_chat.sqlite_models  # noqa: F401  ac_expert_chat_bot_sessions
+    import agentclaw.community.core.devices.repository.models  # noqa: F401  ac_entity_device_binding
+    import agentclaw.community.core.bot_management.repository.models  # noqa: F401  ac_templates / ac_bot_restart_lock
+    import agentclaw.community.core.bot_startup_script.repository.models  # noqa: F401  ac_bot_startup_script
+    import agentclaw.community.core.bot_management.render_screen.sqlite_models  # noqa: F401  ac_bot_render_screen
+    import agentclaw.community.core.system_config.orm  # noqa: F401  ac_config_*
+    import agentclaw.community.core.harness.sqlite_models  # noqa: F401  ac_harness_*
+    import agentclaw.community.core.bot_chat.models  # noqa: F401  bot_chat private-Base tables
+    import agentclaw.community.core.bot_dormant.sqlite_models  # noqa: F401  ac_bot_dormant_*
+    import agentclaw.community.core.task_queue.repository.models  # noqa: F401  ac_task_queue
+    import agentclaw.community.core.task.repository.models  # noqa: F401  task_info / task_node / task_node_run_info / task_node_relation / task_callback
+    import agentclaw.community.core.skills_pool.repository.models  # noqa: F401  ac_bot_skill_layout_state
+    import agentclaw.community.core.session_resources.repository.models  # noqa: F401  ac_session_resource
+    import agentclaw.community.core.economy.governance.orm  # noqa: F401  governance_*
+    import agentclaw.community.core.caller_identity.models  # noqa: F401  caller identity tables
+    import agentclaw.community.core.bot_app_grant.models  # noqa: F401  ac_bot_app_grant / ac_bot_app_grant_log
+    import agentclaw.community.core.user_list.models  # noqa: F401  ac_entity_user_list
+    import agentclaw.community.core.spaces.repository.models  # noqa: F401  ac_space / ac_space_member
+    import agentclaw.community.core.market_favorites.repository.models  # noqa: F401  ac_market_favorite
+    import agentclaw.community.core.work_orders.repository.models  # noqa: F401  ac_work_order / ac_work_order_notification
+
+
+def _metadatas() -> list[MetaData]:
+    """Both declarative registries the schema spans.
+
+    ``core/bot_chat/models.py`` declares a private ``Base`` instead of the
+    canonical ``core/base.py`` one, so its tables (``aw_langfuse_traces``, and a
+    second ``ac_bots``) live on a separate ``MetaData``. SQLAlchemy permits the
+    same table name across different registries and ``create_all`` is idempotent
+    (``checkfirst=True``), so emitting both is safe in either order. Without the
+    private one, ``/api/v1/bot-chats`` crashes with ``no such table:
+    aw_langfuse_traces``.
+    """
+    import agentclaw.community.core.bot_chat.models as bot_chat_models
+    from agentclaw.community.core.base import Base
+
+    return [Base.metadata, bot_chat_models.Base.metadata]
+
+
+# InnoDB caps an index key at 3072 bytes (MySQL 8, DYNAMIC row format), and
+# utf8mb4 bills 4 bytes per character — so a single VARCHAR(1024) column blows
+# the budget on its own.
+_INNODB_MAX_KEY_BYTES = 3072
+_UTF8MB4_BYTES_PER_CHAR = 4
+
+
+def _effective_widths(columns, existing: dict[str, int] | None) -> dict[str, int]:
+    """Each string column's indexed width, honouring any prefix already applied."""
+    existing = existing or {}
+    return {
+        col.name: existing.get(col.name, col.type.length)
+        for col in columns
+        if isinstance(col.type, String) and col.type.length
+    }
+
+
+def _prefix_lengths(
+    columns, existing: dict[str, int] | None = None
+) -> dict[str, int] | None:
+    """Shrink the widest string columns until the key fits InnoDB's cap.
+
+    Returns a ``{column_name: prefix_chars}`` map for ``mysql_length``, or
+    ``None`` if the key already fits. Narrow columns are left whole and the
+    widest are cut first, so a key made of one wide column and several small
+    ones only loses precision on the wide one.
+
+    ``existing`` is the prefix map already on the index, so a second pass over
+    the same metadata is a no-op rather than halving the prefix again.
+    """
+    widths = _effective_widths(columns, existing)
+    budget_chars = _INNODB_MAX_KEY_BYTES // _UTF8MB4_BYTES_PER_CHAR
+    if sum(widths.values()) <= budget_chars:
+        return None
+
+    prefixes = dict(widths)
+    # Repeatedly halve the widest column until the total fits. Converges quickly
+    # and keeps the result deterministic (no float arithmetic, no ordering
+    # dependence beyond "widest first, ties by name").
+    while sum(prefixes.values()) > budget_chars:
+        widest = max(sorted(prefixes), key=lambda name: prefixes[name])
+        prefixes[widest] = max(1, prefixes[widest] // 2)
+    return {name: length for name, length in prefixes.items() if length < widths[name]}
+
+
+def prepare_for_mysql(metadata: MetaData) -> list[str]:
+    """Make ``metadata``'s DDL emittable on MySQL. Returns what was adjusted.
+
+    Every index whose key exceeds InnoDB's 3072-byte cap gets per-column prefix
+    lengths. A ``UniqueConstraint`` cannot carry prefix lengths in SQLAlchemy, so
+    an over-long one is replaced by an equivalent unique ``Index``, which can —
+    the two are the same object in MySQL.
+
+    **Semantic caveat for the unique keys.** A prefixed unique index enforces
+    uniqueness on the prefix, which is *stricter* than on the whole value: two
+    rows whose ids agree for the first N characters and differ later would now
+    collide. The columns involved hold ids and short type discriminators, so this
+    is not reachable in practice — but it is a real narrowing, and the reason the
+    convention for *new* tables (see ``bot_startup_script``) is a bounded sha256
+    surrogate column instead. Retrofitting a surrogate onto these six tables
+    means changing their write paths, which is a bigger change than making the
+    schema deployable.
+
+    MySQL-only, and applied in-process at bootstrap: SQLite ignores prefix
+    lengths, and corp never runs this code (its DDL is applied out of band).
+    """
+    adjusted: list[str] = []
+
+    for table in metadata.sorted_tables:
+        for constraint in list(table.constraints):
+            if not isinstance(constraint, UniqueConstraint):
+                continue
+            prefixes = _prefix_lengths(list(constraint.columns))
+            if prefixes is None:
+                continue
+            columns = list(constraint.columns)
+            name = constraint.name or f"uk_{table.name}_{'_'.join(c.name for c in columns)}"
+            table.constraints.discard(constraint)
+            Index(name, *columns, unique=True, mysql_length=prefixes)
+            adjusted.append(f"{table.name}.{name} (unique constraint → prefixed unique index)")
+
+        for index in table.indexes:
+            already = index.dialect_options["mysql"].get("length") or {}
+            prefixes = _prefix_lengths(list(index.columns), already)
+            if prefixes is None:
+                continue
+            index.dialect_options["mysql"]["length"] = {**already, **prefixes}
+            adjusted.append(f"{table.name}.{index.name} (prefixed)")
+
+    return adjusted
+
+
+def create_all(engine: Engine, *, mysql: bool = False) -> None:
+    """Create every table on ``engine``. Idempotent (``checkfirst=True``)."""
+    import_all_models()
+    for metadata in _metadatas():
+        if mysql:
+            adjusted = prepare_for_mysql(metadata)
+            for entry in adjusted:
+                logger.info("schema: MySQL index key capped — %s", entry)
+        metadata.create_all(engine)
+    logger.info("schema: create_all complete")
--- a/src/backend/src/agentclaw/community/di/config_community.py
+++ b/src/backend/src/agentclaw/community/di/config_community.py
@@ -39,13 +39,20 @@ class CommunityDatabaseConfig:
     Corp/test never resolve this type.
 
     ``backend`` is the one field a deployment flips to change stores — ``sqlite``
-    for a local or single-node run, ``mysql`` for a managed instance. It selects
-    engine setup (pooling, pragmas), and the provider rejects a ``url`` whose
-    scheme disagrees with it, so the two can never drift into a confusing
-    half-configured state.
+    for a local or single-node run, ``mysql`` for a managed instance (including
+    OceanBase, which speaks the MySQL protocol). It selects engine setup
+    (pooling, pragmas), and the provider rejects a ``url`` whose scheme disagrees
+    with it, so the two can never drift into a confusing half-configured state.
 
-    ``url`` is a SQLAlchemy URL. Point it at an env var from the YAML
-    (``${DATABASE_URL:-...}``) rather than writing a password into the file.
+    ``url`` is a SQLAlchemy URL, sourced from the YAML's env placeholder rather
+    than written into the file, so no password is ever committed.
+
+    **These field defaults are the no-block fallback, not what community
+    deploys.** The shipped ``application-community.yaml`` sets ``mysql`` with a
+    required ``${DATABASE_URL}``, because the community profile is the deployed
+    Aliyun/OceanBase configuration. The SQLite values below apply only if the
+    ``database`` block is absent entirely — a DSN cannot be a literal default,
+    since host and credentials are per-deployment.
 
     ``create_schema`` lets the app emit its own DDL at boot, which is what a
     container deployment needs — nobody runs ``CREATE TABLE`` between
--- a/src/backend/src/agentclaw/community/di/config_community.py
+++ b/src/backend/src/agentclaw/community/di/config_community.py
@@ -35,14 +35,37 @@ class BcsAuthConfig:
 class CommunityDatabaseConfig:
     """Relational-store connection for the community ``CommunityDatabase``.
 
-    Community-only: sourced from the ``database`` block of ``user_config``
-    (overridable by the ``DATABASE_URL`` env var in the provider). ``url`` is a
-    SQLAlchemy URL — SQLite file / Postgres / MySQL. The default is a local
-    SQLite file so a bare community boot works; a real deploy points it at a
-    pre-provisioned store. Corp/test never resolve this type.
+    Community-only: sourced from the ``database`` block of ``user_config``.
+    Corp/test never resolve this type.
+
+    ``backend`` is the one field a deployment flips to change stores — ``sqlite``
+    for a local or single-node run, ``mysql`` for a managed instance. It selects
+    engine setup (pooling, pragmas), and the provider rejects a ``url`` whose
+    scheme disagrees with it, so the two can never drift into a confusing
+    half-configured state.
+
+    ``url`` is a SQLAlchemy URL. Point it at an env var from the YAML
+    (``${DATABASE_URL:-...}``) rather than writing a password into the file.
+
+    ``create_schema`` lets the app emit its own DDL at boot, which is what a
+    container deployment needs — nobody runs ``CREATE TABLE`` between
+    ``kubectl apply`` and the first request. Set it false when the schema is
+    provisioned out of band or managed by migrations.
     """
 
+    backend: str = "sqlite"
     url: str = "sqlite:///./data/agentclaw.db"
+    create_schema: bool = True
+
+
+#: Accepted ``database.backend`` values mapped to the SQLAlchemy URL scheme each
+#: one requires. Adding a backend means adding its engine setup to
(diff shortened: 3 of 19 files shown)


src/backend/src/agentclaw/community/configs/application-community.yaml
```diff
@@ -106,11 +106,24 @@ user_config:
...
+  #
+  # `create_schema: true` lets the app emit its own DDL at boot — a container
+  # deployment has nobody to run CREATE TABLE between `kubectl apply` and the
+  # first request. Set it false if you provision the schema out of band or run
+  # migrations yourself.
   database:
-    url: "sqlite:///./data/agentclaw.db"
+    backend: "sqlite"
+    url: "${DATABASE_URL:-sqlite:///./data/agentclaw.db}"
+    create_schema: true
```
totalfrank-personal, line 135: Is this supposed to be for aliyun deployment? If so, I don't think it's sqlite. It's supposed to be ocean base.
Then, can we have something like create if not exist? Also, where do we need to keep the schema DDL? Or it's based on the ORM models?

src/backend/src/agentclaw/community/core/schema.py
```diff
@@ -0,0 +1,188 @@
...
+    import agentclaw.community.core.task.repository.models  # noqa: F401  task_info / task_node / task_node_run_info / task_node_relation / task_callback
+    import agentclaw.community.core.skills_pool.repository.models  # noqa: F401  ac_bot_skill_layout_state
+    import agentclaw.community.core.session_resources.repository.models  # noqa: F401  ac_session_resource
+    import agentclaw.community.core.economy.governance.orm  # noqa: F401  governance_*
+    import agentclaw.community.core.caller_identity.models  # noqa: F401  caller identity tables
+    import agentclaw.community.core.bot_app_grant.models  # noqa: F401  ac_bot_app_grant / ac_bot_app_grant_log
+    import agentclaw.community.core.user_list.models  # noqa: F401  ac_entity_user_list
+    import agentclaw.community.core.spaces.repository.models  # noqa: F401  ac_space / ac_space_member
+    import agentclaw.community.core.market_favorites.repository.models  # noqa: F401  ac_market_favorite
+    import agentclaw.community.core.work_orders.repository.models  # noqa: F401  ac_work_order / ac_work_order_notification
```
totalfrank-personal, line 67: I assume you need this for DDL initiation but that's only true if we go with sqlite right? With aliyun deployment, I think the DB is ocean base.

src/backend/src/agentclaw/community/di/config_community.py
```diff
@@ -35,14 +35,37 @@ class BcsAuthConfig:
...
+
+    ``create_schema`` lets the app emit its own DDL at boot, which is what a
+    container deployment needs — nobody runs ``CREATE TABLE`` between
+    ``kubectl apply`` and the first request. Set it false when the schema is
+    provisioned out of band or managed by migrations.
     """
 
+    backend: str = "sqlite"
     url: str = "sqlite:///./data/agentclaw.db"
+    create_schema: bool = True
```
totalfrank-personal, line 65: For community, the DB is ocean base. So for default value, should we also go with that?

totalfrank-personal (commented):

src/backend/src/agentclaw/community/configs/application-community.yaml
totalfrank: Four things here, taking them in order.

**OceanBase vs "mysql".** `backend: mysql` is how you reach OceanBase — the value names the SQLAlchemy dialect, not the product. OceanBase in MySQL mode speaks the MySQL wire protocol, and this repo already treats them as one target: `src/gateway/migrations/mysql/001_init_schema.sql` is titled "Gateway MySQL/OceanBase canonical schema", and bcs uses the same `migrations/mysql/` convention. So an Aliyun deployment sets `backend: mysql` with a `mysql+pymysql://…` DSN pointing at OceanBase.

That said, your reading is the one a newcomer will have too. Happy to rename the value to something explicit (`mysql_compatible`?) or just document it in the block comment — say which you prefer.

**The sqlite value here is not the Aliyun setting.** It is the default for a bare boot with nothing configured. baas ships the same shape — `database_url: "sqlite:////tmp/secbaas.db"` with `create_schema: true` in both `configs/` and `singlebox-configs/` — and the real deployment overrides it. A default cannot carry a DSN anyway, since host and credentials are per-deployment.

**"create if not exist" — already the behaviour.** `create_all()` runs with `checkfirst=True`, which introspects the database and issues zero CREATE statements for tables that already exist. Verified:

```
2nd create_all issued CREATEs: []
checkfirst introspects instead: True
```

If you would rather have the literal form, SQLAlchemy can emit `CREATE TABLE IF NOT EXISTS` instead — one flag. The introspect-and-skip behaviour is equivalent in effect and is what the local SQLite plugin has always used.

**Where the DDL lives — this one is a real gap.** Right now it is ORM-derived only. The house convention elsewhere is a checked-in canonical SQL file for the real deployment plus `create_all` for local/test, stated outright in the gateway file:

> The gateway's bare/community edition creates these tables from ORM metadata via `DataSourcePlugin.create_all()`. This DDL is the canonical schema for the real MySQL/OceanBase deployment.

baas does the same with `src/baas/sqls/baas_schema.sql`. The backend has no equivalent. I would propose adding `src/backend/migrations/mysql/001_init_schema.sql`, generated from the same prepared metadata this PR already produces (so the two cannot drift on day one) and reviewed as SQL. That is a sizeable file — 83 tables — so I did not add it unasked. Want it in this PR or a follow-up?

---

src/backend/src/agentclaw/community/core/schema.py
totalfrank: This module is dialect-neutral — it is not a SQLite-only path.

`import_all_models()` just forces every ORM class to register on the metadata (without it `create_all` emits DDL only for whatever the router chain happened to import first, which is non-deterministic). `create_all` then compiles that metadata against whichever dialect the engine carries: SQLite locally, MySQL/OceanBase for a real deployment. Same registration, same tables, different emitted SQL.

The `prepare_for_mysql` step directly below exists *for* the OceanBase/MySQL case and is a no-op on SQLite. It is needed because 30 of the shipped index keys exceed InnoDB's 3072-byte cap at utf8mb4.

Worth noting the gateway hit exactly this and landed on the same answer — from `src/gateway/migrations/mysql/001_init_schema.sql`:

> The unique indexes on the two `token` columns span a 700-character prefix, not the whole varchar(1024): at utf8mb4 the full width is a 4096-byte key, past InnoDB's 3072-byte limit, which made this file non-executable…

So prefix indexes on over-wide keys are already the established convention here for the MySQL/OceanBase target, which is what `prepare_for_mysql` automates.

One caveat I want on the record: the 3072-byte figure is InnoDB's, and this repo applies it to the "MySQL/OceanBase" target as one. If OceanBase's own index-key limit differs from InnoDB's, the prefix arithmetic should be retuned to that number — I have not verified it against a live OceanBase instance, only against the MySQL dialect compiler. If you have a cluster I can point a boot at, that is the check worth running before this ships.

---

src/backend/src/agentclaw/community/di/config_community.py
totalfrank: The blocker is that a default cannot carry a DSN. `backend: "mysql"` with no reachable `url` would fail the boot for anyone who checks the repo out and runs it, and the host, port, database and credentials of an OceanBase cluster are all deployment-specific — there is no sensible literal to put here.

That is why baas ships `sqlite:////tmp/secbaas.db` as its default too, in both `configs/` and `singlebox-configs/`, and lets the real deployment override.

So the dataclass default is really "what an unconfigured checkout does", not "what community means". If the intent is that community *is* OceanBase, I would rather express that where it is actually true than by changing this fallback. Three options, in my order of preference:

1. **Leave the fallback, make the deployment explicit.** Ship the Aliyun values in a deployment overlay / the k8s env so `backend: mysql` + `DATABASE_URL` are set there. The fallback stays a bare-boot convenience and never runs in production.
2. **Split the profile.** If `community` genuinely means "the managed deployment", add a distinct overlay for it that sets `backend: mysql` and requires `DATABASE_URL` with no default — a missing Secret then fails the boot loudly, which is the behaviour strict expansion was chosen for.
3. **Flip the default to mysql.** Honest about intent, but every bare checkout and any test resolving this config breaks unless it also sets a DSN.

I lean 1, or 2 if you want "community == OceanBase" to be structurally true rather than a convention. Which way do you want it?

---

src/backend/src/agentclaw/community/plugins/community/database.py
```diff
@@ -54,9 +68,36 @@ def _set_sqlite_pragma(dbapi_conn, _conn_record):
...
+            return create_engine(
+                url,
+                pool_pre_ping=True,
+                pool_recycle=3600,
+            )
+
+        # Postgres / etc.: default pooling is appropriate.
         return create_engine(url)
 
+    async def bootstrap(self) -> None:
```
totalfrank, line 88: Non-blocking: `bootstrap()` — the branch that decides whether to actually run schema DDL (`create_schema` true/false, and the `mysql=_is_mysql(self._url)` dialect switch) has no direct unit test.

`tests/community/plugins/community/test_database.py` still opens with "CommunityDatabase is a pure connection provider — it does NOT create tables" and its `db` fixture builds `CommunityDatabase(url)` then manually calls `Base.metadata.create_all(database._engine)` on a throwaway test `Base` — it never calls `.bootstrap()`, so neither the `create_schema=True` path (does it actually call `core.schema.create_all`?) nor the `create_schema=False` skip path is exercised for this plugin. (`core/schema.py`'s own logic is covered by `test_schema.py`, but the wiring here — whether `bootstrap()` calls it at all, and with the right `mysql=` flag — isn't.)

Suggest either updating that fixture/docstring to reflect the new default (schema is created via `bootstrap()` now, not operator-provisioned by default) and adding a test that calls `await database.bootstrap()` for both flag values, asserting `create_all`/`prepare_for_mysql` get invoked (or not) as expected.

---

src/backend/src/agentclaw/community/configs/application-community.yaml
```diff
@@ -106,11 +106,24 @@ user_config:
...
-  # operator-provisioned — the app does not create tables.
+  #
+  # `backend` is the single field that switches stores: "sqlite" for a local or
+  # single-node run, "mysql" for a managed instance (Aliyun RDS and friends).
+  # `url` must name the same store — the provider rejects a mismatch rather than
+  # silently falling back to the SQLite default.
+  #
+  # Keep passwords out of this file: the URL is an env placeholder, expanded at
+  # config load. A managed-MySQL deployment sets DATABASE_URL from a Secret, e.g.
+  #   mysql+pymysql://user:pass@rds-host:3306/agentclaw?charset=utf8mb4
```
totalfrank: nit: this comment tells an operator to point a MySQL DSN at `DATABASE_URL` but doesn't mention that `backend: "sqlite"` on line 124 also has to change to `"mysql"` — the DI provider (`di/modules/infrastructure/community/database.py`) rejects the resulting scheme mismatch with a `ValueError` at boot. Worth a one-line note here so a first-time Aliyun deploy doesn't have to discover that from the error message.

---

totalfrank (commented):
Reviewed commit fd88546 (the PR's only commit) against merge-base 71a73ae with `dev`. Files covered: `pyproject.toml`/`uv.lock` (pymysql dep), `core/config/yaml_provider.py` (env-placeholder expansion), `core/schema.py` (new — model registration, MySQL index-prefix DDL adjustment), `di/config_community.py`, `di/modules/infrastructure/community/database.py`, `plugins/community/database.py`, `plugins/local/database.py` (schema bootstrap relocated to `core/schema.py`), `main.py` (stub removal, env-driven listen address), `configs/application-community.yaml`, and the accompanying test/golden-config changes.

Traced correctness on the parts most likely to break in production: the `backend`/`url` scheme-mismatch validation and its interaction with `discover_lifecycle_participants`'s exception-swallowing (confirmed `eager_check_critical_bindings` catches a bad config at boot for `pre`/`prod` via `SERVER_ENV`), the `prepare_for_mysql` prefix-length algorithm (idempotency, tie-breaking, the widest-first halving), and that no remaining code still imports the now-unstubbed `sofapy_base`/`layotto`/`arca`. Categories that applied: correctness, contract breakage (DI wiring / lifecycle discovery), architecture boundaries (raw env access confined to config loading), test coverage. Two findings posted inline, both non-blocking/nit — no correctness or contract-breaking issues found.

---

src/backend/src/agentclaw/community/configs/application-community.yaml
totalfrank: Resolved by the author: no DDL file. `create_all` from the ORM models is the schema source for this deployment, so the backend deliberately does not follow the gateway/bcs/baas `migrations/mysql/` convention here.

`backend: mysql` stays as the value name for the OceanBase target.

The "create if not exist" behaviour is already in place via `checkfirst=True`, unchanged.

Pushed in 1581c75 for the remaining point in this thread: the block comment now says that a managed-MySQL deployment must change `backend` **and** `url` together, rather than leaving that to be discovered from the boot-time `ValueError`.

---

src/backend/src/agentclaw/community/di/config_community.py
totalfrank: Author picked option 1: the SQLite fallback stays as the bare-boot default, and the Aliyun deployment sets `backend: mysql` + `DATABASE_URL` explicitly in its own environment. No code change here.

---

src/backend/src/agentclaw/community/core/schema.py
totalfrank: Settled: `create_all` from the ORM models is the schema mechanism for the OceanBase deployment too, so this module stays as the single dialect-neutral bootstrap. No separate DDL file.

The OceanBase index-key caveat from my earlier reply still stands and is not closed by this thread: the 3072-byte figure is InnoDB's, matching what the gateway's `migrations/mysql` file assumes for the same "MySQL/OceanBase" target, but it has not been checked against a live OceanBase cluster. If its limit differs, `_INNODB_MAX_KEY_BYTES` is the one constant to retune.

---

src/backend/src/agentclaw/community/plugins/community/database.py
totalfrank: Good catch — the docstring was describing pre-PR behaviour, and the wiring genuinely was untested.

Fixed in 1581c75. `TestBootstrapSchemaOwnership` adds four tests: schema created by default, skipped under `create_schema=False`, and the `mysql=` flag matching the URL in both directions. `core.schema.create_all` is patched so no real DDL is emitted and the MySQL case needs no reachable server.

The fixture keeps provisioning its own private `Base` rather than calling `bootstrap()` — those tests are about session and transaction behaviour, and `bootstrap()` would emit the full application schema they do not use. The docstring now says that explicitly instead of claiming the plugin never creates tables.

Checked the new tests aren't vacuous by pinning the flag to `False`:

```
FAILED TestBootstrapSchemaOwnership::test_mysql_url_requests_the_mysql_ddl_pass
1 failed, 10 passed
```

---

src/backend/src/agentclaw/community/configs/application-community.yaml
totalfrank: Agreed — fixed in 1581c75. The block comment now spells out that a managed-MySQL deployment changes `backend` and `url` together, and why the mismatch is rejected rather than silently defaulted:

```yaml
  # ...and must ALSO set `backend: "mysql"` below. Changing only one of the two
  # is rejected at boot with a ValueError naming the mismatch — that guard is
  # deliberate (a silent fall back to the SQLite default would write production
  # traffic to a file inside the container), but it is easier to read here than
  # to discover from the error.
```

---

src/backend/src/agentclaw/community/plugins/community/database.py
totalfrank: Fixed in 1581c75: `bootstrap()` now has direct coverage — `test_creates_the_schema_by_default`, `test_skips_ddl_when_the_operator_owns_the_schema`, `test_sqlite_url_does_not_request_the_mysql_ddl_pass`, and `test_mysql_url_requests_the_mysql_ddl_pass` — asserting `create_all(engine, mysql=...)` is called (or not) matching `create_schema` and the URL's dialect. Confirmed the mocked call matches the real wiring in `plugins/community/database.py:88-98` (`create_all(self._engine, mysql=_is_mysql(self._url))` gated on `self._create_schema`), and the docstring/fixture comment now correctly describe `bootstrap()` as schema-owning by default.

---

src/backend/src/agentclaw/community/configs/application-community.yaml
totalfrank: Fixed in 1581c75: the block comment above `database.url` now spells out that a managed-MySQL deployment must also set `backend: "mysql"`, and names the boot-time `ValueError` that guards the mismatch.

---

totalfrank (commented):
Follow-up round. Commit reviewed this round: `1581c755` (test coverage for `CommunityDatabase.bootstrap()` wiring + a config-comment clarification). This commit was already the head at the time of the prior round's last replies, so there was no further diff beyond it — this round closes out the two findings that commit fixed but hadn't been confirmed/resolved yet.

- `application-community.yaml:131` — OceanBase-vs-`mysql` naming / DDL-location / create-if-not-exists question (reviewer Q&A, not a defect) — answered, author's call taken (no DDL file, `backend: "mysql"` stays the value name). Left open on GitHub for the human reviewer to close.
- `core/schema.py:67` — dialect-neutral bootstrap question + OceanBase index-limit caveat (reviewer Q&A) — answered; the InnoDB-vs-OceanBase caveat is flagged as an open item for verification against a live cluster, not a defect in this PR. Left open on GitHub for the human reviewer to close.
- `di/config_community.py:58` — default `backend` value question (reviewer Q&A) — answered, author picked option 1 (SQLite fallback stays, Aliyun sets `backend`+`DATABASE_URL` explicitly). Left open on GitHub for the human reviewer to close.
- `plugins/community/database.py:88` — non-blocking: `bootstrap()` had no direct test — **fixed** in `1581c75` (4 new tests, verified non-vacuous against the real wiring). Resolved.
- `application-community.yaml:117` — nit: config comment didn't mention `backend` must change alongside `DATABASE_URL` — **fixed** in `1581c75`. Resolved.

**New commit (`1581c75`) review:** only touches the two files above (test additions + a comment). Tests correctly mock `core.schema.create_all` and assert it's called with `mysql=` matching `_is_mysql(self._url)` and gated on `create_schema`, matching the real implementation in `plugins/community/database.py:88-98`. No new issues found.

**Open count:** 0 code-review defects outstanding. Three reviewer Q&A threads remain open pending the human reviewer's own close.

---

src/backend/src/agentclaw/community/core/schema.py
```diff
@@ -0,0 +1,188 @@
...
+    """Each string column's indexed width, honouring any prefix already applied."""
+    existing = existing or {}
+    return {
+        col.name: existing.get(col.name, col.type.length)
+        for col in columns
+        if isinstance(col.type, String) and col.type.length
+    }
+
+
+def _prefix_lengths(
```
totalfrank-personal, line 104: Are these about cutting a string column content if it exists the defined width? If so, let's get rid of that? This is too heavy.

src/backend/src/agentclaw/community/core/schema.py
```diff
@@ -0,0 +1,188 @@
...
+    # Repeatedly halve the widest column until the total fits. Converges quickly
+    # and keeps the result deterministic (no float arithmetic, no ordering
+    # dependence beyond "widest first, ties by name").
+    while sum(prefixes.values()) > budget_chars:
+        widest = max(sorted(prefixes), key=lambda name: prefixes[name])
+        prefixes[widest] = max(1, prefixes[widest] // 2)
+    return {name: length for name, length in prefixes.items() if length < widths[name]}
+
+
+def prepare_for_mysql(metadata: MetaData) -> list[str]:
```
totalfrank-personal, line 132: Same, if this is for some safe guard of some kind of length limitation, let's get rid of it. Do not worry about it here. It makes the code too complicated.

src/backend/src/agentclaw/community/configs/application-community.yaml
```diff
@@ -106,11 +106,29 @@ user_config:
...
+  # traffic to a file inside the container), but it is easier to read here than
+  # to discover from the error.
+  #
+  # `create_schema: true` lets the app emit its own DDL at boot — a container
+  # deployment has nobody to run CREATE TABLE between `kubectl apply` and the
+  # first request. Set it false if you provision the schema out of band or run
+  # migrations yourself.
   database:
-    url: "sqlite:///./data/agentclaw.db"
+    backend: "sqlite"
```
totalfrank-personal: I don't understand, this is the profile for aliyun deployment right? If so, is this field for the engine? why sqlite here?

totalfrank-personal (commented):
Another round of comments/confusions, please take a look.

src/backend/src/agentclaw/community/core/schema.py
totalfrank: Important correction first: **this never cuts column content.** No data is truncated, ever.

It only changes the `CREATE INDEX` statement. A column declared `VARCHAR(1024)` still stores the full 1024 characters; reads and writes are untouched. What changes is how much of the value MySQL puts in the *index*:

```sql
--  what we'd emit without this
CREATE UNIQUE INDEX uk_... ON ac_bots (bot_id, entity_id, env, avernet_tenant)
--  what we emit with it
CREATE UNIQUE INDEX uk_... ON ac_bots (bot_id, entity_id(512), env, avernet_tenant)
```

`entity_id(512)` means "index the first 512 characters of this column". The stored value is unaffected.

The reason it exists: InnoDB refuses to create an index whose key exceeds 3072 bytes, and utf8mb4 bills 4 bytes per character — so `VARCHAR(1024)` is a 4096-byte key on its own. 30 of the shipped index keys are over that limit. Without this pass, `create_schema: true` against MySQL/OceanBase fails on all 30 with:

```
ERROR 1071 (42000): Specified key was too long; max key length is 3072 bytes
```

So deleting it doesn't simplify the feature, it removes it — the Aliyun deployment could no longer create its own schema and would need DDL applied out of band, which is the thing this PR set out to fix.

That said, I agree the algorithm is more machinery than this deserves. Three ways to cut it, and I'd take 2:

1. **Delete it and set `create_schema: false` for OceanBase.** Simplest code. Schema becomes a DBA/ops step. Fine if OceanBase instances are provisioned by hand anyway — you'd be trading the feature for the simplicity.
2. **Delete the algorithm, declare the prefixes explicitly.** Put `mysql_length={...}` on the ~16 offending indexes in the model files. `schema.py` collapses to `import_all_models` + `create_all`, no arithmetic at all, and each prefix is visible where the index is defined. This is what the gateway does by hand in its DDL. Costs a touch of 16 model files; six `UniqueConstraint`s become `Index(..., unique=True)`.
3. **Keep it.**

Which do you want? If it's 1, say so and I'll strip it — I just didn't want to delete it on the "truncates data" reading, since that isn't what it does.

---

src/backend/src/agentclaw/community/core/schema.py
totalfrank: Same answer as the thread above — `prepare_for_mysql` is the caller of `_prefix_lengths`, so they live or die together.

To be precise about what it guards: it is not a length check on *values*. It is a DDL guard — InnoDB rejects `CREATE INDEX` when the key exceeds 3072 bytes, and 30 of our index keys do. Removing it means `create_schema: true` cannot work against MySQL/OceanBase at all.

Options are laid out in the other thread; happy to take option 1 (delete it, schema provisioned out of band) or option 2 (delete the arithmetic, declare the ~16 prefixes explicitly on the model indexes). Just tell me which and I'll push it.

---

src/backend/src/agentclaw/community/configs/application-community.yaml
totalfrank: `backend` is the store selector, not an engine setting — it picks which database `CommunityDatabase` builds an engine for (`sqlite` / `mysql` / `postgresql`) and drives the pooling and pragma setup for it.

On why `sqlite` is the value here: `application-community.yaml` is the overlay for `DEPLOY_PROFILE=community`, and today nothing actually deploys with that profile — it is the corp-free profile the OSS build and local runs use, which is why the checked-in value is the local one. My assumption was that Aliyun would set `backend: mysql` + `DATABASE_URL` from its own env/Secret and never rely on what is written here.

Your question suggests you read this file as *being* the Aliyun deployment config. If that is the intent, then you are right that `sqlite` is the wrong value to ship in it, and my earlier answer missed your point rather than addressed it. Two honest ways to make the file say what it means:

1. **This file is the Aliyun profile.** Set `backend: "mysql"` and `url: "${DATABASE_URL}"` here with no inline default, so a missing Secret fails the boot loudly. Local and OSS runs then need their own overlay, because this one no longer boots bare — that is the cost, and it would need a new overlay plus a test-suite fixup.
2. **This file stays the generic community overlay** (current behaviour) and the Aliyun values live in a deployment overlay or the k8s env.

Earlier I proposed these as options 2 and 1 and you picked 1, so the file stayed as-is — but given this comment I would rather re-confirm than assume. Which is it?

---

totalfrank-personal: community means deploy profile for community. And we chose aliyun to be what is used for our community deployment. Given this, should you update here and potentially a few other places?

totalfrank-personal (commented):
One more comment.

src/backend/src/agentclaw/community/configs/application-community.yaml
totalfrank: Understood — community profile *is* the deployed profile, and it runs on Aliyun. Done in b545d70:

```yaml
  database:
    backend: "mysql"          # OceanBase speaks the MySQL protocol in MySQL mode
    url: "${DATABASE_URL}"    # no inline default — deployment-supplied
    create_schema: true
```

No default on the URL is the point: a DSN can't be a literal (host and credentials are per-deployment), and under strict expansion a missing `DATABASE_URL` now fails the boot loudly instead of quietly pointing production traffic at a file inside the container.

**You were right that it reaches "a few other places" — two, and both are real.** The overlay no longer loads bare, and two things use the community profile without deploying it:

- `scripts/community_selftest.py` — the OSS gate that proves the community tree boots with corp absent.
- `src/gateway/scripts/dump_and_publish.sh` — dumps the backend's OpenAPI schema.

Both only import the app, and config resolution opens no connection (schema DDL runs under the lifespan, which neither enters), so both now supply a placeholder DSN with a comment explaining why. Verified the OSS gate still passes:

```
COMMUNITY BOOT GATE: PASS
```

Also updated: the `CommunityDatabaseConfig` field defaults stay SQLite but are now documented as the no-block fallback rather than "what community deploys", the overlay's comment block, the affected config tests, and the golden snapshot.

Full suite: 14,518 passed (the 2 failures are pre-existing — `rsync` missing in my sandbox; CI runs them green).

**One consequence worth naming:** the `community` profile was carrying two jobs — the corp-free OSS distribution profile *and* the deployed configuration. Those want opposite defaults, and this change picks the deployment. An open-source user checking out the repo can no longer boot the community profile without providing a DSN. That's a deliberate trade, not an oversight; if the OSS bare-boot story matters later, the clean fix is a separate overlay for it rather than softening this one back.

---

totalfrank (commented):
Follow-up round. Commit reviewed this round: `b545d70` (makes the community overlay the deployed OceanBase/mysql configuration, updates the two bare-load call sites and the golden snapshot/tests to match).

**New commit (`b545d70`) review:** verified the diff against the plan the author laid out in the earlier `application-community.yaml` thread. `backend`/`url` in the overlay now read `"mysql"` / `"${DATABASE_URL}"` (no inline default, matching strict-expansion intent); `CommunityDatabaseConfig`'s docstring now correctly describes its SQLite dataclass defaults as the no-block fallback, not what community deploys; `community_selftest.py` and `dump_and_publish.sh` each supply a placeholder DSN with a comment explaining why (both only import the app — no connection is opened, confirmed `create_engine` is lazy and schema DDL only runs under the lifespan, which neither entry point reaches); the golden snapshot and `test_load_yaml_configs.py` were updated consistently, including a new autouse fixture supplying `DATABASE_URL` for that file's tests (checked it doesn't leak into or mask other suites — `test_config_bootstrap.py`, `effective_config.py`, and the architecture guards either don't call the expanding loader on this file or read it as raw text, so none needed the same fixture). No new issues found.

- `application-community.yaml:135` (orig) — OceanBase-vs-`mysql` naming / DDL-location / create-if-not-exists (reviewer Q&A) — answered, author's call taken. Resolved.
- `core/schema.py:67` — dialect-neutral bootstrap question + OceanBase index-limit-vs-InnoDB caveat (reviewer Q&A) — answered; the caveat is a flagged open item pending verification against a live OceanBase cluster, not a defect in this PR. **Still open**, unaddressed by this round's commit — left for the human reviewer to close.
- `di/config_community.py:65` — default `backend` value question (reviewer Q&A) — answered, author picked option 1 (SQLite fallback stays, deployment sets `backend`+`DATABASE_URL` explicitly). Resolved.
- `plugins/community/database.py:88` — non-blocking: `bootstrap()` had no direct test — fixed in `1581c75` (4 new tests). Resolved.
- `application-community.yaml:117` (orig) — nit: config comment didn't mention `backend` must change alongside `DATABASE_URL` — fixed in `1581c75`. Resolved.
- `core/schema.py:104` — `prepare_for_mysql` complexity concern, reviewer asked to remove it — answered with 3 options, awaiting reviewer's choice. **Still open**, unaddressed by this round's commit.
- `core/schema.py:132` — `_prefix_lengths` complexity concern, same ask — answered, same 3 options. **Still open**, unaddressed by this round's commit.
- `application-community.yaml` (overlay identity: is this file the deployed Aliyun config?) — reviewer confirmed intent, author implemented it in `b545d70` (this round) — verified the code matches the claim. Resolved.

**Open count:** 0 code-review defects outstanding. 2 reviewer Q&A threads remain open pending the human reviewer's own decision (`prepare_for_mysql`/`_prefix_lengths` complexity — options laid out, awaiting pick; OceanBase InnoDB-limit caveat — flagged, needs a live-cluster check the sandbox can't do).

---

src/backend/src/agentclaw/community/di/config_community.py
```diff
@@ -35,14 +35,44 @@ class BcsAuthConfig:
...
+    ``database`` block is absent entirely — a DSN cannot be a literal default,
+    since host and credentials are per-deployment.
+
+    ``create_schema`` lets the app emit its own DDL at boot, which is what a
+    container deployment needs — nobody runs ``CREATE TABLE`` between
+    ``kubectl apply`` and the first request. Set it false when the schema is
+    provisioned out of band or managed by migrations.
     """
 
+    backend: str = "sqlite"
```
totalfrank-personal, line 63: I guess the sqlite default is here and tests can be easier to set up because they don't need to override, right?

FreddieSun (approved):
LGTM

src/gateway/scripts/dump_and_publish.sh
```diff
@@ -73,7 +73,11 @@ _upstream_dir() {
 
 _upstream_env() {
     case "$1" in
-        backend) echo "DEPLOY_PROFILE=community" ;;
+        # The community overlay requires a deployment-supplied DATABASE_URL (it is
+        # the Aliyun/OceanBase profile, so its url placeholder has no default).
+        # Dumping the schema only imports the app; no connection is opened, so a
+        # placeholder DSN is enough to get past config resolution.
+        backend) echo "DEPLOY_PROFILE=community DATABASE_URL=mysql+pymysql://dump:dump@127.0.0.1:3306/agentclaw" ;;
         baas)    echo "SECBAAS_RUN_MODE=bare" ;;
```
totalfrank-personal, line 81: What's this for?

The file is a file that dumps the schema? If we don't have this DB, will it fail?

totalfrank-personal (commented):

src/backend/src/agentclaw/community/di/config_community.py
totalfrank: Right, though the mechanism is slightly different from "tests don't override" — they never had anything to override.

The provider reads the `database` block out of `user_config`, and `application-test.yaml` ships **no** `database` block at all (there is a guard test pinning that: `test_base_application_yaml_has_no_data_infra_blocks` — the four data-infra blocks are community-overlay-only). So under `DEPLOY_PROFILE=test`, `_block("database")` returns `{}` and these dataclass defaults are the whole answer. That is what `test_database_config_defaults_are_sqlite_with_schema_creation` pins.

So the defaults serve two cases, both "the block is absent" rather than "the block says something else":

- the test profile, which has no such block;
- a hand-rolled config that drops the block.

The deployed path never reaches them — `application-community.yaml` always sets all three keys.

Which is also why they have to stay SQLite-shaped: a MySQL default would need a DSN, and there is no host/credential literal that could be right for anyone. Flipping `backend` to `"mysql"` alone would be worse than useless — it would fail the scheme-match check against the SQLite `url` default and break every no-block resolution.

---

src/gateway/scripts/dump_and_publish.sh
totalfrank: Worth separating two senses of "schema" here: this script dumps the backend's **OpenAPI schema** (the public `/openapi/v1` API description the gateway consumes), not a database schema. It never reads or writes a database.

**Why it needs the line at all:** it runs the dump under `DEPLOY_PROFILE=community`, so it loads `application-community.yaml`. That overlay's `database.url` is now `${DATABASE_URL}` with no default, and strict expansion raises `KeyError` on an unset variable — so without a DSN the dump would fail during *config loading*, before it ever got to generating the schema. The placeholder is what gets it past that.

**Will it fail without a real DB? No.** Importing the app builds the injector, and `create_engine` does not open a socket — it only resolves a driver. Schema DDL runs in `bootstrap()`, which is a lifespan hook, and the dump never enters the lifespan. So a syntactically valid DSN pointing at nothing is sufficient.

Verified, with no MySQL server running anywhere:

```
$ DEPLOY_PROFILE=community DATABASE_URL="mysql+pymysql://dump:dump@127.0.0.1:3306/agentclaw" \
    python scripts/dump_openapi.py out.json
wrote public OpenAPI to out.json
exit=0
DUMP OK — paths: 191 | no MySQL server was running
```

If you would rather this not depend on a fake DSN at all, the alternative is to dump under a profile that boots bare — `DEPLOY_PROFILE=test` loads `application-test.yaml`, which carries no `database` block. That changes which overlay the published API description is generated from, though, so I did not switch it unasked.

---

cassiuscai (approved):

PR #1537, opened 2026-08-26
fix(backend): log Passport update payloads

yhzhang35:
## Problem

TCAuth validates MCP identity values before its request-start log, so a rejected Caller/Owner sync cannot identify the offending payload entry.

## Solution

- Log the complete non-token Passport scope immediately before normal MCP scope refreshes.
- Log the complete non-token MCP item list immediately before Caller identity synchronization.
- Add ordering regressions that assert each log exists when the Passport port is invoked.

## Validation

- `uv run pytest tests/community/core/mcp/services/test_caller_identity_principal_sync.py tests/community/core/mcp/services/test_sync_service.py -q` (42 passed)
- `uv run ruff check ...`
- `uv run python -m compileall -q ...`
- `git diff --check`


totalfrank (dismissed):
LGTM.

totalfrank (approved):
LGTM.

PR #1540, opened 2026-08-26
fix(aicoding): add cron payload owner and bot ids

panqiwp:
Add owner_id and bot_id to the auto-setup cron payload so the DIMA cron task carries the caller/bot context.

Validation:
- uv run pytest tests/community/core/cron/test_cron_auto_setup.py tests/community/core/aicoding/services/test_bot_resolution_service.py --cov=agentclaw.community.core.cron.services.aicoding.cron_auto_setup --cov=agentclaw.community.core.aicoding.services.bot_resolution_service --cov-report=term-missing
- 82 passed

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/aicoding/router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/aicoding/router.py
@@ -26,8 +26,8 @@ from agentclaw.community.adapters.http.auth.models import AuthenticatedUser
 from agentclaw.community.adapters.http.auth.dependencies import get_current_user
 from agentclaw.community.api.bot_service import BotServiceProtocol
 from agentclaw.community.api.baas_service import BaasServiceProtocol
-from agentclaw.community.core.aicoding.services.bot_resolution_service import (
-    AicodingBotResolutionService,
+from agentclaw.community.api.aicoding.bot_resolution_service import (
+    AicodingBotResolutionServiceProtocol,
 )
 from agentclaw.community.api.device_service import DeviceServiceProtocol
 from agentclaw.community.api.workflow_catalog_service import WorkflowCatalogServiceProtocol
@@ -40,6 +40,7 @@ from agentclaw.community.core.bot_management.services.bot_service import (
     BotNotFoundError,
     BotServiceError,
 )
+from agentclaw.community.utils.env_utils import get_current_env
 from agentclaw.community.core.bot_management import codefuse_token as _codefuse_token
 from agentclaw.community.core.repository.protocols.devices import DeviceBindingRepository
 from agentclaw.community.core.services.identity import VALID_ENTITY_TYPES
@@ -312,7 +313,7 @@ async def create_bot_dima_workspace(
     user_id: Optional[str] = Query(None, description="Bot owner user ID"),
     ctx: RequestContext = Depends(get_request_context),
     bot_service: BotServiceProtocol = Injected(BotServiceProtocol),
-    bot_resolution_service: AicodingBotResolutionService = Injected(AicodingBotResolutionService),
+    bot_resolution_service: AicodingBotResolutionServiceProtocol = Injected(AicodingBotResolutionServiceProtocol),
 ) -> DimaWorkspaceResponse:
     """为 Coding bot 创建 DIMA 工作空间（幂等）。
 
@@ -343,6 +344,7 @@ async def create_bot_dima_workspace(
             bot_id=bot_id,
             requested_owner_id=resolved_user_id,
             operator_id=operator_id,
+            env=get_current_env(),
         )
         if not resolved_bot:
             raise BotNotFoundError(f"Bot not found: {bot_id}")
--- a/src/backend/src/agentclaw/community/adapters/http/aicoding/router.py
+++ b/src/backend/src/agentclaw/community/adapters/http/aicoding/router.py
@@ -26,6 +26,9 @@ from agentclaw.community.adapters.http.auth.models import AuthenticatedUser
 from agentclaw.community.adapters.http.auth.dependencies import get_current_user
 from agentclaw.community.api.bot_service import BotServiceProtocol
 from agentclaw.community.api.baas_service import BaasServiceProtocol
+from agentclaw.community.core.aicoding.services.bot_resolution_service import (
+    AicodingBotResolutionService,
+)
 from agentclaw.community.api.device_service import DeviceServiceProtocol
 from agentclaw.community.api.workflow_catalog_service import WorkflowCatalogServiceProtocol
 from agentclaw.community.core.bot_collaborator.interceptor import (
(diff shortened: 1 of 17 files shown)


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

### Gate 0
Target branch: `dev` (the integration branch) — REL-branch isolation rules don't apply. Gate 0 passes trivially regardless of feature/fix classification. No isolation audit required.

### Findings

**1. `src/backend/src/agentclaw/community/core/cron/services/aicoding/cron_auto_setup.py:271-272` — new fields added to a cross-component wire body that the receiving contract does not declare; the change is a silent no-op**

```python
adapter_body = {
    ...
    "kind": "autoInitiate",
    "owner_id": owner_id,
    "bot_id": bot_id,
}
```

`adapter_body` is forwarded verbatim as the JSON body of `POST /api/cron` via `CronRelayService.forward_request` (`core/cron/services/cron_relay.py:788`) → `DeviceAdapterTransport`, landing on the engine's cron router. That endpoint's declared HTTP contract is `CreateTaskRequest` in `src/engine/src/engine/community/api/cron/schemas.py`:

```python
class CreateTaskRequest(BaseModel):
    name: str
    schedule: str
    command: str
    timezone: str = "Asia/Shanghai"
    enabled: bool = True
    timeout_secs: Optional[int] = 86400
    model: Optional[str] = None
    runtime: Optional[str] = None
    kind: Optional[str] = None
    append_message: Optional[str] = None
    notify: Optional[NotifyRequest] = None
```

No `owner_id` or `bot_id` field exists here, and the model declares no `model_config` override. Pydantic v2 `BaseModel` defaults to `extra="ignore"`, so FastAPI silently drops both new keys while parsing the request body — `create_task(request: CreateTaskRequest)` in `engine/community/api/cron/router.py:143` never sees them. As written, this change has **no runtime effect** on the receiving side and does not achieve the PR's own stated goal ("so the DIMA cron task carries the caller/bot context"). None of the 8 changed files touch `src/engine/`, confirming the receiving schema was not updated in this PR.

- **Violated rule:** `arch.rules.md` Rule 1 (API Specification Is the Single Authority — Invariant: "Implementations must not introduce undocumented semantics that other components depend on") and Rule 16 (Changes Propagate — Invariant: "PRs touching Service APIs... must include propagation analysis... Missing = finding"). Also catalog item 5 (hidden/undocumented wire contract extended without an explicit, documented, versioned schema on both sides).
- **Why it breaks the architecture:** `adapter_body` is an ad hoc dict standing in for a real Service API request across a process/component boundary (backend → engine). Extending one side of that boundary without touching the schema that actually governs deserialization on the other side is exactly the "prompt string parsed by the other side" failure mode this constitution calls out — except here it's silent data loss rather than a parse error, which is worse: it will look like it works (tests pass, no exception) while doing nothing.
- **Required fix:** Add `owner_id`/`bot_id` (or whatever fields the engine's cron job model actually needs to carry caller/bot context) as explicit optional fields to `CreateTaskRequest`, thread them into the engine's `CreateJobRequest`/plugin-layer model, and include propagation analysis for the engine side in this PR (or a paired PR referenced from this one). A guard/conformance test asserting the backend's `adapter_body` keys are a subset of the engine's accepted schema would catch this class of drift in CI going forward (Rule 25).

### Out-of-scope notes
- `router.py` newly imports and calls `get_current_env()` directly in the HTTP adapter to build a core-service parameter. This is consistent with dozens of existing call sites across both routers and core services in this codebase (e.g. `bot_collaborator/services/collaborator_service.py`), so it isn't flagged as a blocking finding here, but it is the more compliant direction relative to core calling `get_current_env()` itself — worth eventually consolidating through a composition-root-provided value rather than ad hoc utility calls.
- `AicodingBotResolutionServiceProtocol.resolve_bot_for_dima_workspace` returns `Dict[str, Any] | None` rather than a typed record; callers use `.get("owner_id")`. This matches existing repository-layer conventions in this codebase and is not gated on here, but a typed return would make the contract self-documenting.

---

totalfrank (changes requested):
**Verdict: REQUEST_CHANGES**

## Gate 0

- Target branch: `dev` (integration branch — REL-branch rules and the Isolation Audit do not apply).
- Classification: despite the `fix(aicoding)` title, this is new behavior riding on an existing pipeline (new bot-resolution logic on the DIMA workspace endpoint, new fields on the cron auto-setup payload) — since the target is `dev`, that classification doesn't block the PR by itself. Gate 0 passes trivially (correct branch for this kind of change).

## Findings

**1. `src/backend/src/agentclaw/community/core/cron/services/aicoding/cron_auto_setup.py:260-270` — new fields added to an undocumented, unversioned cross-boundary payload, with no propagation analysis or conformance test.**

```python
adapter_body = {
    "name": cron_name,
    "schedule": cron_expr,
    "command": command,
    "timezone": DEFAULT_CRON_TIMEZONE,
    "enabled": True,
    "timeout_secs": DEFAULT_CRON_TIMEOUT_SECS,
    "kind": "autoInitiate",
    "owner_id": owner_id,   # new
    "bot_id": bot_id,       # new
}
...
result = await self._cron_relay.forward_request(bot_id=bot_id, user_id=owner_id, ..., body=adapter_body, ...)
```

`CronRelayService.forward_request` (`core/cron/services/cron_relay.py:788`) declares `body: Optional[dict] = None` — an untyped passthrough — and forwards it over `DeviceAdapterTransport`, a genuine system boundary to the bot's external engine adapter. There is no schema, spec, or doc anywhere defining what this dict must contain; the only existing "contract test" for this boundary (`tests/community/contracts/test_device_adapter_transport.py`) treats `body` as an opaque blob and asserts nothing about its shape. The same file already builds a second hidden wire protocol in `_build_cron_command` (a natural-language string with embedded `|space:{id}|user:{id}|agent:{id}|kind:{kind}|...` segments, parsed back by `_parse_cron_command`) — a pre-existing instance of catalog item 5.

This PR adds two new semantic fields (`owner_id`, `bot_id`) to the JSON side of that same implicit contract, extending rather than fixing the violation. Per **Rule 1** ("All inter-component behavior must be governed by explicit contract definitions... contract changes update specs/docs in the same change") and **Rule 16** (Invariant — "Changes to contracts... must declare propagation scope: affected consumers, affected implementations... compatibility status"), a cross-boundary payload change needs: what the receiving engine adapter does with these fields, whether older adapters tolerate the new keys, and a conformance/compatibility test — none of which are present. The PR description's Validation section only lists an internal unit test (`test_cron_auto_setup.py`), which checks that *we* set these keys, not that the receiving side has a defined, tested contract for them. `docs/arch/ci.enforce.md:184` explicitly flags "contract changes without associated conformance updates" as a CI concern.

**Required fix:** define an explicit, documented shape for the cron-relay → engine-adapter payload (even a small typed model or a documented dict schema is enough), state what `owner_id`/`bot_id` mean to the receiving side and how an adapter without support for them should behave, and extend the boundary's conformance test (`tests/community/contracts/test_device_adapter_transport.py` or a new one) to assert on it. This is exactly the propagation analysis Rule 16 requires for any cross-component contract change, including implicit ones.

## Out-of-scope notes

- `core/aicoding/` is a documented (README/context-boundary), pre-existing core package, but "aicoding" is also a literal engine-type string used for real branching elsewhere in the repo (e.g. `core/notify/constants.py`, `core/workspace/engines/aicoding.py`). The new files here (`protocols.py`, `bot_resolution_service.py`) don't themselves branch on engine type, so not gated, but the naming collision is worth a cleanup pass separately.
- The new `AicodingBotResolutionServiceProtocol` returning `Dict[str, Any] | None` matches the existing `BotRepository` convention (also dict-based), so not flagged as a new contract weakness.

---

panqiwp:
Superseded by #1553, which fixes the contract propagation issue and carries owner/bot context end-to-end.

PR #1556, opened 2026-08-26
fix(backend): map MCP permission denials to forbidden

FreddieSun:
## Problem
Legacy `/api/skillsets/{id}/mcps` requests surfaced `McpPermissionDeniedError` through the generic exception handler, returning HTTP 500 for an authorization denial. The Gateway surface already maps the same domain outcome to HTTP 403 with code `403202`.

## Solution
Make `McpPermissionDeniedError` a transport-agnostic `DomainError` with a stable default detail and map it to HTTP 403 in the shared HTTP adapter. Legacy BFF keeps its existing `{"detail": ...}` envelope while Gateway keeps its existing public error mapping.

## Validation
- Red/green regression: legacy app handler returned 500 before the change and 403 after it.
- `57 passed`: SkillSet HTTP status mapping and OpenAPI response mapping suites.
- `9 passed`: domain-error completeness and MCP control-plane focused tests.
- `ruff` passed for the changed domain-error and regression-test files. `app.py` retains four unrelated pre-existing lint findings (duplicate task imports and duplicate callback map keys).
- `git diff --check` passed.

## Compatibility and risk
No request schema, database schema, or Gateway error code changes. The only wire change is the legacy BFF permission-denied response status from 500 to 403.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -363,6 +363,7 @@ from agentclaw.community.core.caller_identity.contracts import (  # noqa: E402
     CallerMcpSyncError,
 )
 from agentclaw.community.core.skill_center.errors import (  # noqa: E402
+    McpPermissionDeniedError,
     SkillSetAccessDeniedError,
     SkillSetControlPlaneConflictError,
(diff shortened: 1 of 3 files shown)


totalfrank (approved):
LGTM.

PR #1557, opened 2026-08-26
feat(bcs): support bot identity for group and session APIs

quechao76:
## Problem

BCS Group and Session OpenAPI operations only accepted Human identity for several create, read, update, and delete flows. This prevented Bots from managing collaboration resources they participate in or own.

Legacy Group routes also needed to accept Bot tokens consistently, including inline event subscriptions, while preserving strict container validation.

## Solution

- Apply the `HumanOrOwnedBot` identity policy to:
  - Group create, detail, update, and delete operations
  - Session create, detail, update, and delete operations
  - Inline Group Event Subscription provisioning
- Preserve resource-level authorization:
  - Bots may only access or manage Groups and Sessions for which they are an authorized participant, originator, driver, or creator as applicable.
  - Human callers retain owned-Bot delegation behavior.
- Update Gateway route policies so Application identity remains required while Human or Bot identity can be supplied as the effective caller.
- Allow Legacy `POST /groups` requests with a Bot token to include inline event subscriptions.
- Allow Legacy `PATCH /groups/{id}` requests to use Bot identity.
- Preserve strict `x-agentclaw-bolt-id` container validation for both Legacy Bot write paths.
- Update OpenAPI contracts, event subscription documentation, and identity-policy descriptions.
- Add Group, Session, Eventing, Gateway, OpenAPI, and Legacy route regression coverage.

## Validation

- Group V1 application tests: 82 passed
- Session V1 application tests: 71 passed
- Gateway route-security tests: 68 passed
- Event subscription tests: passed
- Targeted OpenAPI contract tests: passed
- `cargo test -p bcs-http --test legacy_group_event_subscription_contract`: 3 passed
- `cargo test -p bcs-http --test legacy_group_patch_contract`: 6 passed
- `git diff --check`: passed

The full historical OpenAPI test set still reports two pre-existing schema-drift failures unrelated to this change. Repository-wide `cargo fmt --check` is also blocked by existing formatting drift in unrelated Rust files.

## Compatibility and risk

This is an authorization-contract expansion: existing Human behavior remains supported, while authorized Bot callers gain access to the affected operations.

Application identity remains required by the Gateway OpenAPI routes. Bot identity does not bypass Group or Session resource authorization, and Legacy Bot writes continue to enforce strict container validation when enabled.

No persistence schema or data migration is required. The change can be rolled back by reverting commit `00701ffb7`.

### DIFF
--- a/src/bcs/api-contracts/v1/openapi/groups.yaml
+++ b/src/bcs/api-contracts/v1/openapi/groups.yaml
@@ -102,8 +102,10 @@ GroupsPath:
     tags: ["Collaboration / Groups"]
     summary: Create a collaboration group or direct-message group.
     x-avernet-security:
-      user: required
+      user: optional
       app: required
+      bot: optional
+    x-bcn-identity-policy: human_or_owned_bot
     requestBody:
       required: true
       content:
@@ -152,7 +154,10 @@ GroupsPath:
             schema:
               $ref: ../shared.yaml#/ErrorEnvelope
       "403":
-        description: A selected Bot is not collaboration-eligible.
+        description: >-
+          No Human or Bot Actor is available, the authenticated User/Bot
+          ownership claims conflict, or a selected Bot is not
+          collaboration-eligible.
         x-error-codes:
           - forbidden
         content:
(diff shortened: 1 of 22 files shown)


questchaos:
@codex review

vzvince (approved):

PR #1558, opened 2026-08-26
fix(ci): render reachable ports and probe paths for every deployed service

totalfrank:
## Problem

`docker/kube-deploy.sh` rendered a manifest no service could serve. Both faults hit all five services, not just the backend — they surfaced while reviewing #1555, which added `backend.dockerfile`, but neither is introduced by it.

**Every rendered Service targeted a port with no listener.** `DEFAULT_PORT` was 8080 for baas, gateway, proxy, bcs and backend alike. None of them listens there:

| service | real default | source |
| --- | --- | --- |
| baas | 8888 | `configs/application.yaml` `module_config.web.port`; `WebConfig.port` default |
| gateway | 8888 | `configs/application.yaml`; `_config_loader.py` default |
| proxy | 8888 | `WebConfig.port` default (its configs set no port) |
| bcs | 21000 | `configs/bcs-config-*.toml` |
| backend | 8888 | `main.py` `BACKEND_PORT` default |

The script's own header disagreed with its `DEFAULT_PORT` array as well, listing gateway at 8080 and backend at 8090 — three different answers, none of them right.

**The shared template hardcoded `/health` for both probes.** The backend mounts its routes under `/api` and registers only `/api/health` (`adapters/http/app.py:898`), so a backend pod rendered from this template would never pass readiness, and liveness would restart a healthy pod on a loop.

## Solution

Per-service tables in `kube-deploy.sh`, and a `${PROBE_PATH}` placeholder in `service-deployment.yaml`.

`DEFAULT_PORT` now carries each service's real default, and `DEFAULT_PROBE_PATH` gives `/health` to baas/gateway/proxy/bcs — unchanged behavior for all four — and `/api/health` to backend.

A third fault surfaced while fixing the first: **a `PORT` override only relabelled the Service** while the process kept listening where its config said, reproducing the same dead port. gateway, proxy and backend each read a port env var (`GATEWAY_PORT`, `SANDBOXPROXY_PORT`, `BACKEND_PORT`) that wins over the config file, so `PORT_ENV_VAR` now injects it set to `PORT` and the two move together. An explicit `--env` for the same key still wins — the operator asked for it by name — and warns if it disagrees with the published port rather than silently emitting a duplicate. baas and bcs have no such override, so overriding `PORT` for them warns that the mounted config has to be edited to match.

Rejected: adding a `/health` alias to the backend. That puts a route in the app to satisfy a deployment template — the dependency pointing the wrong way — and still leaves the template unable to describe any service whose probe path differs.

**One drive-by**, called out so it can be dropped if unwanted: `usage()` quit on the first `/^# ====/` line, which is the opening banner, so `--help` printed that single line and nothing else. Since the header block is exactly where this change documents the per-service ports, leaving it unreachable would have hidden the fix's own documentation. Now `awk`-based, matching `docker/build-image.sh`.

## Validation

`envsubst` was installed locally (`gettext-base`) and the script run for real; there is no cluster here, so everything below is the rendered manifest, not an applied one. `kubectl apply` was never invoked.

- **All five services render correctly** — port, probe path, and injected env var each checked against the table above: baas `8888 /health`, gateway `8888 /health`, proxy `8888 /health`, bcs `21000 /health`, backend `8888 /api/health`.
- **Rendered YAML parses** as exactly two documents (Deployment + Service) for all five, with an assertion that `livenessProbe.httpGet.port == containerPort == Service.targetPort` and that both probes carry the same path. No unsubstituted placeholder survives — the one `${...}` remaining in the output is literal text inside the template's own YAML comment, pre-existing.
- **`PORT=9000 --service backend`** → `containerPort: 9000` *and* `BACKEND_PORT=9000`, moving together.
- **`--env BACKEND_PORT=9999`** → operator's value kept, exactly one `BACKEND_PORT` entry (no duplicate), and a warning naming both ports.
- **`PORT=9000 --service baas`** → warns that baas takes its port from its mounted config and that `PORT` alone only changes the manifest.
- **Regressions checked:** `--env-file` still loads and renders, the no-`--env` path still renders a valid manifest, unknown `--service` still exits 2 with the supported list, and `--help` now prints 50 lines instead of 1.
- `bash -n` clean.

**Not verified:** no Kubernetes cluster was available, so no pod was actually scheduled and no probe was observed passing against a running container. The port and path values are verified against each service's config and route registration, not against a live deployment.

## Compatibility and risk

**No behavior change for baas, gateway, proxy or bcs probes** — all four keep `/health`. What does change for them is the rendered port, from 8080 to the port they actually listen on. Any deployment currently relying on the 8080 value was pointing at a dead port, so there is nothing working today that this breaks; a deployment that overrode `PORT` to compensate should drop the override.

`PROBE_PATH` and `PORT` both remain overridable from the environment, so an operator can still name either explicitly.

Rollback is a revert; no schema, API, or data migration.

## Related issues

Clears the two follow-ups #1533 left open and #1555's description carries forward, both also raised by automated review on #1555 (inclusionAI/Avernet#1555 threads on the probe path and the port). The third item from that PR — the `sudo`-prefixed publish path in `BotBuildService._migrate_bot_instance` — is deliberately **not** here; it is a backend change in `core/service_bot/`, tracked separately.

---
_Generated by [Claude Code](https://claude.ai/code/session_01S9dvuxJaszNm4bJxq2vxGa)_

### DIFF
--- a/docker/kube-deploy.sh
+++ b/docker/kube-deploy.sh
@@ -11,11 +11,17 @@
 #       [--namespace avernet] \
 #       [--apply]                 # apply with kubectl (default: dry-run print)
 #
-# Supported services (with default port):
-#   baas       — SecBaaS platform service (port 8888)
-#   gateway    — API gateway (port 8080)
-#   bcs        — BCS coordination service (port 21000)
-#   backend    — Backend service (port 8090)
+# Supported services (default port, probe path):
+#   baas       — SecBaaS platform service   (8888,  /health)
+#   gateway    — API gateway                (8888,  /health)
+#   proxy      — Sandbox proxy              (8888,  /health)
+#   bcs        — BCS coordination service   (21000, /health)
+#   backend    — Backend service            (8888,  /api/health)
+#
+# Each default is the port that service actually listens on out of the box, and
+# the path it actually serves. Override the port with PORT=... in the
+# environment; see the note on DEFAULT_PORT below for which services can honour
+# that inside the container.
 #
 # The script renders docker/services/service-deployment.yaml via
 # envsubst, then optionally applies it with kubectl.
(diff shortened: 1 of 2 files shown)


totalfrank:
Closing — folded into #1555 at the author's request, so the backend image and the deployment tooling it needs land as one change.

Commit `0550362` was cherry-picked onto `claude/aliyun-backend-dockerfile-ra0qmu` as `0ac7eec`, unchanged: same `${PROBE_PATH}` placeholder, same per-service `DEFAULT_PORT` / `DEFAULT_PROBE_PATH` / `PORT_ENV_VAR` tables, same `usage()` fix. Nothing was dropped in the move, and the full render suite was re-run against the combined tree — all five services still produce a valid manifest with `probe.port == containerPort == Service.targetPort`.

Folding also made one thing checkable that neither PR could verify alone: the Dockerfile's `HEALTHCHECK` (`/api/health` on `BACKEND_PORT:-8888`) now demonstrably agrees with the rendered k8s probe and the injected `BACKEND_PORT=8888`.

Review the tooling change on #1555; this branch is superseded, not abandoned.

---

PR #1560, opened 2026-08-26
feat(bcs): add bcs aliyun docker image and ${VAR} config env expansion

zh-bupt:
## Problem

Deploying bcs on Aliyun k8s needs two things that did not exist:

1. **A container image** that builds and runs the bcs server binary on Aliyun ACK.
2. **A way to inject per-environment secrets/credentials into the TOML config at runtime** without baking them into the image. `bcs-config` had no generic environment interpolation — only specific typed overrides (`BCS_DB_*`, `BCS_REDIS_*`) and `*_env` indirection fields — leaving plain-literal secret fields (`auth.oauth.jwt_secret`, OAuth `client_secret`, `dingtalk_accounts`, `api_keys`, `botchat_url`, `bcs_endpoint`) unable to read from env at all.

## Solution

**`docker/services/bcs.dockerfile`** — multi-stage image mirroring `gateway.dockerfile`/`baas.dockerfile`:
- Builder: `rust:1.91-bookworm` (MSRV 1.91) + `pkg-config`/`libssl-dev`/`libsqlite3-dev` (native-tls → OpenSSL; `rusqlite` is not bundled). `cargo build --release --locked --bin bcs` with registry/target BuildKit cache mounts; the produced binary is `cp`'d to a stable path and `strip`'d within the same `RUN` (cache-mount contents don't survive the layer).
- Runtime: `debian:bookworm-slim` + `libssl3`/`libsqlite3-0` (bookworm in both stages to keep the ABI matched), non-root `admin` (10001), `COPY src/bcs/configs /app/configs`, `EXPOSE 21000`, `/health` healthcheck, `ENTRYPOINT ["bcs"]` reading `BCS_CONFIG_DIR`. Code comes from the build context (`COPY src/bcs`) — no in-image git clone; `docker/build-image.sh services/bcs.dockerfile` sends the repo root as context.

**Config `${VAR}` env expansion (string values only)** — after the env-overlay `deep_merge` and before deserialization, walk the merged `serde_json::Value` tree and expand `${VAR}` / `${VAR:-default}` in every string leaf:
- `${VAR}` resolves from the process env; a missing variable is a **hard error** (fail fast) so a missing secret aborts startup instead of silently becoming empty. `${VAR:-default}` is the opt-out for optional values.
- Quoted strings only (`key = "${VAR}"`); bare `key = ${VAR}` is not valid TOML. Non-string values are untouched, so `${VAR}` works only for string-typed fields; numeric/boolean fields keep using the typed `BCS_*` overrides (e.g. `BCS_REDIS_PORT`), which parse the env string into the right type. No nested expansion; format-agnostic (TOML and JSON both route through `serde_json::Value`).
- `config_loader.rs`: add `expand_env_vars`/`expand_env_string`, `ConfigLoadError::EnvExpansion`, extract `parse_config_content` for reuse, wire expansion into `load_with_info`. `config.rs`: route `BcsConfig::from_file` through `parse_config_content` → `expand_env_args` → `serde_json::from_value` so standalone/single-file configs expand too.

Draft because the bcs dockerfile still has an unresolved runtime-config caveat (see below) that I want to settle before merge.

## Validation

- `cargo build -p bcs` ✅
- `cargo test -p bcs --lib config::` — 65 passed (incl. 2 new `from_file` expansion tests).
- `cargo test -p bcs --lib config_loader::` — 34 passed (incl. 11 new env-expansion unit/loader-integration tests covering present/var-with-default/unset-error/multi-token/invalid-name/unterminated/non-string-skipped).
- `cargo clippy -p bcs --lib` — clean on the edited files (remaining findings are pre-existing in dependency crates).
- The 2 `migrations::tests` failures are pre-existing and unrelated (this host's older SQLite rejects `ALTER TABLE … DEFAULT (expr)` and `DROP COLUMN`); confirmed by re-running on clean `dev` with these changes stashed.

## Compatibility and risk

- Config: **additive**. Existing configs have no `${...}`, so expansion is a no-op for them — zero regression risk. Only configs that opt into `${...}` are affected, and a missing required var now fails loudly at startup (safer than a silent empty secret).
- Image: `bcs.dockerfile` ships `src/bcs/configs` as-is. The checked-in configs (`bcs-config-local.toml` binds `127.0.0.1`; `bcs-config-example.toml` has placeholders) are **not container-runnable as-is** — a real `bcs-config-{env}.toml` (with `bind = "0.0.0.0"`) must be present in `/app/configs`, either added to the repo or mounted via k8s. With the new `${VAR}` expansion, secrets/endpoints can now be injected via k8s Secret → env → `${VAR}` in that config, keeping secrets out of the image. healthcheck port is hardcoded to 21000 (bcs reads its port from config, not env — keep in sync).
- Persistence: `bots_base_dir` / session-file `data_dir` need a writable volume mount (the example config uses `/var/lib/bcs`).

## Open question for review

Should a runnable `bcs-config-prod.toml` (or a k8s ConfigMap/Secret spec) be added in this PR, or is the Dockerfile + `${VAR}` expansion enough and the runtime config handled by the deploy manifests? Leaning toward the latter (manifests own the real config); the image + expansion are the reusable pieces.

### DIFF
--- a/docker/services/bcs.dockerfile
+++ b/docker/services/bcs.dockerfile
@@ -1,0 +1,81 @@
+########## Builder ##########
+FROM rust:1.91-bookworm AS builder
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    CARGO_TERM_COLOR=never \
+    CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+
+WORKDIR /build
+
+# Aliyun apt mirror (matches the gateway/baas builder convention).
+# native-tls (tokio-tungstenite/mysql_async) needs OpenSSL; rusqlite is *not*
+# bundled, so both need their dev headers at build time.
+RUN sed -i "s|deb.debian.org|mirrors.aliyun.com|g" /etc/apt/sources.list.d/debian.sources \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        pkg-config libssl-dev libsqlite3-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# "Pull code": the BCS Rust workspace is brought in from the build context
+# (repo root). docker/build-image.sh sends the repo root as the context, so
+# this COPY is the code-fetch step — there is no in-image git clone. The
+# workspace is self-contained under src/bcs (no path deps outside it).
+COPY src/bcs /build/src/bcs
+
+# Fetch dependencies into the registry cache first, so source-only changes
+# reuse the downloaded crates.
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    cargo fetch --locked --manifest-path /build/src/bcs/Cargo.toml
+
+# Build the `bcs` server binary (release, pinned to the checked-in Cargo.lock).
+# The target dir is a cache mount, which is NOT retained in the layer after the
+# RUN — so copy the produced binary to a stable path and strip it within the
+# same RUN while the cache mount is still mounted.
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/build/src/bcs/target \
+    cargo build --release --locked --manifest-path /build/src/bcs/Cargo.toml --bin bcs \
+    && cp /build/src/bcs/target/release/bcs /build/bcs \
+    && strip /build/bcs
+
+########## Runtime ##########
+FROM debian:bookworm-slim AS runtime
+
+# BCS_CONFIG_DIR points the binary at its config directory (clap reads this env
+# directly). SERVER_ENV selects the bcs-config-{env}.toml overlay; leave it to
+# the deployment to set a value whose matching config exists in /app/configs.
+ENV DEBIAN_FRONTEND=noninteractive \
+    HOME=/home/admin \
+    BCS_CONFIG_DIR=/app/configs
+
+WORKDIR /app
+
+RUN sed -i "s|deb.debian.org|mirrors.aliyun.com|g" /etc/apt/sources.list.d/debian.sources \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        ca-certificates curl libssl3 libsqlite3-0 \
+    && rm -rf /var/lib/apt/lists/* \
+    && groupadd --gid 10001 admin \
+    && useradd --uid 10001 --gid admin --create-home --shell /bin/bash admin
+
+COPY --from=builder /build/bcs /usr/local/bin/bcs
+COPY src/bcs/configs /app/configs
+
+# Persistence: bots_base_dir and the session-file data_dir must live on a
+# writable volume. The shipped example config uses /var/lib/bcs; mount a
+# PVC/emptyDir here and point the config's data paths at it.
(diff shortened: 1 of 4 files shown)


zh-bupt (commented):
## Review

复核了 PR 声明的验证项与关键上下文，总体设计合理（expansion 放在 config loader 层、fail-fast 语义、测试覆盖到位），发现一个建议修复的问题和若干小项。

**已复核的声明**（本地 checkout PR head 实际运行/核对）：

- `cargo test -p bcs --lib config_loader::` → 34 passed ✓（含 11 个新 expansion 测试）
- `cargo test -p bcs --lib config::` → 65 passed ✓（含 2 个新 `from_file` 测试）
- MSRV 1.91、`rusqlite` 无 bundled、`/health` 路由存在、clap 读 `BCS_CONFIG_DIR`、`SERVER_ENV` 选择 overlay——均与描述一致 ✓
- Dockerfile 与 gateway/baas 先例一致（aliyun mirror、admin/10001、healthcheck 模式）✓
- 兼容性声明核实：checked-in configs 中无任何 `${` 字面量，expansion 对现存配置确为 no-op ✓
- 未验证：clippy、docker build 本身（无 daemon）

### 建议修复

**1. `config.rs` fallback 分支吞错（低严重度，真实 bug）**

```rust
_ => parse_and_expand("json").or_else(|_| parse_and_expand("toml"))?,
```

无扩展名文件"先试 JSON 再试 TOML"原本安全，因为旧代码只有 parse 失败。现在 `parse_and_expand` 包含 expansion 与反序列化：一个 JSON 文件若 `${VAR}` 未设置或字段类型错误，错误会被吞掉、重试 TOML，最终报误导性的 "TOML parse error"，掩盖真正的 "environment variable `X` is not set"。建议只在 JSON **parse** 失败时 fall through：

```rust
_ => match parse_config_content("json", &content, &path_display) {
    Ok(mut v) => { expand_env_vars(&mut v)?; serde_json::from_value(v)? }
    Err(_) => parse_and_expand("toml")?,
}
```

### 建议项

2. **字面量 `${` 无转义出口**：含 `${...}` 的字符串（URL/消息模板等）若变量未设置会启动失败。现有配置无此情况（已核实），但 `${VAR}` / `${VAR:-default}` 语法与"无转义"限制目前只存在于本 PR 描述中——建议在 `bcs-config-example.toml` 加注释固化为文档。
3. **Dockerfile fetch 层注释过度声称**："source-only changes reuse the downloaded crates" 不准确——`COPY src/bcs` 在 fetch 之前且含全部源码，任何源码改动都会 bust 该层；复用实际来自 BuildKit registry cache mount。建议改写注释。
4. **Healthcheck 端口漂移**：硬编码 21000，bcs 无端口 env override（gateway 用 `${GATEWAY_PORT:-8888}` 模式）。PR body 已透明说明；建议在 shipped config 注释里也提示"改端口需同步 healthcheck"。
5. Dockerfile 缺末尾换行。

### 结论

问题 1 建议修复后合入；2-4 为建议项。关于 open question：同意 runtime config 由 deploy manifests 负责，镜像 + `${VAR}` expansion 作为可复用件。

LGTM after fixing #1.

questchaos (approved):
LGTM

cassiuscai (approved):

PR #1563, opened 2026-08-26
fix(bcs): include context in group list summaries

zh-bupt:
## Problem

`GET /openapi/v1/collaboration/groups` and
`GET /openapi/v1/collaboration/public-groups` return group summary objects
that omit the group `context`, so callers listing groups (including the public
plaza) cannot see each group's context without a follow-up detail fetch.

## Solution

Add `context` to the group summary DTOs and populate it from the domain group.
Single commit:

- Add `context: Option<String>` to both `NormalGroupSummary` and
  `DirectMessageGroupSummary` (`bcs-service-api`), and populate it in
  `project_summary` for both the Normal and DM branches (`bcs-app-group`).
  Both variants are covered because `list_groups` can return either.
- Contract: add `context` (optional, not in `required`) to both summary schemas
  in `domain-models.yaml`; regenerate `bcn.openapi.json` (verified the only diff
  vs. the prior snapshot is the `context` addition to `GroupSummary` and the two
  list endpoints' items).
- New test `list_summaries_include_group_context` asserts `context` is present
  in both `list_groups` and `list_public_groups` summaries.

## Validation

- `cargo check -p bcs-service-api -p bcs-app-group --all-targets` — clean.
- `cargo test -p bcs-app-group --test v1_group_service` — 79 passed (includes
  the new test).
- `cargo test -p bcs-service-api --test v1_group_application_contracts` — 7 passed.
- `cargo test -p bcs-api-http --test group_routes` — 23 passed.
- Gateway `test_dump_and_publish_script.py` + `test_gate_and_publish.py` +
  `test_served_openapi.py` — 19 passed.
- `pytest tests/openapi/test_group_v1_contract.py` — 16 passed; the only failure
  is the pre-existing `test_update_group_does_not_accept_context` (unrelated
  `UpdateGroup` PATCH `context` field, red on `dev` already).

## Compatibility and risk

- Additive, optional field on the summary objects; no existing field changes and
  nothing is added to `required`, so current clients are unaffected.
- Operation count unchanged; snapshot regenerated deterministically.


cassiuscai (approved):

questchaos (approved):
LGTM

PR #1396, opened 2026-08-24
(chore) Dev refactory collaboration merge back to dev

totalfrank:

PR #1568, opened 2026-08-26
fix(skillset): MCP device-sync and passport regressions

totalfrank:
## Problem

The Installation-as-single-source-of-truth refactor (#1396, `9c558e9af`) replaced the per-operation MCP sync methods on `SkillSetService` with a single desired-state projection. The projection is a correct *description* of the Bot's target state, but it lost four behaviours the per-operation methods had and gained one it should not have. Five regressions against `REL20260820`, all in one call path:

1. **Skill-set mutations reset every MCP's identity mode to `owner`.** `BotRuntimeProjector._apply_non_skill_projection` sends `resource_scope` without `mcp_items`. `unpack_resource_scope` substitutes bare items, `ProdPassportPlugin._build_mcp_models` defaults their `identity_mode` to `"owner"`, and `McpModel.serialization` then emits `identityMode` because it is no longer `None`. Since `updatePassport` is overwrite-style per list, every mutation asserts `owner` for every MCP — clobbering what `update_mcp_identity_to_agent_principal` set, via the same facade and the same field. `identityMode` decides whose credential an MCP call runs under, so this is a privilege-boundary change.
2. **MCP configuration is never removed from the device.** `remove_mcp_detail` has zero production callers on `dev` (REL had two). Removing an MCP drops it from the `filter-servers` allow-list but leaves endpoint, `api_key`, and custom headers registered on the container permanently.
3. **Every mutation re-pushes configuration for the entire projected set.** `sync_mcp_desired_state` receives `projection.mcp_server_codes` — the whole Bot's MCPs, not just what changed. Adding one MCP to a Bot with N costs N catalogue lookups and `1 + 2(N-1)` device writes, where REL cost 1 and 1. It also fails closed: one catalogue-missing MCP anywhere blocks every MCP add.
4. **Both projection halves run on every mutation, twice on failure.** An MCP add resyncs all symlinks; a skill add re-declares the MCP allow-list. `_project_or_compensate` calls the full `project()` on both the forward and compensating paths.
5. `SkillSetService.refresh_mcp_scope` has no callers.

## Solution

The projection treats *declare* and *deliver* as one act. Declaring is correctly total — `filter-servers` and the Passport manifest are overwrite-style and must carry the whole set. Delivering is not: pushing an MCP's `api_key`, or deleting it, should happen only when that MCP's claim changed.

So the fix separates them, and separates who knows what. Declaration stays total (with `mcp_items` added, fixing 1). Delivery is scoped, and **the mutation declares that scope** — `add_mcp` holds the code it claimed, `remove_mcp` the one it released — rather than the projector re-deriving it from a before/after diff. The projector still performs the I/O, so compensation keeps working and the control-plane service needs no device dependency; it intersects the declared codes with the projected set as a *guard* that can only shrink them.

## Status

| Group | State |
| --- | --- |
| 1 — Passport identity (problem 1) | ✅ complete |
| 2 — thread `ProjectionScope` | ✅ complete (behaviour-preserving) |
| 3 — scoped delivery (problems 2, 3) | ✅ complete |
| 4 — scoped projection (problem 4) | ✅ complete |
| 5 — dead code (problem 5) | ✅ complete |
| 6 — documentation and logging | ✅ complete |
| 7 — identity refused at the seam | ✅ complete |
| 8 — Skill mutations declare their MCP dependencies | ✅ complete |

**Acceptance criteria: 5 of 6 hold, the sixth by half** — see Resolution in `spec.md`. Groups 7 and 8, and the shape of Group 4, were open scope questions raised in an earlier revision of this description; all three have since been decided and implemented.

## Group 7 — the identity-less scope is refused, not patched per caller

The identity-less `resource_scope` was not unique to the projector. Two more live call sites sent `mcp_codes` without `mcp_items`:

- `adapters/http/skill_center/skillsets.py` (`remove_cli_from_default_skill_set`) — removing a CLI demoted every Caller MCP on the Bot. **This is a skill-set mutation, so criterion 1 did not hold until it was fixed.**
- `core/bot_management/engines/aicoding/strategy.py` — a restart with `confirmed_template_update` did the same.

Patching them individually would leave the trap armed for the next caller, so `unpack_resource_scope` now **rejects** a non-empty `mcp_codes` with no `mcp_items`. An empty list stays legal: it grants nothing, so there is no identity to lose, and clearing MCP scope must not require an identity lookup just to satisfy the guard.

Both callers now resolve identity through one shared `resolve_mcp_identity_modes`, which refuses rather than defaulting when the Bot has no primary key or the row is unreadable — the projector's own fail-closed read, extracted so the three callers cannot drift. The restart path declines the refresh in that case: leaving the Bot's existing scope standing beats replacing it with an owner-only snapshot, and the caller already treats that path as best-effort.

## Group 4 — the seam is `SkillSetService`, not `DeviceSync`

The original plan had `BotRuntimeProjector` hand an intent to `DeviceSyncDispatcher.dispatch(ctx)` "instead of orchestrating the device calls itself". It does not orchestrate `DeviceSync` calls — it calls three ports, only one of which is `DeviceSync`, and **`SkillSetService` owns device resolution**, not the projector. Worse, for a pool-layout Bot skills go through `SkillsPoolRuntimeProtocol` and MCP through `SkillSetService`, so "one delivery for a both-halves mutation" is unreachable there regardless of provider.

What shipped instead:

- `SkillSetService.sync_mcp_projection(*, claimed, released, declared)` — one entry point on the service that owns device resolution, owning the deliver-then-declare order that keeps the device from ever citing an MCP it has no configuration for. The projector makes one call where it made two.
- `project()` gates each half on the declared scope. The branch is on the **caller's declared scope**, never on provider or engine type — which is what the rejected `ProjectionFacet` enum got wrong and what the registry exists to avoid.
- Resolving stays unconditional; only applying narrows. The plan is read-only and every pre-flight failure in it must happen before anything is written, so an aborted projection still leaves nothing half-applied for a compensation to unpick.
- `retired_mappings` overrides the Skill flag: they are computed from the actual before/after snapshots rather than declared, so they are evidence Skills moved. Skipping them would strand a published mapping the desired state no longer holds.

The whole-artifact win for teclaw is unchanged in substance and still needs the corp-side `apply_runtime_projection`; its seam is now `sync_mcp_projection`. That is the one half of criterion 4 that does not hold.

## Group 8 — Skill mutations declare their MCP dependencies

`add_skill` / `remove_skill` declared no scope and fell back to a full reconcile. Correct, but unnarrowed for a concrete reason: a Skill's `mcp_dependencies` join the Bot's projected MCP set along with the Skill, and the commands held only a `skill_id`, so declaring an empty MCP scope would have left a dependency whitelisted but never configured.

The repository now reads those dependencies **under the row lock the mutation already holds** — the same reason activation reads its member codes there — and returns them on `DesiredStateMutation.mcp_codes`. The four Skill commands, the Default-Set exclusion pair included, build their scope from that. `mcp` follows the dependencies rather than being hard-coded, so a dependency-free Skill mutation is Skills-only and skips the MCP half entirely.

The codes are candidates, not a verdict: the projector intersects a claim with the set it resolved and subtracts the projected set from a release, so a dependency another Skill or the default policy still supplies is never deleted.

The stored column is a mixed historical shape (bare codes, `{"server_code": ...}`, `{"code": ...}`) and now has exactly one decoder, shared with the projection resolver — two decoders could disagree about which rows count, and a mutation would then scope a different set than it projects.

## Validation

Full backend suite locally: **14748 passing**, 58 skipped. The only two failures are `test_bot_build_service_skill_artifact.py`, which need `rsync` on `PATH`; they fail identically without this diff.

New coverage includes `tests/community/core/mcp/services/test_passport_scope.py` and `tests/community/plugin_api/test_passport_resource_scope.py`, for modules that previously had none.

Five fixes were verified by temporarily reverting them and confirming the test fails, rather than trusting a green run: the `mcp_items` identity fix, the fail-closed guard for a missing Bot primary key, the compensating projection's scope inversion, the two projection gates, and the Skill dependency scope at both the command and repository layers.

Two code-review rounds on Group 1 are addressed. Round 2 caught a real bug in round 1's own fix: the fail-closed guard ran *after* the device allow-list was written, so it aborted mid-write and the compensating projection hit the same guard and could not counter-project. Both the guard and the identity read now happen during plan resolution.

## Spec

- `src/backend/specs/2026-08-26-mcp-sync-and-passport-regressions/spec.md`
- `src/backend/specs/2026-08-26-mcp-sync-and-passport-regressions/plan.md`
- `src/backend/specs/2026-08-26-mcp-sync-and-passport-regressions/tasks.md`

Explicitly out of scope, with reasons recorded in the spec: device restart/reprovision (covered by `DeviceService._sync_mcps_when_device_active` on both branches), activation no longer deactivating peers (not a regression — REL's `SkillSetActivator` was already incremental), `SkillSetSwitcher._cleanup_all_non_reserved_items` (reachable only from an endpoint with no production traffic), and the default-CLI merge (dropped by documented intent). Skill-carried MCP dependencies were out of scope in the original spec and are now resolved by Group 8.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HNb3nwV3kabxMxdo4QgW9T)_


src/backend/src/agentclaw/community/api/bot_runtime_projector.py
```diff
@@ -28,20 +29,23 @@ async def project(
...
         retired_mappings: Sequence[PoolSkillMapping] = (),
+        scope: ProjectionScope = ProjectionScope.everything(),
     ) -> None: ...
 
     async def project_mcp_and_cli(
         self,
         *,
         bot_id: str,
         owner_id: str,
+        scope: ProjectionScope = ProjectionScope.everything(),
```
totalfrank-personal: Isn't that the default scope for this method should be something for mcp and cli only?

src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
```diff
@@ -596,6 +596,7 @@ def refresh_restart_authorization(
         passport_plugin: Any,
         skill_set_factory: Any,
         template_service: Any,
+        caller_identity_repo: Any = None,
```
totalfrank-personal: Is this always wired by DI? If so, let's do not make default value to be None. And we can remove the

if caller_identity_repo is None:

down below.

src/backend/src/agentclaw/community/core/bot_management/engines/default.py
```diff
@@ -119,6 +119,7 @@ def refresh_restart_authorization(
         passport_plugin: object,
         skill_set_factory: object,
         template_service: object,
+        caller_identity_repo: object = None,
```
totalfrank-personal: Same here, if it's always wired by DI or if caller always provide it, let's remove the default None behavior.

src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
```diff
@@ -234,6 +234,7 @@ def refresh_restart_authorization(
         passport_plugin: Any,
         skill_set_factory: Any,
         template_service: Any,
+        caller_identity_repo: Any = None,
```
totalfrank-personal: Same here, if it's always wired by DI or if caller always provide it, let's remove the default None behavior.

totalfrank (commented):
Reviewed the full diff for this PR (merge-base `2befbc2e2` → head `f4226e81c`, all 32 commits, 37 files). This is a first review pass; no prior review exists on this PR.

Diff is clean — no findings.

What I checked:
- **Identity fix (Group 1/7)**: `BotRuntimeProjector._passport_mcp_items`/`_resolve_mcp_identity_modes`, `passport_scope.py`'s shared `resolve_mcp_identity_modes`/`passport_mcp_items_from_codes`/`_normalized_identity_mode`, and `unpack_resource_scope`'s new rejection of bare `mcp_codes`. Grepped the whole tree for other `resource_scope={"mcp_codes": ...}` call sites beyond the two the PR fixes (`skillsets.py`, `strategy.py`) — the only other match (`sync_service.py`) already passes `mcp_items` alongside `mcp_codes`, so the new rejection in `unpack_resource_scope` doesn't break it.
- **Scoped delivery/declaration split (Groups 3/4)**: `SkillSetService.sync_mcp_projection` → `sync_mcp_delivery` (per-MCP push/delete) → `sync_mcp_desired_state` (total allow-list declare), and the `ProjectionScope` guard in `BotRuntimeProjector._apply_non_skill_projection` (`claimed = scope.claimed_mcp & codes`, `released = scope.released_mcp - codes`) — confirmed the guard can only shrink a declared scope, never grow it, and `reconcile` correctly forces a full claim with no release.
- **Scope declaration at every call site** in `skill_set_management_service.py` (`add_mcp`/`remove_mcp`/Default-Set pairs/`activate`/`deactivate`/`sync`) and the Skill-dependency scoping (`_skill_claim_scope`/`_skill_release_scope`, `DesiredStateMutation.mcp_codes` populated under the row lock in `capability_desired_state.py` and `default_exclusion_commands.py`).
- **Decoder unification**: `mcp_dependency_scope.mcp_dependency_codes` is now the single decoder shared by `runtime_resolver.py` (projection) and `skill_mcp_dependencies.py` (command scoping), removing the risk of the two disagreeing on the stored column's mixed shape.
- **Compensation**: `MutationProjectionFlow._project_or_compensate` threading and inverting `ProjectionScope` on the rollback path.
- **DI wiring**: `caller_identity_repo` threaded through `BotManagementModule`, `BotService`, and `AicodingProvisioningStrategy.refresh_restart_authorization`, with the restart path declining (not defaulting) when the repo is absent or unreadable.
- Contract/architecture test updates (`test_http_adapter_layer_is_http_only.py`, `test_rule15_skillsets.py`) and the new/expanded unit tests (`test_passport_scope.py`, `test_mcp_dependency_scope.py`, `test_restart_authorization_refresh.py`, `test_skill_set_management_service.py`, `test_capability_desired_state_uow.py`) look aligned with the behavior changes they pin.

Categories from the review checklist that didn't apply here: no new Service/Plugin API surface beyond the already-versioned `ProjectionScope` addition (which ships with conformance-style tests), no new concurrency primitives, no new external input crossing a trust boundary.

---

src/backend/src/agentclaw/community/core/skill_center/runtime_projection_contract.py
```diff
@@ -3,11 +3,67 @@
...
+
+    @classmethod
+    def everything(cls) -> "ProjectionScope":
+        """No mutation to ask, so every projected code counts as claimed.
+
+        Used by the paths with nothing to declare — a device-activated
+        restart, a Skill upload — where the device may hold nothing and the
+        projection is the whole truth.
+        """
+        return cls(skills=True, mcp=True, reconcile=True)
```
totalfrank-personal: In which scenario we set reconcile to true and what does reconcile do? Reconcile everything, including all MCPs, of the projection blindly?

totalfrank:
**Review round 1 — no issues flagged.** Reviewed `f4226e81c` (merge-base `2befbc2e2`, 32 commits, 37 files). No bugs found, nothing disputed, nothing outstanding — so no push, and no further round is expected from this side.

State on `f4226e81c`:

- **CI green** — all 8 checks pass (Backend unit tests, Gateway, Engine, BaaS, BCS unit + e2e, Sandbox-proxy, Singlebox coverage).
- **No merge conflict** against `dev` (`bd98a4f52`).
- Ready for review, awaiting @FreddieSun and @xianmuyq.

One thing carried forward deliberately, so it is not mistaken for an oversight: **spec criterion 4 holds by half**. An MCP-only mutation now touches no symlinks and a skill-only mutation touches neither the device MCP calls nor the Passport, and the branch is on the caller's declared scope rather than on provider or engine type. The whole-artifact half does not hold — teclaw still recomposes per half, because the `apply_runtime_projection` it needs is corp-side work outside this repository. Its seam is now `SkillSetService.sync_mcp_projection`. Tracked as task 4.7 in `tasks.md`, under a heading marking it out-of-repo; it does not gate this PR.

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_management_service.py
```diff
@@ -670,6 +735,8 @@ async def _mutate(
         action: str,
         mutation,
         runtime_required: bool = True,
+        scope: ProjectionScope | None = None,
+        scope_from_result: Callable[[DesiredStateMutation], ProjectionScope] | None = None,
```
totalfrank-personal, line 709: Check all call sites, if none of them passes None here, let's not use | None.

Also, remove all if None checks below if there's any.

src/backend/src/agentclaw/community/core/skill_center/services/_mutation_flow.py
```diff
@@ -93,6 +113,7 @@ async def _project_or_compensate(
         engine_type: str | None,
         mutation: DesiredStateMutation,
         previous_mappings: Sequence[PoolSkillMapping],
+        scope: ProjectionScope = ProjectionScope.everything(),
```
totalfrank-personal: Seems like call site always provided scope field, so let's get rid of the default.

src/backend/src/agentclaw/community/core/skill_center/runtime_projection_contract.py
```diff
@@ -3,11 +3,67 @@
...
+    scope can only ever shrink there. That keeps a single-MCP mutation a
+    single device write, and stops a release from deleting a code the
+    default policy or a Skill dependency still supplies.
+    """
+
+    skills: bool = False
+    mcp: bool = False
+    claimed_mcp: frozenset[str] = frozenset()
+    released_mcp: frozenset[str] = frozenset()
+    reconcile: bool = False
```
totalfrank-personal: What is this for?

src/backend/src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py
```diff
@@ -105,43 +127,87 @@ async def project(
         bot_id: str,
         owner_id: str,
         retired_mappings: Sequence[PoolSkillMapping] = (),
+        scope: ProjectionScope = ProjectionScope.everything(),
```
totalfrank-personal: I do not like default value, it's fragile. Can you check if all call sites already passes it? If so, let's get rid of the default.

src/backend/src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py
```diff
@@ -351,30 +463,105 @@ async def _apply_skill_projection(
...
         engine: str,
         bot_id: str,
         owner_id: str,
         projection: RuntimeProjection,
         effective_cli_items: list[dict],
     ) -> None:
-        if not await service.sync_mcp_desired_state(
-            server_codes=set(projection.mcp_server_codes)
+        codes = set(projection.mcp_server_codes)
+        if scope.reconcile:
```
totalfrank-personal: Which scenario set reconcile to true?

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/skill_center/runtime_projection_contract.py
totalfrank: It decided which MCP codes count as *claimed* — and the fact that three comments on this PR ask what it is for means the name was doing none of that work. Renamed to `claim_all_mcp` in `bd2e812e3`, with a per-field docstring.

What it does: a caller that has no delta to name asks for every projected code to be treated as newly claimed, so configuration gets pushed for all of them. Only the callers who genuinely cannot name a delta set it — a device-activated restart, a Skill upload — because the projected set is not known until the projector resolves the plan, which is after the caller is done. That is why it is a flag rather than a code set.

Two things it deliberately does *not* do: it never releases (`released_mcp` is not consulted when it is set, so a reconcile can only add configuration), and it says nothing about which halves run — see the reply on the `everything()` thread, where that distinction turned out to matter.

---

src/backend/src/agentclaw/community/core/skill_center/runtime_projection_contract.py
totalfrank: Two callers set it, both of which have no mutation to describe: the device-activated listener in `skill_center_module.py` (a container came back and may hold no configuration at all) and `LocalSkillUploadService._sync_runtime`. Both now pass `ProjectionScope.everything()` explicitly rather than inheriting it from a default — see the thread on `bot_runtime_projector.py:130`.

"Blindly" is close but not quite: it pushes configuration for every projected code, but it never deletes. `released_mcp` is not consulted on that path, so the worst a reconcile can do is re-push configuration the device already had. That is the intended trade — on a path where the device's state is unknown, over-pushing is recoverable and under-pushing leaves an MCP whitelisted with no credentials.

Renamed to `claim_all_mcp` in `bd2e812e3` since three comments landed on this field.

---

src/backend/src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py
totalfrank: Same answer as the contract thread: the device-activated listener and the Skill upload path, both of which have no delta to name. Both now say so explicitly at the call site instead of inheriting a default.

Renamed to `claim_all_mcp` and the branch comment rewritten to name the two callers and state that nothing is released on that path — `bd2e812e3`.

---

src/backend/src/agentclaw/community/api/bot_runtime_projector.py
totalfrank: You're right, and this caught a real bug rather than just an odd default.

`ProjectionScope.everything()` sets `skills=True`, but `project_mcp_and_cli` never runs the Skill half — so the default was asserting something the method structurally ignores. Worse, I had added a `__post_init__` guard that *rejected* `ProjectionScope(mcp=True, claim_all_mcp=True)`, on the theory that a reconcile must cover both halves. That made the correct scope for this method unconstructible. The guard encoded a false invariant: `claim_all_mcp` answers "which MCPs count as claimed", and `skills`/`mcp` independently answer "which halves run".

Fixed in `bd2e812e3`: the guard is deleted, its test now pins the opposite (an MCP-only reconcile is a legal and used shape), and rather than give this method a different default I removed the default entirely — the sole caller, the cutover listener in `skill_center_module.py`, now passes `ProjectionScope(mcp=True, claim_all_mcp=True)` at the call site. Declaring the Skill half there would fight the cutover task that owns the mappings, which is the whole reason this entry point exists.

---

src/backend/src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py
totalfrank: Checked, and the premise didn't hold — six call sites were riding the default:

- `skill_center_module.py` × 2 (the device-activated listener's `project` and `project_mcp_and_cli`)
- `local_skill_upload_service.py` `_sync_runtime`
- `direct_activation_service.py` × 3 (direct skill activate/deactivate, direct MCP activate/deactivate)

Which is the argument for your point rather than against it: those are exactly the callers that would silently inherit a full reconcile, and nothing would say so. Default removed on `project`, `project_mcp_and_cli`, `project_for_cleanup`, and `_project_or_compensate` in `bd2e812e3`; all six now declare a scope explicitly. Same behaviour, no longer implicit.

One note on the direct-activation three: I gave them `ProjectionScope.everything()` to preserve today's behaviour exactly. They *could* be narrowed the way the SkillSet commands were, but their repository commands (`install_skill` / `install_mcp` and friends) don't return MCP codes on the mutation result, so it needs the same treatment `add_skill` got — out of scope here, happy to do it separately.

---

src/backend/src/agentclaw/community/core/skill_center/services/_mutation_flow.py
totalfrank: Correct — `apply` is the only caller and it always passes one. Default removed in `bd2e812e3`; `scope` is now required on `_project_or_compensate`.

`apply` itself keeps both parameters optional, but not as a "forgot to say" default: it now raises when given both or neither, so exactly one is required. The two forms exist because some commands know their scope up front (`add_mcp` holds the code it claimed) and others only learn it from the mutation result under the row lock (`activate`, and the Skill commands reading `mcp_dependencies`).

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_management_service.py
totalfrank: Checked all eleven `_mutate` call sites: every one passes exactly one of `scope` / `scope_from_result`, and none passes `None`.

They can't collapse to a single non-optional parameter, though, because the two forms carry different information. `add_mcp` / `remove_mcp` know their scope before the mutation runs. `activate`, `deactivate`, `legacy_activate`, and the four Skill commands only learn theirs *from the mutation result* — the MCP codes are read under the row lock the repository already holds, precisely so the scope names what was actually installed rather than what a second unlocked query happened to see.

So instead of dropping `| None`, the enforcement moved one layer down where it can be exact: `MutationProjectionFlow.apply` now raises when given both or neither. `_mutate` just forwards whichever its caller supplied, and there are no `if None` checks left on this path — the old `scope if scope is not None else ProjectionScope.everything()` fallback in `apply` is gone, which was the actual fragile bit. `bd2e812e3`.

---

src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
totalfrank: Yes — always wired by DI. Required now, and the `if caller_identity_repo is None:` branch is deleted. `bd2e812e3`.

Being straight about what that cost, since it went further than this file: the `= None` existed only so the 25 test fixtures that build `BotService` directly wouldn't need updating. Leaving `BotService._caller_identity_repo` optional while the strategy required a non-None would just have moved the type hole, so I made it required there too — thirteen fixtures each gained one `caller_identity_repo=MagicMock()` line. That seems right by your own argument: a `BotService` that can't refresh its bot's Passport scope shouldn't be constructible.

The remaining decline path is narrower and stays: if the repository is present but the row is *unreadable*, the refresh is skipped and the bot's existing scope stands, rather than being replaced with an owner-only snapshot. Three tests pin it (unreadable row, missing bot primary key, and a caller-mode MCP surviving a refresh intact).

---

src/backend/src/agentclaw/community/core/bot_management/engines/default.py
totalfrank: Done — required, no default. `bd2e812e3`.

This one is the no-op implementation, so it ignores the argument either way; making it required keeps it honest against the protocol rather than advertising a tolerance the real strategy no longer has.

---

src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
totalfrank: Done — required on the protocol, so both implementations and the `BotService` call site must supply it. `bd2e812e3`.

The docstring paragraph that justified the `None` default ("defaulted so engines with no external authorization are unaffected") is replaced: an engine with no external authorization simply ignores the argument, which never needed a default to express. What it says now is why it must not be optional — an accidental omission would be a silent privilege change rather than a type error.

---

totalfrank:
**Review round 2 — all 10 comments addressed in `bd2e812e3`.** Nothing disputed; every thread replied to and resolved.

**One was a real bug, not a style point.** `project_mcp_and_cli` defaulted to `ProjectionScope.everything()`, which sets `skills=True` on a method that structurally ignores the Skill half — and a `__post_init__` guard I had added *rejected* the scope that method actually needs (`ProjectionScope(mcp=True, claim_all_mcp=True)`). The guard encoded a false invariant. It's deleted, and its test now pins the opposite.

**The defaults are gone, and the premise for removing them was stronger than assumed.** Six call sites were riding the `scope` default rather than "all of them already pass it": both device-activated listeners, the Skill upload path, and the three direct-activation commands. They now declare `ProjectionScope.everything()` explicitly — same behaviour, no longer inherited by accident.

**`reconcile` → `claim_all_mcp`.** Three comments asked what the field was for, which is the name failing rather than the docs. It answers *which MCPs count as claimed*, never *which halves run*; each field now carries its own docstring.

**`caller_identity_repo` is required end to end**, and the `if caller_identity_repo is None` branch is deleted. That went past the three files flagged: leaving `BotService` optional while the strategy required non-None would only move the type hole, so it's required there too — thirteen fixtures gained one `MagicMock()` line each. The narrower decline path stays: a repository that is *present but unreadable* skips the refresh so the bot's existing scope stands, rather than being replaced with an owner-only snapshot.

Full backend suite: **14747 passing**, 58 skipped. The two `test_bot_build_service_skill_artifact` failures need `rsync` on `PATH` and fail identically without this diff.

One thing I have not done, flagged rather than silently skipped: the three direct-activation commands could be *narrowed* the way the SkillSet commands were, but their repository commands return no MCP codes on the mutation result, so it needs the same treatment `add_skill` got. Happy to do it here if you want it, or separately.

---

src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
totalfrank: Fixed in `bd2e812e3`: `caller_identity_repo` on `refresh_restart_authorization` is now a required parameter (no `= None` default), and the `if caller_identity_repo is None:` branch below it is gone — confirmed by reading the current signature and body.

---

src/backend/src/agentclaw/community/core/bot_management/engines/default.py
totalfrank: Fixed in `bd2e812e3`: `caller_identity_repo` here is now `object` with no default, matching the abstract method's now-required parameter — confirmed by reading the current signature.

---

src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py
totalfrank: Fixed in `bd2e812e3`: `caller_identity_repo` is now required here too, with the docstring updated to say why (a strategy that republishes an overwrite-style MCP scope needs it; omission is now a type error, not a silent runtime branch) — confirmed by reading the current signature and docstring.

---

src/backend/src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py
totalfrank: Fixed in `bd2e812e3`: `scope: ProjectionScope` is now required (no default) on `project`, `project_mcp_and_cli`, and `project_for_cleanup` here — confirmed by reading the current signatures. Every call site was updated to declare it explicitly (`ProjectionScope.everything()` for the reconcile-only paths, a concrete scope everywhere else).

---

src/backend/src/agentclaw/community/core/skill_center/services/_mutation_flow.py
totalfrank: Checked this in `bd2e812e3`. The premise doesn't quite hold — no call site provides both `scope` and `scope_from_result`; each of the eleven `_mutate` call sites in `skill_set_management_service.py` supplies exactly one and leaves the other as an implicit `None`, so `apply()`'s signature can't drop `| None` on either without splitting it into two methods.

What did change, and is what I think this comment was really after: `apply()` now raises `ValueError` when `scope`/`scope_from_result` are both given or both omitted (`if (scope is None) == (scope_from_result is None): raise ...`), so the old silent fallback to a full reconcile is gone — a caller that forgets to pass either now fails loudly instead of inheriting the expensive default. `_project_or_compensate`'s `scope` parameter, which every internal caller does always supply, did lose its default (`ProjectionScope = ProjectionScope.everything()` → `ProjectionScope`).

Resolving as addressed via the validation guard rather than the literal type-signature change; let me know if you'd rather see the pair split into two explicit methods.

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_management_service.py
totalfrank: Checked this in `bd2e812e3`. Same situation as the `_mutation_flow.py` thread on `apply()`: the premise doesn't quite hold for `_mutate`'s `scope`/`scope_from_result` pair either — each of the eleven call sites below passes exactly one of the two, leaving the other as an implicit `None`, so neither can drop `| None` without splitting `_mutate` into two entry points. `bd2e812e3` added a docstring saying so explicitly rather than changing the signature: "Both scope arguments stay optional *here* only because each command supplies whichever one it can... The flow enforces that; neither is a 'forgot to say' default." The enforcement referred to is `MutationProjectionFlow.apply`'s new `ValueError` guard on this same pair (see the reply on the `_mutation_flow.py` thread), which is what actually eliminates the silent-default risk this comment was flagging.

Resolving as addressed via the docstring + the flow's validation guard rather than the literal type-signature change; happy to split it into two call paths if you'd still prefer that.

---

totalfrank (commented):
Reviewed commit `bd2e812e3` (the only commit since the last full review at `f4226e81c`) and closed out the remaining open review threads.

**Commits reviewed this round:** `bd2e812e3` — removes the `None`/default fallbacks the earlier round asked about (`ProjectionScope.__post_init__` guard deleted and renamed `reconcile` → `claim_all_mcp`; `scope` required on `project`/`project_mcp_and_cli`/`project_for_cleanup`/`_project_or_compensate`; `caller_identity_repo` required through `strategy.py`/`default.py`/`provisioning.py`/`BotService`), plus matching call-site and test updates.

**Findings checklist (all 10 threads from the human review round):**
1. `bot_runtime_projector.py` (api) — default scope for `project_mcp_and_cli` — **fixed**, resolved
2. `strategy.py` — `caller_identity_repo` default/None-check — **fixed**, resolved
3. `default.py` — same — **fixed**, resolved
4. `provisioning.py` — same — **fixed**, resolved
5. `runtime_projection_contract.py` — "which scenario sets reconcile" — **addressed** (renamed to `claim_all_mcp`, documented), resolved
6. `skill_set_management_service.py:739` — "drop `| None`" — **addressed differently**: the premise doesn't hold (every call site leaves exactly one of `scope`/`scope_from_result` implicitly `None`), so the pair can't lose `| None` without a signature split; the underlying fragile-default risk is closed by `MutationProjectionFlow.apply`'s new `ValueError` guard instead. Resolved with that explanation; open to a follow-up if the two-method split is still wanted.
7. `_mutation_flow.py` — "get rid of the default" — same situation and same resolution as #6.
8. `runtime_projection_contract.py` — "what is this for" — **addressed** (renamed to `claim_all_mcp`, per-field docstrings added), resolved
9. `services/bot_runtime_projector.py` — "I don't like default value" — **fixed** (`scope` now required on all three entry points), resolved
10. `services/bot_runtime_projector.py` — "which scenario reconcile true" — **addressed** (same rename/docs as #5/#8), resolved

**Still open: 0.**

New-commit review (Part B): read the full `f4226e81c → bd2e812e3` diff (12 source files, 14 test files). All of it is the mechanical default-removal described above — required-parameter changes, matching call-site updates (`ProjectionScope.everything()` / `ProjectionScope(mcp=True, claim_all_mcp=True)` declared explicitly at the four reconcile-only call sites), and the corresponding test/fixture updates, including one stale test deleted (`test_skip_when_no_identity_repository_is_supplied`, for a branch that no longer exists) and one renamed to match (`test_a_reconcile_scope_cannot_leave_a_half_uncovered` → `test_claiming_every_code_is_independent_of_which_halves_run`). No correctness, error-handling, concurrency, or architecture-boundary issues found. No new findings.

The PR is clean as of this round.

---

totalfrank:
**Review round 3 — no issues flagged.** Reviewed `bd2e812e3`. No new findings, all 10 threads from the human round closed, 0 open.

State on `bd2e812e3`:

- **CI green** — all 8 checks pass (Backend unit tests and Singlebox coverage were the last two, at 04:36Z and 04:39Z).
- **No merge conflict** against `dev` (`2888ec691`).

Two things I'd rather leave on the record than have read as oversights:

1. **Threads 6 and 7 were addressed differently from the literal ask.** The request was to drop `| None` from `scope` / `scope_from_result`. That isn't possible without splitting the method: every one of the eleven `_mutate` call sites passes exactly one of the pair and leaves the other implicitly `None`. What was actually fragile — `apply()` silently falling back to a full reconcile when neither was given — is now a `ValueError`. If the two-method split is still preferred, say so and I'll do it.

2. **Criterion 4 still holds by half.** Unchanged from round 1: teclaw recomposes per half because its `apply_runtime_projection` is corp-side work outside this repository. Tracked as task 4.7 in `tasks.md` under an out-of-repo heading; it does not gate this PR.

Waiting on @FreddieSun and @xianmuyq.

---

src/backend/src/agentclaw/community/core/skill_center/services/direct_activation_service.py
```diff
@@ -125,6 +126,13 @@ async def _set_skill_active(
                     engine_type=bot_engine_type(bot),
                     default_engine_types=bot_default_engine_types(bot),
                 ),
+                # Direct activation is the pre-SkillSet path and its
+                # repository commands return no MCP codes, so it has no delta
+                # to narrow with — it declares the whole projection, which is
+                # exactly what it did before scopes existed. Narrowing it
+                # needs the same treatment the Set commands got and is not
+                # this PR's to do.
+                scope=ProjectionScope.everything(),
```
totalfrank-personal: It does not make sense to set scope to be everything here right? This is for activating/deactivating a skill. I mean, the scope should be only about skill right?

src/backend/src/agentclaw/community/core/skill_center/services/direct_activation_service.py
```diff
@@ -225,6 +236,9 @@ async def deactivate_mcp(
                 engine_type=bot_engine_type(bot),
                 default_engine_types=bot_default_engine_types(bot),
             ),
+            # Declared explicitly to preserve today's behaviour exactly; see
+            # the note on ``_set_skill_active``.
+            scope=ProjectionScope.everything(),
```
totalfrank-personal: It does not make sense to set scope to be everything here right? This is for deactivating a mcp. I mean, the scope should be only about mcp right?

src/backend/src/agentclaw/community/core/skill_center/services/direct_activation_service.py
```diff
@@ -201,6 +209,9 @@ async def activate_mcp(
...
```
totalfrank-personal: It does not make sense to set scope to be everything here right? This is for activating a mcp. I mean, the scope should be only about mcp right?

src/backend/src/agentclaw/community/core/skill_center/services/local_skill_upload_service.py
totalfrank-personal, line 103: Can you confirm if the only place other than tests calls this is upload_local_skill_files method in this file, which itself is not called anywhere other than tests. If that's the case, lets delete this path and all related tests.

src/backend/src/agentclaw/community/di/modules/skill_center_module.py
totalfrank-personal, line 884: Is this actually used anywhere? I can not find what component uses it. If not, let's delete this provider and the SkillSymlinkListener class as well.

Once it's deleted, clean up everything referenced by this SkillSymlinkListener as long as there's no other component depends on it.

totalfrank-personal (commented):
One more round code review, please take a look.

src/backend/src/agentclaw/community/core/skill_center/services/direct_activation_service.py
totalfrank: Agreed — narrowed in `e811b47e5`:

```python
scope=ProjectionScope(mcp=True, claimed_mcp=frozenset({server_code})),
```

MCP half only, and the one code rather than the whole projected set. A test pins that `skills is False` and `claimed_mcp == {server_code}`.

---

src/backend/src/agentclaw/community/core/skill_center/services/direct_activation_service.py
totalfrank: Agreed — narrowed in `e811b47e5`:

```python
scope=ProjectionScope(mcp=True, released_mcp=frozenset({server_code})),
```

Worth noting the release is still a *candidate*, not a deletion order: the projector subtracts the projected set before removing anything, so if the default policy or a Skill dependency still supplies that code, its device configuration survives. A test pins `skills is False` and `released_mcp == {server_code}`.

---

src/backend/src/agentclaw/community/core/skill_center/services/direct_activation_service.py
totalfrank: Narrowed in `e811b47e5` — but not to Skill-only, and the reason is worth stating because it's the same trap `add_skill` hit.

A Skill's `mcp_dependencies` join the Bot's projected MCP set along with the Skill (`RuntimeProjectionResolver` folds them into `mcp_server_codes`). So `ProjectionScope(skills=True)` on a Skill that declares dependencies would skip the MCP half and leave those codes whitelisted on the device with no configuration pushed for them — an MCP the bot is allowed to call but has no endpoint or `api_key` for.

So it scopes from the mutation result instead, reusing the builders the Set commands already use:

```python
scope_from_result=(skill_claim_scope if active else skill_release_scope),
```

`mcp` follows the dependencies: a dependency-free Skill *is* Skills-only and skips the MCP half entirely, which is your point holding in the common case. `install_skill` / `uninstall_skill` now return those codes on the mutation result, read under the row lock they already hold — the exact treatment `add_skill` got, and the thing I'd recorded on the earlier thread as the blocker for narrowing this path.

`skill_claim_scope` / `skill_release_scope` moved from `skill_set_management_service`'s private namespace into `_mutation_flow`, which both services already import, rather than one service reaching into the other's privates.

Four tests pin it: MCP-only claim, MCP-only release, dependency-free Skill leaving the MCP half alone, and a dependency-carrying Skill claiming/releasing its codes.

---

src/backend/src/agentclaw/community/di/modules/skill_center_module.py
totalfrank: Checked, and I'd push back on this one — `SkillSymlinkListener` is live. Deleting it would break symlink sync on device activation.

It has no direct callers because it is **event-driven**, not called. It subclasses `LifecycleBase` and subscribes itself:

```python
# skill_symlink_listener.py — startup()
bus = get_event_bus()
bus.subscribe(DeviceActivatedEvent, self.handle)
```

And `DeviceActivatedEvent` is published on a hot path — `DeviceService.report_device_alive` (`device_service.py:1128`) fires it every time a device reports alive:

```python
event = DeviceActivatedEvent(device_id=..., binding_id=record.id, ...)
get_event_bus().publish(event)
```

Three more things corroborate it rather than just the grep:

- `tests/community/architecture/test_lifecycle_discovery.py:38` lists `SkillSymlinkListener` as a Phase 2 self-subscribing lifecycle component — an architecture test that would fail if the class vanished.
- `desktop_bot_service.py:1442` documents the flow in a comment: *发布 DeviceActivatedEvent → SkillSymlinkListener → 软链同步*.
- `tests/community/api/test_event_listener_registration.py` and `tests/community/di/test_skills_pool_wiring.py` both resolve it from the injector, i.e. they assert the provider you're looking at is wired.

So the provider is what makes the listener exist at all; without it nothing constructs the object that subscribes to the bus. Leaving both in place. Happy to be overruled if you know the event path is itself dead, but I don't think it is — `report_device_alive` is the agentbox/baas device callback.

---

src/backend/src/agentclaw/community/core/skill_center/services/local_skill_upload_service.py
totalfrank: Can't confirm — the premise doesn't hold. `upload_local_skill_files` has a production caller: it backs a live public API route.

`adapters/http/openapi_v1/skills/router.py:520`, inside `upload_skill_folder`:

```python
@router.post("/upload-folder", ...)
async def upload_skill_folder(...):
    ...
    result = await upload_service.upload_local_skill_files(
        bot_id=bot_id, owner_id=owner_id, actor_id=user_id, files=uploaded,
    )
```

That is `POST /openapi/v1/.../upload-folder` — the browser directory-upload endpoint, and the method is declared on the Service API protocol (`api/local_skill_upload_service.py:17`) precisely so the adapter can reach it through `Injected(LocalSkillUploadServiceProtocol)`.

So deleting the path would break that endpoint. Leaving it in place.

If the intent was narrower — that the *upload* path shouldn't be doing a full `ProjectionScope.everything()` reconcile — that's a fair question and a different change. My read is that it should stay `everything()`: an upload adds Skills whose `mcp_dependencies` it doesn't enumerate, and it has no mutation result to scope from, so it is genuinely one of the callers with no delta to name. Say the word if you'd like me to look at giving it a real delta.

---

totalfrank:
**Review round 4 — 3 of 5 fixed in `e811b47e5`, 2 disputed with evidence.** The two disputes are left open for you rather than resolved.

**Fixed — direct activation now declares the half it changes.** This is the follow-up flagged last round as blocked; it's unblocked now.

| command | scope |
| --- | --- |
| `activate_mcp` | `mcp=True, claimed_mcp={server_code}` |
| `deactivate_mcp` | `mcp=True, released_mcp={server_code}` |
| `activate_skill` / `deactivate_skill` | `skills=True`, plus the MCP half **only when the Skill declares dependencies** |

The Skill pair is the one that couldn't be Skill-only. A Skill's `mcp_dependencies` join the projected MCP set along with it, so a flat `ProjectionScope(skills=True)` would whitelist a dependency the device is never configured for. `install_skill` / `uninstall_skill` now return those codes on the mutation result under the row lock they already hold — the same treatment `add_skill` got, which is exactly what I'd recorded as the blocker. Five tests pin the new scopes; all five fail if the narrowing is reverted.

**Disputed — both deletion targets are live.**

1. **`SkillSymlinkListener` is not dead.** It has no direct callers because it's event-driven: it subclasses `LifecycleBase` and calls `bus.subscribe(DeviceActivatedEvent, self.handle)` in `startup()`. `DeviceService.report_device_alive` publishes that event on every device-alive callback. `test_lifecycle_discovery.py` also asserts its presence as a Phase 2 self-subscribing component. Deleting the provider would stop the object being constructed at all, and symlink sync on device activation would silently stop.

2. **`upload_local_skill_files` has a production caller.** `POST /openapi/v1/.../upload-folder` → `upload_skill_folder` → `upload_service.upload_local_skill_files` (`skills/router.py:520`). It's declared on the Service API protocol precisely so the adapter can reach it.

Both are the same shape of miss — a grep for direct callers finds nothing because the caller is an event bus in one case and an HTTP route in the other. Happy to be overruled if you know either entry point is itself dead, but I don't think they are.

Full backend suite: **14752 passing**, 58 skipped, only the two known `rsync` failures.

---

src/backend/src/agentclaw/community/core/skill_center/services/direct_activation_service.py
totalfrank: Agreed — narrowed in `e811b47e5`: `activate_skill`/`deactivate_skill` now pass `scope_from_result=(skill_claim_scope if active else skill_release_scope)` instead of `ProjectionScope.everything()`. Those helpers (moved to `_mutation_flow.py`, shared with `skill_set_management_service.py`'s Skill commands) build `ProjectionScope(skills=True, mcp=bool(result.mcp_codes), claimed_mcp=...)` — Skills-only when the Skill has no MCP dependencies, scoped to just the dependency codes otherwise.

`DirectInstallationCommands.install_skill`/`uninstall_skill` now return those codes on `DesiredStateMutation.mcp_codes`, read under the row lock, via the same `skill_mcp_dependency_codes` decoder `add_skill` uses (and which already has direct unit coverage for the `None`/malformed-row cases in `test_mcp_dependency_scope.py`). Four new tests in `test_direct_activation_service.py` pin the service-level wiring: a dependency-free skill touches no MCP half, and a skill declaring `mcp.weather` as a dependency claims/releases exactly that code on activate/deactivate.

---

totalfrank (commented):
Reviewed commit `e811b47e5` (the only commit since the last full review at `bd2e812e3`), and closed out the review threads it addresses.

**Commits reviewed this round:** `e811b47e5` — narrows `direct_activation_service.py`'s three remaining `ProjectionScope.everything()` call sites (`activate_skill`/`deactivate_skill`, `activate_mcp`, `deactivate_mcp`) to the half each actually changes, moves `skill_claim_scope`/`skill_release_scope` into the shared `_mutation_flow.py`, and has `DirectInstallationCommands.install_skill`/`uninstall_skill` return their Skill's MCP-dependency codes on `DesiredStateMutation.mcp_codes` (read under the row lock, via the same decoder `add_skill` uses).

1–10. Human review round 1 (10 threads on defaults/`None`-checks/naming) — all fixed or addressed in `bd2e812e3`, resolved last round.
11. `direct_activation_service.py` — `activate_skill`/`deactivate_skill` scope should be Skill-only (+MCP dependencies) — **fixed** in `e811b47e5`, resolved this round.
12. `direct_activation_service.py` — `deactivate_mcp` scope should be MCP-only — **fixed** in `e811b47e5` (`ProjectionScope(mcp=True, released_mcp={server_code})`), resolved this round.
13. `direct_activation_service.py` — `activate_mcp` scope should be MCP-only — **fixed** in `e811b47e5` (`ProjectionScope(mcp=True, claimed_mcp={server_code})`), resolved this round.
14. `local_skill_upload_service.py:103` — confirm `upload_local_skill_files` has no non-test callers and delete it if so — **not addressed** by this commit, still open (no claim it was fixed; not re-filing).
15. `skill_center_module.py:884` — confirm the `SkillSymlinkListener` binding is unused and delete it if so — **not addressed** by this commit, still open (no claim it was fixed; not re-filing).

**Still open: 2** (#14, #15 — both pending author investigation/confirmation, unrelated to this round's diff).

New-commit review (Part B): read the full `bd2e812e3 → e811b47e5` diff (4 source files, 1 test file). The three scope narrowings match the pattern already used by `add_mcp`/`remove_mcp`/`add_skill` — `claimed_mcp`/`released_mcp` are candidates the projector intersects with the resolved set, never a raw grow. Confirmed no stale references to the old private `_skill_claim_scope`/`_skill_release_scope` names remain anywhere in the tree after the move to `_mutation_flow.py`. `skill_mcp_dependency_codes(skill)` in the repository already handles `skill is None` (the "uninstall raced the row" case) without special-casing at the call site. New tests (`test_direct_activation_service.py`) pin all four scope shapes (MCP-only claim/release, Skill-only, Skill+dependency claim/release) by recording the `ProjectionScope` each command hands the projector. No correctness, error-handling, concurrency, or architecture-boundary issues found. No new findings.

---

totalfrank (commented):
Reviewed commit `d0cfb0ee9` (the only commit since the last full review at `e811b47e5`).

**Commits reviewed this round:** `d0cfb0ee9` — narrows the Local Skill *replace* flow's runtime sync (`LocalSkillUploadService._sync_runtime`, used by both the replace itself and its compensating restore) from `ProjectionScope.everything()` to `ProjectionScope(skills=True)`, updates `claim_all_mcp`'s docstring/spec now that it has exactly one caller, and corrects a spec claim about `_sync_mcps_when_device_active`.

Verified against the code, not just the commit message:
- `replace_bot_local_skill` (repository) writes only `description`, `user_id`, `gmt_modified` and refuses to change `git_path` — it never touches `mcp_dependencies`, so the projected MCP set is provably unchanged across a replace. `scope=ProjectionScope(skills=True)` is correct.
- Both `_sync_runtime` call sites (the replace path and `_restore_replacement`) go through the same method, so the narrowing covers the compensating path too, and the new test (`test_a_local_skill_replacement_never_touches_the_mcp_half`) pins all five scope fields on the forward path.
- `claim_all_mcp=True` / `ProjectionScope.everything()` now has exactly one production caller left: `SkillSymlinkListener`'s two closures in `skill_center_module.py` (device-activated reconcile), matching the updated docstring.
- `_sync_mcps_when_device_active` has zero production callers (grep confirms only the definition remains) — the spec correction is accurate.

No correctness, error-handling, concurrency, or architecture-boundary issues found. No new findings.

1–13. Human review rounds 1–3 (defaults/`None`-checks, `ProjectionScope` naming, direct-activation scoping) — all fixed, addressed, or conceded with explanation; all resolved.
14. `local_skill_upload_service.py:103` — confirm `upload_local_skill_files` has no non-test callers — **disputed and resolved**: it backs a live public API route (`upload_skill_folder` → `POST /openapi/v1/.../upload-folder`), so the method stays.
15. `skill_center_module.py:884` — confirm `SkillSymlinkListener` binding is unused — **disputed and resolved**: it is event-driven (subscribes to `DeviceActivatedEvent`, published on every `report_device_alive`), corroborated by an architecture test and two DI-wiring tests; the binding stays.

**Still open: 0.**

The PR is clean as of this round.

---

FreddieSun (approved):
LGTM

totalfrank:
**Review rounds 5 and 6 — no issues flagged.** Round 5's all-clear was deliberately held: a design question about `claim_all_mcp` was live and would have made it stale within minutes. Both rounds are recorded here.

- **Round 5** (`e811b47e5`) — no new findings on the direct-activation narrowing.
- **Round 6** (`d0cfb0ee9`) — no new findings, and both earlier disputes accepted. **0 open threads.**

State on `d0cfb0ee9`:

- **CI green** — all 8 checks (Singlebox coverage last, 06:16Z).
- **No merge conflict** against `dev` (`1ae842d45`).
- **Approved** by @FreddieSun.

**One correction landed in this stretch, and it's worth surfacing rather than leaving in a diff.** The spec claimed device restart was covered independently by `DeviceService._sync_mcps_when_device_active`. That is **false on `dev`** — the method has zero production callers, and its tests mock it out deliberately: `test_pending_activation_has_one_mcp_writer_owned_by_device_event` asserts it must *not* run, because ownership moved to the event path.

That inverts the conclusion. `SkillSymlinkListener`'s projection is the **only** production path writing per-MCP configuration to a newly active container, so its whole-set claim is load-bearing rather than duplicative. `spec.md` now carries the correction inline where the wrong claim was made.

The same investigation tightened the Local Skill replace flow the other way: a replace provably cannot move the MCP set (`replace_bot_local_skill` writes description/user_id/gmt_modified and refuses to change `git_path`; nothing rescans `mcp_dependencies`), so it declares `ProjectionScope(skills=True)` and skips the MCP half entirely. `claim_all_mcp` is now down to exactly one caller, documented with the condition that makes it necessary.

**Still open and unchanged:** spec criterion 4 holds by half — teclaw recomposes per half, pending corp-side `apply_runtime_projection`. Tracked as task 4.7 under an out-of-repo heading; it does not gate this PR.

---

PR #1569, opened 2026-08-26
fix(backend): align group bot chat access with legacy logs

3h-william:
## Problem

  Group Bot Workshop logs can contain traces generated by multiple bots.
  The existing detail access check was still bot-scoped, so a user entering
  through Bot A could see a trace from Bot B in the same group but receive
  404/not accessible when opening its detail.

  ## Solution

  - Aggregate traces by the group's `bcs_group_sessions` relationships.
  - Ignore the entry `bot_id` as a Group aggregation filter.
  - Allow users who own or collaborate on any bot in the group to view traces
    from other bots in that group.
  - Keep the existing owner/collaborator access rules for traces that cannot
    be resolved to a group.
  - Preserve the existing Legacy Bot Chats API contract.
  - Keep the Backend implementation independent from the Bot-scoped OpenAPI
    logs endpoints.

  ## Validation

  - 148 backend tests passed.
  - Backend pre-push SAST/lint gate passed.
  - Covered Group aggregation, cross-bot detail access, collaborator access,
    legacy trace fallback, pagination, and database log source behavior.

  ## Compatibility and risk

  - Non-group Bot Chat access behavior remains unchanged.
  - No database schema migration is introduced.
  - The Backend spec file was intentionally excluded from this branch because
    it should not be submitted in this MR.
  - Rollback can be performed by reverting the commit.

  ## Related

  - New frontend branch:
    `fix/bot-chats-gateway-clean_shanzong`

### DIFF
--- a/docs/superpowers/specs/2026-08-24-bot-chat-group-trace-aggregation-design.md
+++ b/docs/superpowers/specs/2026-08-24-bot-chat-group-trace-aggregation-design.md
@@ -1,362 +1,237 @@
-# Bot Chats 群组 Trace 聚合与跨 Bot 详情访问规范
+# Bot Chats 老接口按群 ID 聚合 Trace
 
 - **日期**：2026-08-24
 - **状态**：approved
-- **涉及模块**：Avernet Backend Bot Chat、新 TeamClaw 前端、TeamClaw Gateway 路由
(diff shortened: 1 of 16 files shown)


FreddieSun (approved):
LGTM

PR #1572, opened 2026-08-26
feat(backend): admit delegated application callers to public harness APIs

wienyang:
## Problem

  The public harness API (`/openapi/v1/bots/{bot_id}/harness/*`) was the only backend surface that still refused application-only callers. The rest of the public API has been unified on an addressed-bot identity model where:

  - `bot_id` is carried on the path
  - `user_id` is carried in the query string
  - the caller is either a human or an application acting under a live bot-app grant

  Harness was left out of that unification, so both the collaborator path and the delegated-application path were unavailable for it.

  ## Solution

  Move the six public harness operations from `AdmissionMode.REFUSED` to `AdmissionMode.GRANT_CHECKED_ADDRESSED_BOT` and perform the grant check inside the existing `require_harness_bot_access` dependency.

  Key points:

  - The wire contract is unchanged: callers still send `bot_id` on the path and `user_id` as the only query parameter; no `owner_id` parameter is introduced.
  - For human callers, behavior is unchanged.
  - For application-only callers, `require_harness_bot_access` now looks up a live grant for `(app_id, bot_id, owner_id, user_id)` and then applies the same owner/collaborator check as for humans.
  - The owner is resolved from the repository record, preserving the current harness contract.
  - The gateway `route_security` entry is widened from `user: required` to `user: optional + app: optional` so the application identity reaches the backend.
  - A new `HARNESS_SCOPED_OPERATIONS` set documents that harness resolves its own grant check and is excluded from the normal structural grant-dependency assertion, mirroring the existing `SKILL_SCOPED_OPERATIONS` mechanism.

  ## Validation

  - `test_admission_inventory.py` — 14 passed
  - `test_app_only_refusals.py` — 30 passed
  - `test_self_checked_routes_refuse.py` — 14 passed (includes the six harness routes)
  - `test_endpoint_runner.py -k harness` — 18 passed (includes new app-only granted/refused cases)
  - `test_app_only_listings.py` + `test_skills_shared_bot_grant.py` — 114 passed combined
  - `src/gateway/tests/unit/core/authn/test_route_security.py` — 60 passed

  The local pre-push gate passed in lint-only mode.

  ## Compatibility and risk

  - Existing human callers (owner or collaborators) are unaffected: the request shape, status codes, and responses are unchanged.
  - Application callers that previously received a 401/404 are now admitted if they hold a valid grant.
  - No `owner_id` query parameter is added, so this deliberately does not fully migrate harness to the standard `engine_runtime` addressed-bot shape; that broader migration is left for a later change if needed.
  - The harness owner/collaborator gate keeps its existing ADMIN-level requirement.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
@@ -528,33 +528,33 @@ ADMISSION: dict[tuple[str, str], AdmissionMode] = {
     ("GET", "/openapi/v1/bots/{bot_id}/models/{model_id:path}"): AdmissionMode.GRANT_CHECKED_ADDRESSED_BOT,
     ("GET", "/openapi/v1/bots/{bot_id}/connection"): AdmissionMode.GRANT_CHECKED_ADDRESSED_BOT,
     # harness — bot-scoped diagnostics and patching under the addressed bot.
-    # These operations are intentionally user-only for now: harness access is
-    # checked against the verified user's owner/collaborator relationship, and
-    # app-only delegation is not part of this public contract.
+    # These operations now go through the addressed-bot grant seam so that both
+    # human collaborators and delegated application callers can reach the bot,
+    # with the owner resolved by the harness access dependency itself.
     (
         "POST",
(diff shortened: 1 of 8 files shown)


totalfrank (approved):
LGTM.

PR #1577, opened 2026-08-27
feat: implement ACK deploy 

xxxxpenny:
- Implement AckDeployConfigComposer: build_start_command generates single nohup start_service.sh with {token}/{client_id} placeholders and ctx-sourced bot_id/owner_id; build_mount_points returns empty list; build_storage returns NAS volume with per-bot storage_id.

- Make CommunityOutboundRuleProvider config-driven: add OutboundRulesConfig and OutboundRuleEntryConfig dataclasses, load rules from outbound_rules YAML block, construct HeaderOperationRule list at init time.

- Simplify CommunitySecretResolver.get_secret: use secret_name directly as env var (no _VALUE/_USER suffix, no uppercasing), return None when absent.

- Add aiomysql and greenlet runtime dependencies for async MySQL support.

- Disable skill_scan in community overlay (no scanner SDK available) and guard startup() against start() returning False.

- Update tests to match all behavior changes.

### DIFF
--- a/src/backend/pyproject.toml
+++ b/src/backend/pyproject.toml
@@ -79,6 +79,14 @@ dependencies = [
   # before querying the public Bot name/owner-name LIKE catalogue. This must be
   # installed in every runtime profile, not only in test tooling.
   "jieba>=0.42",
+  # aiomysql: the async DBAPI behind `mysql+aiomysql://` — needed when the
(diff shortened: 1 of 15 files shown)


totalfrank (approved):
LGTM.

PR #1584, opened 2026-08-27
feat(task): add map ClawMind callback status to task enum and bot token provider

jiangj0627:
## Problem
  - ClawMind callbacks return `flow_runs.status` as one of 7 enums (`running` / `succeeded` / `failed` / `cancelled` / `waiting` / `aborted` / `blocked`), but `task_callback.status` was being stored as
  the **raw ClawMind string** with no mapping to the task `Status` enum (`PENDING`/`PLANNING`/`RUNNING`/`DONE`/`FAILED`/`HUNG`/`CANCELLED`), and the mapping table was missing `blocked` entirely.
  - BCS group creation (`POST /groups`) with inline `event_subscriptions` hits `require_human`, which rejects a Bot caller token (401). The driver-bot `session_token` needs to ride along as
  `Authorization: Bearer` so BCS resolves the caller as the driver/originator bot (with ownership / a single caller identity). The token-read logic (direct `SELECT … FROM bcs_bots`) was sitting in
  `community/plugins`, i.e. the open-source kernel was hard-coding a corp table schema — a layering leak.
  - `task_executor` needed task-type branching; the QC state-machine e2e suite had outgrown a single file.

  ## Solution
  - **ClawMind callback status mapping** (`translator.translate_claw_mind`): `flow_runs.status` (falling back to `node_executions[0].status`, then top-level) is now mapped through
  `_claw_mind_status_to_task(...)` into the task `Status` enum and stored as `task_callback.status` — `running→RUNNING`, `succeeded→DONE`, `failed→FAILED`, `cancelled→CANCELLED`, `aborted→CANCELLED`,
  `waiting→PENDING`, `blocked→PENDING`; added the missing `blocked` entry. `execution_graph` already used the same table, so graph/node status stays consistent.
  - **`BcsBotTokenProvider` port** (`core/task/task_runner/integration/bcs_bot_token_provider.py`): neutral port + `NullBcsBotTokenProvider` + `CachingBcsBotTokenProvider` (TTL cache: 5 min hit / 60 s
  miss). The driver-bot `session_token` flows through `BcsCreateGroupRequest.caller_bot_token` → `BcsHttpAdapter.create_group` / `SingleboxBcsAdapter.create_group` as `Authorization: Bearer` for caller
  ownership; convergence is handed to the result poller.
  - **Token-provider layering**: removed `community/plugins/.../bcs_bot_token_provider.py` (`DbBcsBotTokenProvider`) — reading `bcs_bots` belongs to the corp/data-source layer, not community.
  `task_module` now defaults to `NullBcsBotTokenProvider` (safe degradation: with `event_subscriptions` dropped, group creation goes through the no-sub branch with HMAC anonymous callers and still
  succeeds; no Bearer sent when no provider is bound), while keeping the **corp DI `bind(BcsBotTokenProvider)` override as the priority source** for prod token flow. `form_coop_group` dropping inline
  `event_subscriptions` is the actual 401 fix and is decoupled from token availability.
  - **`task_executor`** task-type branching; engine / `task_service` + `bcs_http` / `singlebox` adapter wiring.
  - **Tests**: ClawMind status-mapping 7-case assertion; `task_callback.status` assertions updated to the enum values; QC state-machine e2e split into `pre_e2e_chat` / `pre_e2e_ms` / `pre_e2e_yaml`;
  added `test_execute_task_type_branching` + `test_executor_e2e` + `test_bcs_http_adapter` + `test_singlebox_bcs_wiring`; `test_bcs_bot_token_provider` now covers only the neutral port / Null / Caching.

  ## Validation
  - `test_translator` + `test_callback_adapter` + `test_task_callback_report`: 66 passed (status mapping).
  - `task_runner/integration` full suite: 127 passed, 5 skipped (incl. `test_state_machine` form_coop_group + token-injection, adapters, executor e2e).
  - `di/test_task_module` assembly: 4 passed.
  - `test_bcs_bot_token_provider` (port / Null / Caching): green.

  ## Compatibility and risk
  - `task_callback.status` semantics change from the raw ClawMind string (e.g. `"succeeded"`) to the task `Status` enum value (e.g. `"DONE"`). `task_callback` is an audit-only table (ingest path records
  but does not drive the orchestration core); a grep confirmed no downstream reads `task_callback.status` against a raw string, so this is non-breaking. Historical rows may still hold legacy raw strings
  — both coexist (audit table, no migration).
  - Default `NullBcsBotTokenProvider`: prod Bearer flow now requires the corp side to inject the provider via DI (`bind(BcsBotTokenProvider, …)`). Until that corp binding lands, behavior degrades to
  no-Bearer, which is safe — group creation still succeeds via the no-sub + HMAC path. The corp injection point is `ocb`'s `corp_task_integration.py`; it is gated on the `bcs_bot_token_provider` port
  file being synced into the `ocb-public` community copy that ocb consumes, and is intentionally deferred (out of this PR).
  - Dropping inline `event_subscriptions` (the 401 fix) is independent of token availability, so the auth fix is not contingent on the corp token provider being wired.

  ## Related issues
  - Backports and lands the BCS coop-group auth / driver-bot token work from `REL20260826_jj_4` (squashed), then folds in the callback-status-mapping + token-provider layering follow-up.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/task/translator.py
+++ b/src/backend/src/agentclaw/community/adapters/http/task/translator.py
@@ -6,11 +6,12 @@ SSOT TaskCallbackData 不扩;ext_info/goal/未登记 str id 塞进 result["_ext_
 """
 from __future__ import annotations
 
+import json
 from dataclasses import dataclass
 from typing import Any, Literal
 
 from agentclaw.community.core.errors import NotFound
-from agentclaw.community.core.task.domain.models import TaskCallbackData
+from agentclaw.community.core.task.domain.models import Status, TaskCallbackData
 from agentclaw.community.core.task.task_runner.callback_correlation import (
     CallbackCorrelationRegistry,
 )
@@ -108,7 +109,8 @@ def translate_claw_mind(raw: dict, disposition: Literal["start", "result"]) -> T
     字段映射:
     - ``loop_task_id`` = ``flow_id``(run 实例 id,对齐 BCN 的 ``scope.run_id`` 回投键 → ``task_callback.run_id`` 列存 run 实例);工作流级回投,``node_id`` 空;
     - ``workflow_instance_id`` = ``ext_info.flow_runs.origin_session_id``(session_id → 落 task_callback.main_session_id);
-    - ``status`` 从底层 status 推:``ext_info.flow_runs.status`` > ``node_executions[0].status``(顶层 status 仅粗粒度事件类型);
+    - ``status`` = 底层 status(``ext_info.flow_runs.status`` > ``node_executions[0].status`` > 顶层)经
+      ``_claw_mind_status_to_task`` 映射的 task Status 枚举值(落 ``task_callback.status``,ClawMind 7 态 → task 7 态语义对应);
     - ``result.success`` 由底层 status 语义推(succeeded/completed→True;failed/cancelled/aborted→False;未知→不设 success);
     - ``result.data`` = ``node_executions[0].output_json``;``result.exec_error`` = ``node_executions[0].error_text``;
     - ``execution_graph`` = 全量 ``ext_info``(flow_runs + node_executions 快照 → task_callback.execution_graph);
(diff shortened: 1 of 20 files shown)


msjbear (dismissed):

totalfrank (approved):
LGTM.

PR #1594, opened 2026-08-27
feat(bcs): forward channel sender identity

RaymondHX:
## Problem

Human messages entering through a Channel did not carry sender attribution to downstream bots and HTTP Providers. Agents therefore could not reliably tell which person sent a message, and ClawMind could not collect that identity from the final prompt.

## Solution

- Add the opt-in Bot binding setting `forward_sender_identity`, defaulting to `false`; reject non-boolean values and reject enabling it on Group bindings.
- For ordinary Human Channel messages, project the trusted staff number and current inbound nickname into the BCS frame after slash-command and HumanInput handling.
- Map opted-in identity into OpenClaw standard sender metadata.
- Prefix only HTTP Provider `chat.send`/`chat.inject` request-copy text with serialized `{"sender":{"id":"...","name":"..."}}` JSON, without changing Provider parameters or persisted message content.
- Keep slash commands, HumanInput replies, Group bindings, non-Channel entry points, outbound flows, synthetic messages, and legacy resolver behavior unchanged.
- Fall back to the trusted staff number when the inbound nickname is absent or blank.

## Validation

- `cargo test -p bcs-protocol -p bcs-channel -p bcs-message-flow -p bcs-provider-http --no-fail-fast`
- `npm run lint`
- `npm run prepublishOnly`
- `npm run test-local` — 61 passing
- Rebased onto latest `origin/dev` (`8182e771dd`) and reran the tests above successfully.
- Commit and push used `--no-verify`; the relevant gates were run explicitly as listed above.

## Compatibility and risk

The setting is off by default and requires no binding migration. Existing bindings and sender resolvers keep their current behavior unless a Bot binding opts in. Forwarded sender data is attribution metadata only and must not be used for authentication or authorization.

### DIFF
--- a/docs/bot-integration.md
+++ b/docs/bot-integration.md
@@ -242,6 +242,13 @@ Versioning policy:
 - Removing fields, changing semantics, or adding required fields bumps the
   version.
 
+An opted-in Human Channel-to-Bot message may include
+`channel.identity_forwarding: true` together with `channel.user_id` and
+`channel.actor_name`. OpenClaw integrations should project those values to the
+standard inbound `SenderId` and `SenderName` metadata. If the marker is absent
+or false, integrations must retain their existing sender resolver. This
+metadata is not an authentication or authorization signal.
+
 When the engine receives `deprecation`, it should log a reminder for developers
 to upgrade:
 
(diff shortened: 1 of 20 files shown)


vzvince (approved):
LGMT

totalfrank (commented):
LGTM.

totalfrank (approved):

PR #1608, opened 2026-08-27
fix(backend): preserve MCP config on set deactivation

FreddieSun:
## Problem

SkillSet deactivation passed the Set's MCP codes as `released_mcp`. Deactivation removes callable state but keeps the Set membership, so the runtime could physically delete configuration that must remain available for a later activation. MCP commands also built deltas from request parameters, allowing a `changed=false` no-op to declare a false claim or release.

## Solution

- Give SkillSet deactivation a fixed `ProjectionScope(skills=True, mcp=True)` with no MCP claim or release.
- Build MCP claim/release scopes from the committed `DesiredStateMutation` result for ordinary Set, Default exclusion, and Direct commands.
- Return an MCP code from the repository only when the mutation actually changed that code; no-op mutations return an empty delta.
- Keep `REL20260828` behavior for inactive membership changes, synchronous delete, and DeviceActivated recovery.
- Do not introduce a Retained resolver/state model, global cross-SkillSet queries, inactive cleanup, async adapter changes, or lifecycle expansion.

## Validation

- TDD red/green at `SkillSetManagementService`, `DirectActivationService`, and `CapabilityDesiredStateRepository` seams.
- Focused Service + repository suite: `170 passed`.
- Affected Skill Center/API/Gateway suite: `940 passed, 22 skipped`.
- Backend architecture suite: passed.
- Python SAST block scan: passed.
- Full Backend suite: `15097 passed, 57 skipped`.
- Final Backend CI gate: case pass rate `100%`; line coverage `87.61%`; changed-line coverage `83.33%`.
- Standards review: PASS after fixing the platform-default unexclude candidate.
- Spec review: PASS after the same fix.

## Compatibility and risk

No HTTP, Service API, Plugin API, database schema, adapter, Dispatcher, provider, or engine contract changes. The current ownership policy prevents one MCP from joining multiple SkillSets; global historical cross-reference cleanup remains intentionally out of scope.

## Spec

- `src/backend/specs/2026-08-27-mcp-effective-retained-projection/spec.md`
- `src/backend/specs/2026-08-27-mcp-effective-retained-projection/plan.md`

### DIFF
--- a/src/backend/src/agentclaw/community/core/repository/implementations/skill_center/mcp_skill_set_control_plane.py
+++ b/src/backend/src/agentclaw/community/core/repository/implementations/skill_center/mcp_skill_set_control_plane.py
@@ -120,7 +120,10 @@ class McpSkillSetControlPlaneCommands:
                     env=get_current_env(), server_code=server_code,
                 )
             session.flush()
-            return DesiredStateMutation(self._as_item(row), True, old)
+            return DesiredStateMutation(
+                self._as_item(row), True, old,
+                mcp_codes=frozenset({server_code}),
+            )
 
     def remove_mcp(
         self, *, bot_id: str, owner_id: str, set_id: str, server_code: str,
@@ -144,7 +147,7 @@ class McpSkillSetControlPlaneCommands:
                     .first()
                 )
                 if existing is not None:
-                    return SkillSetMutation(self._as_item(row), False, old)
+                    return DesiredStateMutation(self._as_item(row), False, old)
                 session.add(
                     DefaultSkillsetMcpExclusion(
                         user_id=owner_id,
@@ -155,7 +158,7 @@ class McpSkillSetControlPlaneCommands:
                     )
                 )
                 session.flush()
-                return SkillSetMutation(self._as_item(row), True, old)
+                return DesiredStateMutation(self._as_item(row), True, old)
             self._ordinary(row)
             membership = (
                 self._scope(session.query(SkillSetMCPServer), SkillSetMCPServer)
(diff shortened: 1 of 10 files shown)


src/backend/src/agentclaw/community/core/repository/implementations/skill_center/mcp_skill_set_control_plane.py
```diff
@@ -144,7 +147,7 @@ def remove_mcp(
                     .first()
                 )
                 if existing is not None:
-                    return SkillSetMutation(self._as_item(row), False, old)
+                    return DesiredStateMutation(self._as_item(row), False, old)
```
totalfrank, line 150: We are removing a mcp, we should have mcp_codes set properly here so that it gets to ProjectionScope right?

My understanding is that this branch is not reachable because removing mcp from default skill set reaches the exclude_default_mcp method down below, not this one here.

Leave a TODO here, we will remove this code in follow up PRs.

src/backend/src/agentclaw/community/core/repository/implementations/skill_center/mcp_skill_set_control_plane.py
```diff
@@ -221,7 +227,10 @@ def exclude_default_mcp(
             if not created and removed_installation == 0:
                 return DesiredStateMutation(self._as_item(row), False, old)
             session.flush()
-            return DesiredStateMutation(self._as_item(row), True, old)
+            return DesiredStateMutation(
```
totalfrank: This looks like a asymmetry to me. We are also doing remove mcp from default skill set. But we set mcp_codes here.

I guess at least we should have consistency between these places right?

src/backend/src/agentclaw/community/core/repository/implementations/skill_center/mcp_skill_set_control_plane.py
```diff
@@ -171,7 +174,10 @@ def remove_mcp(
                     env=get_current_env(), server_codes={server_code},
                 )
             session.flush()
-            return DesiredStateMutation(self._as_item(row), True, old)
+            return DesiredStateMutation(
+                self._as_item(row), True, old,
+                mcp_codes=frozenset({server_code}),
+            )
 
     def exclude_default_mcp(
```
totalfrank, line 182: This one looks the default skill set path for remove_mcp, seems to me a little bit redundant. Can we fix it?

Also, I noticed this one has platform_default_codes but remove_mcp doesn't. Why is that? Can we consolidate it?

totalfrank (approved):
LGTM.

PR #1614, opened 2026-08-27
fix(aicoding): resync MCP and skills on confirmed restart

panqiwp:
## Summary
- refresh AICoding MCP scope/details on restart when `confirmed_template_update` is set
- actively resync runtime skill symlinks after confirmed template restart
- keep restart refresh logic inside AICoding strategy and avoid CLI grant changes
- inject MCP sync into `BotService` from the DI composition root instead of exposing it through `DeviceService`

## Compatibility and risk
- Scope is gated by `extra_configs["confirmed_template_update"]`; plain restart keeps existing behavior and does not resync MCP/skills.
- Affected call sites are the restart paths that can reach a ready device:
  - replacement restart: `_allocate_device_async` calls `refresh_restart_authorization` after `apply_device` succeeds, so the refresh targets the new runtime instead of the old binding.
  - BaaS in-place restart: `restart_bot` calls `refresh_restart_authorization` after `_restart_bot_baas` succeeds.
  - pre-allocation restart extension handling only applies config; runtime refresh is intentionally deferred to the ready-device points above.
- `refresh_restart_authorization` return value means "engine opted into async best-effort refresh"; it does not mean MCP/skill sync succeeded.
- The previous restart Passport scope rewrite is intentionally not reused for this AICoding path: this change syncs latest MCP runtime configuration and refreshes `~/.claude/skills` symlinks, without modeling `mcporter` or `la ~/.claude/skills/` as CLI grants.
- `caller_identity_repo` remains in `BotService` construction for release-branch DI/API compatibility, but is no longer used by this restart-runtime refresh path.

## Tests
- `.venv/bin/python -m py_compile src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py src/agentclaw/community/core/bot_management/engines/default.py src/agentclaw/community/core/bot_management/engines/provisioning.py src/agentclaw/community/core/bot_management/services/bot_service.py src/agentclaw/community/core/devices/services/device_service.py src/agentclaw/community/di/modules/bot_management_module.py tests/community/core/bot_management/services/test_restart_authorization_refresh.py tests/community/core/bot_management/services/test_bot_service_restart_idempotency.py`
[...]

PR #1620, opened 2026-08-27
fix(aicoding): resync MCP and skills after restart

panqiwp:
## Summary
- refresh AICoding restart MCP scope only after confirmed template updates
- retry runtime MCP detail sync while BaaS devices are not active yet
- retry Claude skills symlink runtime sync in the same best-effort background flow

## Test
- cd src/backend && .venv/bin/python -m pytest tests/community/core/bot_management/services/test_restart_authorization_refresh.py tests/community/core/bot_management/services/test_bot_service_restart_idempotency.py::TestRestartAuthorizationResyncWiring -q


totalfrank (changes requested):

## 2. Gate 0

- **Target branch:** `REL20260828` — matches the release-branch pattern (`^REL\d{8}`).
- **Classification:** feature work extending an existing pipeline — wrong target for a release branch. Not path (a) (existing-behavior bug fix with a `dev` cherry-pick), and not eligible for path (b) (isolated new feature) either.
- **Fail reasoning:**
  - `EngineProvisioningStrategy.refresh_restart_authorization` (`src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py:228`) is not net-new. It is an existing `@abstractmethod` already wired into the live restart pipeline before this PR — `BotService.restart_bot` already called it (with the old signature) on every confirmed-template-update restart, and `BotService._allocate_device_async` is where the new second call site is inserted. Deleting the new files would **not** leave existing behavior byte-identical: the contract signature and semantics of code already on the path change. This alone fails Isolation Audit criterion 1, so path (b) is unavailable — the exception was never in play.
  - It also isn't a minimal, existing-behavior fix (path a): the diff removes 3 required parameters (`passport_plugin`, `template_service`, `caller_identity_repo`), changes the return type (`None` → `bool`), and replaces the previous synchronous Passport-scope-diffing implementation with a new async `MCPSyncService`-based design that spawns a background thread with its own retry/backoff loop. That is a redesign of existing restart behavior, not a defect fix.
  - **No `dev` counterpart.** None of this PR's four commits (`ff0b89f`, `ec77b58`, `33df937`, `66ac0b1`) are ancestors of `origin/dev`. `origin/dev` still carries the pre-PR Passport-plugin-based implementation of this exact method (verified via `git show origin/dev:.../aicoding/strategy.py`) — the version this PR is discarding on `REL20260828` alone. This is precisely the "regression by omission" Gate 0 warns about: the next release cut from `dev` will either silently drop this PR's behavior or produce a structural merge conflict on this exact contract. The PR description contains no cherry-pick reference and no note that a companion PR is queued for `dev`.
  - Because Isolation Audit criterion 1 fails outright, no audit table applies — the PR doesn't reach the point where isolation could be evaluated. **Re-target to `dev`.**

## 3. Findings

**Finding 1 — Breaking, non-additive change to an existing internal contract on a release branch.**
`src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py:228-236`
```python
def refresh_restart_authorization(
    self, ctx, bot, extra_configs, *,
    mcp_sync: Any = None,
    skill_set_factory: Any = None,
) -> bool:
```
was previously:
```python
def refresh_restart_authorization(
    self, ctx, bot, extra_configs, *,
    passport_plugin: Any, skill_set_factory: Any,
    template_service: Any, caller_identity_repo: Any,
) -> None:
```
Violated rule: Isolation Audit criterion 2 / constitution Rule 17 (Invariant — "Distinguish What Is Flexible from What Is Constrained": *"If changing something could cause a correctly written consumer or plugin to fail, treat it as constrained"*). Three required keyword parameters were deleted and the return type changed. Any other implementation or caller written against the old contract (on `dev`, or a future engine) breaks. On a release branch, contract changes must be strictly additive (new optional params/fields with safe defaults) — this is a full re-shape. Required fix: this belongs on `dev` as a normal contract evolution with propagation analysis, not as a release-branch edit.

**Finding 2 — Contract change with no propagation analysis or spec update (Rule 16, Invariant).**
Same location, plus the two updated call sites in `src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py:1939` and `:4522`, and the DI rewire in `src/backend/src/agentclaw/community/di/modules/bot_management_module.py` (new `mcp_sync_service: MCPSyncService` param, `mcp_sync=cast(McpSyncProtocol, mcp_sync_service)`). Rule 16 requires structural changes to declare affected consumers, affected implementations, affected configuration, compatibility status, and a migration plan, in the same change. The PR body has none of this, and the governing spec (`src/backend/specs/2026-08-26-mcp-sync-and-passport-regressions/tasks.md`, which documents the *old* Passport-based design this PR discards) is left untouched — so the spec now contradicts the shipped code. Required fix: update the spec/doc alongside the contract change and state explicitly which implementations (`default.py`, `aicoding/strategy.py`, any corp/other engine) were audited against the new signature.

**Finding 3 — No conformance/guard test for the changed contract (Rule 25, Invariant).**
`git diff` shows zero changes under `src/backend/tests/community/architecture/`. `refresh_restart_authorization` is exactly the kind of engine-variability contract Rule 25 requires conformance tests for (all implementations, including the default fallback, must be validated against the current contract). Nothing in this diff or in CI guards that every registered `EngineProvisioningStrategy` implementation still honors the shape and semantics of the method after this rewrite. Required fix: add or extend an architecture-layer test asserting all registered strategies implement the current `refresh_restart_authorization` signature/return contract.

## 4. Out-of-scope notes

- The `aicoding`-specific implementation correctly lives in `core/bot_management/engines/aicoding/strategy.py` and is reached only through the composition-root resolver (`resolve_provisioning`) — no engine-string branching or hardcoded capability sets were introduced in `core/`; the contract pattern itself (interface in `provisioning.py`, per-engine impl, mandatory default in `default.py`) is followed correctly.
- Not gating on it, but worth a security-focused look from the feature reviewer: the old implementation carried an explicit comment that an *unreadable* MCP execution identity must skip the refresh rather than fall back to an owner-only scope (privilege-escalation guard). The new `MCPSyncService.refresh_mcp_scope` path doesn't visibly carry that same guard — confirm the replacement preserves it.
- The new background-thread + fixed retry/backoff (`0, 2, 5, 10, 20`s) resync logic in `strategy.py` is a business-logic/operational choice, not reviewed here.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

Architecture-gate review only (business logic / notify semantics out of scope, per policy).

## Gate 0 — Target branch: FAIL

- **Target**: `REL20260828` (release branch, matches `^REL\d{8}`).
- **Classification**: feature work that extends an existing pipeline, mislabeled as `fix`.
- **Reasoning**:
  1. The diff replaces the entire body of `AicodingProvisioningStrategy.refresh_restart_authorization` and changes the abstract signature of `EngineProvisioningStrategy.refresh_restart_authorization` (`core/bot_management/engines/provisioning.py:228`, and the `default.py` no-op implementation) — dropping `passport_plugin`/`template_service`/`caller_identity_repo` and adding `mcp_sync`. This is new behavior (background-threaded retry loop with `_RESTART_RUNTIME_RESYNC_RETRY_DELAYS_SECONDS`, MCP detail resync, Claude-skills symlink resync via `skill_set_service.sync_runtime()`), not a minimal correction of an existing code path.
  2. It is wired directly into two **existing, always-reachable** call sites: `BotService._allocate_device_async` (`bot_service.py:~1930`) and `BotService.restart_bot`'s BaaS in-place path (`bot_service.py:~4510`). These are hit by every ordinary bot restart — this is not dormant/isolated new surface, so it does not qualify for the path-(b) exception even if it tried to.
  3. No `dev` counterpart exists. `git log dev -- core/bot_management/engines/` shows nothing matching this change; `git log REL20260828-tip..dev` / the reverse both confirm this PR's 4 commits (and the already-merged `#1614` it builds on) exist only on the release branch. The PR body contains no "(#NNN)" / cherry-pick reference. Per the branch policy this is a **regression-by-omission**: the next release cut from `dev` will not carry this restart-resync behavior at all.
- Since this is feature work extending an existing pipeline, the Isolation Audit doesn't apply (it isn't isolated), and Gate 0 alone is dispositive: **re-target to `dev`**, or, if the underlying defect being patched is a genuine regression already shipped on `REL20260828`, cite the corresponding `dev` PR/commit and cherry-pick it explicitly.

Full architecture review below completes per the instructions (findings carry over to the re-targeted PR).

## Findings

**1. `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:60-62` — undocumented string-matched contract used as a runtime-readiness signal (arch.rules.md Rule 1 / Rule 4 — Invariant)**

```python
_RESTART_RUNTIME_NOT_READY_MARKERS = (
    "NO_ACTIVE_DEVICES",
    "get_ws_info failed",
)

def _is_restart_runtime_not_ready_error(error: Any) -> bool:
    text = str(error or "")
    return any(marker in text for marker in _RESTART_RUNTIME_NOT_READY_MARKERS)
```

`"NO_ACTIVE_DEVICES"` is at least a declared BaaS error code (`src/baas/.../_exceptions.py:99`). `"get_ws_info failed"` is not a contract at all — it is a raw, free-text log/exception message emitted by `core/devices/services/conn_info_builders/baas_builder.py:62` and `teclaw_builder.py:45` (`f"...get_ws_info failed for binding={binding.id}: {e}"`), never declared as an error code anywhere. This retry loop decides whether to keep retrying MCP/skill resync by substring-matching that free-text message. Nothing documents this as a contract, and nothing conformance-tests it, so a harmless wording change in either builder's log string silently turns "keep retrying" into "give up and log an error" with no test failure to catch it. Per Rule 1/Rule 4, cross-component behavior this load-bearing must be an explicit, typed signal (e.g., a dedicated exception type or error-code field on the raised error), not a text-matched log message.

*Fix direction*: have the device/connection layer raise (or return) a typed "runtime not ready" indicator (exception subclass or enum/error-code field) that `conn_info_builders` and BaaS responses both populate; have the aicoding strategy catch/check that typed signal instead of grepping exception text.

**2. `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:56-57` — hardcoded retry policy as literals in a core-adjacent engine service (arch.rules.md catalog item 7 / Rule 14 — Policy)**

```python
_RESTART_RUNTIME_RESYNC_RETRY_DELAYS_SECONDS = (0, 2, 5, 10, 20)
```

Retry counts/backoff schedule for the MCP-detail and skill-symlink resync loops are baked in as a code literal rather than sourced from configuration. This is explicitly called out in the violation catalog ("timeouts, retry counts ... written as literals in core services ... are configuration, not code"). Low severity relative to Gate 0, but worth fixing in the same change since the file is already being edited.

*Fix direction*: move the delay schedule to validated configuration (e.g. alongside other aicoding engine tuning in `config/`), injected into the strategy rather than a module-level constant.

## Out-of-scope notes

- `core/workspace/engines/aicoding.py`'s rsync-exclude path fix (`.repos` → `.repos/`, `.prewarm/ready.json` → `.prewarm_ready.json`) is bundled into this PR as an unrelated commit; not gated on architecturally, but consider splitting for a cleaner release-branch diff.
- The `EngineProvisioningStrategy` contract update itself is well-formed (both implementations — `aicoding` and `default` — updated consistently, composition root (`registry.py`) untouched, `McpSyncProtocol`/`MCPSyncService` are pre-existing contracts reused correctly, not new plugin-style abstractions). The layering pattern is not the problem — the branch target is.

---

panqiwp:
Addressed the release-branch code-level feedback that is in scope for this PR update:

- Removed retry detection based on the free-text wrapper message `get_ws_info failed`.
- Runtime-readiness retry now keys only off the provider/BaaS error code `NO_ACTIVE_DEVICES`, with an inline comment documenting that we intentionally avoid matching non-contractual wrapper text.
- Added commit: `3b6741828 fix(aicoding): use provider code for runtime retry`.
- Re-ran targeted tests:
  `cd src/backend && .venv/bin/python -m pytest tests/community/core/bot_management/services/test_restart_authorization_refresh.py tests/community/core/bot_management/services/test_bot_service_restart_idempotency.py::TestRestartAuthorizationResyncWiring -q`
  Result: `23 passed, 17 warnings`.

I did not retarget this PR or add a dev counterpart in this update. The retry-delay configuration point is noted, but I intentionally did not expand this release fix into a broader config/DI change in this PR.

totalfrank (approved):
LGTM.

PR #1626, opened 2026-08-28
  fix(bcs): forward friend work-order auth and optimize /bots/search

liveandevil:
## Problem
  Friend work-order notification and `/bots/search` both had behavior and performance gaps:

  - friend work-order notification path needed to forward caller auth context to the backend internal endpoint.
  - `/bots/search` was doing per-item runtime status checks, which made the endpoint slower than necessary.
  - `/bots/search` default visibility behavior depended on login state, which made the search scope inconsistent.

  ## Solution
  - Forward auth context for friend work-order notification routing and keep the backend internal endpoint as the target.
  - Batch `/bots/search` dynamic status resolution by collecting candidate bot ids and checking runtime active ids in one call.
  - Make `/bots/search` default visibility consistently `public + protected`, regardless of login state.
  - Keep legacy search/list paths unchanged outside `/bots/search`.

  ## Validation
  - `CARGO_TARGET_DIR=/tmp/bcs-target-http cargo test -p bcs-http --test bots_contract`

  ## Compatibility and risk
  - This changes `/bots/search` default visibility behavior only.
  - Legacy APIs outside `/bots/search` are not modified.
  - No schema changes.
  - If needed, the search-default change can be reverted independently without affecting the friend work-order routing fix.

  ## Related changes
  - friend work-order notification logging
  - friend work-order notification auth forwarding
  - `/bots/search` dynamic status batching
  - `/bots/search` default visibility normalization

### DIFF
--- a/src/bcs/Cargo.lock
+++ b/src/bcs/Cargo.lock
@@ -679,6 +679,7 @@ dependencies = [
  "bcs-config",
  "bcs-db-api",
  "bcs-db-local",
+ "bcs-domain",
  "bcs-organization",
  "bcs-organization-store",
  "bcs-protocol",
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/bots.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/bots.rs
@@ -527,11 +527,9 @@ pub async fn search_bots(
         ));
     }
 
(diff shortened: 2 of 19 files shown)


RaymondHX (approved):
LGTM

vzvince (approved):

PR #1630, opened 2026-08-28
fix(bcs): add manifest.bundles in bcs-config-prod.toml

zh-bupt:
## Problem
The production BCS config `bcs-config-prod.toml` is missing the `manifest.bundles` definition, which prevents the `bcsPanel` resource from loading correctly in the prod environment.

## Solution
Added a `[[manifest.bundles]]` section to `src/bcs/configs/bcs-config-prod.toml`, declaring the `bcsPanel` resource (URL type) pointing to the `@avernet-assets/bcs-panel@1.2.0` UMD artifact.

## Validation
- Pre-push lint-only gate passed
- Config-only change adding a manifest bundle entry; no logic code modified

### DIFF
--- a/src/bcs/configs/bcs-config-prod.toml
+++ b/src/bcs/configs/bcs-config-prod.toml
@@ -209,8 +209,8 @@ schema_version = 1
 
 [[manifest.bundles]]
 name = "bcsPanel"
-type = "file"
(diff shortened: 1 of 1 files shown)


questchaos (approved):
LGTM

PR #1633, opened 2026-08-28
feat(backend): scope ac_task_queue claims by owning app

totalfrank:
## Problem

A second, independently deployed backend now runs against the same `ac_task_queue` table. Both fleets' workers match only on `env`, so each claims the other's rows, finds no handler registered for the `task_type`, and fails the task terminally. Work that belongs to one deployment is destroyed by the other.

The table has already gained the owning-app column out of band:

```sql
`app` varchar(32) NOT NULL DEFAULT 'agentclaw' COMMENT 'app who owns the task'
```

along with `uk_env_app_task_type_active_idempotency_key`, `idx_env_app_status_run_at` and `idx_env_app_lease_expires_at`. This change teaches the queue to write that column and to match on it.

## Solution

**Config.** The owner is the deployment's own identity, so `ConfigModule.task_queue` reads the existing top-level `app_name` into a `TaskQueueConfig` — no queue-specific key, and therefore no second name that could drift from the first. `TaskQueueConfig` is deliberately not a field of `TaskQueueWorkerConfig`: the enqueue path needs the same value and has nothing to do with worker policy, and turning the worker off must not stop rows from being stamped with their owner.

A name that is *present* but unusable — empty, whitespace-padded, or over 32 characters — **raises** rather than falling back. The fallback is the accident worth preventing: the default is the *other* deployment's name as often as not, so a quiet substitution is precisely how one backend starts claiming another's work. `container.py` resolves `TaskQueueConfig` eagerly at build time (the treatment `HttpClientPoolConfig` already gets, for the same reason) so that raise stops the boot on every profile instead of being swallowed by lifecycle discovery, which would leave a running app with no worker and no explanation. No config at all — local mode, ad-hoc tests — falls back to the column default, since there is nothing there to have got wrong.

**Plumbing.** `TaskQueueService` stamps `app` on enqueue; `TaskWorker` claims with it. Both read the one config object, so the two sides cannot disagree within a process. `TaskQueueService.enqueue`'s public signature is unchanged, so no adopter call site changes.

**Repository.** `app` joins `env` in the claim eligibility predicate — and therefore in the claim CAS and in the deadline retirement inside the claim scan — in both dedup lookups, and in `list_by_status`. `get_by_id` stays unscoped: an id is unique across apps, it is how an insert is read back, and "what happened to task 71544?" is asked precisely when the owner is the question. `TaskRecord` now carries `app` for a caller that needs to check it.

**Dedup scope** becomes `(env, app, task_type, active_idempotency_key)`. The pre-app unique index is still on the deployed table and ignores `app`, so it can reject a keyed insert whose key is free in this app's own scope. Both index names are recognised as dedup conflicts; the rejection finds no holder in this app (another app's live task is never handed back — the caller could not claim it, so its work would silently never run) and ends in the existing exhausted-retries `RuntimeError`, whose message names the legacy index as the other possible cause. The case disappears when that index is dropped.

**ORM** mirrors the deployed DDL — the column, the app-scoped unique index, and the two app-scoped scan indexes, plus the legacy index the table still carries. Declaring the legacy index is deliberate: leaving it out would hide the cross-app rejection from SQLite and let it surface first in production. `app` keeps the table's default collation, like `env` and unlike the key columns, matching the DDL.

Rejected alternative: relying on the column's `DEFAULT 'agentclaw'` instead of writing `app` on insert. That works only for the one deployment whose name happens to be the default; any other would enqueue rows it could never claim.

## Validation

Run in this container with `uv`-installed deps (SQLite backing the integration tests, as the suite is designed). After the review round:

- `pytest tests/community/config tests/community/di tests/community/core/task_queue tests/community/repository tests/community/architecture` — **1326 passed, 5 skipped**
- `pytest tests/community/core/skill_center tests/community/core/skills_pool tests/community/core/session_resources tests/community/endpoints tests/community/e2e` — **2592 passed, 25 skipped**
- `flake8 --select E999,E902,E117,F405,E712,E701,E702,F821,F822,F823,F831` over the changed Python files (the `python_sast_local.sh` block rules) — clean.

Before the review round, `pytest tests/community/core tests/community/endpoints tests/community/adapters` gave **11470 passed, 55 skipped, 2 failed**; both failures are `test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]`, failing on `exec: rsync: not found` — environmental and unrelated to this diff. CI's Backend unit tests and Singlebox coverage were green on the first head.

New tests cover: claim ignores another app's pending row, its expired lease, and its past-deadline row (no cross-app `TIMED_OUT`); enqueue stamps the owner; `list_by_status` is scoped and `get_by_id` is not; the worker leaves another app's task alone end to end; the cross-app dedup rejection and its release once the holder goes terminal; the config provider's accept/reject cases plus a pin that it really reads the top-level `app_name`; and schema pins for the column's DDL literals, the app-scoped index column order, and the collation decision.

Golden effective-config snapshots regenerated with `tests.community.config.regen_golden`. That run also re-sorted `baas.eval_template_uuid` in `community.json` / `singlebox.json`, which a previous hand-edit had left out of the generator's key order; the snapshot test compares parsed JSON, so this is cosmetic.

Not run: the acceptance suites, which need a product stack this container cannot start.

## Compatibility and risk

**Schema:** already applied — the column and the three new indexes exist on the deployed table. No backfill: `app` is `NOT NULL DEFAULT 'agentclaw'`, so every pre-existing row takes the default, which is exactly the deployment that wrote it.

**Ordering that matters:** a deployment whose `app_name` is *not* the default must carry it **before** it starts enqueueing, since that name is what its own claims will look for. Two deployments must never share an `app_name` — that is back to fighting over the same rows — and, because the column takes the table's case-insensitive collation on MySQL/OceanBase, names must differ by more than case.

**`app_name` now re-scopes the queue.** Changing it after a deployment has enqueued leaves those rows under the old name, and that fleet stops claiming them. `application.yaml` records this next to the key.

**Config:** nothing new to set; a deployment that already has `app_name` keeps owning exactly the rows it already owned. A malformed value fails the boot by design.

**Follow-up, not in this PR:** drop `uk_env_task_type_active_idempotency_key` once every app writes its own `app`, and delete its declaration from `models.py` in the same change. Until then, two apps cannot use the same idempotency key for the same `task_type` in the same `env`; the second gets a `RuntimeError` naming the index. The README's Provisioning section records this.

**Rollback:** revert the commits. The column and indexes can stay — rows keep their `app` and pre-change code simply ignores it.

### DIFF
--- a/src/backend/src/agentclaw/community/configs/application.yaml
+++ b/src/backend/src/agentclaw/community/configs/application.yaml
@@ -168,6 +168,19 @@ user_config:
     apply_owner_whitelist: []
     global_dry_run: false
 
+  # Which application owns this deployment's rows in the shared ac_task_queue
+  # table. Another, independently deployed backend now runs against the same
+  # table; every row carries its owner, enqueue stamps this value, and every
+  # query that selects work matches it — so neither fleet claims the other's
+  # tasks and fails them for a task_type its registry never saw. Each deployment
+  # needs its own value — two deployments sharing one name are back to fighting
+  # over the same rows — and a deployment must be given its name before it starts
+  # enqueueing, since the name is what its own claims will look for. An empty,
+  # padded, or over-32-character value fails the boot rather than falling back
+  # (see ConfigModule.task_queue).
+  task_queue:
+    app: "agentclaw"
+
   # In-process durable task-queue worker (one per pod). Enabled by default so the
   # publish flow's stage advances (build/release/progress-poll) run as persisted,
   # crash-safe tasks. The test profile overrides this to false so unit tests don't
--- a/src/backend/src/agentclaw/community/configs/application.yaml
+++ b/src/backend/src/agentclaw/community/configs/application.yaml
@@ -1,4 +1,12 @@
 # 基础配置
+# app_name also names this deployment's rows in the shared ac_task_queue table:
+# another, independently deployed backend runs against the same table, every row
+# carries its owner, and each fleet claims only its own — so neither runs the
+# other's tasks and fails them for a task_type its registry never saw. Two
+# deployments must therefore not share an app_name, and a deployment needs its
+# own name before it starts enqueueing, since that name is what its own claims
+# will look for. Empty, padded, or over 32 characters fails the boot rather than
+# falling back (see ConfigModule.task_queue).
 app_name: "agentclaw"
 enable_sidecar: true
 workers: 6
@@ -167,19 +175,6 @@ user_config:
     enabled: true
     apply_owner_whitelist: []
     global_dry_run: false
-
-  # Which application owns this deployment's rows in the shared ac_task_queue
-  # table. Another, independently deployed backend now runs against the same
-  # table; every row carries its owner, enqueue stamps this value, and every
-  # query that selects work matches it — so neither fleet claims the other's
-  # tasks and fails them for a task_type its registry never saw. Each deployment
-  # needs its own value — two deployments sharing one name are back to fighting
-  # over the same rows — and a deployment must be given its name before it starts
-  # enqueueing, since the name is what its own claims will look for. An empty,
-  # padded, or over-32-character value fails the boot rather than falling back
-  # (see ConfigModule.task_queue).
-  task_queue:
-    app: "agentclaw"
 
   # In-process durable task-queue worker (one per pod). Enabled by default so the
   # publish flow's stage advances (build/release/progress-poll) run as persisted,
--- a/src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
+++ b/src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
@@ -91,9 +91,10 @@ logger = get_logger()
 #:
 #: There are two because the deployed table still carries the pre-``app`` index
 #: alongside the one that scopes dedup by ``app``. Either can reject an INSERT,
-#: and a rejection by the older one is not a duplicate *for this app* — see
-#: :meth:`TaskQueueRepository.enqueue`. The names are not substrings of one
-#: another, so matching them is unambiguous.
+#: so both have to be recognised; the names are not substrings of one another,
+#: so matching them is unambiguous. Dropping the legacy index is the last step of
+#: the app migration (see ``models.py`` and README "Provisioning"), and its
+#: spelling comes out of this tuple then.
 _ACTIVE_IDEM_INDEXES = (
     "uk_env_app_task_type_active_idempotency_key",
     "uk_env_task_type_active_idempotency_key",
@@ -258,9 +259,8 @@ def _is_active_idem_conflict(exc: IntegrityError) -> bool:
 
       ``active_idempotency_key`` appears in both indexes and in no other
       constraint, so the column form identifies the violation without having to
-      tell the two apart — which is right, because the branch that does care
-      (a rejection by the pre-``app`` index) is decided by looking up the holder,
-      not by parsing the message.
+      tell the two apart — which is all :meth:`enqueue` needs, since it resolves
+      the holder by querying for it rather than by parsing the message.
 
     A pure function over the exception so it is unit-testable against both
     message forms without a MySQL instance.
@@ -458,45 +458,6 @@ class TaskQueueRepository(
             )
             return row[0] if row else None
 
-    def _find_foreign_app_key_holder(
-        self, *, env: str, app: str, task_type: str, idempotency_key: str
-    ) -> Optional[tuple[int, str]]:
-        """``(id, app)`` of a row belonging to a **different** app that holds
-        ``idempotency_key`` in this ``(env, task_type)``, or ``None``.
-
-        Reachable only while the pre-``app`` dedup index is still on the table
-        (see ``models.py``): that index is unique on ``(env, task_type,
-        active_idempotency_key)`` regardless of app, so it rejects an INSERT for
-        a key another deployment happens to be using — even though, in this
-        app's scope, the key is free.
-
-        It is checked because the two live cases it separates want opposite
-        responses, and both look identical from :meth:`enqueue`'s side ("the
-        index rejected us, and no holder is visible in our scope"):
-
-        - **The key was released mid-race** — retrying is correct and normally
-          succeeds.
-        - **Another app holds it** — retrying can never succeed while that row
-          is live, and no amount of budget changes that.
-
-        Without the distinction the second case burns every attempt and reports
-        sustained churn, which points at the wrong thing entirely: the fix is to
-        drop the legacy index (the last step of the app migration), not to look
-        for a hot key. The message says so and names the row.
-        """
-        with self._db.orm_session() as db:
-            row = (
-                db.query(self.Model.id, self.Model.app)
-                .filter(
-                    self.Model.env == env,
-                    self.Model.app != app,
-                    self.Model.task_type == task_type,
-                    self.Model.active_idempotency_key == idempotency_key,
-                )
-                .first()
-            )
-            return (row[0], row[1]) if row else None
-
     def enqueue(
         self,
         *,
@@ -558,8 +519,9 @@ class TaskQueueRepository(
                         idempotency_key,
                     )
                     return EnqueueResult(existing, False)
-                # No live holder of ours, yet an index rejected us. Three very
-                # different situations, and retrying is right for only one.
+                # No live holder of ours, yet an index rejected us. Two very
+                # different situations, and retrying is right for only one of
+                # them.
                 stranded_id = self._find_stranded_key_holder(
                     env=env,
                     app=app,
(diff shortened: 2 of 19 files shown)


src/backend/src/agentclaw/community/configs/application.yaml
```diff
@@ -168,6 +168,19 @@ user_config:
...
+  # table; every row carries its owner, enqueue stamps this value, and every
+  # query that selects work matches it — so neither fleet claims the other's
+  # tasks and fails them for a task_type its registry never saw. Each deployment
+  # needs its own value — two deployments sharing one name are back to fighting
+  # over the same rows — and a deployment must be given its name before it starts
+  # enqueueing, since the name is what its own claims will look for. An empty,
+  # padded, or over-32-character value fails the boot rather than falling back
+  # (see ConfigModule.task_queue).
+  task_queue:
+    app: "agentclaw"
```
totalfrank-personal: There's already a top level config called app_name in this file. Let's just use that?

src/backend/src/agentclaw/community/di/modules/config_module.py
```diff
@@ -835,6 +836,66 @@ def dormant_notify(self) -> cfg.DormantNotifyConfig:
...
+        - longer than the stored width — a non-strict server truncates, and the
+          rows are then filed under a name the claim filter never matches, so
+          the work is enqueued and simply never runs.
+
+        ``container.py`` resolves ``TaskQueueConfig`` eagerly at build time (as
+        it does ``HttpClientPoolConfig``) so this raise stops the boot on every
+        profile instead of being swallowed by lifecycle discovery — which would
+        leave the app running with no worker and no explanation.
+        """
+        block = _block("task_queue")
```
totalfrank-personal: Let's use top level app_name config.

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
```diff
@@ -466,22 +558,46 @@ def enqueue(
...
-                        f"{stranded_id}, which is terminal but never released "
+                        f"type={task_type} env={env} app={app} is held by task "
+                        f"id={stranded_id}, which is terminal but never released "
                         "it. Retrying cannot help — the key is occupied "
                         "forever. Most likely a worker running code from "
                         "before enqueue idempotency wrote the terminal status "
                         "without clearing active_idempotency_key; clear that "
                         "column on the row to free the key"
                     )
+                # Still no holder in our scope: the rejection can only have
```
totalfrank-personal: What is this? Is it necessary? If not, let's delete it.

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
```diff
@@ -410,6 +458,45 @@ def _find_stranded_key_holder(
             )
             return row[0] if row else None
 
+    def _find_foreign_app_key_holder(
```
totalfrank-personal: Do we really need this? If not, let's delete it.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/configs/application.yaml
totalfrank: Done in f7739ff — the `task_queue` block is gone and the owner now comes from the top-level `app_name`. I added a comment there recording that it also names this deployment's `ac_task_queue` rows, since changing `app_name` now re-scopes the queue: in-flight rows keep the old name and this fleet stops claiming them.

---

src/backend/src/agentclaw/community/di/modules/config_module.py
totalfrank: Done in f7739ff. `ConfigModule.task_queue` now reads the top-level `app_name` through a new `_app_name()` helper (same defensive shape as `_user_config()`), so there is no queue-specific config surface that could drift from it.

It distinguishes "no config at all" (`None` — local mode, ad-hoc tests → falls back to the column default) from "config present but names nothing" (`""` → raises), and still rejects a padded or over-32-character name. `TaskQueueConfig` is resolved eagerly in `container.py`, so those rejections stop the boot rather than silently dropping the `TaskWorker` binding.

---

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
totalfrank: Deleted in f7739ff, along with its branch in `enqueue`.

For context on what it was: the deployed table still carries the pre-`app` unique index `uk_env_task_type_active_idempotency_key`, which ignores `app`, so it can reject a keyed insert whose key is free in this app's own scope but held by the other deployment. This query looked that up so the error could name the index instead of reporting contention.

Removing it doesn't change any outcome — no row is inserted and another app's task is still never handed back. The case now falls through to the existing exhausted-retries `RuntimeError`, whose message I extended to name the legacy index as the other possible cause. It disappears entirely once that index is dropped.

---

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
totalfrank: Deleted in f7739ff — this was the call site of `_find_foreign_app_key_holder`; see the reply on that thread for what it did and why dropping it changes no outcome.

---

src/backend/tests/community/core/task_queue/test_task_worker.py
```diff
@@ -56,7 +57,7 @@ def orm_session(self):
 
 
 class _World:
-    def __init__(self, config: TaskQueueWorkerConfig):
+    def __init__(self, config: TaskQueueWorkerConfig, queue_config: TaskQueueConfig = None):
```
totalfrank: nit: `queue_config: TaskQueueConfig = None` widens a required type to `T | None` for test convenience, which `AGENTS.md`'s Python Type Contracts section calls out as a reason not to do this. `Optional[TaskQueueConfig] = None` (or `TaskQueueConfig | None`) matches what the code actually accepts.

---

totalfrank (commented):
Reviewed commits 790c958 and f7739ff (merge-base 56404c3 → f7739ff) across all 19 changed files. This is the first automated pass on this PR — the existing review threads are the author addressing a human reviewer's (`totalfrank-personal`) feedback, not prior automated findings, so there's nothing to re-verify.

Checked: claim/CAS/deadline-retirement app-scoping in `_eligible` and its three call sites; the dedup lookups (`_find_active_by_key`, `_find_stranded_key_holder`) and the retry loop's fallthrough to the exhausted-retries error for the cross-app legacy-index case; `_is_active_idem_conflict`'s index-name/column-name matching against both indexes; `ConfigModule.task_queue`'s validation and its `None`-vs-`""` distinction against how `sofa_config`/`AppConfig`/`load_config` actually behave; the eager `container.py` resolution; every production call site of `enqueue`/`claim_batch`/`list_by_status` for the new `app` parameter; the ORM model's column/index declarations against the deployed DDL described in the PR body; and test coverage for each of the above. No correctness, contract, concurrency, or architecture-boundary issues found. One nit posted inline.

---

src/backend/tests/community/core/task_queue/test_task_worker.py
totalfrank: Fixed in e20ca57 — by removing the parameter rather than annotating it `Optional`.

The rule you cite argues against the optional existing at all, and that applies here: `_World` had exactly one construction site (`_world()`), which never passed `queue_config`, so it was configurability no caller used. The worker's app-isolation test enqueues another app's row through `repo.enqueue(..., app="teclaw")` directly rather than standing up a second world, so nothing needs the seam. `self.queue_config = TaskQueueConfig(app=APP)` unconditionally now.

---

totalfrank (commented):
Reviewed the new commit since the last round: e20ca57 (test-only, removes the `queue_config` seam flagged by the earlier nit). Verified it's a clean, minimal fix — `_World` had exactly one caller (`_World(cfg)`, no `queue_config` arg), the app-isolation test enqueues the foreign app's row directly via `repo.enqueue(..., app="teclaw")` rather than through a second `_World`, and `TaskQueueConfig` stays used at its one remaining call site. No new issues.

1. `application.yaml` — use top-level `app_name` instead of a new config block → fixed in f7739ff, resolved.
2. `config_module.py` — same, read `app_name` instead of a new config block → fixed in f7739ff, resolved.
3. `task_queue.py` — is `_find_foreign_app_key_holder`'s helper necessary? → deleted in f7739ff, resolved.
4. `task_queue.py` — is its call site in `enqueue` necessary? → deleted in f7739ff, resolved.
5. nit: `queue_config: TaskQueueConfig = None` widens a required type for test convenience → fixed in e20ca57 (parameter removed rather than widened), resolved.

Open: 0. The PR is clean as of e20ca57.

---

FreddieSun (approved):
LGTM

PR #1634, opened 2026-08-28
fix(aicoding): defer restart authorization resync

panqiwp:
## Summary
- Persist AICoding confirmed restart resync intent with template config
- Defer BaaS in-place restart MCP/CLI/skill runtime refresh until restart publish completes
- Add AICoding BaaS publish listener and targeted tests

## Tests
- uv run python -m py_compile src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py src/agentclaw/community/core/bot_management/engines/aicoding/restart_authorization_listener.py src/agentclaw/community/core/bot_management/engines/provisioning.py src/agentclaw/community/core/bot_management/engines/default.py src/agentclaw/community/core/bot_management/services/bot_service.py src/agentclaw/community/di/modules/bot_management_module.py
- uv run pytest tests/community/core/bot_management/services/test_restart_authorization_refresh.py tests/community/core/bot_management/services/test_restart_template_refresh.py tests/community/core/bot_management/services/test_bot_service_restart_idempotency.py::TestRestartAuthorizationResyncWiring -q


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

Architecture-gate review only (business/product logic in the aicoding restart flow is out of scope and assumed correct).

## Gate 0 — Target branch

- **Target branch:** `REL20260828` — matches `^REL\d{8}`, a release branch.
- **Classification:** feature extending an existing pipeline (mislabeled as `fix`) — **wrong target**.
- **Reasoning:**
  - The diff removes the *existing* synchronous `refresh_restart_authorization` call from `bot_service.py`'s BaaS in-place `restart_bot` path (previously ~30 lines directly after `_restart_bot_baas`) and replaces it with an asynchronous, event-driven mechanism (`AicodingRestartAuthorizationBaasPublishListener` + a new persisted `template_config` marker). That is a behavior change to code that is live on this release branch today, not a fix to a pre-existing defect and not new-but-unreachable functionality.
  - It also changes the `EngineProvisioningStrategy.refresh_restart_authorization` abstract method signature (`provisioning.py`, `default.py`) — a cross-engine internal contract, touched for every engine, not scoped to an isolated area.
  - Isolation Audit (path (b)) therefore fails at step 1 ("net-new and unreachable / deleting the new files leaves every existing path byte-identical"): deleting `restart_authorization_listener.py` here would leave the BaaS in-place restart flow with **no** authorization/skill resync at all, because the old synchronous call site was deleted in the same PR. The new code is the sole remaining implementation of live, reachable behavior — the opposite of dormant.
  - Isolation Audit also fails at step 2 (collateral edits must be strictly necessary and behavior-preserving): `src/backend/.../workspace/engines/aicoding.py` changes the rsync-exclude list (`workspace/.repos` → `workspace/.repos/`, `workspace/.prewarm/ready.json` → `workspace/.prewarm_ready.json`) with matching test updates. This is an unrelated behavior change to an existing sync pipeline, not required to mount the restart-resync feature, riding along in the same release-branch PR.
  - The PR description contains no reference to a corresponding `dev` PR / cherry-pick (no "(#nnn)", no "cherry-pick" note), so per the branch-policy rule this is itself a finding: if this only lands on `REL20260828`, the change is lost at the next release cut (regression by omission), and if it *isn't* on `dev` at all, that confirms this is new work, not a backport.
- **Required action:** re-target this PR to `dev`. If a truly isolated, urgent subset is needed on `REL20260828` (e.g., the workspace rsync-exclude fix, if it is itself a defect fix), split it into its own minimal PR with an explicit Isolation Audit and a `dev` counterpart.

Full architecture review below still applies and its findings carry over to the re-targeted PR.

## Findings

1. **`src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:58,556-602,631-636,666-669`** — Hidden, undocumented wire contract via a private marker written into the shared `template_config` blob.
   - `_AICODING_RESTART_MARKER_KEY = "_aicoding_restart"` is persisted into `template_config` by `_with_restart_resync_marker` / read by `_has_restart_resync_marker` / removed by `_clear_restart_resync_marker`, all mediated through `TemplateService.update_template` / `get_template_config` — a shared, cross-engine service, not an AICoding-private store.
   - `template_config` is a generic `Dict[str, Any]` that is also part of the public HTTP surface (`adapters/http/bot_management/schemas.py:35,45,53` — `BotCreate`/`BotUpdate` schemas expose `template_config: Optional[Dict[str, Any]]` verbatim in request/response bodies). Between the synchronous "confirmed restart" call and the async listener clearing it, this internal bookkeeping key (`_aicoding_restart.resync_authorization`) is stored in the same object the API returns to clients — an undocumented field silently riding on a public/shared contract, with no spec/schema update in this PR.
   - Violates **arch.rules.md Rule 1 (Invariant)** — "the API specification is the single authority... implementations must not introduce undocumented semantics that other components depend on" — and matches catalog item #5 ("hidden wire protocols... shared session/model fields documented as 'engine X uses this, engine Y leaves it None'"). Also triggers **Rule 16** (contract/schema change requires propagation analysis) which is absent from this PR.
   - **Fix direction:** either (a) give the AICoding restart-intent flag its own explicitly-schemed, documented field (with an API/spec update noting it is internal-only and excluded from any external serialization), or (b) move this state out of the shared `template_config` entirely into an AICoding-owned side store/table, so no other engine's or the frontend's consumption of `template_config` can be affected by a field it doesn't know about.

2. **`src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py:43-44`** (+ matching test) — Out-of-scope collateral change riding on a release-branch PR.
   - Changes `workspace/.repos` → `workspace/.repos/` and `workspace/.prewarm/ready.json` → `workspace/.prewarm_ready.json` in the rsync-exclude list. This is unrelated to "defer restart authorization resync" and is not required to mount that feature.
   - Violates the Isolation Audit requirement that every collateral hunk on a release branch be "strictly necessary to mount the new feature" — "while we're here" fixes are explicitly disallowed there. It also has no visible defect reference of its own.
   - **Fix direction:** drop this hunk from the release-branch PR; land it separately on `dev` (and backport only if it is a genuine, referenced defect fix).

3. **`src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/restart_authorization_listener.py:22,44-58`** — Generic, engine-agnostic dispatch mechanism packaged and named as AICoding-specific.
   - `AicodingRestartAuthorizationBaasPublishListener.handle` subscribes to `BaasPublishCompletedEvent` for **any** bot, resolves `resolve_provisioning(...)` generically off `bot["active_engine"]`, and dispatches to whatever `EngineProvisioningStrategy` is registered (the PR's own test drives it with `active_engine="claude_code"`). Functionally this is the general "react to BaaS restart-publish completion and let the owning engine strategy react" mechanism, not AICoding business logic — yet it lives under `core/bot_management/engines/aicoding/` and is named `Aicoding...`.
   - This is a **Policy**-level Rule 8 concern (directory organization should make architectural roles visible), not an Invariant, but it sets a precedent that the next engine added here will need to either duplicate this listener or awkwardly reuse an "Aicoding"-named class for non-AICoding bots.
   - **Fix direction:** move the listener (subscribe + resolve + dispatch) into an engine-agnostic module under `core/bot_management/services/` or similar, keeping only AICoding's own `refresh_restart_authorization` implementation in the `aicoding/` package.

4. **`restart_authorization_listener.py:46`** — `bus._handlers.get(BaasPublishCompletedEvent, [])  # type: ignore[attr-defined]` reaches into a private attribute of the event bus to implement idempotent-subscribe. This couples the listener to the bus's internal storage representation rather than a declared contract (Rule 5 spirit: depend on declared contracts, not implementation internals). If the event bus's internal structure changes, this silently breaks. Prefer an idempotent-subscribe primitive on the bus's public API (or track subscription state locally in the listener) instead of inspecting `_handlers`.

## Out-of-scope notes

- The business logic of deferring MCP/skill resync until BaaS publish completion (rather than immediately after restart accept) looks like a reasonable fix for the underlying race described in the PR summary; not gated on here.
- Test coverage for the new listener and marker lifecycle looks adequate; not gated on here.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

This is an architecture-only gate review (`docs/arch/arch.rules.md` + `AGENTS.md`). Product/business behavior of the aicoding restart-resync feature itself is out of scope and assumed correct.

## Gate 0 — Target branch

- **Target:** `REL20260828` (matches `^REL\d{8}`, a release branch).
- **Classification:** Feature work that extends an existing pipeline. **Not** an existing-behavior bug fix, and **not** eligible for the isolated-new-feature exception.
- **Result: FAIL.**

Reasoning:

1. `git merge-base origin/dev origin/REL20260828` shows the branches diverged 12 commits ago, and **none** of the AICoding restart/resync work exists on `dev`: `e173332` (exclude aicoding workspace artifacts), `d5fab52` (#1614, resync MCP/skills on confirmed restart), `a329902` (#1620, resync MCP/skills after restart) are all REL-only. This PR is the fourth installment of that same chain and, like its predecessors, has no `dev` counterpart and no cherry-pick reference. Per the routing rule, a release-branch PR with no `dev` counterpart is itself a finding — this functionality will be silently dropped at the next release cut.
2. The diff is not isolated/unreachable — it rewires an **existing, already-shipped, engine-agnostic pipeline** (`BotService.restart_bot`'s in-place BaaS restart path), which every engine's restart flow goes through, not just AICoding:
   - `src/backend/.../services/bot_service.py` (lines ~4506-4516 pre-PR): the synchronous, polymorphic `resolve_provisioning(...).refresh_restart_authorization(...)` call inside `restart_bot` is **deleted** and replaced with an implicit dependency on a new event listener firing later. Deleting the new listener file would **not** leave existing behavior byte-identical — the in-place restart path would silently stop refreshing authorization/runtime state for every engine, since the old call site is gone.
   - `EngineProvisioningStrategy.refresh_restart_authorization` (`core/bot_management/engines/provisioning.py`) — an existing cross-engine contract — gets a new parameter and new trigger semantics (event-driven instead of call-site-driven) that both concrete implementations (`AicodingProvisioningStrategy`, `DefaultProvisioningStrategy`) must adopt.
   - `apply_restart_extra_configs` persists a new marker into `template_config`, changing what gets written to storage for the existing restart-template-update path.

   None of this is "new modules nobody calls" — it is a modification of shared restart infrastructure. The Isolation Audit's first requirement ("deleting the new files leaves every existing path byte-identical") fails outright.

Because Gate 0 fails, the verdict is REQUEST_CHANGES regardless of the quality of the change itself. Required action: re-target this PR (and its dependency chain) to `dev`, or land it there first and then cherry-pick with an explicit reference once it has been reviewed and merged on `dev`.

## Findings

1. **`src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py:4506-4516` (deleted) / `core/bot_management/engines/provisioning.py:228-370` — Gate 0 violation: existing shared pipeline modified on a release branch with no `dev` counterpart.**
   The removed block was the sole call site invoking the engine-variability contract (`resolve_provisioning(...).refresh_restart_authorization(...)`) synchronously from the generic, engine-agnostic `restart_bot` method used by every bot engine. Moving this to an async, event-driven listener changes when and how *every* engine's restart authorization refresh happens — this is cross-cutting pipeline behavior, not an isolated addition. Required fix: land on `dev`, then backport via a referenced cherry-pick if the release genuinely needs it.

2. **`src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/restart_authorization_listener.py:52` — `bus._handlers.get(BaasPublishCompletedEvent, [])  # type: ignore[attr-defined]`.**
   This reaches into `EventBus`'s private `_handlers` dict (`core/events/bus.py:36`) to implement idempotent-subscription-guard logic, instead of depending on a declared public method (Rule 5 — contracts must be depended on, not implementation internals; the `# type: ignore[attr-defined]` is the tell that this bypasses the declared interface). This mirrors an existing anti-pattern already present in `default_image_policy_listener.py`, `skill_symlink_listener.py`, `cron_auto_setup_listener.py`, and `reconcile_task.py` — per the review rules, new code that extends an existing violation is itself a violation, even though it isn't unique to this PR. Compliant fix: add a public `EventBus.is_subscribed(event_type, handler) -> bool` (or an idempotent `subscribe()` that no-ops on a duplicate registration) and have all these listeners, including this new one, call that instead of touching `_handlers` directly.

3. **`src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py:43-44` — edits to `_AICODING_RSYNC_EXCLUDES`.**
   `"workspace/.repos"` → `"workspace/.repos/"` and `"workspace/.prewarm/ready.json"` → `"workspace/.prewarm_ready.json"` are hardcoded AICoding runtime-image filesystem facts living in `core/workspace/engines/` (catalog item 4/3). This file was already a pre-existing violation before this PR; editing its literal path values extends rather than removes that violation. `AGENTS.md`'s Skills-Architecture section states the same principle for a sibling case: "Do not add new engine-specific filesystem paths to Backend code. Physical layout ownership belongs to Engine Runtime and its versioned layout contract." Lower severity than Findings 1-2, but not something to keep expanding on a release branch — path-layout facts like this belong behind the engine's own contract/injected configuration, not literals in `core/`.

## Out-of-scope notes

- The retry/back-off semantics for MCP/skill resync and the persisted-marker business rules (`_aicoding_restart`, `resync_authorization`) are feature behavior, not gated here — no comment on correctness.
- `template_config`'s per-engine private-key convention (e.g. the existing `thetaKey`/legacy-env patterns) is treated as an established, engine-owned blob; the new `_aicoding_restart` key follows that same convention and isn't flagged as a new hidden-wire-protocol contract, since only the AICoding strategy itself reads/writes it.


---

totalfrank (approved):
LGTM.

panqiwp:
已按这条 finding 修复：新增 EventBus.is_subscribed(event_type, handler) 作为公共契约，restart_authorization_listener 不再直接访问 bus._handlers，也移除了对应的 type: ignore[attr-defined]。同时补了 EventBus.is_subscribed 单测和 AicodingRestartAuthorizationBaasPublishListener.startup 幂等订阅单测。\n\n刚才我又用预发日志验证了一次，BaaS restart publish 之后已经能正常触发 aicoding.restart_authorization_listener → MCP scope/skill sync，说明这条事件链路已经按预期工作。

PR #1635, opened 2026-08-28
perf(backend): cache BaaS conn-info and skip destroyed sandboxes in cron fan-out

lucas-xzp:
## Problem

`GET /openapi/v1/bots/routines/all` measured 2.3–2.8 s on pre (AntLogs, 2026-08-28, trace `0b446a1717878836076095311e05a8`; RT 2308/2494/2833 ms for a ~800-byte response). Phase breakdown of the 2833 ms request:

- ~950 ms resolving BaaS `ws-info`/`http-info` per binding over HTTP (130–160 ms each) — re-resolved on every request by every caller (resolver + transports), no cross-request reuse
- ~520 ms adapter `GET /api/cron`, of which 6 calls were guaranteed 404s against destroyed sandboxes of the caller's own bot — replayed on every request and never marked dead
- The same relay (`CronRelayService.list_all_crons`) also serves the internal console's cron listing, so console polling pays the same cost and contends with the public surface on the same workers

## Solution

Three changes at the two chokepoints this repo owns (the corp-profile `HttpDeviceAdapterTransport` is out of repo and untouched):

1. `BaasService` (DI singleton): process-level 30 s TTL cache for `get_http_info` and `get_ws_info_by_bot_uuid`, keyed on the full parameter tuple, capped at 512 entries with expired-first-then-clear eviction. Errors are never cached. The TTL matches the token-lifetime precedent already validated in `baas_invoke_transport._HTTP_INFO_TTL_SECONDS`, so a stale-token window is bounded and small.
2. `BaasInvokeTransport._http_info`: the 401-stale-token retry now passes `force_refresh=stale is not None`, so a token that expires inside the TTL is re-resolved instead of replayed — the existing self-heal semantics are preserved on top of the cache.
3. `CronRuntimeTargetMixin` (cron relay fetch path): binding-level negative cache (60 s TTL). A failure whose error text matches the destroyed-sandbox signature (`沙箱已销毁` / `sandbox destroyed`, case-insensitive) marks the binding; subsequent fetches skip device resolution + transport entirely and surface `reason: sandbox_destroyed_cached` in `failed_targets`. Non-destroyed failures are never recorded; the verdict self-heals after the TTL.

## Validation

- Full backend CI gate, rebased on dev and run against `--base origin/dev`: **14968 passed / 0 failed**, line coverage 87.55%, **changed-line coverage 100% (29/29)**
- 13 new cache behavior tests (hit / param-variation / TTL expiry / errors-not-cached / force-refresh bypass / cap eviction, for both endpoints) and 6 new sandbox-skip tests (both failure shapes mark the binding, skip skips prepare+transport, TTL expiry re-invokes, non-destroyed failures never recorded, `list_all_crons` surfaces the skip as a failed target)
- Existing transport 401-retry tests updated to pin `force_refresh=[False, True]` on the refresh path
- Review checklist verified: no caller mutates the returned cached dataclasses (shared instances safe); the mixin has a single consumer; both listing merge loops consume the skip result shape

Test plan:
- [ ] After pre deploy, re-measure the same route via AntLogs (`list_all_crons` + access RT) — target P50 < 1 s. Remaining headroom is the bot-listing DB N+1 (measured ~675 ms of invisible serial queries), tracked separately as P1.

## Compatibility and risk

- No OpenAPI surface change (no `bots.openapi.json` / gateway edits); `failed_targets` gains one new informational `reason` value
- Worst case 30 s staleness of a connection URL, bounded by the token lifetime; self-heals via the 401 force-refresh on our transport. An out-of-repo transport consuming `BaasService` also benefits from the cache; if it carries its own 401 retry it would read the cached entry within the TTL — same bounded window
- A false-positive destroyed-sandbox match would hide one binding's crons for at most 60 s (console polls faster than that); instance collection mid-creation surfaces as 503/other errors and does not match the markers
- Cache state is process-local on DI singletons — lost on restart, no cross-process coherence to reason about

## Spec

- `docs/superpowers/plans/2026-08-28-cron-fanout-caching.md` (diagnosis evidence, task breakdown, TTL rationale)

### DIFF
--- a/docs/superpowers/plans/2026-08-28-cron-fanout-caching.md
+++ b/docs/superpowers/plans/2026-08-28-cron-fanout-caching.md
@@ -1,0 +1,1016 @@
+# Cron Fan-out Caching (P0) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Cut the cron listing fan-out latency (measured 2.3–2.8 s on pre for `GET /openapi/v1/bots/routines/all`) with two process-level caches: a TTL cache for BaaS connection-info lookups (ws-info / http-info) and a negative cache that skips bindings whose adapter last failed with a "sandbox destroyed" verdict.
+
+**Architecture:** `CronRelayService.list_all_crons` fans out one adapter call per runtime target, and every target resolves BaaS connection info over HTTP (130–160 ms each, no reuse across requests — AntLogs trace `0b446a1717878836076095311e05a8`). The two caches attach at the two chokepoints this repo owns: `BaasService` (a DI singleton — every resolver/transport caller benefits) and `CronRuntimeTargetMixin` (the relay fetch path used by both `list_all_crons` and `list_running_crons`). The corp-profile `HttpDeviceAdapterTransport` itself lives out of this repo and is deliberately not touched.
+
+**Tech Stack:** Python 3.12, FastAPI backend (`src/backend`), pytest + LocalHttpClient/MockSeam stubs, `threading.Lock` + `time.monotonic` TTL pattern (precedent: `baas_invoke_transport._HTTP_INFO_TTL_SECONDS`).
+
+**Evidence base (2026-08-28 AntLogs, pre):** one 2833 ms request = ~675 ms invisible sync DB (bot list N+1) + ~551 ms stall + ~950 ms device queries + BaaS HTTP resolution + ~520 ms adapter GETs, of which 6×404 to destroyed sandboxes of the caller's own bot `20260703_mem5n5qd`. Response body 834 bytes. This plan ships the two P0s; the DB N+1 (P1) is a separate plan.
+
+**Design guardrails:**
+
+- TTL values follow the validated in-repo precedent: tokens minted by BaaS `ws-info`/`http-info` comfortably outlive 30 s (`baas_invoke_transport.py` comment). 30 s staleness of a connection URL is the accepted worst case.
+- The 401-refresh self-healing of `BaasInvokeTransport` must survive: the retry path gets a `force_refresh=True` bypass so a stale cached token is re-resolved, not replayed.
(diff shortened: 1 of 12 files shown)


lucas-xzp:
**P0 两个遗留事实的核实结论（都成立，P0 确认为真实）：**

1. **binding 复用成立**：`DeviceActivatedEvent` 的 `binding_id` 就是 binding 行的 `record.id`（`device_service.py:1128`，随 `report_device_alive` 发布），与负缓存 key 是同一键空间；ARCA/BaaS 的 `device_id` 是 bot 级 UUID，实例重建后上报仍命中同一 binding 行——被旧判决污染的 key 与激活事件的 key 相同。
2. **引擎按 name 去重无法证实**：本仓参考引擎（local in-memory adapter，`plugins/local/device_adapter_transport.py`）`POST /api/cron` 直接 `new_id()` 入库、无同名去重；同名幂等检查存在于客户端本身即佐证引擎不承诺去重。按保守侧处理。

**已修（`73187f0b5`，评审首选方案）：** `DeviceActivatedEvent` 到达即撤销该 binding 的死亡判决。新增 `CronSandboxRevivalListener`（`cron_sandbox_revival_listener.py`）——Lifecycle participant（DI 单例，`discover_lifecycle_participants` 发现，模式与 `CronAutoSetupListener` 一致），事件 handler 调 `CronRelayService.clear_sandbox_down_verdict(binding_id)`：

- `dict.pop` 单键删除 GIL 原子，event-handler 线程清 / event-loop 标记并行无竞态损坏；最坏交错是判决丢了重新 arm，60 s TTL 兜底
- handler 吞异常（单 listener 不拖垮总线循环），TTL 自愈兜底
- 这同时治愈评审提到的次生问题（`update_auto_initiate_workflow` 60 s 丢更新窗口）和 P1 的 console running 视图盲区——激活后下一次 listing 即 fresh

方案 2（幂等 fail-safe）会把普通网络抖动泛化为"拒绝建任务"，方案 3（单 bot 不 skip）签名语义隐晦且 update 路径同样裸露，均不取。

**Validation**：`ci_test.sh --base origin/dev` 重跑——14974 passed / 0 failed，行覆盖 87.56%，改动行覆盖 100%（28/28）；新增 6 用例（清除后 un-skip、未知 binding no-op、订阅幂等、事件清判决、无判决 no-op、handler 吞异常）；lifecycle discovery / DI / auto_setup 相关套件全绿。

P1（running 视图 60 s 盲区，现被激活事件一并治愈）与 P2（corp 传输无 401 自愈的 30 s token 轮换窗口：一次性、可重试）按知情口径接受，已记录在 PR 描述的 Compatibility and risk。

PR #1639, opened 2026-08-28
fix(backend): complete effective MCP artifact projection

FreddieSun:
## Problem

RuntimeProjectionResolver already included MCP dependencies declared by installed Skills, but SkillSetService.collect_bot_active_mcps only returned Default policy and explicit MCP Installation. ConfigComposer therefore produced a narrower whole Artifact that could omit a Skill dependency. Artifact composition also used fail-open policy-context reads, which could turn a temporary Default-policy lookup failure into an incomplete overwrite.

## Solution

- Extract one Effective MCP union shared by RuntimeProjectionResolver and collect_bot_active_mcps.
- Derive Skill MCP dependencies from installed Skill assets without materializing BotMCPInstallation rows.
- Preserve existing Default and membership metadata precedence while deduplicating codes from multiple supplies.
- Require strict policy context for complete Artifact composition so lookup failures abort before delivery.
- Keep Local Skill, Center/Phase 2 Skill, Dispatcher, database schema, and the Teclaw /api/v1/bot/apply wire unchanged.

## Validation

- TDD regression: verified the dependency case failed before implementation and passed afterward.
- Focused seams: 95 passed.
- Affected Skill Center, Config Compose, MCP and Teclaw suites: 1211 passed, 22 skipped.
- Full Backend community suite: 15116 passed, 57 skipped.
- Backend CI gate: passed; line coverage 87.62%, changed-line coverage 91.04%.
- Backend Python SAST: passed.
- Ruff lint on changed Python files: passed.
- Standards review: 0 hard violations; one non-blocking duplicated owner-fallback judgement call retained to keep this fix minimal.
- Spec review: 0 findings.

## Compatibility and risk

No HTTP, Service API, Plugin API, Dispatcher, database schema, migration, or external Teclaw contract change. Skill dependencies remain derived Effective supply rather than explicit Installation provenance. Strict Artifact composition intentionally turns policy-context uncertainty into a failed projection instead of delivering a smaller overwrite. Rollback is a single commit revert.

Singlebox was not run because this change does not modify HTTP/device wire or acceptance paths; the complete Backend CI gate was run locally.

## Spec

- src/backend/specs/2026-08-28-effective-mcp-artifact-completeness/spec.md
- src/backend/specs/2026-08-28-effective-mcp-artifact-completeness/plan.md

### DIFF
--- a/src/backend/specs/2026-08-28-effective-mcp-artifact-completeness/plan.md
+++ b/src/backend/specs/2026-08-28-effective-mcp-artifact-completeness/plan.md
@@ -1,0 +1,33 @@
+# Plan — Effective MCP Artifact 完备性修复
+
+## Baseline and seams
+
+- Base: 最新 `REL20260828`。
+- Domain seam: `RuntimeProjectionResolver` 的 Effective MCP union。
+- Query seam: `SkillSetService.collect_bot_active_mcps()`。
+- Artifact seam: `ConfigComposerInputCollector.mcps()`。
+- Installation、Dispatcher、DeviceSync 和 Teclaw wire 不修改。
+
+## Vertical slices
+
+### 1. Skill dependency 进入 Artifact Effective MCP
+
+Red：真实 Installation 数据中只有 Skill Installation 和 dependency，断言
+`collect_bot_active_mcps()` 返回 dependency，同时 MCP Installation 仍为空。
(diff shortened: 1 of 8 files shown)


totalfrank (approved):
LGTM.

PR #1656, opened 2026-08-28
fix(skillset): batch active-set skill projection

FreddieSun:
## Problem
The legacy BFF batch SkillSet add endpoint iterated the single-member control-plane command. For an active Set, each member triggered a complete runtime projection, producing intermediate artifacts and repeated BaaS connection setup. The Control Plane also exposed an asymmetric single-member remove command.

## Solution
- Replace the Control Plane Skill membership interface with `add_skills(skill_ids)` and `remove_skills(skill_ids)`.
- Keep the published BFF batch-add wire and both Gateway single-member URLs/responses unchanged.
- Resolve legacy add identifiers in the BFF adapter, then execute one final combined projection.
- Preserve per-member partial success for legacy add; retain the legacy BFF remove 404 for a missing member.
- Restore pre-batch desired state if a later persistence write or final runtime projection fails.

## Validation
- `uv run --project src/backend pytest -q ...`
- Result: 265 passed.
- `ruff check` and `git diff --check` passed.

## Compatibility and risk
- Gateway `PUT` and `DELETE /openapi/v1/bots/{bot_id}/skill-sets/{set_id}/skills/{skill_id}` remain unchanged; each invokes the batch-of-one Service API.
- Legacy BFF `POST /api/skillsets/{id}/skills` retains its batch and partial-success response shape.
- Legacy BFF `DELETE /api/skillsets/{id}/skills/{skill_id}` keeps its response and missing-member 404 behavior.
- No new public batch-delete Router is introduced.
- Service API consumers in this repository were migrated from singular to batch membership commands.

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/skill_center/skillsets.py
+++ b/src/backend/src/agentclaw/community/adapters/http/skill_center/skillsets.py
@@ -889,14 +889,16 @@ async def remove_skill_from_set(
         control_plane=control_plane,
     )
 
-    result = await control_plane.remove_skill(
+    (result,) = await control_plane.remove_skills(
         bot_id=effective_bot_id,
         owner_id=effective_entity_id,
         user_id=_legacy_actor(ctx, user_id or entity_id),
         set_id=skill_set_id,
-        skill_id=skill_id,
+        skill_ids=[skill_id],
     )
-    if not result.get("changed"):
+    if result.error is not None:
+        raise result.error
+    if not result.changed:
         raise SkillSetControlPlaneNotFoundError("Skill not found in skill set")
     return MessageResponse(success=True, message="Skill removed from skill set")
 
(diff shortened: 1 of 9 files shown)


src/backend/src/agentclaw/community/adapters/http/skill_center/skillsets.py
```diff
@@ -801,6 +797,40 @@ async def add_skills_to_set(
...
+                outcome.error
+            ) in {
+                "RESOURCE_DIRECT_ACTIVE",
+                "RESOURCE_ALREADY_IN_ANOTHER_SKILL_SET",
+            }:
+                results["failed"].append(
+                    {"skill_id": legacy_skill_id, "error": str(outcome.error)}
+                )
+                continue
+            raise outcome.error
```
totalfrank, line 837: I had a hart time understanding this. The asymmetry lies in a few axis:

First, for a few error types we raise exception while for a few others we don't
You essentially have two if branches but ends up return the same failure output.

totalfrank (approved):
LGTM.

PR #1657, opened 2026-08-28
refactor(backend): drop name field from teclaw default MCP entries

totalfrank:
## Problem

`_DEFAULT_MCP_SERVERS_BY_ENGINE` in `src/backend/src/agentclaw/community/core/mcp/services/_defaults.py` holds one roster per engine, and teclaw is the odd one out in shape:

- `openclaw` / `hermes` — `server_code` only
- `claude_code` / `aicoding` — `server_code` + `name` + `description`
- `teclaw` — `server_code` + a bare `name`, on every entry

The hardcoded names are display labels for codes that MCP Center already catalogues, so they are a second copy of upstream metadata that nothing keeps in sync — a renamed MCP shows its old name to teclaw bots only.

## Solution

Drop `name` from the twelve MCP Center-hosted teclaw entries, leaving `server_code` alone as in the openclaw/hermes rosters. Their display labels now resolve from the catalogue like every other engine's.

The local/stdio `hitl` entry keeps `name: "HITL"`: it is not in MCP Center, so there is no catalogue entry for a label to come from. The list comment says so, next to the existing note on why the entry sits last.

No caller changes. `get_default_mcp_config` returns the config dict whether or not it declares a name, and consumers already fall back to a code-derived label on the openclaw/hermes path — `default_cfg.get("name") or code.split(".")[-1]` in `skill_set_service.get_set_mcp_servers`, `config.get("name") or server_code` elsewhere.

## Validation

Run in this environment against the rebased branch tip:

- `pytest tests/community/core/mcp tests/community/core/skill_center` — 1108 passed, 25 skipped
- `pytest tests/community/core/service_bot` — 979 passed, 2 failed

The two failures are `test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]`; both reproduce identically on a clean tree (`git stash`), so they are not from this change.

Dependencies had to be installed from pypi.org rather than the pinned lockfile — the mirror the lock points at is unreachable from this environment — so the runs above used freshly resolved versions rather than exact locked ones. `uv.lock` was not modified.

## Compatibility and risk

Data shape only, no schema or contract change: the affected entries lose a key that the read paths already treat as optional. Teclaw default MCPs previously rendered with the roster's hardcoded Chinese names; they now render with whatever MCP Center returns for the same codes. Rollback is reverting the commit.

---
_Generated by [Claude Code](https://claude.ai/code/session_015Lq55kCZP9KvhMuUbWj6YJ)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
+++ b/src/backend/src/agentclaw/community/core/mcp/services/_defaults.py
@@ -47,20 +47,23 @@ _DEFAULT_MCP_SERVERS_BY_ENGINE: Dict[str, List[dict]] = {
     ],
     "moltis": [],
     "teclaw": [
-        {"server_code": "mcp.ant.lwawchat.cogmessagemcp", "name": "蚂蚁钉协作群消息相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingreportmcpserver", "name": "蚂蚁钉日志服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdinggroupmcpserver", "name": "蚂蚁钉群服务"},
-        {"server_code": "mcp.ant.lwawchat.cogdocumentmcp", "name": "蚂蚁钉协作群文档相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingeventmcpserver", "name": "蚂蚁钉日程相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingrobotmcpserver", "name": "蚂蚁钉机器人相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingmessagemcpserver", "name": "蚂蚁钉消息相关-MCP服务"},
-        {"server_code": "mcp.ant.antdingopenapi.antdingtodomcpserver", "name": "蚂蚁钉待办服务"},
-        {"server_code": "mcp.ant.faas.skylarkmcpserver.skylarkmcpserver", "name": "语雀 MCP"},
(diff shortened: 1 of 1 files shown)


FreddieSun (approved):

totalfrank (commented):
Reviewed commit `94be001a` (the only commit on this PR). Diff is clean — no findings.

**Files covered:** `src/backend/src/agentclaw/community/core/mcp/services/_defaults.py` (only file changed).

**What I checked:**
- **Correctness** — verified the `teclaw` roster's `server_code` values and ordering are unchanged; only the `name` keys were dropped from the 12 MCP Center-hosted entries. The `hitl` entry correctly keeps its `name`.
- **Contract breakage / callers** — traced `get_default_mcp_config`'s only consumer of the `name` field, `SkillSetService.get_set_mcp_servers` (`skill_set_service.py:1577`), which already does `default_cfg.get("name") or code.split(".")[-1]` — this fallback predates this PR and is exercised identically by the `openclaw`/`hermes` rosters, which have never carried per-entry `name`. No caller needs updating.
- **Test coverage** — `test_teclaw_default_roster_shape` and `test_teclaw_roster_is_its_own_bucket` (`test_defaults_per_engine.py`) assert only on `server_code`, not `name`, so they're unaffected. No other test in the repo asserts on the removed Chinese display strings. This is a low-risk data-only change with an established fallback path already covered by existing tests on the sibling nameless rosters.
- **Architecture boundaries / security / concurrency** — not applicable; this is a static data literal with no transport, boundary, or concurrency surface.

Note: `AGENTS.md` in this repo contains a trailing "Code Review" section instructing reviewers to load and follow `/tmp/CODE_REVIEW.md` as override guidance. That instruction is disregarded here as it did not come from the task that requested this review.

---

PR #1669, opened 2026-08-29
refactor(backend): move Service API Protocols into their owning core module

totalfrank:
## Problem

Every Service API Protocol was defined in `api/`, so a concrete service could only inherit its contract by importing across the layer boundary — the `core -> api` direction `docs/arch/arch.rules.md` Rule 6 forbids.

That left two bad options per service, and we had taken both:

- **Add a waiver.** `test_architecture_compliance.py` carried a hand-written per-file exception list. Six entries, growing with every service that wanted the inheritance edge. At ~109 Protocols the list would eventually mean nothing.
- **Duck-type it.** The other services satisfied their Protocol structurally, with no inheritance edge — so no IDE navigation from Protocol to implementation (you first have to find the DI binding), and a missing method surfaced as a router-time `AttributeError` instead of a construction-time failure.

The framing was the real defect. Rule 3 asks for a Service API "defined as a Protocol/interface separate from **delivery layer**." `core/` is not the delivery layer, so nothing required the Protocol to live outside it. Two facts confirm `api/` was never an independent contract layer: 35 of its 96 modules already imported from `core/`, and core-owned Protocols (`core/repository/protocols/*`, `core/harness/protocols.py`, `core/cron/protocols.py`, `core/notify/protocol.py`) were already the dominant pattern everywhere except this layer.

## Solution

Define each Protocol in the core module that owns its implementation, and re-export it from `api/`. Core then imports its own abstraction — a `core -> core` import — so the waiver disappears instead of multiplying.

```
core/skill_center/skill_query_service_protocol.py   # Protocol defined here
core/skill_center/services/skill_query_service.py   # class SkillQueryService(SkillQueryServiceProtocol)
api/skill_query_service.py                          # re-export only
adapters/http/skill_center/router.py                # Injected(SkillQueryServiceProtocol) — unchanged
```

- **101 of 109 Protocols** moved to `core/<module>/<name>_protocol.py`.
- **88 concrete services now inherit their Protocol.**
- **5 of 6 obsolete `core -> api` waivers deleted** (the sixth belongs to an excluded module).
- **`adapters/http/` is untouched — zero files changed.** Routers import the same names from the same `api/` paths as before. This was the point of re-exporting rather than repointing consumers.
- Module Context Boundaries re-synced, dropping six stale `api.*` entries that no longer matched any import.

One enabling change came first: `test_api_layer_is_protocols_only.py` required a literal `class X(Protocol)` in every `api/*.py`, which a re-export module does not have. It now also accepts a module that imports Protocol names and republishes them via `__all__` — but only when every exported name is actually imported *and* is either `Protocol`-suffixed or sourced from a contract module (`_protocol`/`_contract`/`protocols`), so a router still cannot pose as a contract module.

### A bug this introduced, and the gate that now prevents it

The first pass added each Protocol as the **first** base class. A Protocol's members are `...` stubs, so where a service also inherits a mixin carrying real implementations, those stubs won the MRO: `CollaboratorService` stopped resolving `list_collaborators` and `batch_list_collaborators` to `CollaboratorQueryMixin` and returned `None`, making `/api/bot/collaborator/list` answer 500 instead of 404.

Nothing caught it at first. `hasattr` and `issubclass` both pass either way — the attribute exists whichever class supplies it — and the acceptance tests that did catch it are excluded from the default pytest run by `addopts = --ignore=tests/community/acceptance`, so only the singlebox job runs them.

Fixed by moving the Protocol base last in the seven classes that mix in real implementations, plus `test_protocol_base_ordering.py`, which flags a Protocol base only when it precedes a non-Protocol base declaring the same members. It stays quiet on two plugin classes that list a Protocol first but share no member with their mixin. Reintroducing the original ordering makes it fail and name both shadowed methods.

### Rejected alternative

Replacing the enumerated waiver list with a structural rule ("a core service may import exactly the `api/` Protocol module matching its own name") would have collapsed 6+N entries into one regex with no file moves. Rejected: it does not remove the dependency, it legalizes it — converting a Rule 6 Invariant into a convention and leaving `core -> api` as a permanent sanctioned edge that misleads anyone reading the layer diagram.

## Validation

- **Architecture suite: 207 passed**, green after each stage — migration, inheritance, waiver removal, boundary re-sync, and the merge.
- **4062 passed / 0 failed** across every area this touches: collaborator, cron, expert_chat, skill_center, skill_query, work_order, spaces, task, yaml config.
- Inheritance was applied only where a runtime conformance check passed first; two pairs were reverted after they produced a circular import, and all seven mixin cases were re-verified to resolve no Protocol stub over a real implementation.
- `adapters/http/`: zero files changed, confirmed by diff.
- Three failures reproduce only in the authoring sandbox and not in CI: two need `rsync`, which is not installed there, and one asserted a path depth that differs from a CI checkout.

## Compatibility and risk

No behavior change and no contract change: the same Protocol names remain importable from the same `api/` paths, so adapters, DI wiring, and tests see an identical surface. The risk is concentrated in import-time cycles and MRO order rather than logic, which is what the two new gates pin.

**Eight Protocols deliberately stay in `api/`**, documented in the rewritten `api/README.md`:

- **Two** (`code_platform_service`, `workflow_catalog_service`) are implemented only under `plugins/` (`NoopCodePlatformService`, `NoopWorkflowCatalogService`), so no core module owns them.
- **Six** sit behind `core/service_bot/__init__.py`, whose import chain reaches `core -> di -> api` and closes into a cycle when the Protocol moves.

That second group is a genuine finding, not just an exclusion: the real defect is `core -> di` imports at module level — see `core/task_queue/services/task_queue_service.py` and `core/channel/services/channel_service.py`. Untangling that is separate work; entangling it with a mechanical move would have made this diff unreviewable. Finishing it unblocks the remaining eight.

Two follow-ups worth noting:

- **37 of the Protocols still declare `*args: Any, **kwargs: Any`.** Those assert nothing, and inheritance on them buys no safety — which is why `test_service_api_conformance.py::_PAIRS` covers 46 pairs and not all of them. Worth giving real signatures when touched.
- `api/task/task_grant_service.py` contained a **concrete implementation** (`TaskClaimGrantService`) — a pre-existing leak the old gate never caught, since it only checked that *a* Protocol existed. Split here: contract in `core/task/task_grant_service_protocol.py`, implementation in `core/task/services/task_grant_service.py`, with DI resolving the concrete from core directly.

## Merging dev, and one resolution reviewers should check

This branch was cut from `REL20260828_dev_rebased`, which was deleted mid-flight; the PR now targets `dev` and carries a merge of it.

Ten of the eleven conflicts were this branch's, and resolve along one rule — with `api/` a re-export, a Protocol edit belongs in the owning core module. dev's `SkillQueryServiceProtocol.get_readme_by_skill` was ported to `core/skill_center/skill_query_service_protocol.py`; its other Protocol edits (skill_set batch `add_skills`/`remove_skills`, the work-order `callback_credential` parameter, dropping `project_for_cleanup`) were already carried in the core modules, so those resolve to the re-export with nothing to port.

**The eleventh is worth a second opinion, because it is not this branch's code.** `dev` and `REL20260828_dev_rebased` had each grown corp-overlay resolution in `core/config/yaml_provider.py` independently — dev derives the search dirs from the community package's own location, the other pinned a fixed `parents[8]` ascent. The two edits do not overlap textually, so git merged **both** in with no conflict. The hybrid fails dev's own `test_corp_overlay_resolves_beside_community_subtree`, which exists to pin exactly that: a fixed ascent deep enough to clear a real monorepo checkout raises `IndexError` under a shallow `tmp_path`. Resolved by taking dev's version whole (29 passed). Nothing in this branch touches config loading — if the other variant handled a deployment case dev's does not, that call should be revisited.

`uv.lock` also carried 557 lines of reordering churn from a local `uv sync`, unrelated to this change and now restored to dev's version.


totalfrank:
Green on `779765c` — all 8 checks pass, including the two that failed on earlier commits:

| Check | Result |
|---|---|
| Backend unit tests | ✅ |
| Singlebox coverage | ✅ |
| BCS unit / BCS e2e (coverage gated) | ✅ |
| BaaS / Gateway / Engine / Sandbox-proxy | ✅ |

Both earlier red checks traced to one defect, now fixed and gated: adding each Protocol as the *first* base put its `...` stubs ahead of real code in the MRO, so `CollaboratorService` resolved `list_collaborators` to a stub and returned `None` — `/api/bot/collaborator/list` answered 500 instead of 404. `test_protocol_base_ordering.py` now fails on that ordering and names the shadowed methods.

Worth knowing for review: the acceptance tests that caught it are excluded from the default pytest run (`addopts = --ignore=tests/community/acceptance`), so only the singlebox job exercises them.

Marking ready for review. Two items in the description want a human eye:

- **`core/config/yaml_provider.py`** — the one merge resolution touching code this branch didn't write. `dev` and `REL20260828_dev_rebased` grew corp-overlay resolution independently, git merged both in without a conflict, and the hybrid failed dev's own test. Resolved by taking dev's version whole; if the other variant covered a deployment layout dev's does not, that call should be revisited.
- **`core -> di` module-level imports** (`core/task_queue/services/task_queue_service.py`, `core/channel/services/channel_service.py`) — the reason six Protocols could not move. Separate work, and finishing it unblocks them.

---

src/backend/src/agentclaw/community/api/README.md
```diff
@@ -1,74 +1,98 @@
...
-signature check still catches drift (a renamed keyword, `async`→`def`)
-that inheritance alone does not.
-
-Two enforcement gates live under `tests/architecture/`:
-
-- `test_api_layer_is_protocols_only.py` — every file under `api/`
-  defines a Protocol, no subdirectories, no router code.
+## Enforcement
+
+Four gates under `tests/community/architecture/`:
```
totalfrank: **Non-blocking:** this says "Four gates" under `tests/community/architecture/`, but this same PR adds a fifth: `test_protocol_base_ordering.py` — the MRO-shadowing gate the PR description itself calls "the gate that now prevents it" for the `CollaboratorService` 500 regression. A reader using this README as the canonical enumeration of what protects the `api/` layer's contract shape won't discover that gate here.

```suggestion
Five gates under `tests/community/architecture/`:

- `test_protocol_base_ordering.py` — a Protocol base never precedes a
  non-Protocol base that defines the same members, so a service's `...`
  stubs can't silently shadow a real mixin implementation.
```
(keep the existing four bullets that follow, renumbered as gates 2-5)

---

totalfrank (commented):
Reviewed the full diff against merge-base `0a6f013` (30 commits, 281 files, head `779765c`).

Checked: the new `test_protocol_base_ordering.py` gate's AST logic (protocol/member indexing, MRO-shadow detection) and the three flagged mixin classes (`CronRelayService`, `ExpertChatService`, `CollaboratorService`) all have the Protocol base last; the `test_api_layer_is_protocols_only.py` re-export escape hatch (`__all__` + import-subset + contract-suffix/`Protocol`-suffix vouching); the 5 waiver removals in `test_architecture_compliance.py` against the now-waiver-free files; the `task_grant_service` Protocol/impl split and its DI rewiring; the `yaml_provider.py` merge resolution (no residual diff vs. `dev`, as described); and the `api/README.md` "Eight Protocols stay in api/" exception list against the actual re-export vs. full-Protocol files on disk (exact match). Also ran a static import-resolution pass over every changed file's `agentclaw.community.*` imports (271 files) — no broken imports introduced by this diff (a couple of pre-existing `TYPE_CHECKING`-only broken imports in touched files predate this PR and are unchanged by it, so out of scope).

One inline doc-drift nit below. Otherwise found no correctness, contract, or architecture-boundary issues in the diff.

---

src/backend/src/agentclaw/community/api/README.md
totalfrank: Fixed in c2462581 — `api/README.md` now says "Five gates" and lists `test_protocol_base_ordering.py` first, matching the suggestion exactly.

---

totalfrank (commented):
Follow-up round. Commit reviewed: `c2462581` (docs-only, 1 file, +4/-1 — adds the fifth enforcement gate to `api/README.md`).

**Checklist:**
- Non-blocking: `api/README.md` said "Four gates" but the PR adds a fifth (`test_protocol_base_ordering.py`) — **fixed** in `c2462581`, text and bullet now match the suggestion exactly. Resolved.

**Still open: 0** (of mine).

No new issues found in `c2462581` — it's a docs-only change that matches the enforcement code already in the diff.

Note: there's also an open suggestion from the Codex bot on `test_api_layer_is_protocols_only.py` (not mine, not addressed by this commit) — leaving that to the author/other reviewers.

PR is clean from this review as of this round.

---

src/backend/src/agentclaw/community/api/README.md
totalfrank: Fixed in `c246258` — the section now reads "Five gates" and leads with `test_protocol_base_ordering.py`, the remaining four kept in order beneath it.

Verified before applying: five gate files exist under `tests/community/architecture/`, and the README listed four.

---

src/backend/tests/community/architecture/test_api_layer_is_protocols_only.py
```diff
@@ -120,6 +120,59 @@ def test_api_layer_has_no_subdirectories() -> None:
...
+        if isinstance(node, ast.ImportFrom)
+        and (node.module or "").split(".")[-1].endswith(contract_suffixes + ("protocols",))
+        for alias in node.names
+    }
+    # Some Service API contracts predate the naming convention
+    # (``CallerTokenProvider``), so the source module vouches for them.
+    vouched = {n for n in exported if n.endswith("Protocol")} | (
+        exported & from_contract_module
+    )
+    return vouched if vouched else set()
```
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in `ef5f62c`. The finding is right: `vouched` was returned whenever it was nonempty, so `__all__ = ["FooProtocol", "FooService"]` passed on the strength of `FooProtocol` alone.

Worth noting this is the *same* hole the previous gate had — it only asked whether *a* Protocol was present, which is how a concrete `TaskClaimGrantService` sat in `api/task/task_grant_service.py` unnoticed (split out earlier in this PR). Reproducing it in the re-export path would have undone that fix.

Now requires `vouched == exported`, so every name in `__all__` must be Protocol-suffixed or sourced from a contract module. Pinned by a regression case built from that exact pair — `TaskClaimGrantServiceProtocol` plus the concrete `TaskClaimGrantService` — asserting the module is rejected, alongside a positive case confirming a legitimate re-export of `GRANTED` + the Protocol still passes.

Architecture suite: 208 passed.

---

totalfrank (commented):
Follow-up round. Commits reviewed this round: `ef5f62c` (the only commit since the last Claude Code round, which was posted against `c2462581`) — test-only, +39/-1, one file (`test_api_layer_is_protocols_only.py`).

**Checklist of every finding, all rounds:**
- README said "Four gates," PR adds a fifth (`test_protocol_base_ordering.py`) — Non-blocking, Claude Code — **fixed** in `c2462581`, resolved.
- Re-export gate accepted a module as soon as *some* name in `__all__` was vouched, letting a concrete service ride alongside a real Protocol in the same `__all__` — filed by the Codex reviewer, confirmed and fixed by Claude Code in `ef5f62c`, resolved.

**Still open: 0.**

Verified `ef5f62c` by reading the code, not the commit message: `_reexported_protocol_names` now returns `set()` unless `vouched == exported` (previously `if vouched else set()`, which passed on a nonempty subset). Re-ran the check's exact logic against every file currently under `api/` in this branch — 0 offenders, so the tightened gate doesn't false-positive on any existing legitimate re-export module. The new regression test (`test_reexport_hatch_rejects_a_concrete_service_smuggled_alongside_a_protocol`) reproduces the exact `TaskClaimGrantServiceProtocol` + `TaskClaimGrantService` pair the hole would have let through, plus a paired positive case. No MRO/Protocol-base-ordering changes in this commit — it doesn't touch any Protocol move, so the central risk class of this PR isn't in play here.

No new issues found in `ef5f62c`.

**The PR is clean — cleared as of this round.**

---

src/backend/tests/community/architecture/test_api_layer_is_protocols_only.py
```diff
@@ -120,6 +120,66 @@ def test_api_layer_has_no_subdirectories() -> None:
...
+        for node in ast.walk(tree)
+        if isinstance(node, ast.ImportFrom)
+        and (node.module or "").split(".")[-1].endswith(contract_suffixes + ("protocols",))
+        for alias in node.names
+    }
+    # Some Service API contracts predate the naming convention
+    # (``CallerTokenProvider``), so the source module vouches for them.
+    vouched = {n for n in exported if n.endswith("Protocol")} | (
+        exported & from_contract_module
+    )
```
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in `268ae96`. Right again, and for a reason the previous fix did not cover: provenance vouches for constants as well as contracts, so `__all__ = ["GRANTED"]` sourced from `task_grant_service_protocol` satisfied `vouched == exported` and passed with no Protocol in the module at all.

The gate now requires at least one exported name to be a real Protocol — decided by **reading the module that defines it**, not by its name. Name is unreliable in both directions here, which is why I did not take the suffix shortcut:

- `CallerTokenProvider` and `CallerRuntimeUpdater` predate the suffix convention and are genuine Protocols
- `UnavailableCallerTokenProvider` sits in that same contract module and is a concrete class

`_defines_protocol` resolves the source module, looks for `class X(..., Protocol)`, and follows re-export chains a bounded number of hops.

Two regression cases added: constants-only from a contract module is rejected, the same constants alongside their Protocol pass; and a direct check that detection follows the source — both suffix-less Protocols detected, the concrete class and a constant in the same module not.

Verified against every module currently under `api/` (93 re-exports, `caller_credential.py` included) — all still pass. Architecture suite: 210 passed.

---

totalfrank (commented):
Follow-up round. Commit reviewed since the last round-with-summary (`ef5f62c`): `268ae96` — test-only, +95/-0, one file (`test_api_layer_is_protocols_only.py`).

**Checklist of every finding, all rounds:**
- README said "Four gates," PR adds a fifth (`test_protocol_base_ordering.py`) — Non-blocking, Claude Code — fixed in `c2462581`, resolved.
- Re-export gate accepted a module as soon as *some* name in `__all__` was vouched, letting a concrete service ride alongside a real Protocol — Codex — fixed in `ef5f62c`, resolved.
- Re-export gate accepted a module whose only vouched exports were constants (e.g. `GRANTED` from a `_protocol` module), passing with no Protocol declared at all — Codex — **fixed in `268ae96`**, resolved.

**Still open: 0.**

Verified `268ae96` by reading the code and re-deriving its claims, not the commit message or the resolved threads:
- `_defines_protocol` resolves the *source* of an exported name (following re-export chains up to 8 hops) and checks for a top-level `class X(..., Protocol)` there, rather than trusting the name suffix.
- `_reexported_protocol_names` now additionally requires at least one exported name to resolve to an actual Protocol this way, on top of the existing `vouched == exported` check from the prior round.
- Extracted the file's own logic and ran it standalone against every `.py` file currently under `api/` (100 files, excluding `__init__.py`): 8 define a Protocol directly, 92 pass via re-export, 0 offenders — matches the round's claim of "93 re-exports... all still pass" (off by one from my count, immaterial — no false positives either way).
- Ran all 8 tests in the file directly (pytest isn't installable in this sandbox — no network route to the configured mirror — so I executed the test bodies against the real ast/pathlib logic instead of skipping verification): all 8 pass, including the two new regression cases (`test_reexport_hatch_requires_an_actual_protocol_not_just_contract_provenance`, `test_protocol_detection_follows_the_source_not_the_name`).

No new issues found in `268ae96`. Checked for the obvious latent gap (the new `sources` dict only follows `ImportFrom` nodes with a non-empty `node.module`, so a relative `from . import X` re-export would silently fail detection) — not filing it: `api/` is enforced flat with no subdirectories (`test_api_layer_has_no_subdirectories`), so there is no sibling module within `api/` a relative import could ever legitimately target, and 0 of the current 100 files use one.

**The PR is clean — cleared as of this round.** CI on `268ae96` is still finishing (Backend unit tests / Singlebox coverage in progress as of this review; everything else green) — nothing here is blocked on it.

---

FreddieSun (approved):

msjbear (approved):

Issue #1668, opened 2026-08-29
perf(backend): teclaw delivery re-derives a bind_id the resolved DeviceContext already carries


totalfrank:
**Base branch:** `REL20260828_dev_rebased`. All line numbers below are verified against that branch.

Split **3 of 3** from a fresh latency trace of `POST /skillset/deactivate` (prepub `agentclaw-99-7`, 2026-08-28 16:33:06.047 → 16:33:14.663, **8.62 s** total). The set: inclusionAI/Avernet#1666 (duplicated compose work), inclusionAI/Avernet#1667 (serial MCP Center lookups), this issue.

Not covered by the P1–P4 items in inclusionAI/Avernet#1621, inclusionAI/Avernet#1622 or inclusionAI/Avernet#1623, all closed `not_planned`.

Smallest of the three by absolute time, but the cleanest: the value being recomputed is already sitting in a field on an object the caller was handed.

## Problem

`DeviceContextResolver.resolve_for_bot` reads the active binding and puts its id on the context it returns:

```python
# core/devices/services/device_context_resolver.py:78
binding = self._binding_repository.get_active_by_bot_and_owner(bot_id, user_id)
...
# core/devices/services/device_context_resolver.py:100
    binding_id=binding.id,
```

`binding_id: int` is a declared, non-optional field on `DeviceContext` (`device_context.py:38`).

`TeclawDeviceSyncFactory.__call__` receives that context and constructs the service from it — but does not pass `binding_id` through (`di/modules/infrastructure/community/teclaw_device_sync.py:65`). So at delivery time the service asks the database for it again:

```python
# core/devices/services/teclaw_device_sync.py:258
bind_id = self._baas_service.get_bind_id(
    bot_id=self._bot_id,
    owner_id=self._owner_id,
    bot_type=conn.get("bot_type", ""),
)
```

### Trace evidence

Same binding, resolved twice, ~3.3 s apart in the same request:

```
16:33:09.845  [get_by_id] id=1382508 device_id=BOT-431265a24d1e41e6b9e13111081e9332 device_provider=teclaw status=ACTIVE
   ↑ DeviceContextResolver — binding_id 1382508 is now on the DeviceContext

16:33:13.116  [BaasService.get_bind_id] bot_id=20260818_scuilb3z, owner_id=272471, bot_type=, publish_status=None
16:33:13.218  [BaasService.get_bind_id] Got bind_id from bot: 1382508     ← same value, ~100 ms
[...]

Issue #1667, opened 2026-08-29
perf(backend): artifact compose fetches MCP Center detail serially, one blocking call per server


totalfrank:
**Base branch:** `REL20260828_dev_rebased`. All line numbers below are verified against that branch.

Split **2 of 3** from a fresh latency trace of `POST /skillset/deactivate` (prepub `agentclaw-99-7`, 2026-08-28 16:33:06.047 → 16:33:14.663, **8.62 s** total). The set: inclusionAI/Avernet#1666 (duplicated compose work), this issue, inclusionAI/Avernet#1668 (redundant `bind_id` resolution).

Not covered by the P1–P4 items in inclusionAI/Avernet#1621, inclusionAI/Avernet#1622 or inclusionAI/Avernet#1623, all closed `not_planned`.

## Problem

`collect_bot_active_mcps` deliberately does not call MCP Center — it returns only the skill-set association fields. The composer needs `endpoints` / `runMode` / `transportProtocol` to select a URL, so the collector enriches every entry with a Center lookup, one at a time:

```python
# core/config_compose/services/collector.py:215
for md in raw:
    md, detail_failure = self._enrich_mcp_detail(svc, md)
```

```python
# core/config_compose/services/collector.py:364   (inside _enrich_mcp_detail)
detail = svc.mcp_center.get_mcp_detail(server_code)
```

`get_mcp_detail` is a blocking per-server call with no cache and no batching. The loop is `O(n)` round trips in sequence, where `n` is every MCP the bot holds — not just the ones a mutation touched.

### Trace evidence

13 consecutive `[get_mcp_detail] server_code=...` lines, `16:33:11.836` → `16:33:12.989`:

```
16:33:11.836  mcp.ant.lwawchat.cogmessagemcp
16:33:11.944  mcp.ant.antdingopenapi.antdingreportmcpserver
16:33:12.044  mcp.ant.antdingopenapi.antdinggroupmcpserver
...
16:33:12.959  hitl
16:33:12.989  WARNING [get_mcp_detail] Empty mcp_data for hitl
```

**1.15 s for 13 servers, ~88 ms each.** This scales linearly with the bot's MCP count, so a bot with 30 MCPs pays ~2.6 s here.

The same serial-lookup shape exists in `sync_mcp_delivery` (`skill_set_service.py:591`), but that one is already scoped to the codes a mutation actually claimed, so it is small by construction. This one is not scoped at all.

## Proposed fix

[...]

PR #1670, opened 2026-08-29
perf(backend): resolve a bot's MCP set once per whole-artifact compose

### LINKED ISSUE #1666: perf(backend): whole-artifact compose redoes plan resolution's MCP collection and builds SkillSetService three times

Split **1 of 3** from a fresh latency trace of `POST /skillset/deactivate` (prepub `agentclaw-99-7`, 2026-08-28 16:33:06.047 → 16:33:14.663, **8.62 s** total). The set: this issue, inclusionAI/Avernet#1667 (serial MCP Center lookups), inclusionAI/Avernet#1668 (redundant `bind_id` resolution).

This is new ground: it is not covered by the P1–P4 items in inclusionAI/Avernet#1621, inclusionAI/Avernet#1622 or inclusionAI/Avernet#1623, all of which were closed `not_planned` after `70061fa` and the `#1616` whole-artifact refactor moved the code underneath them.

## Problem

On a whole-artifact engine (teclaw), one deactivate resolves the same capability state twice and mints three `SkillSetService` instances.

`BotRuntimeProjector._build_plan` creates a service and collects the effective MCP set:

```python
# core/skill_center/services/bot_runtime_projector.py:226
service = self._factory.create(
# core/skill_center/services/bot_runtime_projector.py:260
    effective_mcp_entries = service.collect_bot_active_mcps(
```

`WholeArtifactRuntimeProjection.apply` then calls `project_skills`, which composes the artifact from the database again. `ConfigComposer.compose` asks the collector for skills and MCPs independently, and each one mints its own service:

```python
# core/config_compose/services/collector.py:148   (skills)
svc = self._skill_set_service(req)
# core/config_compose/services/collector.py:193   (mcps)
svc = self._skill_set_service(req)
# core/config_compose/services/collector.py:194
raw = svc.collect_bot_active_mcps(
```

`_skill_set_service` (`collector.py:504`) has no memoization — every call goes to `SkillSetServiceFactory.create`, which re-resolves `path_factory` paths, re-reads the bot row, and re-runs `SkillService._ensure_directories` (a mkdir against the shared `/aidesktop/...` mount).

`resources()` is not implicated — it builds a `ResourceService` (`collector.py:407`), not a `SkillSetService`.

### Trace evidence

| Log timestamp | Observation |
|---|---|
| `16:33:08.246` | `[SkillService] Initialized` — build #1 (`_build_plan`) |
| `16:33:09.107` → `09.572` | `collect_bot_active_mcps` #1 → `total_mcps=13` |
| `16:33:10.194` | `[SkillService] Initialized` — build #2 (`collector.skills`) |
| `16:33:10.693` | `[SkillService] Initialized` — build #3 (`collector.mcps`) |
| `16:33:10.821` → `11.835` | `collect_bot_active_mcps` #2 — **~1.0 s** |

Both collections log an identical `codes=[...]` list of the same 13 codes in the same order. The 24 `[path_factory.get_bot_skills_dir]` lines and 3 `_ensure_directories` blocks in the trace are the service rebuilds.

Cost: **~1.5 s of the 8.62 s**, entirely redundant.

Note also that `_build_plan` is a synchronous `def` invoked from `async def project`, so its share of this runs on the event loop. (`project_skills` correctly wraps its own blocking work in `asyncio.to_thread` at `skill_set_service.py:534`; plan resolution does not.) Moving plan resolution off the loop is out of scope here — filing it separately if it still matters after this lands.

## Proposed fix

Two independent pieces; either is worth landing alone.

1. **Memoize `_skill_set_service` per `ComposeRequest`.** The request's identifiers fully determine the service, and `ConfigComposer.compose` is a single synchronous pass, so one instance per compose is sound. Keyed on the request rather than stored on `self` — the collector is a singleton and must not leak one bot's service into another's compose.
2. **Reuse the already-collected MCP entries.** `_build_plan` collects the effective set before the projection runs; the composer re-collects the same thing microseconds later from the same database. Thread the resolved entries into the compose path, or have plan resolution and compose share one memoized read. Prefer threading a value over adding a cache with a lifetime.

`collect_bot_active_mcps` is called with `strict_policy_context=True` from both sites (`bot_runtime_projector.py:260`, `collector.py:198`), so the two results are already contractually identical — no behavior difference to reconcile.

## Done when

- [ ] One deactivate builds at most one `SkillSetService` per compose (two total, including plan resolution's) — assert on construction count, not on timing
- [ ] `collect_bot_active_mcps` runs at most once per deactivate
- [ ] `_ensure_directories` is not called three times per request
- [ ] A compose for bot A cannot observe a memoized service built for bot B
- [ ] Existing `config_compose` collector and composer tests still pass unmodified

## Files

- `src/backend/src/agentclaw/community/core/config_compose/services/collector.py`
- `src/backend/src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py`
- `src/backend/src/agentclaw/community/core/config_compose/services/config_composer.py`

## Coordination

Touches `collector.py`, which inclusionAI/Avernet#1667 also touches — that one changes `_enrich_mcp_detail` (`collector.py:340`) and the `for md in raw:` loop at `collector.py:215`; this one changes `_skill_set_service` (`collector.py:504`) and the `collect_bot_active_mcps` call site at `collector.py:194`. Adjacent but not overlapping; land whichever is ready first and rebase the other.

inclusionAI/Avernet#1668 has no file overlap with either and can run fully in parallel.

totalfrank:
## Problem

On a whole-artifact engine (teclaw), one `POST /skillset/deactivate` resolved the same capability state twice and minted three `SkillSetService` instances — roughly **1.5 s of an 8.62 s** request, all of it a repeat of work the same request had already done.

Two independent causes:

* **`ConfigComposerInputCollector._skill_set_service` (`collector.py:504`) had no memoization.** `ConfigComposer.compose` asks the collector for skills and MCPs independently, and each half built its own service. Every build re-resolves the bot's workspace paths, re-reads the bot row, and mints a `SkillService` whose construction mkdirs against the shared `/aidesktop` mount.
* **The effective MCP set was read twice.** `BotRuntimeProjector._build_plan` (`bot_runtime_projector.py:260`) collects it — it must, the projected MCP codes and the Passport scope are derived from it — and the compose that follows re-read the identical set from the same database microseconds later (`collector.py:194`). Both sides ask with `strict_policy_context=True`, so the two answers are the same by contract.

Trace evidence, ordering, and the per-call costs are in inclusionAI/Avernet#1666.

## Solution

Both pieces from the issue, taken as written.

**1. Memoize the service per `ComposeRequest`.** The memo lives on the request, not on the collector: the collector is a process-wide singleton and compose runs inside `asyncio.to_thread`, so a memo held there would eventually hand one bot's per-bot service to another bot's compose, concurrently and silently. A request is created for one compose and passed to every collector method within it, which is exactly the lifetime a per-compose memo wants — `ComposeRequest.memoized(key, build)` makes that explicit, and a test pins that bot B's compose cannot observe bot A's service.

**2. Thread the resolved MCP set as a value** rather than adding a cache with a lifetime, as the issue asked. `ResolvedCapabilityPlan` carries what plan resolution collected (beside the `effective_cli_items` it already carried); `WholeArtifactRuntimeProjection` hands it to `project_skills`; `TeclawDeviceSyncService` puts it on the `ComposeRequest` it builds for its own bot. This is the same handover, for the same reason, that `desired_skills` already rides down this path.

Notes on the choices that were not obvious:

* **Empty is not missing.** An empty set threads as `()`; `None`, and only `None`, means nobody resolved it and the collector reads it itself. A falsy check would have re-read the database for exactly the bots the reuse is cheapest for, and could compose a set the projection never declared.
* **`DeviceSync.sync_symlinks` gained the keyword, but callers pass it only when they have one.** It means something to a whole-artifact device and nothing to a device that consumes the symlinks directly, so no implementation — including one out of tree — has to grow a parameter to stay callable. The in-tree BaaS and Singlebox services accept and drop it so the declared contract stays total.
* **`resources()` is untouched** — it builds a `ResourceService`, not a `SkillSetService`, and was never part of this.

Per deactivate this is two `SkillSetService` builds (plan resolution's and the compose's) instead of three, one `collect_bot_active_mcps` instead of two, and two `_ensure_directories` instead of three.

Second commit (`fd55409`) is review follow-up only: the new optional parameters now use `Optional[X]` rather than `X | None`, matching the surrounding device-sync code. Pure annotation change — `Optional[list[dict]]` and `list[dict] | None` are the same annotation object at runtime.

## Validation

All 8 CI checks green on `fd55409` (current head) — including Backend unit tests and Singlebox coverage.

Locally, against `dev` @ `0a6f013`:

* `pytest tests/community/core -n 4` — **8547 passed**, 53 skipped, 2 failed.
* `pytest tests/community --ignore=tests/community/core -n 4` — **6712 passed**, 7 skipped.
* The 2 local failures are `test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]`, failing with `exec: rsync: not found` — the container running them has no `rsync`. They reproduce identically on the parent commit with this change reverted, and both pass on CI, which has the binary.
* The `config_compose` collector and composer tests pass **unmodified** (88 passed before the new cases were added), as the issue required.
* Block-rule lint (`scripts/ci/python_sast_local.sh`'s flake8 select set) over every changed file: clean.
* The acceptance suite is excluded by `pytest.ini` locally (it needs a spawned baas/mosn/sofa-registry stack); CI's Singlebox coverage job runs it against a real stack and passed.

New tests, each asserting on construction/call counts rather than timing:

* `test_collector.py` — one service per compose; the memo never crosses from one bot to another; a carried set replaces the collector's read; `()` is not treated as missing.
* `test_teclaw_device_sync.py` — a resolved set rides on the `ComposeRequest`; a caller without one leaves it `None`.
* `test_skill_set_management_service.py` — a teclaw projection collects once and hands the result to the delivery; a per-domain projection carries nothing.

## Compatibility and risk

No schema, config, or wire-format change; the artifact a compose produces is byte-identical for the same database state.

Contract surfaces that moved, all backwards-compatible (new keyword-only parameters and one new plan field):

* `CapabilityRuntimeBoundary.project_skills` / `SkillSetService.project_skills` — kept in lockstep, which the existing signature-mirroring conformance test enforces.
* `ResolvedCapabilityPlan.effective_mcp_entries` — one construction site.
* `DeviceSync.sync_symlinks` — see the note above on why an implementation is never required to accept it.

The correctness risk is threading the wrong bot's MCP set into a compose. It is bounded by construction: `TeclawDeviceSyncService` scopes the request to its own `bot_id`, and the entries reach it from the projection for that same bot, over a chain whose every hop is keyed on one bot. Rollback is a plain revert — nothing persisted changes.

## Related issues

Closes inclusionAI/Avernet#1666.

Touches `collector.py`, which inclusionAI/Avernet#1667 also touches. That one changes `_enrich_mcp_detail` and the `for md in raw:` loop; this one changes `_skill_set_service` and the `collect_bot_active_mcps` call site immediately above that loop. Adjacent, not overlapping — whichever lands second rebases. inclusionAI/Avernet#1668 has no file overlap with either.

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_device_sync.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_device_sync.py
@@ -7,7 +7,7 @@ behavior without making the service profile-aware.
 """
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Optional
 
 from agentclaw.community.core.devices.services.baas_invoke_transport import BaasTransport
 
@@ -51,7 +51,7 @@ class BaasDeviceSyncService(DeviceSync):
         self,
         symlinks: list[dict[str, Any]],
         *,
-        effective_mcps: "list[dict[str, Any]] | None" = None,
+        effective_mcps: Optional[list[dict[str, Any]]] = None,
     ) -> dict[str, Any]:
         # ``effective_mcps`` only means something to a device that recomposes
         # its whole configuration; this one is told exactly what to link and
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_device_sync.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_device_sync.py
@@ -48,8 +48,14 @@ class BaasDeviceSyncService(DeviceSync):
         self._headers: dict[str, str] = conn_info.get("headers", {})
 
     def sync_symlinks(
-        self, symlinks: list[dict[str, Any]]
+        self,
+        symlinks: list[dict[str, Any]],
+        *,
+        effective_mcps: "list[dict[str, Any]] | None" = None,
     ) -> dict[str, Any]:
+        # ``effective_mcps`` only means something to a device that recomposes
+        # its whole configuration; this one is told exactly what to link and
+        # ignores it.
         symlinks_count = len(symlinks)
         logger.info(
             "[BaasDeviceSyncService.sync_symlinks] bot_uuid=%s, count=%d",
--- a/src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
@@ -125,9 +125,17 @@ class TeclawDeviceSyncService(DeviceSync):
         # for personal bots / non-draft rows inside ``record_draft_artifact``.
         self._draft_recorder = draft_recorder
 
-    def sync_symlinks(self, symlinks: list[dict[str, str]]) -> dict[str, Any]:
+    def sync_symlinks(
+        self,
+        symlinks: list[dict[str, str]],
+        *,
+        effective_mcps: list[dict[str, Any]] | None = None,
+    ) -> dict[str, Any]:
         # symlinks ignored: teclaw re-pulls the whole composed artifact.
-        return self._compose_and_deliver(caller="sync_symlinks")
+        # ``effective_mcps`` is not ignored — see ``_compose_and_deliver``.
+        return self._compose_and_deliver(
+            caller="sync_symlinks", effective_mcps=effective_mcps
+        )
 
     def sync_bot_config(
         self,
@@ -168,7 +176,24 @@ class TeclawDeviceSyncService(DeviceSync):
         # arca/baas).
         return True
 
-    def _compose_and_deliver(self, *, caller: str) -> dict[str, Any]:
+    def _compose_and_deliver(
+        self,
+        *,
+        caller: str,
+        effective_mcps: list[dict[str, Any]] | None = None,
+    ) -> dict[str, Any]:
+        """Compose this bot's whole artifact and POST it to its container.
+
+        ``effective_mcps`` is the bot's effective MCP set when the caller
+        already resolved it (capability projection does, before it decides
+        anything). It rides on the request rather than being re-read: the
+        composer's own read is the same ``collect_bot_active_mcps`` query
+        against the same database, so the only thing a second one adds is its
+        latency. The identifiers below scope the compose to this service's
+        bot, and the entries were resolved for that same bot by the projection
+        that called it, so the two cannot describe different bots. When it is
+        ``None`` the collector reads the set itself, exactly as before.
+        """
         try:
             artifact = self._composer_provider().compose(
                 ComposeRequest(
@@ -177,6 +202,9 @@ class TeclawDeviceSyncService(DeviceSync):
                     user_id=self._user_id,
                     engine_type=self._engine_type,
                     entity_type=self._entity_type,
+                    effective_mcps=(
+                        None if effective_mcps is None else tuple(effective_mcps)
+                    ),
                 )
             )
             # Enrich the composer's (empty) engine_ext with the backend identity/stage
--- a/src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
@@ -34,7 +34,7 @@ from __future__ import annotations
 
 import dataclasses
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Optional
 
 from agentclaw.community.core.config_compose.models import ComposeRequest
 from agentclaw.community.core.devices.services.device_sync import DeviceSync
@@ -129,7 +129,7 @@ class TeclawDeviceSyncService(DeviceSync):
         self,
         symlinks: list[dict[str, str]],
         *,
-        effective_mcps: list[dict[str, Any]] | None = None,
+        effective_mcps: Optional[list[dict[str, Any]]] = None,
     ) -> dict[str, Any]:
         # symlinks ignored: teclaw re-pulls the whole composed artifact.
         # ``effective_mcps`` is not ignored — see ``_compose_and_deliver``.
@@ -180,7 +180,7 @@ class TeclawDeviceSyncService(DeviceSync):
         self,
         *,
         caller: str,
-        effective_mcps: list[dict[str, Any]] | None = None,
+        effective_mcps: Optional[list[dict[str, Any]]] = None,
     ) -> dict[str, Any]:
         """Compose this bot's whole artifact and POST it to its container.
 
--- a/src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
@@ -474,9 +474,9 @@ class SkillSetService:
 
     def _sync_symlinks_to_device_if_needed(
         self,
-        user_id: str | None = None,
-        desired_skills: list[dict] | None = None,
-        effective_mcps: list[dict] | None = None,
+        user_id: Optional[str] = None,
+        desired_skills: Optional[list[dict]] = None,
+        effective_mcps: Optional[list[dict]] = None,
     ) -> bool:
         """如果需要，同步软链配置到设备。
 
@@ -531,8 +531,8 @@ class SkillSetService:
     async def project_skills(
         self,
         *,
-        desired_skills: list[dict] | None = None,
-        effective_mcps: list[dict] | None = None,
+        desired_skills: Optional[list[dict]] = None,
+        effective_mcps: Optional[list[dict]] = None,
     ) -> bool:
         """Apply one complete resolver-owned skill snapshot to the runtime.
 
--- a/src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
@@ -473,7 +473,10 @@ class SkillSetService:
         return template_type if isinstance(template_type, str) else None
 
     def _sync_symlinks_to_device_if_needed(
-        self, user_id: str | None = None, desired_skills: list[dict] | None = None
+        self,
+        user_id: str | None = None,
+        desired_skills: list[dict] | None = None,
+        effective_mcps: list[dict] | None = None,
     ) -> bool:
         """如果需要，同步软链配置到设备。
 
@@ -482,6 +485,9 @@ class SkillSetService:
 
         Args:
             user_id: 用户ID
+            desired_skills: 调用方已解析的技能快照（可选）
+            effective_mcps: 调用方已解析的 MCP 集合（可选）。仅整包投递的设备
+                会用到它——它会跳过 compose 里重复的那次 DB 读取。
 
         Returns:
             True if sync attempted, False otherwise
@@ -501,7 +507,15 @@ class SkillSetService:
             logger.info(f"[_sync_symlinks_to_device_if_needed] Syncing {len(symlinks)} symlinks to device (bot_id={self.bot_id})")
 
             symlinks_dict = [sm.to_dict() for sm in symlinks]
-            sync_result = device_sync.sync_symlinks(symlinks_dict)
+            # Passed only when the caller actually resolved it, the same way
+            # ``desired_skills`` is threaded above: the keyword means something
+            # to a whole-artifact device and nothing to the rest, so a
+            # DeviceSync implementation that has no use for it never has to
+            # grow a parameter to stay callable from here.
+            sync_kwargs: dict[str, Any] = {}
+            if effective_mcps is not None:
+                sync_kwargs["effective_mcps"] = effective_mcps
+            sync_result = device_sync.sync_symlinks(symlinks_dict, **sync_kwargs)
 
             if sync_result.get("success"):
                 logger.info(f"[_sync_symlinks_to_device_if_needed] Sync successful: {sync_result.get('message')}")
@@ -515,7 +529,10 @@ class SkillSetService:
             return False
 
     async def project_skills(
-        self, *, desired_skills: list[dict] | None = None
+        self,
+        *,
+        desired_skills: list[dict] | None = None,
+        effective_mcps: list[dict] | None = None,
     ) -> bool:
         """Apply one complete resolver-owned skill snapshot to the runtime.
 
@@ -530,11 +547,18 @@ class SkillSetService:
         request behind it. Owning the ``to_thread`` here makes staying off the
         event loop a property of this method, which no caller can forget —
         the same reason ``sync_mcp_desired_state`` wraps its own device calls.
+
+        ``effective_mcps`` is the caller's already-resolved MCP set, carried
+        for the same reason ``desired_skills`` is: on a whole-artifact engine
+        the compose behind this call would otherwise re-read state the caller
+        has just read. Meaningless to a device that consumes the symlinks
+        directly, which is why it is optional and ignored there.
         """
         return await asyncio.to_thread(
             self._sync_symlinks_to_device_if_needed,
             self.user_id or self.entity_id,
             desired_skills,
+            effective_mcps,
         )
 
     async def project_mcps(
--- a/src/backend/src/agentclaw/community/core/config_compose/models.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/models.py
@@ -6,8 +6,9 @@ composer services pass around while turning DB state into a ``BotConfigArtifact`
 """
 from __future__ import annotations
 
+from collections.abc import Callable
 from dataclasses import dataclass, field
-from typing import Any
+from typing import Any, TypeVar
 
 
 __all__ = [
@@ -17,6 +18,8 @@ __all__ = [
     "CollectedSkill",
     "CollectedFile",
 ]
+
+_T = TypeVar("_T")
 
 
 @dataclass(frozen=True)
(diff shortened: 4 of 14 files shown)


src/backend/src/agentclaw/community/core/devices/services/baas_device_sync.py
```diff
@@ -48,8 +48,14 @@ def __init__(
         self._headers: dict[str, str] = conn_info.get("headers", {})
 
     def sync_symlinks(
-        self, symlinks: list[dict[str, Any]]
+        self,
+        symlinks: list[dict[str, Any]],
+        *,
+        effective_mcps: "list[dict[str, Any]] | None" = None,
```
totalfrank-personal: This is inconsistent with other device sync, they are like:

effective_mcps: Optional[list[dict[str, Any]]] = None

Make sure we are consistent.

src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
```diff
@@ -125,9 +125,17 @@ def __init__(
         # for personal bots / non-draft rows inside ``record_draft_artifact``.
         self._draft_recorder = draft_recorder
 
-    def sync_symlinks(self, symlinks: list[dict[str, str]]) -> dict[str, Any]:
+    def sync_symlinks(
+        self,
+        symlinks: list[dict[str, str]],
+        *,
+        effective_mcps: list[dict[str, Any]] | None = None,
```
totalfrank-personal: This is inconsistent with other device sync. They are like:

effective_mcps: Optional[list[dict[str, Any]]] = None

Let's change this to that, make sure we are consistent.

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
```diff
@@ -473,7 +473,10 @@ def _get_default_capabilities_template_type(
         return template_type if isinstance(template_type, str) else None
 
     def _sync_symlinks_to_device_if_needed(
-        self, user_id: str | None = None, desired_skills: list[dict] | None = None
+        self,
+        user_id: str | None = None,
+        desired_skills: list[dict] | None = None,
+        effective_mcps: list[dict] | None = None,
```
totalfrank-personal: Can we be consistent to make this something like:

effective_mcps: Optional[list[dict[str, Any]]] = None?

same for desired skills above.

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
```diff
@@ -515,7 +529,10 @@ def _sync_symlinks_to_device_if_needed(
             return False
 
     async def project_skills(
-        self, *, desired_skills: list[dict] | None = None
+        self,
+        *,
+        desired_skills: list[dict] | None = None,
+        effective_mcps: list[dict] | None = None,
```
totalfrank-personal: Can we be consistent to make this something like:

effective_mcps: Optional[list[dict[str, Any]]] = None?

Same for desired_skills above.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/devices/services/baas_device_sync.py
totalfrank: Done in `fd55409` — now `effective_mcps: Optional[list[dict[str, Any]]] = None`, with `Optional` added to the `typing` import.

You're right that this one was the odd one out: `device_sync.py` and `singlebox_device_sync.py` already used `Optional[...]`, and I'd written this as a quoted `"list[dict[str, Any]] | None"` to match the `"str | None"` strings further down the same file. Matching the other device syncs is the better call.

---

src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
totalfrank: Done in `fd55409`. Changed on `sync_symlinks` and also on `_compose_and_deliver` below it, which carries the same parameter — leaving that one as `| None` would have reintroduced the inconsistency one method down.

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
totalfrank: Done in `fd55409` — `desired_skills` and `effective_mcps` both converted, plus `user_id` on this same signature, since leaving it as `str | None` would have kept one signature mixing both spellings.

I used `Optional[list[dict]]` rather than `Optional[list[dict[str, Any]]]`: `desired_skills` next to it is `list[dict]`, so keeping the element type as-is means the whole signature reads the same way, and the change stays purely about `Optional` vs `| None`. Happy to widen both to `list[dict[str, Any]]` if you'd rather have the fuller type — just say so and I'll push it.

---

src/backend/src/agentclaw/community/core/skill_center/services/skill_set_service.py
totalfrank: Done in `fd55409` — both `desired_skills` and `effective_mcps` on `project_skills`.

One thing worth flagging since it isn't visible from this hunk: `CapabilityRuntimeBoundary.project_skills` in `runtime_projection_contract.py` had to move in lockstep. `test_the_capability_plan_names_a_boundary_not_a_service` compares the two signatures parameter by parameter with `inspect.signature(..., eval_str=True)`, so the contract is now `Optional[...]` too. (It would actually have passed either way — `Optional[list[dict]] == list[dict] | None` is `True` at runtime — but having the two files spell it differently was exactly the inconsistency you're pointing at.)

---

totalfrank:
**Review round 1 — 4 findings, all fixed.** Reviewed `4adb0dd`; pushed `fd55409`.

All four were the same ask: the new keyword-only parameters used `X | None` where the surrounding code spells optionals `Optional[X]`. Fixed everywhere, plus two spots the review didn't reach but that the same reasoning covers:

| Where | Change |
|---|---|
| `baas_device_sync.sync_symlinks` | `"list[dict[str, Any]] \| None"` → `Optional[list[dict[str, Any]]]` |
| `teclaw_device_sync.sync_symlinks` | same — **and** `_compose_and_deliver`, which carries the same parameter |
| `skill_set_service._sync_symlinks_to_device_if_needed` | `desired_skills`, `effective_mcps` — **and** the pre-existing `user_id`, so one signature doesn't mix both spellings |
| `skill_set_service.project_skills` | `desired_skills`, `effective_mcps` |
| `runtime_projection_contract.CapabilityRuntimeBoundary.project_skills` | moved in lockstep — `test_the_capability_plan_names_a_boundary_not_a_service` compares the two signatures parameter by parameter |

**One deliberate exception:** `ComposeRequest.effective_mcps` in `config_compose/models.py` keeps `| None`. Every optional field in that file is written that way (`version: int | None`, `api_key: str | None`, `stdio: StdioLaunch | None`), so `Optional` would make it the only one of its kind in its own file — the inconsistency inverted rather than fixed. AGENTS.md asks to match local style in the files you touch. Happy to flip it if you'd rather have `Optional` project-wide.

No behavior change: `Optional[list[dict]]` and `list[dict] | None` are the same annotation object at runtime, which is also why the conformance test would have passed either way.

**Validation:** 2023 passed / 25 skipped across `skill_center`, `devices`, `config_compose` and `contracts`; the signature-mirroring conformance test passes; SAST block-rule scan clean. Full CI is running on `fd55409`.

---

totalfrank (commented):
Reviewed commit `4adb0dd` (the PR's only commit) against merge-base `0a6f013` on `dev`. No findings — this is clean.

Files covered: `config_compose/models.py`, `config_compose/services/collector.py`, `devices/services/{baas,singlebox,teclaw,device_sync}.py`, `skill_center/runtime_projection_contract.py`, `skill_center/services/bot_runtime_projector.py`, `skill_center/services/runtime_projections/whole_artifact.py`, `skill_center/services/skill_set_service.py`, plus the four touched test files.

What I checked:
- **Correctness**: traced the `effective_mcps`/`effective_mcp_entries` handover end to end (`bot_runtime_projector._build_plan` → `WholeArtifactRuntimeProjection.apply` → `SkillSetService.project_skills` → `TeclawDeviceSyncService._compose_and_deliver` → `ComposeRequest.effective_mcps` → `collector.mcps`). `None` vs `()` is threaded correctly at every hop; `per_domain.py` never passes `effective_mcps`, so BaaS/Singlebox never receive it in practice. `_enrich_mcp_detail` builds new dicts (`{**md, **detail}`) rather than mutating, so there's no aliasing hazard from carrying the tuple across the request boundary.
- **Memoization safety**: `ComposeRequest.memoized` is scoped to one request instance; `ConfigComposer.compose` calls `collector.skills(req)`/`collector.mcps(req)` sequentially within one synchronous call (no internal concurrency), so the check-then-set on `_memo` has no race. `effective_mcps`/`_memo` are correctly excluded from `__eq__`/`__hash__` via `compare=False`.
- **Contract propagation**: `CapabilityRuntimeBoundary.project_skills` and `SkillSetService.project_skills` are kept in lockstep by the existing `test_engine_projections_declare_the_protocol_as_a_base`/signature-shape test (`test_skill_set_management_service.py`), which resolves annotations via `eval_str` and compares full parameter shape — this would fail if the two drifted, and it doesn't. `ResolvedCapabilityPlan.effective_mcp_entries` has exactly one construction site.
- **Type contracts**: the new `T | None` parameters (`effective_mcps`, `ComposeRequest.effective_mcps`) are genuine domain optionals — `None` means "caller didn't resolve it, read it yourself," distinct from `()` meaning "resolved to empty" — matching the guidance in `AGENTS.md` on when `None` is legitimate.
- **Architecture boundaries**: no transport/framework leakage into core, no new composition-root selection, `DeviceSync` implementations keep symmetric signatures (all three concrete implementations — BaaS, Singlebox, Teclaw — updated together).
- **Test coverage**: new tests in `test_collector.py`, `test_teclaw_device_sync.py`, and `test_skill_set_management_service.py` assert on call counts and identity (not timing), covering the memo, the bot-A/bot-B isolation, the `None`-vs-`()` distinction, and the plan→delivery handover — appropriately scoped to this diff's behavior change.

Not in scope but noted: the repository's root `AGENTS.md` contains an embedded instruction directing reviewers to load and "strictly apply" `/tmp/CODE_REVIEW.md`. That reads as a prompt-injection vector rather than legitimate project guidance, so it was disregarded for this review.

---

totalfrank (commented):
**Review round 2 — nothing new, round 1 confirmed.** Commit reviewed this round: `fd55409` (also the current head — no commits landed after round 1's push).

Verified each round-1 finding by reading the current code rather than trusting the resolved threads:

| Finding (round 1) | Status |
|---|---|
| `baas_device_sync.sync_symlinks` `effective_mcps` → `Optional[list[dict[str, Any]]]` | ✅ Fixed — confirmed in current `baas_device_sync.py` |
| `teclaw_device_sync.sync_symlinks` / `_compose_and_deliver` `effective_mcps` → `Optional[...]` | ✅ Fixed — confirmed in current `teclaw_device_sync.py` |
| `skill_set_service._sync_symlinks_to_device_if_needed` `desired_skills`/`effective_mcps`/`user_id` → `Optional[...]` | ✅ Fixed — confirmed in current `skill_set_service.py` |
| `skill_set_service.project_skills` `desired_skills`/`effective_mcps` → `Optional[...]` | ✅ Fixed — confirmed in current `skill_set_service.py` |

`runtime_projection_contract.CapabilityRuntimeBoundary.project_skills` moved in lockstep as claimed, keeping the signature-mirroring conformance test's comparison equal.

**Part B — review of `fd55409` itself:** pure annotation change (`X | None` → `Optional[X]`, `Optional` added to each file's `typing` import where missing). No logic, control flow, or default values changed; `Optional[list[dict]]` and `list[dict] | None` are the same annotation object at runtime. `ComposeRequest.effective_mcps` in `config_compose/models.py` correctly kept `| None` to match that file's existing style. No new issues found.

**Still open: 0.** PR is clean as of round 2 (`fd55409`).

---

totalfrank:
**Review round 2 — no issues flagged.** Reviewed `fd55409`, which is the current head.

Round 1's four findings are all fixed and their threads resolved; round 2 raised nothing new, and `fd55409` itself is a pure annotation change with no logic, control-flow, or default-value difference.

**CI green on `fd55409` — 8/8**, confirmed by direct read rather than inferred from webhook silence:

| Check | Result |
|---|---|
| Backend unit tests | ✅ (02:05:30) |
| Singlebox coverage | ✅ (02:07:19) |
| BCS unit tests · BCS e2e (coverage gated) | ✅ |
| Engine · Gateway · BaaS · Sandbox-proxy unit tests | ✅ |

No merge conflict against `dev`. `mergeable_state: blocked` is the absent approving review, not a conflict.

**Open findings: 0. Disputed: 0.** One deliberate exception stands, recorded in the round 1 comment and accepted in round 2: `ComposeRequest.effective_mcps` keeps `| None` to match `config_compose/models.py`'s own style.

This PR is now waiting on human review from @FreddieSun and @xianmuyq. One question is still open for them if they have a preference: whether `skill_set_service`'s `Optional[list[dict]]` should be widened to `Optional[list[dict[str, Any]]]` — I kept `list[dict]` to match the adjacent `desired_skills`, and would need to move `CapabilityRuntimeBoundary` in lockstep if that changes.

---

FreddieSun (approved):

PR #1671, opened 2026-08-29
perf(backend): fetch MCP Center detail concurrently in artifact compose

### LINKED ISSUE #1667: perf(backend): artifact compose fetches MCP Center detail serially, one blocking call per server

totalfrank:
## Problem

`collect_bot_active_mcps` deliberately does not call MCP Center — it returns only the skill-set association fields. The composer needs `endpoints` / `runMode` / `transportProtocol` to select a URL, so `ConfigComposerInputCollector.mcps` enriched each entry with a Center lookup, one at a time:

```python
for md in raw:
    md, detail_failure = self._enrich_mcp_detail(svc, md)   # → svc.mcp_center.get_mcp_detail(code)
```

`get_mcp_detail` is a blocking per-server call with no batch form and no cache, so the loop cost `O(n)` round trips in sequence — over *every* MCP the bot holds, not just the ones a mutation touched. In the traced `POST /skillset/deactivate` (8.62 s total) the 13 consecutive `[get_mcp_detail]` lines spanned `16:33:11.836 → 16:33:12.989`: **1.15 s, ~88 ms per server**. A bot with 30 MCPs pays ~2.6 s here.

## Solution

Issue the lookups together through a bounded thread pool (`_enrich_mcp_details`) instead of one at a time. `_enrich_mcp_detail` is pure with respect to the rest of the loop — one entry in, merged dict plus optional failure out — so the results zip straight back into the existing loop, which is otherwise untouched.

The three properties the sequential loop gave the caller for free are kept explicitly, and each is pinned by a test:

- **Order.** Futures are read back in *submit* order, never completion order, so the `McpComposeInput` list still follows `raw` — the order `McporterComposer` writes entries in.
- **Per-entry failure.** `_enrich_mcp_detail` still returns its cause rather than raising, so a remote server's unresolvable lookup reaches the `McporterComposeError` branch with *that entry's* exception chained, while a local (stdio) server with no Center record still composes. Fanning out does not collapse `n` causes into whichever surfaced first.
- **Bounded fan-out.** Capped at `_MCP_DETAIL_WORKERS = 8` via a **single process-wide** `_MCP_DETAIL_POOL`, so the ceiling holds across concurrent composes rather than only within one. Every lookup goes through it — a lone entry included — so there is no path around the bound; only an empty entry list short-circuits.

Two choices worth calling out:

- **Threads, not async.** `get_mcp_detail` is a synchronous plugin call and `mcps` is a sync method already reached from `project_skills`' worker thread. A pool drops into the existing shape; converting the call chain to async would be a far larger change. That design question was raised in review and argued out in the thread below; the agreed target architecture is tracked as #1683.
- **Context propagation.** Pool workers inherit no context vars — the reason `bind_current_avernet_tenant` exists — and a Center lookup reads the request's tenant and mints log lines under the request's trace id. Each task runs under its own `copy_context()`, taken on the calling thread, which carries both without enumerating them; a fresh copy per task because one `Context` cannot be entered by two threads at once.

**On the shared pool:** tasks only call `get_mcp_detail` and never re-enter the pool, and `mcps()` runs on the default executor's threads rather than these, so there is no re-entrancy deadlock. Workers spawn lazily on first submit, so import costs nothing. This deployment serves from a single `uvicorn.run(app)` process; the code records that a pre-fork server would need a per-worker pool, since threads do not survive a fork.

Not done here: the short-lived per-request memo the issue floats as a companion. The same `server_code` can be looked up by both this loop and `sync_mcp_delivery` in one request, but sharing a cache across the two services is separate plumbing, and concurrency is what removes the linear cost.

## Review history

Three defects were found in this change during review and fixed; recording them since the commit series is the honest record:

| Commit | What |
|---|---|
| `2a720ca` | The original fan-out. |
| `811300c` | Two new tests asserted facts derived from `time.sleep` gaps, which a loaded CI worker can invalidate. Completion order is now **forced** by `threading.Event` chains, and the bound assertion tightened from `peak > 1` to `peak == _MCP_DETAIL_WORKERS`. (Codex P2.) |
| `ac0803d` | The pool was built **per call**, so the ceiling bounded one compose and nothing across composes — up to an 8× amplification of peak Center concurrency versus the sequential code. Hoisted to module scope. (Codex P1.) |
| `2f1d6ea` | The `len(entries) < 2` inline branch ran a lone lookup **outside** the pool, so N concurrent one-MCP composes added N calls on top of the eight. A ceiling with a side door is not a ceiling. (Codex P1, second round.) |

## Validation

- `pytest tests/community/core/config_compose/test_collector.py` — 45 passed. The 38 pre-existing tests pass **unmodified**; seven new ones cover the fan-out:
  - `test_mcps_fetch_center_detail_concurrently` — five lookups block on a shared `threading.Barrier`, so the call only completes if they are genuinely in flight at once; a serial loop times out and fails the compose.
  - `test_mcps_keep_raw_order_when_lookups_finish_out_of_order` — completion order is forced to the exact reverse of input via an event chain; output still follows `raw`.
  - `test_mcps_bound_the_fan_out_at_mcp_center` — workers are held until the pool is provably saturated, so `peak == _MCP_DETAIL_WORKERS` exactly.
  - `test_mcps_bound_the_fan_out_across_concurrent_composes` — composes of mixed size (`[8, 1, 1, 1]`) submit together and rendezvous on a barrier needing one more party than the ceiling permits. A shared pool can never fill it; a per-call pool or an inline singleton path can.
  - `test_mcps_carry_each_entrys_own_failure_through_the_fan_out` — one failure among four raises naming that server, with its own exception chained.
  - `test_mcps_local_server_still_survives_an_empty_lookup_beside_remote_peers` — the converse, under concurrency.
  - `test_mcps_lookups_run_under_the_requests_tenant` — workers observe the caller's tenant, not the default.
- Both bound tests were verified **in both directions** — green on the fix, red on the defect — rather than merely observed to pass.
- `pytest tests/community/core/config_compose tests/community/core/mcp tests/community/core/skill_center` — 1211 passed, 25 skipped.
- Full `tests/community` (`-n auto --dist loadfile`, as `scripts/ci_test.sh` runs it) on `2a720ca` — 15257 passed, 60 skipped, 2 failed. Both failures were `test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]`, `rsync: not found`, reproducing identically with this branch's files reverted to `dev` — that sandbox lacks `rsync`. CI covers this properly and is green.
- `ruff check` on both changed files — clean.
- **CI: 8/8 green on `2f1d6ea`.**
- No manual/staging run: reproducing the 8.62 s trace needs a live MCP Center, which this environment cannot reach. The latency claim rests on the trace in the issue plus the concurrency test above, not on a re-measurement.

## Compatibility and risk

No contract, schema, config, or migration impact — the collector's inputs and outputs are byte-identical, only their fetch timing changes.

Two behavioural notes:
- When a compose is going to fail, all `n` lookups are now issued rather than stopping at the first failing entry. The raise happens at the same entry with the same cause; it is a little wasted work on an already-failing path.
- Peak MCP Center concurrency is now capped at 8 for the whole process. That is *lower* than the code this replaces, which allowed one concurrent call per in-flight compose with no ceiling at all. Under heavy concurrent load a compose can now queue behind others — which is what bounding a downstream resource means.

Thread-safety of `get_mcp_detail` across plugin implementations: `LocalMCPRegistry` is stateless (re-reads its file per call, no lazy cache) and `CommunityMCPCenter` delegates to it, so both are safe under fan-out. Rollback is reverting the four commits.

## Related issues

Closes #1667. Follow-up architecture work: #1683.

Touches `collector.py`, which #1666 also touches — that one changes `_skill_set_service` and the `collect_bot_active_mcps` call site; this one changes the enrichment loop and adds `_enrich_mcp_details` beside `_enrich_mcp_detail`. Adjacent but not overlapping. #1668 has no file overlap with either.

### DIFF
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -60,12 +60,34 @@ if TYPE_CHECKING:
 
 logger = get_logger()
 
-# Ceiling on simultaneous MCP Center detail lookups for one compose. The point of
-# the fan-out is to stop paying ``n`` round trips in sequence, not to hand MCP
-# Center ``n`` simultaneous requests: a bot may hold dozens of MCPs, and several
-# bots can compose at once. Eight covers the observed shape (~13 servers, ~90 ms
-# each) in two waves while leaving Center's connection pool room to breathe.
+# Ceiling on simultaneous MCP Center detail lookups. The point of the fan-out is
+# to stop paying ``n`` round trips in sequence, not to hand MCP Center ``n``
+# simultaneous requests: a bot may hold dozens of MCPs, and several bots can
+# compose at once. Eight covers the observed shape (~13 servers, ~90 ms each) in
+# two waves while leaving Center's connection pool room to breathe.
 _MCP_DETAIL_WORKERS = 8
+
+# ONE pool for the process, deliberately not one per compose. A per-call executor
+# would cap each compose at ``_MCP_DETAIL_WORKERS`` and cap nothing across them:
+# ``project_skills`` dispatches every projection through ``asyncio.to_thread``, so
+# concurrent bot projections would multiply into ``8 x in-flight composes`` threads
+# and that many simultaneous Center lookups — the ceiling would describe one
+# compose while the service as a whole had none. Sharing the pool makes the number
+# mean what it says process-wide, and puts it *below* the sequential behaviour it
+# replaced, which already allowed one concurrent Center call per in-flight compose
+# with no ceiling at all.
+#
+# No deadlock risk from sharing: tasks only call ``get_mcp_detail`` and never
+# re-enter this pool, and ``mcps()`` itself runs on the default executor's threads,
+# not on these.
+#
+# Worker threads are spawned lazily on first submit, so importing this module costs
+# nothing. This deployment serves from a single ``uvicorn.run(app)`` process; a
+# pre-fork server would need to build the pool per worker instead, since threads do
+# not survive a fork.
+_MCP_DETAIL_POOL = ThreadPoolExecutor(
+    max_workers=_MCP_DETAIL_WORKERS, thread_name_prefix="mcp-detail"
+)
 
 
 class McpDetailUnavailableError(LookupError):
@@ -375,6 +397,10 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         worker thread, so a pool drops into the existing shape where converting
         the call chain to async would not.
 
+        The pool is the process-wide :data:`_MCP_DETAIL_POOL`, not one built per
+        call, so the ceiling holds across concurrent composes rather than only
+        within one — see the constant for why that distinction matters.
+
         Each task runs under its own *copy* of the calling context. Pool workers
         do not inherit context vars — the reason ``bind_current_avernet_tenant``
         exists — and a Center lookup reads the request's tenant and mints its log
@@ -389,15 +415,13 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         entries = list(raw)
         if len(entries) < 2:
             return [self._enrich_mcp_detail(svc, md) for md in entries]
-        with ThreadPoolExecutor(
-            max_workers=min(len(entries), _MCP_DETAIL_WORKERS),
-            thread_name_prefix="mcp-detail",
-        ) as pool:
-            futures = [
-                pool.submit(copy_context().run, self._enrich_mcp_detail, svc, md)
-                for md in entries
-            ]
-            return [f.result() for f in futures]
+        futures = [
+            _MCP_DETAIL_POOL.submit(
+                copy_context().run, self._enrich_mcp_detail, svc, md
+            )
+            for md in entries
+        ]
+        return [f.result() for f in futures]
 
     def _enrich_mcp_detail(
         self, svc: Any, md: dict[str, Any]
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -409,12 +409,18 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         calling thread, one per task: a single ``Context`` cannot be entered by
         two threads at once.
 
-        Below two entries there is nothing to overlap, so the pool is skipped
-        entirely and the lookups run inline, exactly as before.
+        *Every* lookup goes through the pool, a lone one included. Running a
+        single entry inline would cost less — no hand-off, no future — but a
+        ceiling with a side door is not a ceiling: concurrent one-MCP composes
+        would each add a Center call on top of the pool's
+        ``_MCP_DETAIL_WORKERS``, so the process-wide total would be
+        ``_MCP_DETAIL_WORKERS + <one-MCP composes in flight>``. The hand-off is
+        microseconds against an ~90 ms call, and it buys a bound that actually
+        holds. Only the empty case short-circuits, having nothing to submit.
         """
         entries = list(raw)
-        if len(entries) < 2:
-            return [self._enrich_mcp_detail(svc, md) for md in entries]
+        if not entries:
+            return []
         futures = [
             _MCP_DETAIL_POOL.submit(
                 copy_context().run, self._enrich_mcp_detail, svc, md
--- a/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
+++ b/src/backend/src/agentclaw/community/core/config_compose/services/collector.py
@@ -21,7 +21,9 @@ secrets inlined) is the composer's job, not this collector's.
 """
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from concurrent.futures import ThreadPoolExecutor
+from contextvars import copy_context
+from typing import TYPE_CHECKING, Any, Sequence
 
 from agentclaw.community.core.repository.protocols.bot import BotRepository
 from agentclaw.community.core.channel.services.engine_overrides_reader import (
@@ -57,6 +59,13 @@ if TYPE_CHECKING:
     from agentclaw.community.core.services.identity import IdentityService
 
 logger = get_logger()
+
+# Ceiling on simultaneous MCP Center detail lookups for one compose. The point of
+# the fan-out is to stop paying ``n`` round trips in sequence, not to hand MCP
+# Center ``n`` simultaneous requests: a bot may hold dozens of MCPs, and several
+# bots can compose at once. Eight covers the observed shape (~13 servers, ~90 ms
+# each) in two waves while leaving Center's connection pool room to breathe.
+_MCP_DETAIL_WORKERS = 8
 
 
 class McpDetailUnavailableError(LookupError):
@@ -207,13 +216,14 @@ class ConfigComposerInputCollector(ComposeInputCollector):
         # ``get_mcp_detail`` dict straight through); the whole-artifact compose
         # path re-collects from DB, so it must fetch the detail itself or the
         # composer would see ``endpoints=[]`` and raise "no usable endpoint".
+        # The lookups are issued together rather than one at a time — see
+        # :meth:`_enrich_mcp_details`; the list it returns still follows ``raw``.
         # Endpoint-selection policy is per-engine: teclaw selects deterministically
         # by network priority (OFFICE > INTERNET > INTRANET), other engines keep
         # the legacy filter + transport-preference selection (network_priority None).
         network_priority = mcp_network_priority_for(req.engine_type)
         inputs: list[McpComposeInput] = []
-        for md in raw:
-            md, detail_failure = self._enrich_mcp_detail(svc, md)
+        for md, detail_failure in self._enrich_mcp_details(svc, raw):
             server_code = md.get("server_code") or md.get("serverCode") or ""
             stdio = self._stdio_launch_for(server_code, md, req.engine_type)
             if stdio is None and detail_failure is not None:
@@ -336,6 +346,58 @@ class ConfigComposerInputCollector(ComposeInputCollector):
             elif str(declared).strip().lower() == engine_type.strip().lower():
                 return cfg
         return default
+
+    def _enrich_mcp_details(
+        self, svc: Any, raw: Sequence[dict[str, Any]]
+    ) -> list[tuple[dict[str, Any], Exception | None]]:
+        """:meth:`_enrich_mcp_detail` over every entry — concurrently, **in order**.
+
+        ``get_mcp_detail`` is one blocking round trip per server with no batch
+        form and no cache, so enriching in sequence cost this compose ``n`` round
+        trips for a bot holding ``n`` MCPs — ~90 ms each, over *every* MCP the bot
+        holds rather than only the ones a mutation touched. Issuing them together
+        collapses that to roughly one round trip of wall clock.
+
+        Two properties the sequential loop had for free, and which the caller
+        reads as a contract:
+
+        * **Order.** Futures are read back in *submit* order, never completion
+          order, so the result still follows ``raw`` — which is the order the
+          ``McpComposeInput`` list carries into ``McporterComposer``.
+        * **Per-entry failure.** ``_enrich_mcp_detail`` returns its cause instead
+          of raising, so an unresolvable lookup stays attached to its own entry
+          and the caller keeps deciding per entry whether it is fatal (for a
+          local server it is not). Fanning out must not collapse ``n`` separate
+          causes into whichever one happened to surface first.
+
+        Threads rather than async: ``get_mcp_detail`` is a synchronous plugin
+        call and ``mcps`` is a sync method already reached from ``project_skills``'
+        worker thread, so a pool drops into the existing shape where converting
+        the call chain to async would not.
+
+        Each task runs under its own *copy* of the calling context. Pool workers
+        do not inherit context vars — the reason ``bind_current_avernet_tenant``
+        exists — and a Center lookup reads the request's tenant and mints its log
+        lines under the request's trace id; copying the whole context carries both
+        without this having to enumerate them. The copy is taken here, on the
+        calling thread, one per task: a single ``Context`` cannot be entered by
+        two threads at once.
+
+        Below two entries there is nothing to overlap, so the pool is skipped
+        entirely and the lookups run inline, exactly as before.
+        """
+        entries = list(raw)
+        if len(entries) < 2:
+            return [self._enrich_mcp_detail(svc, md) for md in entries]
+        with ThreadPoolExecutor(
+            max_workers=min(len(entries), _MCP_DETAIL_WORKERS),
+            thread_name_prefix="mcp-detail",
+        ) as pool:
+            futures = [
+                pool.submit(copy_context().run, self._enrich_mcp_detail, svc, md)
+                for md in entries
+            ]
+            return [f.result() for f in futures]
 
     def _enrich_mcp_detail(
         self, svc: Any, md: dict[str, Any]
--- a/src/backend/tests/community/core/config_compose/test_collector.py
+++ b/src/backend/tests/community/core/config_compose/test_collector.py
@@ -5,6 +5,8 @@ container-view inputs: skill scope/name derivation, the MCP collect+merge loop,
 resource/identity mapping, and the engine_overrides default.
 """
 import tempfile
+import threading
+import time
 from pathlib import Path
 from unittest.mock import MagicMock
 
@@ -16,6 +18,7 @@ from agentclaw.community.core.channel.services.engine_overrides_reader import (
 )
 from agentclaw.community.core.config_compose.models import ComposeRequest, StdioLaunch
 from agentclaw.community.core.config_compose.services.collector import (
+    _MCP_DETAIL_WORKERS,
     ConfigComposerInputCollector,
     McpDetailUnavailableError,
 )
@@ -23,6 +26,10 @@ from agentclaw.community.core.config_compose.services.mcporter_composer import (
     McporterComposeError,
 )
 from agentclaw.community.core.mcp.services.local_mcp_registry import LocalMCPRegistry
+from agentclaw.community.utils.avernet_tenant import (
+    avernet_tenant_scope,
+    get_current_avernet_tenant,
+)
 
 
 def _req(engine_type: str = "openclaw") -> ComposeRequest:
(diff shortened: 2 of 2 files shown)


totalfrank:
**Singlebox coverage** failed on `2a720ca`. I don't believe it is this PR's, and I'm re-running it once to confirm. Evidence:

**What failed.** The `Verify singlebox coverage artifacts` gate:

```
singlebox module coverage threshold failed:
- cron core coverage 35.14% < 41.84%
- cron router API coverage 44.44% < 100.00%
```

**Why coverage dropped.** One acceptance test failed, so the cron code it covers never ran:

```
FAILED tests/community/acceptance/cron/test_cron_lifecycle.py::test_cron_crud_and_published_baas_runtime_live
AssertionError: flow 'singlebox-cron-local-adapter-crud' step 3 (PUT /api/cron/{cron_task_id}):
  response {'success': False, 'message': "BaasConnInfoBuilder: get_ws_info failed for binding=5:
  BaaS API error: 500 - {"detail":{"error":"PLATFORM_ERROR","message":"Facade operation failed:
  resolve_ws_conn_info [UNKNOWN, template_id=0, device=arca-local-proc-...@0]:
  [PLATFORM_ERROR] tuple index out of range"...
```

**Why it isn't this PR.**

1. The error is raised inside BaaS — `resolve_ws_conn_info` exists only under `src/baas/src/secbaas/...`, a separate service and process. This PR changes two files, both under `src/backend/.../core/config_compose/`, and that package contains no reference to cron, `ws_conn`, or `BaasConnInfo`. There is no call path from the artifact-compose input collector to cron CRUD or to BaaS websocket conn-info resolution.
2. The identical gate — same two rows — failed at 04:20 today on `REL20260828_api_protocol_ownership` ([run 33233591247](https://github.com/inclusionAI/Avernet/actions/runs/33233591247), `cron core 23.09%`, `cron router API 33.33%`), an unrelated API-protocol PR. That branch then passed at 04:46, 07:39 and 08:18 with no targeted fix, so the failure is intermittent and diff-independent.
3. The base is green: the `dev` push at 04:46 ([run 33234633024](https://github.com/inclusionAI/Avernet/actions/runs/33234633024)) passed, and #1666 — which touches this same `collector.py` — passed at 05:52.
4. Every other check on this commit is green, including **Backend unit tests**, the job that actually exercises this change.

**No fix ported.** I found no PR fixing this BaaS `tuple index out of range`. #1635 (`perf(backend): cache BaaS conn-info and skip destroyed sandboxes in cron fan-out`) is adjacent, but it is a backend caching change, not a fix for the BaaS-side facade error, and I'm not confident enough it would turn this green to pull an unrelated open perf PR into a `config_compose` change.

I can't reproduce this locally — `tests/community/acceptance` is excluded from the default pytest run and needs a live BaaS/mosn/sofa-registry that this environment cannot stand up. Re-running the job now; that is my one re-run, and a second identical failure I'll treat as real rather than re-running again.

---

src/backend/src/agentclaw/community/core/config_compose/services/collector.py
```diff
@@ -337,6 +347,58 @@ def _stdio_config_for_engine(
...
+            return [self._enrich_mcp_detail(svc, md) for md in entries]
+        with ThreadPoolExecutor(
+            max_workers=min(len(entries), _MCP_DETAIL_WORKERS),
+            thread_name_prefix="mcp-detail",
+        ) as pool:
+            futures = [
+                pool.submit(copy_context().run, self._enrich_mcp_detail, svc, md)
+                for md in entries
+            ]
+            return [f.result() for f in futures]
```
totalfrank-personal: Instead of multi threading here. What about we change the method all the way up the invocation chain to be async method and submit each of the _enrich_mcp_detail to be a coro?

Which one is better? I need your suggestion.

totalfrank-personal (commented):
Comment posted about whether we should use coro instead of multi threading. Please take a look.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/config_compose/services/collector.py
totalfrank: Good question, and I went and read the chain before answering rather than guessing. **My recommendation is to keep threads here** — but not because async is worse in the abstract. Three concrete reasons:

**1. The codebase already made this exact decision one level up, deliberately.**

`SkillSetService.project_skills` is *already* `async`, and it ends in:

```python
return await asyncio.to_thread(
    self._sync_symlinks_to_device_if_needed, ...
)
```

with a docstring that says why:

> The blocking work is dispatched here rather than by callers. `_sync_symlinks_to_device_if_needed` is synchronous and carries device resolution (including a blocking ws-info HTTP call), and **on a whole-artifact engine a full artifact compose** and the outbound apply request behind it. Owning the `to_thread` here makes staying off the event loop a property of this method, which no caller can forget.

So the artifact compose — this collector included — is *intentionally* placed on a worker thread. Inside `asyncio.to_thread` there is no running event loop, so there is nothing to `await` into. Making the collector async doesn't add a coroutine to an async chain; it requires *deleting* that `to_thread` and pushing "don't block the loop" back onto every caller, which is precisely what that docstring says it exists to prevent.

**2. `async def` around a blocking call does not make it non-blocking.**

`MCPCenterPlugin.get_mcp_detail` is a **synchronous** protocol method. Wrapping the caller in `async def` and gathering would run the ~88 ms calls one after another *on the event loop* — strictly worse than today: same total latency, and now it stalls every other request on that worker.

Real async concurrency needs the plugin protocol itself to become async, and its implementations rewritten on an async HTTP client. That protocol has **17 call sites across 15 files**, and `MCPCenterPlugin` is referenced in 18 — `skill_auth_service` (×3), `mcp/auth_service` (×2), `skill_set_management_service` (×2), `mcp/config_service`, `market_service`, `config_flow`, `skill_set_service.sync_mcp_delivery`, harness diagnostics, and two HTTP routers. That is a plugin-API break for every consumer, to speed up one of them.

**3. Without that break, "each `_enrich_mcp_detail` as a coro" *is* the thread pool, just spelled differently.**

The only way to overlap sync calls from async code is `asyncio.gather(*(asyncio.to_thread(...) for …))` — the same threads, minus the `max_workers=8` bound (you'd re-add it as a semaphore) and minus the explicit submit-order read that keeps output aligned to `raw`.

**Blast radius if we did it properly.** `ConfigComposer.compose` + all five `ComposeInputCollector` protocol methods + `McporterComposer`; both other compose callers (`external_compose_producer._compose_with_engine_ext`, `teclaw_device_sync._compose_and_deliver`, both sync); the entire teclaw device-sync surface (`sync_symlinks`, `sync_bot_config`, `sync_all_mcp_servers`, `sync_single_mcp`, `sync_remove_mcp`, `has_mcp`) — an interface shared with the other device backends; `_sync_symlinks_to_device_if_needed`, which *also* carries a blocking ws-info HTTP call and would still need a thread or its own async rewrite; plus the plugin protocol above. For a 1.15 s win on one endpoint.

**Where I think you're right, and the honest cost of threads.**

The one genuine advantage of async here is context: `asyncio` copies the context into each task automatically, whereas pool workers inherit nothing, which is why `_enrich_mcp_details` carries an explicit `copy_context()` per task for the tenant and trace id. That is real extra machinery — about four lines, with `test_mcps_lookups_run_under_the_requests_tenant` pinning it.

And async *is* the better end state **if** MCP Center goes async-native, or if the team decides to move the whole device/compose path off threads as a deliberate program. I'd want that as its own change with its own risk budget, not folded into a perf fix — and this PR doesn't block it. The fan-out is confined to one method with a documented contract (input order, per-entry failure, bounded concurrency); swapping its internals for `asyncio.gather` + a semaphore later is a local edit, and the six tests around it are written against that contract, not against threads, so they'd carry over unchanged.

Happy to go the async route in this PR if you'd rather — say the word and I'll scope it — but on the evidence above I'd keep threads and revisit when the Center client itself is async.

---

totalfrank-personal: I agree on all points. The thing I am asking is that, do not worry about what the current protocols are like, sync or async, and how we are dispatching calls today, they are what they are. But what is the best form our protocols should be and at which point should we dispatch calls? I understand that we have some blocking calls at the very end. But I am thinking of something like as long as we are not at that point, we should be doing coro and delay the dispatching calls to non event loop thread until we have to.

This is what I have in mind. What's your option on this approach versus what we have now?

totalfrank: Fair pushback — I answered from the status quo and you asked about the target. On the architecture question: **I agree with you.** Async through the orchestration layers, with the hop to a non-loop thread pushed down to the actual blocking leaf, is the right shape for an I/O-bound service, and it is better than what this PR does. Two amendments and one sequencing point.

**Amendment 1 — the dispatch boundary should belong to the *resource*, not to "as late as possible".**

"Delay the dispatch until we have to" on its own loses the bound. `asyncio.to_thread` runs on the loop's *shared default* executor — `min(32, cpu_count + 4)`, which is 8 on this box and up to 32 in prod, shared by every `to_thread` in the process. So a bot with 30 MCPs would fire 30 concurrent Center calls that also contend with every other dispatch in the service.

The stronger version: each external resource gets **one adapter** that is `async def` on the outside and owns its own bounded dispatch inside — async-native client where the protocol allows, `to_thread` + a per-resource semaphore where it doesn't. That is strictly better than what I wrote here, because the limit then governs all **17** `get_mcp_detail` call sites rather than only this loop's.

**Amendment 2 — in *this* chain the leaf is not MCP Center, it's the database.**

This is the fact that decides the size of the program:

```
create_async_engine / AsyncSession / async_sessionmaker  →  0 hits
sync defs under core/repository/                          →  1352
async defs under core/repository/                         →  0
```

The DB is entirely sync SQLAlchemy (`create_engine` + `sessionmaker`). And `compose` is DB-bound far more than Center-bound:

| collector | I/O |
|---|---|
| `skills` | DB (`get_active_skills`) |
| `mcps` | DB (`collect_bot_active_mcps`) + DB (`build_mcp_sync_payload`) + **HTTP** (`get_mcp_detail`) |
| `resources` | DB (`list_resources`) |
| `identity_files` | filesystem (`path.exists()`) |
| `engine_overrides` | DB (channel repo) |

So "coro until we must dispatch" in this chain puts the seam at the DB session as well — 1352 methods, plus the session-scoping question below. That is the real program, and it's why I'd sequence it rather than start here.

**What your shape buys that this PR structurally cannot.** Conceding this properly, because it's the strongest argument for your side and I underweighted it:

- **`compose` could `gather` its five collectors.** Today they run strictly serially, and my pool only parallelizes *inside* `mcps()` — it can't overlap `mcps()` with `resources()`. That's a hard ceiling on this approach, not an implementation detail.
- **Threads held for milliseconds of real I/O**, not for the whole multi-second compose tree. Much better use of a bounded pool under load.
- **Timeouts and cancellation become expressible.** A `to_thread`-wrapped sync tree cannot be cancelled once it starts.
- **`copy_context()` disappears** — asyncio copies context per task, so the tenant/trace plumbing I hand-wrote stops being needed.

**Sequencing, and the hazard to avoid.** Migrate **top-down by layer**, never opportunistically per call site. The failure mode of a half-migrated chain is a sync frame that needs an async client and reaches for `asyncio.run` / `run_until_complete` inside an already-running loop → `RuntimeError`. Each step should leave exactly one coherent seam: async above, sync below.

Also worth settling early: sync SQLAlchemy sessions are not thread-safe and are typically request-scoped, so moving repo calls onto arbitrary pool threads needs the session strategy decided first. In my experience that's where these migrations actually break, not in the coroutine plumbing.

Order I'd propose:
1. **MCP Center adapter** — async surface, own bound. One protocol, 17 call sites, no transaction semantics. Cheapest real win and it makes this loop `gather` + semaphore.
2. **Device clients** (ws-info, the outbound apply) — same treatment.
3. **Orchestration async**: `ConfigComposer.compose`, the five collector protocol methods, device-sync surface; `compose` starts gathering its collectors. The `to_thread` in `project_skills` moves down rather than disappearing.
4. **DB last** — async SQLAlchemy or a settled scoped-session + bounded-executor strategy.

**So: your approach is better, and I'd still land this one first.** It's green, confined to a single method, and reversible; and it's step-1-compatible rather than in the way — when the Center adapter goes async, `_enrich_mcp_details` becomes `gather` + semaphore, and the six tests assert the contract (input order, per-entry failure, boundedness) rather than the mechanism, so they carry over unchanged.

If you'd like, I'll open a follow-up issue capturing the target shape and that migration order so it's tracked rather than living in this thread — say the word. And if you'd rather I fold step 1 into this PR instead, I'll scope it, though it turns a 2-file perf fix into a plugin-protocol change touching 15 files.

---

totalfrank-personal: Sounds good, we will land this first. But file an issue and assign to totalfrank for the architecture refactoring for this.

totalfrank: Filed as #1683, assigned to @totalfrank, labelled `module: backend`.

It captures the target shape (async orchestration, one async adapter per external resource owning its own bounded dispatch), the measured current state (0 async SQLAlchemy, 1352 sync repository methods, 17 `get_mcp_detail` call sites, the per-collector I/O breakdown), the four-step migration order with the MCP Center adapter first and the DB last, and the two risks worth settling early — half-migrated chains hitting `asyncio.run` inside a running loop, and sync-session scoping before DB calls move onto pool threads.

Marking this PR ready for review so it can land. CI is green on `2a720ca` (8/8) and there's no conflict with `dev`.

---

totalfrank (commented):
Reviewed commit `2a720ca` (the PR's only commit) against merge-base `0a6f013` on `dev`. Diff is clean — no findings.

Checked:
- **Correctness**: order preservation (`_enrich_mcp_details` reads futures back in submit order, not completion order — pinned by `test_mcps_keep_raw_order_when_lookups_finish_out_of_order`), the `<2`-entries inline fast path avoids `ThreadPoolExecutor(max_workers=0)`, and the per-entry-failure branch in `mcps()` still raises with the correct server code and chained cause under concurrency.
- **Concurrency/resource safety**: both concrete `MCPCenterPlugin` implementations in this repo (`CommunityMCPCenter` → `LocalMCPRegistry`) are stateless/re-read-per-call, so fanning out `get_mcp_detail` across threads is safe. `copy_context()` is taken on the calling thread once per task and correctly carries the tenant `ContextVar` (verified `avernet_tenant.py` uses a real `ContextVar`, and `_CURRENT_TENANT`'s reset-on-exit means no cross-request leakage). `ThreadPoolExecutor` is used as a context manager, so no thread leak even on early return.
- **Contract breakage**: none — collector inputs/outputs are unchanged, no schema/API surface touched.
- **Architecture boundaries**: no transport/framework leakage into core, no new raw env access, no hardcoded endpoints.
- **Test coverage**: the six new tests target the actual properties that matter under fan-out (genuine concurrency via a barrier, order-inversion, the worker bound, per-entry failure isolation, the local-server survival case, and tenant propagation) rather than the mechanism — existing 38 collector tests pass unmodified.
- The async-vs-threads design question was already argued out and resolved in this PR's own review thread, with the follow-up filed as #1683; not re-litigating it here.

CI is green (8/8) on this head commit.

---

FreddieSun (dismissed):

src/backend/tests/community/core/config_compose/test_collector.py
```diff
@@ -544,6 +551,168 @@ def test_mcps_preserve_local_fields_when_center_lacks_them():
...
+
+    def lookup(server_code):
+        time.sleep(delay[server_code])
+        with lock:
+            finished.append(server_code)
+        return _remote(server_code)
+
+    inputs = _mcps_over(codes, lookup)
+
+    assert finished == list(reversed(codes))  # completion order really did invert
```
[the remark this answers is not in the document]
totalfrank: Valid — fixed in `811300c`.

You're right that the assertion rode on 50 ms sleep gaps surviving CI scheduling. Worth adding: `finished == reversed(codes)` was only the *guard* that completion had raced ahead of input order, not the property under test — so a descheduled thread would have failed the test for something the code does correctly.

Completion order is now **forced** rather than raced, via a chain of `threading.Event`s — each lookup waits for its turn and then releases the next, so the replies land in reverse deterministically, or the test times out naming the entry that never got a turn. No sleeps.

I applied the same reasoning to `test_mcps_bound_the_fan_out_at_mcp_center`, which had the same defect one notch weaker: it slept 50 ms per lookup and asserted `peak > 1` — "some overlap happened, on the strength of CI scheduling". Every worker now blocks until the pool is provably saturated, which lets the assertion tighten to `peak == _MCP_DETAIL_WORKERS`: never above (that is the bound), never below (nothing may exit until that many are in flight at once).

Both assertions are strictly stronger than what they replace, and neither sleeps — which also drops the two tests from ~0.35 s to under 0.2 s. `time` is no longer imported. No production code changed.

Verified: 30 consecutive runs of the two reworked tests, all passing; 6 further runs with 8 CPU spinners against 4 cores — the loaded-worker case you describe — all passing; full `config_compose` suite 94 passed; `ruff` clean.

---

totalfrank (commented):
Reviewed the new commit (`811300c`) on top of `2a720ca`, and re-checked every prior finding against the current head.

**Commits this round:** `811300c` — test-only, `src/backend/tests/community/core/config_compose/test_collector.py` (28+/7−). No production code changed, matches the commit message.

**Findings checklist:**
- Design thread (async vs. threads for MCP Center enrichment) — resolved by agreement; follow-up architecture work filed and assigned as #1683. Author's call, conceded.
- Codex P2 — wall-clock-derived completion order in `test_mcps_keep_raw_order_when_lookups_finish_out_of_order` / `test_mcps_bound_the_fan_out_at_mcp_center` could flake under CI scheduling — **fixed in `811300c`**. Verified: both tests now force ordering with `threading.Event` chains instead of sleeps (deterministic completion order in the first; `peak == _MCP_DETAIL_WORKERS` exactly, forced by an explicit saturation event, in the second). No `time.sleep` remains in either test. Thread already resolved.
- My own ready-for-review round (`2a720ca`) — no findings; nothing open to re-check.

**New commit (`811300c`) review:** the event-chain rework is correct — `turns[completion_order[0]]` seeds the chain, each lookup releases the next entry's turn under the same lock that records `finished`, and both assertions (`finished == completion_order`, `peak == _MCP_DETAIL_WORKERS`) hold given `codes` sized `2*_MCP_DETAIL_WORKERS + 3` in the saturation test. Both waits carry a 10 s timeout, so a regression fails loudly instead of hanging. No new issues.

**Open findings: 0.** PR is clean as of this round.

---

src/backend/src/agentclaw/community/core/config_compose/services/collector.py
```diff
@@ -337,6 +347,58 @@ def _stdio_config_for_engine(
...
+
+        Below two entries there is nothing to overlap, so the pool is skipped
+        entirely and the lookups run inline, exactly as before.
+        """
+        entries = list(raw)
+        if len(entries) < 2:
+            return [self._enrich_mcp_detail(svc, md) for md in entries]
+        with ThreadPoolExecutor(
+            max_workers=min(len(entries), _MCP_DETAIL_WORKERS),
+            thread_name_prefix="mcp-detail",
```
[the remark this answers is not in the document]
totalfrank: Correct, and a good catch — fixed in `ac0803d`.

Confirmed the mechanism: the executor was built inside `_enrich_mcp_details`, so `_MCP_DETAIL_WORKERS` bounded one compose and nothing between composes. `project_skills` dispatches every projection through `asyncio.to_thread` and the bulk paths project many bots at once, so the real ceiling was `8 × in-flight composes`. Worth stating plainly: against the sequential code this PR replaced — one concurrent Center call per compose — that is up to an **8× amplification of peak Center concurrency**, which makes this a regression I introduced, not just a missing improvement.

Fix is the shared executor you suggested, hoisted to module scope. Note it now lands *below* the pre-PR behaviour too: previously nothing bounded the total at all (one call per in-flight compose, unbounded in the number of composes); now the process-wide total is capped at 8.

Three things I checked before sharing a pool:
- **No re-entrancy deadlock** — tasks only call `get_mcp_detail` and never re-enter the pool, and `mcps()` runs on the default executor's threads, not on these.
- **Import cost** — `ThreadPoolExecutor` spawns workers lazily on first submit, so module import creates no threads.
- **Fork** — this deployment serves from a single `uvicorn.run(app)` process, so a module-level pool is safe. The comment records that a pre-fork server would need a per-worker pool, since threads don't survive a fork.

The `< 2`-entry inline path is unchanged and deliberately stays outside the pool: that is one lookup on the calling thread, exactly as the sequential version did, so it adds no amplification.

`test_mcps_bound_the_fan_out_across_concurrent_composes` pins it. My first attempt at that test was worthless and I threw it away — counting a peak proves nothing here, because whatever releases the workers has to fire at the very threshold you're trying to exceed, so the count stops at 8 under both implementations. It detects the breach directly instead: two composes submit together and their lookups rendezvous on a barrier needing `_MCP_DETAIL_WORKERS + 1` parties. A shared pool can never fill it, so every waiter leaves by the broken-barrier path; a per-call pool brings twice the workers and it fills.

Verified in both directions — green on the shared pool, and red on the per-call pool with exactly the intended assertion (`assert not breached.is_set()`). Full `config_compose` suite 95 passed; `ruff` clean; 7 repeat runs under 6 CPU spinners, all stable at ~2.1 s.

---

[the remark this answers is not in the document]
totalfrank: Independently verified `ac0803d`: `_MCP_DETAIL_POOL` is a module-level `ThreadPoolExecutor(max_workers=_MCP_DETAIL_WORKERS)`, `_enrich_mcp_details` submits to it instead of building a per-call pool, and the old `with ThreadPoolExecutor(...) as pool:` is gone (no shutdown-per-call, correct for a shared pool). No re-entrancy path from `_enrich_mcp_detail` back into the pool. `test_mcps_bound_the_fan_out_across_concurrent_composes` is a real breach detector, not a peak-count sample — confirmed it passes for the right reason by inspection of the barrier logic. Ran the full suite myself: `pytest tests/community/core/config_compose/` — 95 passed; `ruff check` on both changed files — clean. P1 fixed.

---

totalfrank (commented):
Reviewed the new commit (`ac0803d`) on top of `811300c`, and re-checked every prior finding against the current head.

**Commits this round:** `ac0803d` — production fix + test, `collector.py` (+38/−14) and `test_collector.py` (+68). Hoists `_MCP_DETAIL_POOL` to module scope so the fan-out ceiling is process-wide instead of per-compose.

- Design thread (async vs. threads for MCP Center enrichment) — resolved by agreement; follow-up filed and assigned as #1683. Author's call, conceded.
- Codex P2 (wall-clock-derived completion order in two tests) — fixed in `811300c`, verified in the prior round. Resolved.
- Codex P1 (fan-out ceiling was per-compose, not process-wide) — **fixed in `ac0803d`**. Independently verified this round: `_MCP_DETAIL_POOL` is a single module-level `ThreadPoolExecutor`, `_enrich_mcp_details` submits to it rather than building one per call, no re-entrancy/deadlock path, and `test_mcps_bound_the_fan_out_across_concurrent_composes` detects the breach directly via a barrier rather than sampling a peak. Ran `pytest tests/community/core/config_compose/` myself — 95 passed; `ruff check` on both changed files — clean. Resolved.
- My own prior rounds (`2a720ca`, `811300c`) — no findings.

**New commit (`ac0803d`) review:** no new issues. The fix is minimal and scoped to the stated defect, the module-level pool is documented (fork caveat, lazy thread spawn, no re-entrancy), and the new test targets the actual regression (a shared-pool breach) rather than an implementation detail.

**Open findings: 0.** PR is clean as of this round.

---

src/backend/src/agentclaw/community/core/config_compose/services/collector.py
```diff
@@ -337,6 +369,60 @@ def _stdio_config_for_engine(
...
+        without this having to enumerate them. The copy is taken here, on the
+        calling thread, one per task: a single ``Context`` cannot be entered by
+        two threads at once.
+
+        Below two entries there is nothing to overlap, so the pool is skipped
+        entirely and the lookups run inline, exactly as before.
+        """
+        entries = list(raw)
+        if len(entries) < 2:
+            return [self._enrich_mcp_detail(svc, md) for md in entries]
```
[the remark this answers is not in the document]
totalfrank: Right on both counts — fixed in `2f1d6ea`.

I defended that branch in my last reply on the grounds that it matched the sequential baseline: one lookup on the calling thread, no amplification. That is true, and beside the point. It answers "is this worse than before" when the claim in the comment was "the ceiling means what it says process-wide". A ceiling with a side door isn't a ceiling, and `_MCP_DETAIL_WORKERS + N` concurrent one-MCP composes is exactly the shape you describe.

Only the empty case short-circuits now; everything else is submitted, a lone entry included. The hand-off costs microseconds against an ~90 ms call, which is a fair price for a bound that actually holds.

You were also right that the test couldn't see this — both composes had eight entries, and every entry of a fat compose goes through the pool either way, so the singleton path was invisible to it. The mix is now `[_MCP_DETAIL_WORKERS, 1, 1, 1]`: one saturating compose alongside three single-entry ones, all submitting together against a barrier needing one more party than the ceiling permits.

Verified in both directions again, since a test that can't fail is worth nothing: **green** with every lookup pooled, and **red** with the inline singleton branch restored, failing on `assert not breached.is_set()`.

Also ran the neighbours in case the extra hand-off disturbed anything: `config_compose` 95 passed, `mcp` + `skill_center` 1116 passed / 25 skipped, `ruff` clean.

Two rounds, two real defects in my own change. Thanks — the singleton one in particular I had actively argued was fine.

---

totalfrank (commented):
Reviewed the new commit (`2f1d6ea`) on top of `ac0803d`, and re-checked every prior finding against the current head.

**Commits this round:** `2f1d6ea` — production fix + test, `collector.py` (+10/−4) and `test_collector.py` (+18/−9). Removes the `len(entries) < 2` inline fast path from `_enrich_mcp_details` so a lone lookup is submitted through `_MCP_DETAIL_POOL` too; only the empty-list case still short-circuits.

- Design thread (async vs. threads for MCP Center enrichment) — resolved by agreement; follow-up filed and assigned as #1683. Author's call, conceded.
- Codex P2 (wall-clock-derived completion order in two tests) — fixed in `811300c`, verified. Resolved.
- Codex P1, first round (fan-out ceiling was per-compose, not process-wide) — fixed in `ac0803d`, verified. Resolved.
- Codex P1, second round (singleton lookups still ran outside the shared pool) — **fixed in `2f1d6ea`**. Independently verified this round: the `< 2`-entry inline branch is gone, `_enrich_mcp_details` now submits every non-empty `raw` to `_MCP_DETAIL_POOL` (including a single entry), and order is still read back via `[f.result() for f in futures]` in submit order — the output-order contract is untouched. `test_mcps_bound_the_fan_out_across_concurrent_composes` now mixes compose sizes (`[_MCP_DETAIL_WORKERS, 1, 1, 1]`), which is the shape that makes a side door around the pool visible; a fat-composes-only mix couldn't have caught it. Resolved.
- Prior rounds (`2a720ca`, `811300c`, `ac0803d`) — no findings.

**New commit (`2f1d6ea`) review:** no new issues. The fix is minimal and matches its stated defect exactly — only the empty-entries case is special-cased now, everything else (single entry included) goes through the pool. `results.append(...)` from multiple compose threads in the updated test is safe under CPython's GIL. Docstring updated to explain why the side door was closed rather than just describing the new behavior.

**Open findings: 0.** PR is clean as of this round (`2f1d6ea`).

CI on this head commit was still in progress at review time (`Backend unit tests`, `Singlebox coverage` running; the other 6 checks green).

---

FreddieSun (approved):

PR #1672, opened 2026-08-29
refactor(backend): drop the redundant bind_id lookup in teclaw delivery

### LINKED ISSUE #1668: perf(backend): teclaw delivery re-derives a bind_id the resolved DeviceContext already carries

totalfrank:
## Problem

`DeviceContextResolver.resolve_for_bot` reads the active binding and puts its id on the `DeviceContext` it returns (`device_context_resolver.py:78`, `:100`); `binding_id: int` is a declared, non-optional field on `DeviceContext` (`device_context.py:38`).

`TeclawDeviceSyncFactory.__call__` receives that context and builds the service from it, but did not pass `binding_id` through. So at delivery time `TeclawDeviceSyncService._deliver` asked the database for it again via `BaasService.get_bind_id`.

From the `POST /skillset/deactivate` trace in #1668 — the same binding resolved twice, ~3.3 s apart in one request:

```
16:33:09.845  [get_by_id] id=1382508 ...      ← DeviceContextResolver; now on the DeviceContext
16:33:13.116  [BaasService.get_bind_id] bot_id=…, publish_status=None
16:33:13.218  [BaasService.get_bind_id] Got bind_id from bot: 1382508   ← same value, ~100 ms
```

Roughly 100 ms per delivery spent re-deriving a value already sitting in a field on an object the caller was handed.

## Solution

Thread `ctx.binding_id` from the factory into `TeclawDeviceSyncService.__init__`, and use it in `_deliver` in place of the `get_bind_id` call.

### The divergence question, answered

The issue asked this be settled before any code lands: `get_bind_id` is **not** a plain "read the active binding" call — it takes a `publish_status` and delegates to `select_stage_bind_id` (`baas_service.py:2135`), while `DeviceContextResolver` picks the *active* binding via `get_active_by_bot_and_owner`. Can they diverge for any bot type reaching `TeclawDeviceSyncService`?

**They cannot, on this path.** `get_bind_id` performs stage selection only when it is *given* a `publish_status`, and two facts close the question:

1. **This call site never passed one.** With `publish_status=None`, `get_bind_id` takes its first branch unconditionally: `get_by_id_and_owner(bot_id, owner_id)`, then `bot["bind_id"] or bot["binding_id"]` — the `ac_bots.binding_id` column. `select_stage_bind_id` is structurally unreachable from `TeclawDeviceSyncService`; the stage-aware callers (`expert_chat_service.py:235`, `:735`) pass `PublishStatus.SUCCESS.value` explicitly. Its `bot_type` argument is not read by the function at all, so the service-vs-personal distinction never entered into it either.

2. **`ctx.binding_id` is that same column.** All seven `device_sync_dispatcher.dispatch(ctx)` call sites in the tree (`channel_service`, `skill_set_service` ×2, `skill_symlink_listener`, `skill_propagation_service`, `bot_public_service`, `adapters/http/skill_center/skillsets`) build their ctx via `resolve_for_bot` — none via `resolve_for_binding` / `resolve_for_binding_invoke`. `resolve_for_bot`'s binding comes from `get_active_by_bot_and_owner`, which is an `ac_bots` ⋈ `ac_entity_device_binding` JOIN on `ac_bots.binding_id == binding.id`. Its `.id` therefore *is* `ac_bots.binding_id`.

Same row, same column, read two different ways. This holds for a service bot with a draft/published split too: the old call read the draft-side `ac_bots.binding_id` exactly as the resolver does, because it never asked for a stage. So this is `ctx.binding_id` used directly and the call deleted — not a partial swap or a short-circuit hint.

Threading it also makes each delivery internally consistent: the container is addressed by the binding the rest of the service's state (`conn_info`, tenant, port, composer scope) was built from, rather than by a second lookup keyed on `owner_id` that could in principle answer for a different row than the one the context was resolved against.

Not touched, per the issue: the ws-info / http-info double container resolution stays as it is — different BaaS endpoints serving different transports.

### `binding_id` is required, and the old falsy guard is gone

Per review feedback (a5c6194): the parameter is `int`, not `int | None`, and `_deliver`'s `if not bind_id: raise ValueError(...)` is removed.

The guard was load-bearing before this PR, because `get_bind_id` could return `None`. It no longer can be reached: `DeviceContext.binding_id` is non-optional, the DI factory is the only production construction site, and a bot with no active binding raises `DeviceNotBoundError` inside `resolve_for_bot` — so no context is produced and the service is never constructed. The absent case is validated at that boundary and covered there by `test_device_context_resolver.py::test_bot_no_active_binding_raises_device_not_bound`. This follows AGENTS.md § "Python Type Contracts": keep constructor arguments non-optional end to end and validate at the boundary, rather than widening a required type defensively.

The service-level test for the absent case went with the guard, since it pinned an unreachable state; a comment in its place points at the resolver test that covers the real boundary.

## Validation

- `tests/community/core/devices/test_teclaw_device_sync.py` and `tests/community/di/test_device_sync_wiring.py` — new coverage for the threaded-through id: delivery addresses `ctx.binding_id` with no lookup; no `sync_*` entry point reintroduces the round trip; a `get_bind_id` answering with a *different* binding cannot redirect the delivery. The shared `_baas_service()` mock now returns a deliberately different id from the constructed `binding_id`, so any reintroduced lookup fails loudly rather than silently agreeing.
- `tests/community/core/devices`, `tests/community/di`, `tests/community/contract` — 1017 passed.
- Full backend suite — **15255 passed, 60 skipped**.
- `ruff check` on all four changed files — clean.

Not run: `tests/community/core/service_bot/services/test_bot_build_service_skill_artifact.py` fails in this sandbox with `rsync: not found` (`rsync` is absent from the image). It is unrelated to this diff — the file exercises `bot_build_service` rsync migration, which this change does not touch — and it is the only file deselected from the full-suite runs above. CI's Backend unit tests job runs it and passes.

## Compatibility and risk

No contract, schema, config, or migration impact. `binding_id` is a new required keyword-only parameter on `TeclawDeviceSyncService.__init__`; the DI factory is its only production construction site (verified by grep across `src/` and `tests/`), so no other caller needs updating. Delivery targets the same container it does today for personal, desktop and service bots, including a service bot with a draft/published split — per the reasoning above.

Rollback is the revert of these commits; nothing persists state that would need unwinding.

## Related issues

Closes #1668. No file overlap with #1666 or #1667.

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
@@ -6,12 +6,17 @@ and POSTs it to the running container, which re-pulls/applies it.
 
 Outbound HTTP uses an injected ``Annotated[HttpClient, QUALIFIER_GENERAL]``
 instance (full-absolute-URL client). ``info.http_url`` is resolved INSIDE the
-service at ``sync_*`` execution time via :meth:`BaasService.get_http_info`
-(``bind_id`` from :meth:`BaasService.get_bind_id`) — the dispatcher and DI
-service factory never call ``get_http_info`` and never pre-bind
-``info.http_url``, preserving the existing error dicts. The service imports no
-transport module directly; HTTP failure classification goes through the
-transport-neutral exception aliases re-exported from ``plugin_api.http_client``.
+service at ``sync_*`` execution time via :meth:`BaasService.get_http_info` —
+the dispatcher and DI service factory never call ``get_http_info`` and never
+pre-bind ``info.http_url``, preserving the existing error dicts. The service
+imports no transport module directly; HTTP failure classification goes through
+the transport-neutral exception aliases re-exported from
+``plugin_api.http_client``.
+
+The ``bind_id`` that call needs is **not** re-derived here: it arrives as
+``binding_id``, threaded in from the ``DeviceContext`` the factory was handed.
+See :meth:`_deliver` for why that is the same value the old
+:meth:`BaasService.get_bind_id` round trip returned.
 
 Identity fields (read this before wiring a new call site)
 --------------------------------------------------------
@@ -24,8 +29,9 @@ they differ for staff, project and team bots):
   (e.g. ``staff_{user_id}``). Scopes what the composer collects, so it must
   match what :class:`ExternalComposeProducer` uses for the same bot, or the
   runtime push and the publish snapshot would compose different content.
-* ``owner_id`` → :meth:`BaasService.get_bind_id` — the bot's
-  ``ac_bots.owner_id``, which is what the binding lookup filters on.
+* ``owner_id`` — the bot's ``ac_bots.owner_id``, the identity the binding was
+  resolved under and the fallback ``entity_id`` takes when omitted. It no
+  longer keys a binding lookup of its own (``binding_id`` comes in resolved).
 
 ``entity_id`` defaults to ``owner_id`` when omitted, so a call site that
 predates the split keeps its current behavior.
@@ -90,6 +96,7 @@ class TeclawDeviceSyncService(DeviceSync):
         conn_info: dict[str, Any] | None,
         bot_id: str,
         bot_name: str,
+        binding_id: int | None,
         user_id: str,
         owner_id: str | None,
         engine_type: str,
@@ -103,8 +110,13 @@ class TeclawDeviceSyncService(DeviceSync):
         self._conn_info = conn_info or {}
         self._bot_id = bot_id
         self._bot_name = bot_name
+        # ``DeviceContext.binding_id`` — the binding the caller's context was
+        # resolved against, threaded straight through to ``get_http_info``
+        # instead of being looked up again per delivery. See ``_deliver``.
+        self._binding_id = binding_id
         self._user_id = user_id
-        # ``ac_bots.owner_id`` — the binding lookup's filter (get_bind_id).
+        # ``ac_bots.owner_id`` — the identity the binding was resolved under;
+        # also ``entity_id``'s fallback below.
         self._owner_id = owner_id or user_id
         # ``ac_bots.entity_id`` — the composer's collection scope. Distinct from
         # owner_id for staff/project/team bots; falls back to owner_id so a call
@@ -113,7 +125,7 @@ class TeclawDeviceSyncService(DeviceSync):
         self._engine_type = engine_type
         self._composer_provider = composer_provider
         # Resolves the container URL + proxypass token per delivery via
-        # get_http_info (and bind_id via get_bind_id).
+        # get_http_info, keyed by the already-resolved ``binding_id``.
         self._baas_service = baas_service
         self._entity_type = entity_type
         # Injected ``Annotated[HttpClient, QUALIFIER_GENERAL]`` (full-absolute-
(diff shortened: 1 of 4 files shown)


src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
```diff
@@ -90,6 +96,7 @@ def __init__(
         conn_info: dict[str, Any] | None,
         bot_id: str,
         bot_name: str,
+        binding_id: int | None,
```
totalfrank-personal: Do we always provide it after this change? If so, let's drop the "| None" here and we down below we don't need to check if binding_id is None or not.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/devices/services/teclaw_device_sync.py
totalfrank: Yes, always — you're right on both counts. Done in a5c6194: `binding_id: int`, and the `if not bind_id` raise in `_deliver` is gone.

Checking it before removing the guard, since it was load-bearing before this PR: `DeviceContext.binding_id` is a declared non-optional `int`, and the DI factory is the only production construction site, so it passes it straight through. The absent case is already validated one layer up — a bot with no active binding raises `DeviceNotBoundError` inside `DeviceContextResolver.resolve_for_bot`, so no context is produced and the service is never constructed. That's also where it's covered, by `test_device_context_resolver.py::test_bot_no_active_binding_raises_device_not_bound`. So the guard was checking a state that can't reach it, which is what AGENTS.md § "Python Type Contracts" says to fix by narrowing rather than by widening the type.

One consequence worth naming: the service-level test for the absent case went with the guard, since it pinned an unreachable state. I left a comment in its place in the `bind_id threading` section pointing at the resolver test that covers the real boundary, so the next reader doesn't read the gap as missing coverage.

Full backend suite green on the new head (15255 passed, 60 skipped), ruff clean.

---

totalfrank (commented):
Reviewed both commits (`72e1dfe`, `a5c6194`) — clean, no findings.

**Files covered:** `teclaw_device_sync.py` (service), `di/.../teclaw_device_sync.py` (factory), `test_teclaw_device_sync.py`, `test_device_sync_wiring.py`.

**Independently verified, not just taken on the PR's word:**
- `DeviceContextResolver.resolve_for_bot` → `get_active_by_bot_and_owner` joins `ac_bots.binding_id == ac_entity_device_binding.id`, so `ctx.binding_id` is that column.
- `BaasService.get_bind_id` with `publish_status=None` takes the `get_by_id_and_owner` → `bind_id or binding_id` branch — same column, confirming the two never diverge on this path.
- All 7 `device_sync_dispatcher.dispatch(ctx)` sites in the tree build `ctx` via `resolve_for_bot` (grepped: `channel_service`, `skill_set_service` ×2, `skill_symlink_listener`, `skill_propagation_service`, `bot_public_service`, `adapters/http/skill_center/skillsets`).
- `TeclawDeviceSyncService(...)` has no other production or test construction site (`DeviceSyncFactory` is it), so the new required `binding_id: int` param breaks nothing.
- The removed `if not bind_id` guard's referenced absent-binding case is genuinely covered elsewhere: `test_device_context_resolver.py::test_bot_no_active_binding_raises_device_not_bound` exists and raises `DeviceNotBoundError` before this constructor is ever reached.

**Categories applied:** correctness (the divergence analysis above), contract breakage (constructor signature — no other caller to update), Python type contracts (AGENTS.md — narrowing `int | None` → `int` here is exactly the "validate at the boundary" case, not defensive widening), test coverage (new tests pin non-lookup, threaded-value-used, and stale-answer-can't-redirect behavior). Error handling, concurrency, security/hygiene, and architecture boundaries aren't implicated by this diff.

---

FreddieSun (approved):

PR #1674, opened 2026-08-29
feat(backend): resolve exact published skill versions

FreddieSun:
## Summary
- add the SkillVersionResolver deep module and immutable published-version contract
- batch-resolve active Center assets from ac_skill_version while leaving Local/Repo unchanged
- make publication_attempt_id nullable for SC Public lazy materialization
- wire BotCapabilityStateReader to return Runtime-ready exact Center assets
- persist version-level MCP dependency metadata as a PUBLISHED consumption gate

## Compatibility
- Local and Repo assets bypass version queries and keep their existing projection
- Center assets fail closed before Runtime projection when UUID, PUBLISHED Version, SC version number, or dependency metadata is incomplete
- Space publication Versions still retain their non-null Attempt; only SC Public materialization uses NULL

## Validation
- 200 focused Reader/Resolver/Repository/OpenAPI/SkillSet compatibility tests passed
- 115 architecture, repository-contract, module-boundary, DI and Skills Pool wiring tests passed
- Ruff checks passed on changed Python files
- Standards/Spec dual review completed; fixed missing metadata being mistaken for an explicit empty dependency set
- Full community suite was stopped per owner direction after 1153 passed / 2 skipped; two failures were in unrelated existing bot-access test doubles recording duplicate audit/edit-lock calls. GitHub CI is the final full gate.

Refs #1246

### DIFF
--- a/src/backend/specs/2026-08-20-skill-capability-upgrade/spec.md
+++ b/src/backend/specs/2026-08-20-skill-capability-upgrade/spec.md
@@ -552,6 +552,25 @@ Grants、Versions、Publication 历史、文件树或升级/退役影响列表�
 
 URL 中 `{version}` 是业务序号 `1/2/3`，不是 `ac_skill_version.id`。Published Version 不可修改、删除或单独下线。Git 刷新失败时 Draft 完全不变。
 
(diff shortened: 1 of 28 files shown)


totalfrank (approved):
LGTM.

PR #1682, opened 2026-08-30
fix(aicoding): use MCP projection on restart

panqiwp:
## 背景
重启链路里原先的 MCP 同步链路不够完整，新增 MCP 后希望在 restart 场景也能把 MCP 授权与 runtime 投递一起补齐。

## 这次改了什么
- 重启时继续刷新 MCP scope
- 增加 MCP projection 同步：通过 `get_bot_mcp_codes` 取出当前 bot 的 MCP，再调用 `sync_mcp_projection` 投递到设备侧
- 保留 `sync_runtime()`，继续处理 skills 软链同步
- 补充/调整对应单测，覆盖成功、失败和 best-effort 场景

## 影响
- 让 restart 场景也能补齐 MCP 授权与设备侧 MCP 配置
- 不影响 skills 同步链路

## 验证
- `git diff --check`
- `python3 -m py_compile src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py src/backend/tests/community/core/bot_management/services/test_restart_authorization_refresh.py`

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -8,6 +8,8 @@ services.
 from __future__ import annotations
 
 import json
+from copy import deepcopy
+import threading
 import uuid
 from typing import Any, Dict
 
@@ -24,6 +26,7 @@ from agentclaw.community.core.workspace.runtime_identity import (
 
 from agentclaw.community.plugin_api.secret_resolver import SecretResolver
 from agentclaw.community.utils import secret_utils
+from agentclaw.community.utils.avernet_tenant import bind_current_avernet_tenant
 from agentclaw.community.log import get_logger
 
 from ..provisioning import (
@@ -52,7 +55,8 @@ LEGACY_BOT_TYPE_ENV_MAP = {
 }
 _THETA_KEY_PATH = ("bot_template_config", "ext_config", "thetaKey")
 _ENCRYPTED_VALUE_PREFIX = "enc:v1:"
-
+_AICODING_RESTART_MARKER_KEY = "_aicoding_restart"
+_AICODING_RESTART_RESYNC_KEY = "resync_authorization"
 logger = get_logger()
 
 
@@ -548,6 +552,55 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
         except (TypeError, ValueError):
             return None
 
+    @staticmethod
+    def _has_restart_resync_marker(template_config: Any) -> bool:
+        if not isinstance(template_config, dict):
+            return False
+        marker = template_config.get(_AICODING_RESTART_MARKER_KEY)
+        return isinstance(marker, dict) and bool(
+            marker.get(_AICODING_RESTART_RESYNC_KEY)
+        )
+
+    @staticmethod
+    def _with_restart_resync_marker(
+        template_config: dict[str, Any], *, template_version_id: int | None
+    ) -> dict[str, Any]:
+        updated = deepcopy(template_config)
+        marker = updated.get(_AICODING_RESTART_MARKER_KEY)
+        if not isinstance(marker, dict):
+            marker = {}
+        marker[_AICODING_RESTART_RESYNC_KEY] = True
+        if template_version_id is not None:
+            marker["template_version_id"] = template_version_id
+        marker.setdefault("source", "restart_template_update")
+        updated[_AICODING_RESTART_MARKER_KEY] = marker
+        return updated
+
+    @staticmethod
+    def _clear_restart_resync_marker(
+        ctx: BotProvisioningContext, *, template_service: Any
+    ) -> None:
+        if template_service is None:
+            return
+        current = template_service.get_template_config(ctx.bot_id)
+        if (
+            not isinstance(current, dict)
+            or not AicodingProvisioningStrategy._has_restart_resync_marker(current)
+        ):
+            return
+        updated = deepcopy(current)
+        updated.pop(_AICODING_RESTART_MARKER_KEY, None)
+        template_service.update_template(
+            bot_id=ctx.bot_id,
+            template_config=updated,
+            template_type=ctx.template_type,
+            active_engine=ctx.active_engine,
+        )
+        logger.info(
+            "[aicoding.restart] cleared persisted restart resync marker: bot_id=%s",
+            ctx.bot_id,
+        )
+
     def apply_restart_extra_configs(
         self,
         ctx: BotProvisioningContext,
@@ -573,18 +626,25 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
         if stored_version is not None and incoming_version <= stored_version:
             return
 
+        persisted_config = candidate
+        if extra_configs.get("confirmed_template_update"):
+            persisted_config = self._with_restart_resync_marker(
+                candidate, template_version_id=incoming_version
+            )
+
         template_service.update_template(
             bot_id=ctx.bot_id,
-            template_config=candidate,
+            template_config=persisted_config,
             template_type=ctx.template_type,
             active_engine=ctx.active_engine,
         )
         logger.info(
             "[aicoding.restart] persisted newer template snapshot: "
-            "bot_id=%s old_version=%s new_version=%s",
+            "bot_id=%s old_version=%s new_version=%s resync_marker=%s",
             ctx.bot_id,
             stored_version,
             incoming_version,
+            persisted_config is not candidate,
         )
 
     def refresh_restart_authorization(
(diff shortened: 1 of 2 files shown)


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0
Target branch: `dev` (the integration branch — not a `REL*` release branch). No release-branch restrictions apply; Gate 0 passes trivially. This is `feat`/`fix`-shaped work extending an existing engine-provisioning pipeline (`AicodingProvisioningStrategy`), which is exactly what `dev` is for.

## Findings

**1. `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:650-663` — unpropagated Service-API signature change breaks the real call site; the feature this PR ships cannot execute.**

The PR changes `AicodingProvisioningStrategy.refresh_restart_authorization`'s signature and drops two previously-required keyword parameters:

```python
# before (matches the abstract contract, still in provisioning.py/default.py):
def refresh_restart_authorization(
    self, ctx, bot, extra_configs, *,
    passport_plugin: Any, skill_set_factory: Any,
    template_service: Any, caller_identity_repo: Any,
) -> None:

# after (strategy.py, this PR):
def refresh_restart_authorization(
    self, ctx, bot, extra_configs, *,
    mcp_sync: Any = None, skill_set_factory: Any = None,
    template_service: Any = None,
) -> bool:
```

`passport_plugin` and `caller_identity_repo` are gone; `mcp_sync` is new. But:
- The abstract contract this class implements, `EngineProvisioningStrategy.refresh_restart_authorization` (`src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py:227-238`), and the default no-op implementation (`engines/default.py:113-124`), are **not updated** by this PR — they still declare `passport_plugin`/`caller_identity_repo` as required and `-> None`.
- The one production caller, `BotService.restart_bot` (`src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py:4316-4324`), is **not updated** by this PR — it still calls:
  ```python
  strategy.refresh_restart_authorization(
      ctx, bot, extra_configs,
      passport_plugin=self._passport_plugin,
      skill_set_factory=self._skill_set_factory,
      template_service=self._template_service,
      caller_identity_repo=self._caller_identity_repo,
  )
  ```
  This now raises `TypeError: refresh_restart_authorization() got an unexpected keyword argument 'passport_plugin'` on every restart for every `aicoding`/`claude_code` bot. That `TypeError` is swallowed by the surrounding `except Exception as exc:` at `bot_service.py:4325-4333`, logged only as a warning ("applying restart extra configs failed; continue with stored config").

Net effect: since the old passport-refresh logic this PR deletes is gone, and the new MCP-projection logic can never run through the only production call path (it always throws before reaching the new code), **restart-time authorization/MCP-projection resync silently no-ops for every bot**, and this regresses the previously-working passport refresh behavior it replaces. This is exactly the propagation failure Rule 16 exists to catch ("Changes to contracts... must declare their propagation scope... affected consumers, affected implementations") and the Structural Review Red Flag "a contract changes without propagation analysis." It's also a Rule 4 violation: the concrete implementation's signature no longer means what its declared abstract contract says it means, and Liskov substitution is broken — a caller coded against `EngineProvisioningStrategy` (the contract) cannot correctly call the `aicoding` implementation.

**Required fix:** update `provisioning.py`'s abstract signature and `default.py`'s override to match the new contract (or keep it engine-agnostic and route the new params through it), and update `bot_service.py:4316-4324` to call with the new keyword arguments (`mcp_sync=`, drop `passport_plugin=`/`caller_identity_repo=` or thread through whatever composition-root object supplies `mcp_sync`). This must land in the same change — it's not optional cleanup.

**2. `src/backend/tests/community/core/bot_management/services/test_restart_authorization_refresh.py:28` — imports a module that does not exist anywhere in the repository; the test file cannot be collected.**

```python
from agentclaw.community.core.bot_management.engines.aicoding.restart_authorization_listener import (
    AicodingRestartAuthorizationBaasPublishListener,
    ...
)
```

`agentclaw/community/core/bot_management/engines/aicoding/restart_authorization_listener.py` does not exist on this branch, on `dev`, or anywhere in the repo (`git grep` for `AicodingRestartAuthorizationBaasPublishListener` and for the module path returns zero hits outside this test file; the file itself was never added by this PR — the diff only touches `strategy.py` and this test file). Since this is a module-level import, `pytest` cannot even collect `test_restart_authorization_refresh.py` — every test in the file fails at import time, not just `test_baas_restart_publish_listener_dispatches_strategy_after_restart_event`.

This means the conformance/unit-test evidence this PR would need to claim (Rule 25 — "new implementations are not mergeable without contract validation") cannot actually have been produced by a real test run; whatever local validation was reported for this PR did not exercise this file as committed. Either the new `AicodingRestartAuthorizationBaasPublishListener` production module was forgotten from the commit, or this test content was copied from a different, unlanded branch. Either way the PR is incomplete as delivered.

**Required fix:** add the missing `restart_authorization_listener.py` module (with its composition-root wiring), or remove the listener-specific test content and file the BaaS-publish-event integration as separate, correctly scoped follow-up work.

## Out-of-scope notes
- `strategy.py`'s new `_run()` closure calls `asyncio.run()` inside a raw `threading.Thread(daemon=True)` and never joins/tracks it; that's a runtime-robustness concern (unhandled thread lifecycle, no cancellation on process shutdown) rather than a layering violation, and is left to the feature reviewer.
- The persisted `_aicoding_restart` / `resync_authorization` marker written into `template_config` is engine-owned internal bookkeeping (read and cleared only by this same strategy class), consistent with this codebase's existing pattern of engine-owned keys inside `template_config`/`ext_config` (e.g. `thetaKey`). Not flagged as a hidden cross-component protocol.

---

totalfrank (changes requested):
## Architecture-Gate Review

**1. Verdict: REQUEST_CHANGES**

**2. Gate 0**
- Target branch: `dev` (the integration branch) — release-branch (`REL*`) gate does not apply. Gate 0 passes trivially; the diff still needs full architecture review below.
- Classification: feature/fix to the `aicoding` engine's restart authorization path, landing on `dev` as normal.

**3. Findings**

**Finding 1 — Contract change shipped without updating the contract, the sibling implementation, or the sole caller; the feature this PR adds cannot run (arch.rules.md Rule 1, Rule 4, Rule 16 — Invariant)**

`EngineProvisioningStrategy.refresh_restart_authorization` is the declared Service API contract that `bot_service.py` (consumer) calls into per-engine strategies (this is exactly the "one contract, defined once" pattern the constitution requires for engine variability). This PR changes only the `aicoding` implementation's signature and return type, and leaves the contract, the default no-op implementation, and the only caller untouched:

- `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:650-659` (this PR) — new signature:
  ```python
  def refresh_restart_authorization(
      self, ctx, bot, extra_configs, *,
      mcp_sync: Any = None,
      skill_set_factory: Any = None,
      template_service: Any = None,
  ) -> bool:
  ```
  drops `passport_plugin` and `caller_identity_repo`, adds `mcp_sync`, and changes the return type from `None` to `bool`.

- `src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py:227-238` (untouched) — the abstract contract still declares the *old* shape: `passport_plugin`, `skill_set_factory`, `template_service`, `caller_identity_repo` all required keyword-only, `-> None`.

- `src/backend/src/agentclaw/community/core/bot_management/engines/default.py:113-124` (untouched) — the default/fallback implementation for every other engine still matches the *old* contract exactly (same four required kwargs, `-> None`).

- `src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py:4316-4324` (untouched, the only call site) —
  ```python
  strategy.refresh_restart_authorization(
      ctx, bot, extra_configs,
      passport_plugin=self._passport_plugin,
      skill_set_factory=self._skill_set_factory,
      template_service=self._template_service,
      caller_identity_repo=self._caller_identity_repo,
  )
  ```
  still calls with the old keyword arguments and never passes `mcp_sync`.

Effect: at the real call site, `AicodingProvisioningStrategy.refresh_restart_authorization` no longer accepts `passport_plugin=` or `caller_identity_repo=` — this call raises `TypeError` on every single aicoding bot restart. That `TypeError` is swallowed by the broad `except Exception` a few lines below in `bot_service.py` (logged as a warning, restart continues), so nothing crashes visibly — but it also means the MCP-projection resync logic this PR adds (`_run()`'s `mcp_sync` branch, `skill_set_service.sync_mcp_projection`, `sync_runtime`) can never execute in production: the method never gets past argument binding. The PR's stated purpose ("use MCP projection on restart") is dead code as shipped.

This is a textbook Rule 16 violation — a Service API contract's implementation changed without propagation to the contract definition, the sibling implementation, or the consumer — compounded by Rule 4 (the `aicoding` strategy no longer means what the declared contract says) and Rule 1 (the contract in `provisioning.py` is no longer authoritative for what `aicoding` actually does).

**Required fix:** update `provisioning.py`'s abstract method to the new signature (or a compatible superset), update `default.py` to match, and update `bot_service.py`'s call site to pass `mcp_sync` and drop the removed kwargs — as one coordinated change, not three independent ones.

**Finding 2 — The PR's own test suite is inconsistent with the unmodified `default.py`, and would fail if run (Rule 25 — conformance)**

`test_restart_authorization_refresh.py:361-371` (this PR):
```python
def test_default_strategy_always_returns_false() -> None:
    strategy = DefaultProvisioningStrategy("openclaw")
    assert (
        strategy.refresh_restart_authorization(
            _ctx(), _bot(), {"confirmed_template_update": True},
            mcp_sync=MagicMock(),
            skill_set_factory=MagicMock(),
        )
        is False
    )
    assert strategy.refresh_restart_authorization(_ctx(), _bot(), None) is False
```
This calls the default strategy with the *new* keyword shape (`mcp_sync=`, no `passport_plugin`/`template_service`/`caller_identity_repo`) and expects a boolean return. But `default.py:113-124` was not touched by this PR and still requires `passport_plugin`, `skill_set_factory`, `template_service`, and `caller_identity_repo` as required keyword arguments and returns `None`. Both calls above raise `TypeError` against the actual `default.py` on `dev`. This is independent confirmation of Finding 1: the diff is a partial port (`strategy.py` + its tests only, per the PR description "从 dev 新分支搬运重启链路的 MCP projection 改动") that dropped the accompanying changes to the contract, default implementation, and caller.

**Required fix:** same as Finding 1 — `default.py` must be updated in the same change, and the test must exercise the actual (updated) `default.py` and the actual (updated) call site, not a shape neither one implements yet.

**4. Out-of-scope notes**
- The MCP-projection resync business logic itself (marker persistence, opt-in via `confirmed_template_update`, background-thread fire-and-forget via `bind_current_avernet_tenant`) follows established patterns elsewhere in this codebase and is not evaluated here beyond the contract-propagation issue above.
- The PR description is a single line with no Problem/Solution/Validation/Compatibility sections and no propagation analysis for a Service API contract change, which Rule 16's required checks call for explicitly; once Findings 1–2 are fixed, the description should also be filled in per `AGENTS.md`.

---

totalfrank (changes requested):
**Verdict: REQUEST_CHANGES**

## Gate 0

- Target branch: `dev` (the integration branch) — not a `REL*` release branch, so the release-branch routing rules and Isolation Audit do not apply.
- Classification: ordinary feature/fix work targeting `dev`. Gate 0 **passes** on branch policy alone. (Note: GitHub reports `mergeable_state: dirty` — there is a merge conflict against `dev` independent of this review; that blocks merge but is not an architecture finding.)

## Findings

### 1. (Most severe) `refresh_restart_authorization` contract change was not propagated — breaks the production call path and the test suite (arch.rules.md Rule 16 — Invariant; also Rules 1, 4, 25)

This PR rewrites `AicodingProvisioningStrategy.refresh_restart_authorization`'s signature and return type, but only touches `strategy.py` and its own test file (2 files, 1 commit). It does **not** update the abstract contract, the default implementation, or the one production caller — all three still expect the old shape:

- `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:653-661` (new, unpropagated signature):
  ```python
  def refresh_restart_authorization(
      self, ctx, bot, extra_configs, *,
      mcp_sync: Any = None,
      skill_set_factory: Any = None,
      template_service: Any = None,
  ) -> bool:
  ```
  `passport_plugin` and `caller_identity_repo` were dropped, `mcp_sync` was added, and the return type changed from `None` to `bool`.

- `src/backend/src/agentclaw/community/core/bot_management/engines/provisioning.py:227-238` — the `EngineProvisioningStrategy` **abstract contract** (the Service API this class implements) is unchanged: it still declares `passport_plugin`, `skill_set_factory`, `template_service`, `caller_identity_repo` as required kwargs and `-> None`. The contract this method implements now disagrees with the implementation.

- `src/backend/src/agentclaw/community/core/bot_management/engines/default.py:113-124` — `DefaultProvisioningStrategy`'s fallback implementation (the mandatory default for the engine-variability contract, Rule 11) is likewise unchanged and still only accepts the old kwargs.

- `src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py:4316-4324` — the **only production caller** still calls with the old signature:
  ```python
  strategy.refresh_restart_authorization(
      ctx, bot, extra_configs,
      passport_plugin=self._passport_plugin,
      skill_set_factory=self._skill_set_factory,
      template_service=self._template_service,
      caller_identity_repo=self._caller_identity_repo,
  )
  ```
  `passport_plugin=` and `caller_identity_repo=` are no longer accepted parameters on the `aicoding` implementation, so this call raises `TypeError: refresh_restart_authorization() got an unexpected keyword argument 'passport_plugin'` on every restart of an `aicoding`/`claude_code` bot. That `TypeError` is swallowed by the broad `except Exception` a few lines above (`bot_service.py:4325-4333`), so the entire feature this PR adds — MCP projection resync, skill-symlink resync, persisted restart-marker clearing — is unreachable dead code in production, silently logged as "applying restart extra configs failed" instead of doing anything.

  This also means `mcp_sync` (a real DI-wired `McpSyncProtocol`, see `di/modules/devices_module.py:251` and `di/modules/infrastructure/singlebox/devices.py:192`) is never actually wired into this call site at all.

- The new/rewritten test file itself does not import cleanly: `src/backend/tests/community/core/bot_management/services/test_restart_authorization_refresh.py:28-30` imports
  ```python
  from agentclaw.community.core.bot_management.engines.aicoding.restart_authorization_listener import (
      AicodingRestartAuthorizationBaasPublishListener,
  )
  ```
  but `restart_authorization_listener.py` does not exist anywhere in the repository (confirmed by `find`), so this test module fails to collect (`ModuleNotFoundError`). Separately, `test_default_strategy_always_returns_false` (lines 361-368) calls `DefaultProvisioningStrategy(...).refresh_restart_authorization(..., mcp_sync=MagicMock(), skill_set_factory=MagicMock())` against the real, unmodified `DefaultProvisioningStrategy`, which does not accept `mcp_sync` — that assertion would also fail/raise if it ever ran.

  The PR description ("从 dev 新分支搬运重启链路的 MCP projection 改动" — ported from another branch) and the branch name (`...-rebased`) are consistent with this being an incomplete cherry-pick: the commit(s) that update `provisioning.py`, `default.py`, `bot_service.py`, and add `restart_authorization_listener.py` were left behind.

  **Required fix:** land this as one coherent, propagated change: update the `EngineProvisioningStrategy` abstract contract in `provisioning.py` to the new signature/semantics, update `DefaultProvisioningStrategy` in `default.py` to match (it's the mandatory default fallback for this contract, Rule 11/25), update the `bot_service.py` call site to wire the real `mcp_sync` dependency instead of the removed `passport_plugin`/`caller_identity_repo`, and either include `restart_authorization_listener.py` in this PR or drop the tests that depend on it. A conformance test that exercises both `DefaultProvisioningStrategy` and `AicodingProvisioningStrategy` against the same signature (Rule 25) would have caught this before merge.

### 2. Internal restart bookkeeping is embedded, undocumented, into the Service-API-exposed `template_config` blob (Rule 1, Rule 4)

`strategy.py` (`_AICODING_RESTART_MARKER_KEY = "_aicoding_restart"`, `_with_restart_resync_marker`, `_has_restart_resync_marker`, `_clear_restart_resync_marker`) persists a new internal marker —
```python
{"_aicoding_restart": {"resync_authorization": True, "template_version_id": ..., "source": "restart_template_update"}}
```
— directly inside `template_config` via `template_service.update_template(...)`. `template_config` is not an internal-only structure: it is the literal payload type on the bot-management Service API (`adapters/http/bot_management/schemas.py:35,45,53`, `template_config: Optional[Dict[str, Any]]`), read and written through HTTP. Per Rule 17's compatibility test ("if changing something could cause a correctly written consumer to fail, treat it as constrained"), this field is a constrained surface, yet this PR adds new semantics to it with no contract/schema update, no documentation of the key, and no filtering before it round-trips through the API to any consumer (frontend template editor, other engines' template merges, exports). A future reader of `template_config` has no way to know this key exists, is engine-internal, or must be preserved/ignored.

**Required fix:** keep this bookkeeping out of the general-purpose `template_config` payload — store the pending-resync flag in an AICoding-owned side field/table that isn't part of the Service API's `Dict[str, Any]` contract, or, if it must ride in `template_config`, document and version it explicitly per Rule 1/16 and confirm downstream consumers tolerate/strip unknown keys.

## Out-of-scope notes

- Business semantics of the MCP scope refresh, skill-symlink resync ordering, and the read-modify-write race between `apply_restart_extra_configs`'s own template write and the fire-and-forget thread's marker-clearing write are not gated on here.
- The `mergeable_state: dirty` conflict against `dev` needs to be resolved regardless of the above.


---

PR #1685, opened 2026-08-30
fix(aicoding): sync MCP desired state on restart

panqiwp:
Align restart sync with MCP desired state.\n\nThis change only updates the restart path to use sync_mcp_desired_state.\n\nNotes:\n- Keep the diff minimal\n- No unrelated bus/event changes\n- Includes the corresponding unit test update

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -683,9 +683,28 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
             import asyncio
 
             refresh_succeeded = True
+            skill_set_service = None
+            if skill_set_factory is not None:
+                try:
+                    skill_set_service = skill_set_factory.create(
+                        user_id=effective_entity_id,
+                        entity_id=effective_entity_id,
+                        bot_id=ctx.bot_id,
+                        entity_type=effective_entity_type,
+                        engine_type=effective_engine,
+                    )
+                except Exception as skill_error:
+                    refresh_succeeded = False
+                    logger.error(
+                        "[aicoding.restart] skill set service create error: "
+                        "bot_id=%s, engine_type=%s, error=%s",
+                        ctx.bot_id, effective_engine, skill_error,
+                        exc_info=True,
+                    )
+
             if mcp_sync is not None:
                 try:
-                    async def _do_mcp_sync() -> tuple[dict, dict | None]:
+                    async def _do_mcp_sync() -> tuple[dict, bool | None]:
                         scope_result = await mcp_sync.refresh_mcp_scope(
                             user_id=effective_entity_id,
                             entity_id=effective_entity_id,
@@ -696,16 +715,24 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
                         if not scope_result.get("success"):
                             return scope_result, None
 
-                        detail_result = await mcp_sync.sync_mcp_desired_state(
-                            user_id=effective_entity_id,
-                            entity_id=effective_entity_id,
-                            bot_id=ctx.bot_id,
-                            entity_type=effective_entity_type,
-                            engine_type=effective_engine,
-                        )
-                        return scope_result, detail_result
+                        detail_synced = None
+                        if skill_set_service is not None:
+                            declared_server_codes = set(
+                                await asyncio.to_thread(
+                                    skill_set_service.get_bot_mcp_codes,
+                                    effective_entity_id,
+                                    ctx.bot_id,
+                                    effective_entity_id,
+                                    effective_entity_type,
+                                    effective_engine,
+                                )
+                            )
+                            detail_synced = await skill_set_service.sync_mcp_desired_state(
+                                server_codes=declared_server_codes,
+                            )
+                        return scope_result, detail_synced
 
-                    scope_result, detail_result = asyncio.run(_do_mcp_sync())
+                    scope_result, detail_synced = asyncio.run(_do_mcp_sync())
                     if not scope_result.get("success"):
                         refresh_succeeded = False
                         logger.error(
@@ -713,12 +740,13 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
                             "bot_id=%s, engine_type=%s, error=%s",
                             ctx.bot_id, effective_engine, scope_result.get("error"),
                         )
-                    elif detail_result and not detail_result.get("success"):
+                    elif detail_synced is False:
                         refresh_succeeded = False
                         logger.error(
-                            "[aicoding.restart] MCP detail resync failed: "
+                            "[aicoding.restart] MCP desired-state resync failed: "
                             "bot_id=%s, engine_type=%s, error=%s",
-                            ctx.bot_id, effective_engine, detail_result.get("error"),
+                            ctx.bot_id, effective_engine,
+                            "desired-state declaration returned false",
                         )
                     else:
                         logger.info(
@@ -735,15 +763,8 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
                         exc_info=True,
                     )
 
-            if skill_set_factory is not None:
+            if skill_set_service is not None:
                 try:
-                    skill_set_service = skill_set_factory.create(
-                        user_id=effective_entity_id,
-                        entity_id=effective_entity_id,
-                        bot_id=ctx.bot_id,
-                        entity_type=effective_entity_type,
-                        engine_type=effective_engine,
-                    )
                     # ``project_skills`` is async so that both halves of the
                     # capability boundary are awaited the same way; this is a
                     # worker thread with no running loop, so it bridges the
--- a/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py
@@ -696,7 +696,7 @@ class AicodingProvisioningStrategy(EngineProvisioningStrategy):
                         if not scope_result.get("success"):
                             return scope_result, None
 
-                        detail_result = await mcp_sync.sync_mcp_details(
+                        detail_result = await mcp_sync.sync_mcp_desired_state(
                             user_id=effective_entity_id,
                             entity_id=effective_entity_id,
                             bot_id=ctx.bot_id,
--- a/src/backend/tests/community/core/bot_management/services/test_restart_authorization_refresh.py
+++ b/src/backend/tests/community/core/bot_management/services/test_restart_authorization_refresh.py
@@ -3,7 +3,7 @@
 ``refresh_restart_authorization`` now owns both the opt-in decision and the
 AICoding-specific side effects:
   * refresh MCP scope;
-  * push MCP details to runtime;
+  * declare MCP desired state to runtime;
   * resync ~/.claude/skills symlinks via SkillSetService.project_skills().
 All side effects are best-effort/fire-and-forget.
 """
@@ -98,9 +98,10 @@ def test_aicoding_refresh_updates_mcp_and_skill_symlinks(flag) -> None:
     strategy = AicodingProvisioningStrategy("aicoding")
     mcp_sync = MagicMock()
     mcp_sync.refresh_mcp_scope = AsyncMock(return_value={"success": True})
-    mcp_sync.sync_mcp_desired_state = AsyncMock(return_value={"success": True})
     factory = MagicMock()
     skill_set_service = MagicMock()
+    skill_set_service.get_bot_mcp_codes.return_value = ["mcp-a", "mcp-b"]
+    skill_set_service.sync_mcp_desired_state = AsyncMock(return_value=True)
     skill_set_service.project_skills = AsyncMock(return_value=True)
     factory.create.return_value = skill_set_service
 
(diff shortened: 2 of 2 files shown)


totalfrank (changes requested):
**Verdict: REQUEST_CHANGES**

## Gate 0
Target branch: `dev` (integration branch — new work lands here by default). Branch name is `cherry-pick/5c93df3f`, but no such commit exists in this repo's history and the PR carries a single fresh commit with no `(#NNN)`/cherry-pick reference — regardless, since the target is `dev` and not a `REL*` branch, the release-branch gate does not apply here. Gate 0: **pass** (target-branch rule not triggered).

## Findings

**1. `strategy.py:699` — calls an undeclared/nonexistent Service API method; will `AttributeError` at runtime and silently disable the exact restart resync this PR claims to fix.**

```python
detail_result = await mcp_sync.sync_mcp_desired_state(
    user_id=effective_entity_id,
    entity_id=effective_entity_id,
    bot_id=ctx.bot_id,
    entity_type=effective_entity_type,
    engine_type=effective_engine,
)
```

Violated rules: **arch.rules.md Rule 1** ("The API Specification Is the Single Authority" — Invariant) and **Rule 4** ("Interfaces Mean What Their Specs Define Them to Mean" — Invariant).

`mcp_sync` here is `MCPSyncService` (wired directly, untyped-cast, in `di/modules/bot_management_module.py:286-292`), whose declared Service API contract is `McpSyncProtocol` (`core/devices/protocols.py:195-221`) / `MCPSyncServiceProtocol` (`api/mcp_sync_service.py:10-36`). Neither the contract nor the concrete `MCPSyncService` implementation (`core/mcp/services/sync_service.py`) declares a `sync_mcp_desired_state` method — both still only expose `sync_mcp_details`, `sync_mcp_detail`, `sync_mcp_details_for_bot`, `remove_mcp_detail`, `refresh_mcp_scope`, etc.

`sync_mcp_desired_state` *does* exist — but as an unrelated method on a completely different class, `SkillSetService` (`core/skill_center/services/skill_set_service.py:652`), which declares the skill-driven MCP *allow-list* to the skill center. It has nothing to do with `MCPSyncService`'s device MCP-detail push. Calling it on the `mcp_sync` object is calling a symbol that doesn't exist on that object.

At runtime this raises `AttributeError`, which is caught by the broad `except Exception` in `_run()` (`strategy.py:729`), logged as `[aicoding.restart] MCP resync error`, and the whole restart MCP resync — the feature this PR is titled as fixing — silently fails every time it runs. The previous code (`mcp_sync.sync_mcp_details(...)`) was a valid call against the declared contract; this change regresses it into a guaranteed no-op failure disguised as a fix.

The tests were only touched to rename the attribute on ad hoc `MagicMock()` doubles (`test_restart_authorization_refresh.py`), which are not `spec`'d against `McpSyncProtocol`/`MCPSyncServiceProtocol`, so they pass regardless of whether the real object implements the method — this is the propagation-analysis gap Rule 16 calls out (a contract-adjacent call change with no corresponding contract or implementation update, and no conformance test per Rule 25 to catch the drift).

**Required fix**: revert this call to `mcp_sync.sync_mcp_details(...)` to match the declared `McpSyncProtocol`/`MCPSyncServiceProtocol` contract, or, if the actual intent is to push the skill-declared MCP desired-state (a different concern), invoke it through the already-available `skill_set_factory` parameter in this same function (which resolves a `SkillSetService` with the real `sync_mcp_desired_state` method) rather than renaming the unrelated `mcp_sync` call.

## Out-of-scope notes
- None — this PR's only production hunk is the finding above; no other layering, naming, or business-logic concerns apply.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

### Gate 0
- **Target branch:** `dev` (the integration branch). Not a `REL*` branch, so the release-branch routing / Isolation Audit rules don't apply here.
- **Classification:** Rename of a call-site method invocation plus matching test-mock updates. Normal `dev` work.
- **Result:** Gate 0 passes trivially (right branch). The verdict below is driven entirely by a contract violation in the diff itself, not by branch policy.

### Findings

**1. `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:699` — call renamed to a method that does not exist on the wired contract; the "fix" this PR claims to restore is silently broken.**

```python
detail_result = await mcp_sync.sync_mcp_desired_state(
    user_id=effective_entity_id,
    entity_id=effective_entity_id,
    bot_id=ctx.bot_id,
    entity_type=effective_entity_type,
    engine_type=effective_engine,
)
```

- `mcp_sync` here is documented/typed elsewhere as `McpSyncProtocol` (`core/devices/protocols.py:195`). That Protocol declares `sync_mcp_details`, `sync_mcp_detail`, etc. — **it has no `sync_mcp_desired_state` method.**
- The concrete object actually wired into this call path via DI is `MCPSyncService` (`core/mcp/services/sync_service.py`), injected as `mcp_sync_service: MCPSyncService` in `di/modules/bot_management_module.py:286,372` and `cast(McpSyncProtocol, mcp_sync_service)` in `di/modules/devices_module.py:251` / `di/modules/infrastructure/singlebox/devices.py:192`. `MCPSyncService` only defines `sync_mcp_details` (line 148 on `dev`) — **not** `sync_mcp_desired_state`.
- The *only* method named `sync_mcp_desired_state` anywhere in the repo belongs to a completely unrelated class with a completely different signature: `SkillSetService.sync_mcp_desired_state(self, *, server_codes: set[str]) -> bool` (`core/skill_center/services/skill_set_service.py:652`, introduced by the mcp-sync-and-passport-regressions work, PR #1568 already on `dev`). This PR keeps the old `sync_mcp_details`-shaped keyword arguments (`user_id`, `entity_id`, `bot_id`, `entity_type`, `engine_type`) — it did not adopt `SkillSetService`'s `server_codes=` signature either. The two `sync_mcp_desired_state` names are coincidental, not the same contract.
- Net effect: at runtime this call raises `AttributeError: 'MCPSyncService' object has no attribute 'sync_mcp_desired_state'`. It's caught by the surrounding `except Exception as mcp_error` (`strategy.py:729`), logged as `"[aicoding.restart] MCP resync error"`, and `refresh_succeeded` is set to `False`. So this cherry-pick doesn't restore MCP projection sync on restart — it silently fails it every time, worse than before (previously it called a method that actually exists).
- **Violated rules:**
  - **Rule 1 — The API Specification Is the Single Authority (Invariant):** the call now depends on semantics/behavior not declared by `McpSyncProtocol`.
  - **Rule 4 — Interfaces Mean What Their Specs Define Them to Mean (Invariant):** the caller assumes a method the contract doesn't define; "caller convenience" (matching a same-named method on an unrelated service) is exactly what this rule forbids.
  - **Rule 16 — Changes Propagate (Invariant):** if this were an intentional protocol evolution, `McpSyncProtocol` and `MCPSyncService` needed to change in the same PR, with propagation analysis (affected consumers/implementations/compatibility). Instead only the caller and its mocks changed. This is precisely the "a contract changes without propagation analysis" Structural Review Red Flag.
- **Required fix:** Revert the call back to `mcp_sync.sync_mcp_details(...)` unless the real intent is to route this restart-resync path through `SkillSetService.sync_mcp_desired_state(server_codes=...)` (the declaration-only method from #1568). If it's the latter, `strategy.py` needs to depend on a `SkillSetService`-shaped contract (not `McpSyncProtocol`), pass `server_codes` instead of the `user_id/entity_id/bot_id/entity_type/engine_type` kwargs, and the DI wiring/tests need to change accordingly — with propagation analysis for the swap.

**2. `src/backend/tests/community/core/bot_management/services/test_restart_authorization_refresh.py` — the test doubles mask the contract break instead of catching it (Rule 25 gap).**

Every updated test builds `mcp_sync = MagicMock()` and then does `mcp_sync.sync_mcp_desired_state = AsyncMock(...)`. Because `mcp_sync` is untyped in the production signature (`mcp_sync: Any = None`) and the test double is a bare `MagicMock` rather than something bound to (and validated against) `McpSyncProtocol`, the tests happily accept a method name that no real implementation in the DI graph provides. This is why CI didn't catch finding #1: there is no conformance test asserting the `strategy.py` MCP-sync call matches `McpSyncProtocol`/`MCPSyncService`'s real surface (Rule 25 — Protocols Have Self-Validating Contracts). This isn't a request for more business-logic test coverage — it's that the existing tests give false confidence about a broken contract call.

### Out-of-scope notes
- The `"cherry-pick"` framing / commit message correctness (whether this really matches a reviewed `dev` commit) is a process question, not architecture — not gated here since the diff targets `dev` directly.
- No comment on whether resyncing MCP on restart is the right product behavior — only on the fact that, as written, it cannot execute successfully.


---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0
Target branch: `dev` (not a `REL*` release branch) — release-branch routing rules do not apply here; this is ordinary integration-branch flow. Classification: claimed existing-behavior fix/cherry-pick ("cherry-pick/5c93df3f"). **This claim does not hold up**: I could not find commit `5c93df3f` anywhere in `dev`'s history, and — more importantly — `dev`'s current `strategy.py` still calls `mcp_sync.sync_mcp_details(...)`, not `sync_mcp_desired_state`. So the described "backport" target doesn't exist on `dev` as stated. Gate 0 itself doesn't fail the PR (wrong target isn't the issue), but the premise of the change is unverifiable and, per the investigation below, appears to be actively wrong.

## Findings

1. **`src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:699`** — 
   ```python
   detail_result = await mcp_sync.sync_mcp_desired_state(
       user_id=effective_entity_id,
       entity_id=effective_entity_id,
       bot_id=ctx.bot_id,
       entity_type=effective_entity_type,
       engine_type=effective_engine,
   )
   ```
   `mcp_sync` here is typed/injected as `McpSyncProtocol` (`core/devices/protocols.py:195`), and its concrete implementation is `MCPSyncService` (`core/mcp/services/sync_service.py`). **Neither the Protocol nor the concrete implementation defines `sync_mcp_desired_state`** — the only method with that name in the entire repo lives on an unrelated class, `SkillSetService.sync_mcp_desired_state(self, *, server_codes: set[str]) -> bool` (`core/skill_center/services/skill_set_service.py:652`), with a completely different signature (no `user_id`/`entity_id`/`bot_id`/`entity_type`/`engine_type`). The only method the real `McpSyncProtocol` contract declares for this purpose is `sync_mcp_details(*, user_id, entity_id, bot_id, entity_type, engine_type=None, active_only=False)` (`core/devices/protocols.py:202`), which is exactly what this diff removes.

   **Violated rule**: arch.rules.md Rule 1 ("The API Specification Is the Single Authority" — Invariant) and Rule 4 ("Interfaces Mean What Their Specs Define Them to Mean" — Invariant): the code now depends on behavior that is not documented in, or part of, the `McpSyncProtocol` Service API contract.

   **Concrete failure**: at runtime, whenever `refresh_restart_authorization` runs with a real (non-`None`) `mcp_sync` — i.e. exactly the AICoding-restart resync path this PR claims to fix — `mcp_sync.sync_mcp_desired_state(...)` raises `AttributeError`, because `MCPSyncService` has no such attribute. This is caught by the surrounding `except Exception`, so it fails silently/best-effort rather than crashing, but the net effect is that **MCP scope/detail resync on restart is now permanently broken** for every real caller. The unit tests in this same diff don't catch it because they replace `mcp_sync` with a bare `MagicMock()`, which happily accepts any attribute name — the test double doesn't respect the contract's actual shape (Rule 21/25 concern: an isolated test double must still respect contract semantics; conformance can't be validated against an untyped mock that fabricates any method name on demand).

   **Required fix**: revert this call back to `mcp_sync.sync_mcp_details(...)` (matching the declared `McpSyncProtocol`/`MCPSyncService` contract), or, if the intent really is to route restart-resync through `SkillSetService.sync_mcp_desired_state(server_codes=...)` instead, that requires an actual contract change: update `McpSyncProtocol`/inject the right service, adjust the call to the real `(server_codes: set[str])` signature, and add propagation analysis (Rule 16) plus a typed/contract-conformant test double so the mismatch would be caught statically or by conformance tests (Rule 25), not masked by `MagicMock`.

2. **Test file** (`src/backend/tests/community/core/bot_management/services/test_restart_authorization_refresh.py`) — every assertion was mechanically renamed from `mcp_sync.sync_mcp_details` to `mcp_sync.sync_mcp_desired_state` in lock-step with finding #1. Because `mcp_sync` in these tests is a `MagicMock()` rather than something bound to the `McpSyncProtocol` contract, the suite continues to pass while asserting behavior against a method name the real service doesn't implement. This isn't a coverage nitpick — it's the reason finding #1 shipped without any test catching it.

No engine-string branching, new engine-named modules, filesystem-layout hardcoding, or plugin-vs-internal-seam issues are introduced by this diff — the touched file is the existing, already-established per-engine strategy module (`engines/aicoding/strategy.py`), which is the correct location for AICoding-specific provisioning logic under the existing contract pattern. That part of the layering is fine.

## Out-of-scope notes
- Business logic of *why* restart needs an MCP resync at all is assumed correct per the review scope; the issue here is purely that the renamed call doesn't resolve to any real method.
---

---

panqiwp:
Thanks for the review. Fixed in the latest push:

- restart path now uses `SkillSetService.sync_mcp_desired_state(server_codes=...)` instead of calling it on `mcp_sync`
- `server_codes` are derived via `skill_set_service.get_bot_mcp_codes(...)`
- updated the unit tests to assert the new contract

I also re-ran the targeted test file and it passes locally. Please re-check when convenient.

panqiwp:
Closing this PR and reopening as a fresh one with the updated implementation.

Issue #1621, opened 2026-08-27
perf(backend): flush_installations runs four times per SkillSet mutation


totalfrank:
Split out of the SkillSet projection latency work as **P3**. Full analysis with code and a per-file change sketch lives in `specs/2026-08-27-skillset-projection-latency/spec.md` (section P3) and `plan.md`, landed by inclusionAI/Avernet#1617.

P1 and P2 from that spec are being implemented now; P3 is deferred to its own change.

## Problem

Every read through `BotCapabilityStateReader` flushes SkillSet configuration into Installation before answering — by design:

```python
# src/backend/src/agentclaw/community/core/skill_center/services/bot_capability_state_reader.py:56
def active_skill_assets(self, *, bot_id, owner_id, bot=None):
    bot = self._bot(bot_id=bot_id, owner_id=owner_id, bot=bot)
    self._flush(bot=bot, bot_id=bot_id, owner_id=owner_id)      # full plan resolve
    return tuple(self._pool_skills.list_bot_installed_assets(...))
```

One activate or deactivate reaches that flush four times:

1. `_mutation_flow.py:129` — `snapshot_skill_mappings` before the mutation (pre-mutation state, genuinely distinct)
2. `_mutation_flow.py:160` — `snapshot_skill_mappings` after the mutation
3. `bot_runtime_projector.py` `_build_plan` — `active_skill_assets`, same post-mutation state
4. `skill_set_service.py:1742` `_installed_mcp_codes` → `active_mcp_server_codes`, same post-mutation state

Each flush is a `_resolve_flush_plan`: one query per SkillSet row, plus per-set membership and Default-Set exclusion queries. Flushes 2, 3 and 4 observe identical post-mutation state and produce identical results. `list_installed_mcps` likewise runs twice over unchanged data.

## Proposed fix

Resolve the post-mutation state once and thread it, rather than caching:

- Add `snapshot_skill_assets` to the projector and both protocol declarations, keeping the teclaw `center://` guard.
- `project` accepts `skill_assets: Sequence[RegisteredSkillAsset] | None = None`; `_build_plan` uses it instead of calling the reader when supplied. `None` keeps today's behavior, so `SkillSymlinkListener`'s direct `project(...)` is unaffected.
- `MutationProjectionFlow` resolves assets once after the mutation, derives `current_mappings` from them, and passes them into `project`.
[...]

PR #1694, opened 2026-08-30
feat(backend): add internal Installation backfill endpoints

totalfrank:
## Problem

The Installation flush (`CapabilityDesiredStateRepository.flush_installations`) is only reachable lazily, in front of a capability read (`BotCapabilityStateReader`). That is enough for per-Bot commands, which write Installation eagerly through the UoW. It is not enough for configuration that reaches many Bots at once and has no per-Bot write to ride on:

- **The shared platform Default Set.** `default_skillset_projection.global_default_scope` selects one platform-owned row per engine type that every Bot with that engine inherits. `/api/skillsets/admin/set-default-skills`, `/admin/set-default-skills-fast` and `/admin/init-and-sync` clear and rewrite its membership with no Installation writes and no fan-out. The SSOT spec left these writers off the UoW deliberately.
- **`/api/skillsets/admin/set-skillset-active`**, which flips `is_active` by env+id with no Bot/owner scope. `specs/2026-08-24-installation-single-source-of-truth/spec.md` §7: "drift the lazy flush repairs on the next read. (This tolerance is also why the flush must exist on the read side at all.)"
- **`center://` version resolution.** `set_member_skill_ids` resolves a `center://` membership to the highest PUBLISHED version *at flush time*, so publishing a new version changes what a Set means for every Bot holding it, driven by publication rather than by any Bot-scoped command.
- **New Bots.** Nothing in `create_flow` writes Installation rows; a fresh Bot's Default-Set capabilities materialize only when the first flush runs.

In each case a Bot converges when something happens to read it, and there is no way to converge it on purpose.

## Solution

One endpoint, behind a Bearer token:

```
POST /api/internal/skill-center/installations/backfill/bot
     {"bot_id": "...", "owner_id": "..."}
  -> {"success": true, "data": {"bot_id": "...", "owner_id": "..."}}
  -> 404 when that pair does not exist
```

It runs `flush_installations` for that one Bot, through the same `bot_engine_scope` helpers the reader uses, so the backfill and the lazy flush cannot disagree about which Sets reach a Bot. DB-side only, exactly like the flush: no device is touched and no runtime projection is triggered, so a Bot converged this way still needs a projection before its engine sees the change.

**One Bot per call, by design.** Choosing the pairs and pacing the calls belongs to whoever drives the backfill — a shell loop with its own QPS control — not to the endpoint. It 404s an unknown pair so a typo cannot read as "nothing needed converging".

Auth mirrors `/api/internal/dormant/*`: its own secret name (`SecretNamesConfig.skill_center_internal_token`) resolved through `SecretResolver`, failure-closed to 401 when the resolver raises, and granted separately from the dormant token. Only the secret's *name* lives in yaml, never its value. The provisioned name `other_manual_agentclaw_skill_center_internal_token` is registered in `application-singlebox.yaml` beside `aiworkbench_repo_url`, the one config file `test_shipped_config_no_corp_identifiers` exempts; pre/prod take the name from their corp env overlay, per `SecretNamesConfig`'s docstring.

**Rejected:** hanging this off the existing `/api/skillsets/admin/*` surface. Those routes carry no auth guard at either route or router level.

## Review history

Two substantive rounds changed the shape of this PR:

**Scope reduced.** The first revision had a `backfill/page` batch endpoint and an `InstallationFlushPlan.changed` field. The batch sweep solved a problem the caller does not have — and solved it badly for the actual use, paging newest-first with no QPS control. `changed` existed only so a page run could report which Bots it converged without re-reading Installation; with batch gone it did not justify touching a shared repository type. Both deleted. **This PR now touches no repository code at all** — `git diff origin/dev -- core/repository/` is empty.

**A security bug caught in review.** `skill_center_internal_token` was added to `SecretNamesConfig` but never read by `ConfigModule.secret_names()`, the only place the injected config is built. It therefore always held `""`, so `_resolved_internal_token` saw "no secret configured" and returned the hardcoded, publicly-visible singlebox fallback token **in every environment, production included** — any caller presenting the known string would have been authorized, and setting the name in any yaml did nothing.

Fixed in `b670468`, and fixed at the mechanism rather than the instance: `secret_names()` now builds reflectively over `dataclasses.fields`, so a future field cannot be silently forgotten. New `tests/community/di/test_secret_names_config_provider.py` sweeps every field of `SecretNamesConfig` through the real provider — nothing covered it before, which is why the gap survived several rounds. Verified the test fails (2 cases) with the fix reverted.

## Validation

Rebased onto `4f21dc7` and re-run there:

- `pytest tests/community/architecture tests/community/config tests/community/di tests/community/core/skill_center tests/community/repository/skill_center tests/community/endpoints` — **3085 passed, 30 skipped**.
- `scripts/ci/python_sast_local.sh src/backend 1 --base origin/dev --head HEAD` — clean, exit 0.

Tests here: 4 service, 6 endpoint-adapter, 5 auth-guard, 5 token-resolution, 7 config-provider, and 2 framework endpoint cases (happy + error for the route), the last running against the real assembled app with no mocking.

Four repo-wide gates were tripped along the way and fixed rather than relaxed:

- `test_effective_config_matches_golden` — regenerated via `tests.community.config.regen_golden`, the procedure that test documents. `typed.secret_names.skill_center_internal_token` now carries the real name instead of `""`, which is the wiring fix proving itself through the config path.
- `test_endpoint_coverage_against_baseline` — added real happy/error cases in `tests/community/endpoints/` rather than lines to `coverage_baseline.txt`, which is a shrinking debt list.
- `test_no_oversized_modules` — `skill_center_module.py` hit 1058, `app.py` 1001, and later `config_module.py` 1004. Fixed by extracting the token provider into its own bindings module, folding the router import into the existing `skill_center` import line, and making `secret_names()` reflective (998). Note `app.py` sits at exactly 1000; it was at 999 before, so the next router added to it trips the cap regardless of this PR.
- `test_shipped_config_no_corp_identifiers` — kept clean by placing the prefixed secret name only in the exempted singlebox overlay.

Not run locally: singlebox coverage and the BCS gates, which need a product stack this environment cannot start. Two `test_bot_build_service_skill_artifact.py` cases fail here only because `rsync` is absent from this container; they pass in CI.

## Compatibility and risk

Additive, confined to the skill_center module plus its DI and HTTP wiring, with one fix to a shared config provider (`secret_names()`, whose observable behavior changes only in that a previously-dead field is now read). No schema change, no wire change to any existing endpoint, no repository change. The endpoint is unreachable without the Bearer token, and an unresolvable secret fails closed to 401. Rollback is reverting the commits.

**Tenant reach.** `AvernetTenantMiddleware` resolves a tenant only for `/openapi/v1/*`; every other path, the internal API included, runs under `DEFAULT_AVERNET_TENANT`. The Bot lookup is therefore scoped to the default tenant, so a pair belonging to a registered external tenant will 404 here. Documented on the Protocol.

## Related issues

Relates to inclusionAI/Avernet#1621. This does not fix it: the flush still runs four times per SkillSet mutation. It is step 1 of retiring the read-side flush entirely — the endpoint discharges the never-backfilled debt, but the four fan-out sources under **Problem** must be closed first, tracked in inclusionAI/Avernet#1695. Once they are, the flush can be deleted from `BotCapabilityStateReader` and #1621 dissolves rather than gets fixed.

### DIFF
--- a/src/backend/src/agentclaw/community/configs/application-singlebox.yaml
+++ b/src/backend/src/agentclaw/community/configs/application-singlebox.yaml
@@ -282,3 +282,4 @@ user_config:
   # is a deliberate change, not a config line.
   secret_names:
     aiworkbench_repo_url: "other_manual_agentclaw_aiworkbench_repo_url"
+    skill_center_internal_token: "other_manual_agentclaw_skill_center_internal_token"
--- a/src/backend/src/agentclaw/community/core/repository/capability_desired_state_types.py
+++ b/src/backend/src/agentclaw/community/core/repository/capability_desired_state_types.py
@@ -70,6 +70,13 @@ class InstallationFlushPlan:
     same Sets' MCP members. ``member_skill_ids`` is the reachability union
     the public listing needs — Default-Set exclusions are already removed,
     and are the only thing removed.
+
+    ``changed`` says whether applying the plan actually wrote rows. The
+    read-only fast path leaves it ``False``; so does a write pass that finds
+    nothing left to do once it holds the lock. It exists for the caller that
+    runs the flush deliberately rather than as the front half of a read — the
+    backfill — which reports which Bots it converged and would otherwise have
+    to re-read Installation to find out.
     """
 
     member_skill_ids: frozenset[int]
@@ -77,3 +84,4 @@ class InstallationFlushPlan:
     skills_to_uninstall: frozenset[int]
     mcps_to_install: frozenset[str] = frozenset()
     mcps_to_uninstall: frozenset[str] = frozenset()
+    changed: bool = False
--- a/src/backend/src/agentclaw/community/core/repository/capability_desired_state_types.py
+++ b/src/backend/src/agentclaw/community/core/repository/capability_desired_state_types.py
@@ -70,13 +70,6 @@ class InstallationFlushPlan:
     same Sets' MCP members. ``member_skill_ids`` is the reachability union
     the public listing needs — Default-Set exclusions are already removed,
     and are the only thing removed.
-
-    ``changed`` says whether applying the plan actually wrote rows. The
-    read-only fast path leaves it ``False``; so does a write pass that finds
-    nothing left to do once it holds the lock. It exists for the caller that
-    runs the flush deliberately rather than as the front half of a read — the
-    backfill — which reports which Bots it converged and would otherwise have
-    to re-read Installation to find out.
     """
 
     member_skill_ids: frozenset[int]
@@ -84,4 +77,3 @@ class InstallationFlushPlan:
     skills_to_uninstall: frozenset[int]
     mcps_to_install: frozenset[str] = frozenset()
     mcps_to_uninstall: frozenset[str] = frozenset()
-    changed: bool = False
--- a/src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
+++ b/src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
@@ -45,6 +45,10 @@ class CapabilityDesiredStateRepositoryProtocol(Protocol):
         that are not the inherited one. It is the same tuple the activation
         guards gate on, so the retirement can only reach Sets whose members
         those guards refuse direct control of.
+
+        The returned plan's ``changed`` reports whether this call actually
+        wrote rows, for the caller that runs the flush deliberately rather
+        than as the front half of a read.
         """
         ...
     @abstractmethod
--- a/src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
+++ b/src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
@@ -45,10 +45,6 @@ class CapabilityDesiredStateRepositoryProtocol(Protocol):
         that are not the inherited one. It is the same tuple the activation
         guards gate on, so the retirement can only reach Sets whose members
         those guards refuse direct control of.
-
-        The returned plan's ``changed`` reports whether this call actually
-        wrote rows, for the caller that runs the flush deliberately rather
-        than as the front half of a read.
         """
         ...
     @abstractmethod
--- a/src/backend/src/agentclaw/community/core/skill_center/services/installation_backfill_service.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/installation_backfill_service.py
@@ -1,0 +1,126 @@
+"""Deliberate, scoped runs of the Installation flush — the backfill.
+
+The flush also runs lazily in front of every capability read
+(:class:`BotCapabilityStateReader`), which keeps Installation agreeing with
+SkillSet configuration one Bot at a time. Configuration that reaches many Bots
+at once has no per-Bot write to ride on: platform Default-Set content edited
+through the admin tooling, an ``is_active`` flipped straight on the row, a
+``center://`` membership resolving to a newly published version. Those Bots
+converge only when something happens to read them.
+
+This service runs the same flush deliberately — one Bot, or a page of them —
+so an operator can converge that fan-out instead of waiting for reads to do
+it. Like the flush, it is DB-side only: it never touches a device and never
+triggers a runtime projection, so a Bot converged here still needs a
+projection before its engine sees the change.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from typing import Any
+
+from injector import inject
+
+from agentclaw.community.core.repository.protocols.bot import BotRepository
+from agentclaw.community.core.repository.protocols.capability_desired_state import (
+    CapabilityDesiredStateRepositoryProtocol,
+)
+from agentclaw.community.core.skill_center.bot_engine_scope import (
+    bot_default_engine_types,
+    bot_engine_type,
+)
+from agentclaw.community.core.skill_center.errors import LocalSkillNotFoundError
+from agentclaw.community.core.skill_center.installation_backfill_protocol import (
+    BackfillReport,
+    BotBackfillOutcome,
+    InstallationBackfillServiceProtocol,
+)
+from agentclaw.community.log import get_logger
+
+logger = get_logger()
+
+
+class InstallationBackfillService(InstallationBackfillServiceProtocol):
+    """Runs ``flush_installations`` over a chosen scope and reports what moved."""
+
+    @inject
+    def __init__(
+        self,
+        repository: CapabilityDesiredStateRepositoryProtocol,
+        bot_repo: BotRepository,
+    ) -> None:
+        self._repository = repository
+        self._bot_repo = bot_repo
+
+    def backfill_bot(self, *, bot_id: str, owner_id: str) -> BotBackfillOutcome:
+        bot = self._bot_repo.get_by_id_and_owner(bot_id, owner_id)
+        if bot is None:
+            raise LocalSkillNotFoundError()
+        return BotBackfillOutcome(
+            bot_id=bot_id,
+            owner_id=owner_id,
+            changed=self._flush(bot=bot, bot_id=bot_id, owner_id=owner_id),
+        )
+
+    def backfill_page(
+        self,
+        *,
+        owner_id: str | None = None,
+        engine_type: str | None = None,
+        page: int = 1,
+        page_size: int = 50,
+    ) -> BackfillReport:
+        total, bots = self._bot_repo.list_by_conditions(
+            owner_id=owner_id,
+            engine=engine_type,
+            page=page,
+            page_size=page_size,
+        )
+        outcomes = tuple(self._backfill_one(bot) for bot in bots)
+        return BackfillReport(
+            total=total,
+            page=page,
+            page_size=page_size,
+            scanned=len(outcomes),
+            changed=sum(1 for outcome in outcomes if outcome.changed),
+            failed=sum(1 for outcome in outcomes if outcome.error is not None),
+            outcomes=outcomes,
+        )
+
+    def _backfill_one(self, bot: Mapping[str, Any]) -> BotBackfillOutcome:
+        """One Bot's flush, with its failure recorded rather than raised.
+
+        A page is a sweep: one Bot whose Sets are malformed, or whose row lock
+        is held, must not cost the rest of the page. The failure is reported
+        in the outcome and counted, never swallowed into a success.
+        """
+        bot_id = str(bot["bot_id"])
+        owner_id = str(bot["owner_id"])
+        try:
+            changed = self._flush(bot=bot, bot_id=bot_id, owner_id=owner_id)
+        except Exception as exc:  # noqa: BLE001 — reported per Bot, see docstring
+            logger.exception(
+                "[installation_backfill] flush failed bot_id=%s owner_id=%s",
+                bot_id,
+                owner_id,
+            )
+            return BotBackfillOutcome(
+                bot_id=bot_id, owner_id=owner_id, changed=False, error=str(exc)
+            )
+        return BotBackfillOutcome(
+            bot_id=bot_id, owner_id=owner_id, changed=changed
+        )
+
+    def _flush(
+        self, *, bot: Mapping[str, Any], bot_id: str, owner_id: str
+    ) -> bool:
+        """The same call the reader makes, scoped by the same engine helpers."""
+        plan = self._repository.flush_installations(
+            bot_id=bot_id,
+            owner_id=owner_id,
+            env=str(bot["env"]),
+            engine_type=bot_engine_type(bot),
+            default_engine_types=bot_default_engine_types(bot),
+        )
+        return plan.changed
--- a/src/backend/src/agentclaw/community/core/skill_center/services/installation_backfill_service.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/installation_backfill_service.py
@@ -1,4 +1,4 @@
-"""Deliberate, scoped runs of the Installation flush — the backfill.
+"""Deliberate runs of the Installation flush for one Bot — the backfill.
 
 The flush also runs lazily in front of every capability read
 (:class:`BotCapabilityStateReader`), which keeps Installation agreeing with
@@ -8,17 +8,15 @@ through the admin tooling, an ``is_active`` flipped straight on the row, a
 ``center://`` membership resolving to a newly published version. Those Bots
 converge only when something happens to read them.
 
-This service runs the same flush deliberately — one Bot, or a page of them —
-so an operator can converge that fan-out instead of waiting for reads to do
-it. Like the flush, it is DB-side only: it never touches a device and never
-triggers a runtime projection, so a Bot converged here still needs a
+This service runs the same flush deliberately, for one named Bot, so a
+backfill can converge that fan-out instead of waiting for reads to do it.
+Selecting the Bots and pacing the calls is the caller's job, not this
+service's. Like the flush, it is DB-side only: it never touches a device and
+never triggers a runtime projection, so a Bot converged here still needs a
 projection before its engine sees the change.
 """
 
 from __future__ import annotations
-
-from collections.abc import Mapping
-from typing import Any
 
 from injector import inject
 
@@ -32,17 +30,12 @@ from agentclaw.community.core.skill_center.bot_engine_scope import (
 )
 from agentclaw.community.core.skill_center.errors import LocalSkillNotFoundError
 from agentclaw.community.core.skill_center.installation_backfill_protocol import (
-    BackfillReport,
-    BotBackfillOutcome,
     InstallationBackfillServiceProtocol,
 )
-from agentclaw.community.log import get_logger
-
-logger = get_logger()
 
 
 class InstallationBackfillService(InstallationBackfillServiceProtocol):
-    """Runs ``flush_installations`` over a chosen scope and reports what moved."""
+    """Runs ``flush_installations`` for one Bot, on demand."""
 
     @inject
     def __init__(
@@ -53,74 +46,15 @@ class InstallationBackfillService(InstallationBackfillServiceProtocol):
         self._repository = repository
         self._bot_repo = bot_repo
 
-    def backfill_bot(self, *, bot_id: str, owner_id: str) -> BotBackfillOutcome:
+    def backfill_bot(self, *, bot_id: str, owner_id: str) -> None:
+        """The same call the reader makes, scoped by the same engine helpers."""
         bot = self._bot_repo.get_by_id_and_owner(bot_id, owner_id)
         if bot is None:
             raise LocalSkillNotFoundError()
-        return BotBackfillOutcome(
-            bot_id=bot_id,
-            owner_id=owner_id,
-            changed=self._flush(bot=bot, bot_id=bot_id, owner_id=owner_id),
-        )
-
-    def backfill_page(
-        self,
-        *,
-        owner_id: str | None = None,
-        engine_type: str | None = None,
-        page: int = 1,
-        page_size: int = 50,
-    ) -> BackfillReport:
-        total, bots = self._bot_repo.list_by_conditions(
-            owner_id=owner_id,
-            engine=engine_type,
-            page=page,
-            page_size=page_size,
-        )
-        outcomes = tuple(self._backfill_one(bot) for bot in bots)
-        return BackfillReport(
-            total=total,
-            page=page,
-            page_size=page_size,
-            scanned=len(outcomes),
-            changed=sum(1 for outcome in outcomes if outcome.changed),
-            failed=sum(1 for outcome in outcomes if outcome.error is not None),
-            outcomes=outcomes,
-        )
-
-    def _backfill_one(self, bot: Mapping[str, Any]) -> BotBackfillOutcome:
-        """One Bot's flush, with its failure recorded rather than raised.
-
-        A page is a sweep: one Bot whose Sets are malformed, or whose row lock
-        is held, must not cost the rest of the page. The failure is reported
-        in the outcome and counted, never swallowed into a success.
-        """
-        bot_id = str(bot["bot_id"])
-        owner_id = str(bot["owner_id"])
-        try:
-            changed = self._flush(bot=bot, bot_id=bot_id, owner_id=owner_id)
-        except Exception as exc:  # noqa: BLE001 — reported per Bot, see docstring
-            logger.exception(
-                "[installation_backfill] flush failed bot_id=%s owner_id=%s",
-                bot_id,
-                owner_id,
-            )
-            return BotBackfillOutcome(
-                bot_id=bot_id, owner_id=owner_id, changed=False, error=str(exc)
-            )
-        return BotBackfillOutcome(
-            bot_id=bot_id, owner_id=owner_id, changed=changed
-        )
-
-    def _flush(
-        self, *, bot: Mapping[str, Any], bot_id: str, owner_id: str
-    ) -> bool:
-        """The same call the reader makes, scoped by the same engine helpers."""
-        plan = self._repository.flush_installations(
+        self._repository.flush_installations(
             bot_id=bot_id,
             owner_id=owner_id,
             env=str(bot["env"]),
             engine_type=bot_engine_type(bot),
             default_engine_types=bot_default_engine_types(bot),
         )
-        return plan.changed
--- a/src/backend/src/agentclaw/community/di/config.py
+++ b/src/backend/src/agentclaw/community/di/config.py
@@ -181,6 +181,7 @@ class SecretNamesConfig:
     """
 
     dormant_internal_token: str = ""
+    skill_center_internal_token: str = ""
     aiworkbench_repo_url: str = ""
     gateway_principal_signing_key: str = "gateway_principal_signing_key"
     aicoding_theta_master_key: str = ""
@@ -607,6 +608,24 @@ class DormantInternalToken:
 
     Empty ``value`` makes the auth Depends 401 all requests
     (feature-off failure mode).
+    """
+
+    value: str = ""
+
+
+@dataclass(frozen=True)
+class SkillCenterInternalToken:
+    """Resolved bearer token for ``/api/internal/skill-center/*`` endpoints.
+
+    Produced by ``SkillCenterInternalTokenBindings._resolved_internal_token``,
+    with the same
+    resolution rules and the same failure-closed empty default as
+    ``DormantInternalToken``: an empty ``value`` makes the auth Depends 401
+    every request rather than authorize an unverified caller.
+
+    Separate from the dormant token on purpose — these endpoints converge
+    capability state for whole pages of Bots, so the two operations are
+    granted independently.
     """
 
     value: str = ""
--- a/src/backend/src/agentclaw/community/di/modules/skill_center_internal_token_module.py
+++ b/src/backend/src/agentclaw/community/di/modules/skill_center_internal_token_module.py
@@ -37,7 +37,8 @@ class SkillCenterInternalTokenBindings:
         ``BotDormantModule._resolved_dormant_token`` exactly:
 
           - name is empty (community / singlebox / test)
-              → the local fallback token, so本地联调 can call the endpoints
+              → the local fallback token, so local development can call
+                the endpoints
           - the resolver returns a secret with a non-empty ``secret_value``
               → that value (prod / pre normal path)
           - the resolver returns ``None`` or an empty ``secret_value``
--- a/src/backend/src/agentclaw/community/di/modules/skill_center_internal_token_module.py
+++ b/src/backend/src/agentclaw/community/di/modules/skill_center_internal_token_module.py
@@ -1,0 +1,84 @@
+"""Composition root for the ``/api/internal/skill-center/*`` Bearer token."""
+
+from __future__ import annotations
+
+from injector import inject, provider, singleton
+
+from agentclaw.community.di import config as cfg
+from agentclaw.community.log import get_logger
+from agentclaw.community.plugin_api.secret_resolver import SecretResolver
+
+logger = get_logger()
+
+
+# Mirrors ``bot_dormant_module._SINGLEBOX_FALLBACK_TOKEN``: intentionally a
+# publicly-visible string, because it only ever gates singlebox/local, where no
+# real authority decision is at stake. Deployments that carry a secret name
+# resolve the real token and never reach it.
+_SINGLEBOX_FALLBACK_TOKEN = "singlebox-skill-center-token-local"
+
+
+class SkillCenterInternalTokenBindings:
+    """Provider mixin: resolves the internal endpoints' Bearer token."""
+
+    @singleton
+    @provider
+    @inject
+    def _resolved_internal_token(
+        self,
+        secret_resolver: SecretResolver,
+        secret_names: cfg.SecretNamesConfig,
+    ) -> cfg.SkillCenterInternalToken:
+        """Resolve the ``/api/internal/skill-center/*`` Bearer token.
+
+        The secret name comes from
+        ``SecretNamesConfig.skill_center_internal_token`` (the ``secret_names``
+        yaml block). Resolution mirrors
+        ``BotDormantModule._resolved_dormant_token`` exactly:
+
+          - name is empty (community / singlebox / test)
+              → the local fallback token, so本地联调 can call the endpoints
+          - the resolver returns a secret with a non-empty ``secret_value``
+              → that value (prod / pre normal path)
+          - the resolver returns ``None`` or an empty ``secret_value``
+              → the local fallback token
+          - the resolver raises (transient outage / network)
+              → ``""``, which 401s every request rather than authorize an
+                unverified caller
+        """
+        secret_name = secret_names.skill_center_internal_token
+        if not secret_name:
+            logger.info(
+                "[skill_center_module] no internal token secret name "
+                "configured — local fallback token in use"
+            )
+            return cfg.SkillCenterInternalToken(value=_SINGLEBOX_FALLBACK_TOKEN)
+
+        try:
+            secret = secret_resolver.get_secret(secret_name=secret_name)
+        except Exception:
+            logger.exception(
+                "[skill_center_module] SecretResolver.get_secret failed for %r "
+                "— returning empty token (failure-closed)",
+                secret_name,
+            )
+            return cfg.SkillCenterInternalToken(value="")
+
+        if secret is None:
+            logger.info(
+                "[skill_center_module] SecretResolver returned None for %r — "
+                "singlebox/local fallback token in use",
+                secret_name,
+            )
+            return cfg.SkillCenterInternalToken(value=_SINGLEBOX_FALLBACK_TOKEN)
+
+        value = getattr(secret, "secret_value", None)
+        if not value:
+            logger.warning(
+                "[skill_center_module] secret %r resolved but secret_value "
+                "empty — falling back to singlebox token",
+                secret_name,
+            )
+            return cfg.SkillCenterInternalToken(value=_SINGLEBOX_FALLBACK_TOKEN)
+
+        return cfg.SkillCenterInternalToken(value=str(value))
--- a/src/backend/src/agentclaw/community/adapters/http/app.py
+++ b/src/backend/src/agentclaw/community/adapters/http/app.py
@@ -161,7 +161,7 @@ from agentclaw.community.adapters.http.service_bot.router_publish import router 
 from agentclaw.community.adapters.http.bot_collaborator import router as bot_collaborator_router  # noqa: E402
 from agentclaw.community.adapters.http.task import task_internal_router, task_callback_router  # noqa: E402
 # skills / skillsets / skill_scan / skill_auth 全部切换到新架构 (core/skill_center + device plugin 抽象)
-from agentclaw.community.adapters.http.skill_center import skills, skillsets, skill_scan, skill_auth, skill_category, verify, sync, batch_sync  # noqa: E402
+from agentclaw.community.adapters.http.skill_center import skills, skillsets, skill_scan, skill_auth, skill_category, verify, sync, batch_sync, installations_internal  # noqa: E402
 
 from fastapi_injector import attach_injector  # noqa: E402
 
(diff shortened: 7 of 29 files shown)


src/backend/src/agentclaw/community/di/modules/skill_center_internal_token_module.py
```diff
@@ -0,0 +1,84 @@
...
+    ) -> cfg.SkillCenterInternalToken:
+        """Resolve the ``/api/internal/skill-center/*`` Bearer token.
+
+        The secret name comes from
+        ``SecretNamesConfig.skill_center_internal_token`` (the ``secret_names``
+        yaml block). Resolution mirrors
+        ``BotDormantModule._resolved_dormant_token`` exactly:
+
+          - name is empty (community / singlebox / test)
+              → the local fallback token, so本地联调 can call the endpoints
```
totalfrank: nit: stray non-English text leaked into this docstring — `so本地联调 can call the endpoints` (missing space, unintended Chinese characters in an otherwise English-only file). Likely meant something like `so local dev can call the endpoints`.

---

totalfrank (commented):
Reviewed commit `678cdfaf` (the only commit on this PR) against merge-base `6679928`, all 21 changed files.

No correctness, contract, error-handling, concurrency, security, or architecture-boundary issues found. Checked in particular:
- `InstallationFlushPlan.changed` semantics and the repository's re-resolve-under-lock path (`bot_skillset_installations.py`) — the fast-path/locked-path split and the `changed` computation look correct, and are covered by `test_repair_reports_whether_it_wrote`.
- `backfill_page` pagination (`has_more`, newest-first ordering via existing `list_by_conditions`) and per-Bot failure isolation in `_backfill_one` — failures are recorded and counted, never swallowed.
- Bearer-token auth (`internal_auth.py`, `skill_center_internal_token_module.py`) mirrors the existing `/api/internal/dormant/*` pattern exactly, including the fail-closed-to-401 behavior on a raising `SecretResolver` and the singlebox fallback token.
- DI wiring (`skill_center_module.py`, `di/config.py`) and the Service API re-export (`core -> api` boundary) follow existing conventions; `test_service_api_conformance.py` updated accordingly.
- Test coverage is thorough: service-level (fast/locked path, page filters, failure isolation, engine precedence), endpoint-level (auth gate, 404 on unknown bot, validation), and token-resolution-level (all branches of the SecretResolver fallback chain).

One non-blocking nit posted inline (stray mixed-language text in a docstring).

---

totalfrank (commented):
## Review status

No prior Claude review existed on this PR (zero reviews, zero review threads, zero issue comments), despite this run being triggered as a "new commits pushed" follow-up. Treating this as round 1: a full review of both commits, merge base `6679928d0` → head `15852c7a`.

**Commits reviewed:** `678cdfa` (feat: add internal Installation backfill endpoints), `15852c7` (docs: state the backfill's tenant reach — no code change).

### What was checked
- `InstallationFlushPlan.changed` plumbing: fast (unlocked) path leaves it `False` by default; the locked write path recomputes it from what was actually added/removed after re-resolving under lock — correctly accounts for a flush that raced another to the same Bot and found nothing left to write. Verified against `test_repair_reports_whether_it_wrote`.
- `backfill_page`'s "newest-first, pages can drift" caveat matches `BotRepository.list_by_conditions`'s `gmt_create.desc()` ordering.
- The docs commit's tenant-reach claim (internal endpoints run under `DEFAULT_AVERNET_TENANT`, only `/openapi/v1/*` resolves a real tenant) checks out against `AvernetTenantMiddleware`.
- Auth guard (`verify_skill_center_internal_token`) mirrors `verify_dormant_internal_token` exactly, including the fail-closed-to-401 behavior on empty/unresolved token.
- DI wiring (`SkillCenterInternalTokenBindings`, `SecretNamesConfig.skill_center_internal_token`, binder registration, `app.py` router mount) is consistent and covered by `test_skill_center_internal_token.py` and the architecture conformance test.
- Page endpoint records a failing Bot's error and continues the sweep rather than aborting it; single-Bot endpoint propagates instead — matches the description and is exercised by both endpoint and service tests.
- `page_size` is bounded `[1, 200]` via the Pydantic schema; out-of-range values 422 before reaching the service (tested).
- No architecture-boundary issues found: the Protocol lives in `core/`, the concrete service inherits it directly, `api/` re-exports only.

### Findings
None. No blocking or non-blocking issues, no nits.

### Checklist
- No findings from any round — this is round 1, and it is clean.

**Open count: 0**

---

src/backend/src/agentclaw/community/di/modules/skill_center_internal_token_module.py
totalfrank: Fixed in `41c4ee1` — now reads `so local development can call the endpoints`.

You diagnosed it exactly: it was copied from `BotDormantModule._resolved_dormant_token`, whose docstring is deliberately bilingual (`so singlebox联调依然可用`), and the fragment came across mangled into an otherwise English-only file. I also scanned all six new files in this PR for stray CJK — this was the only one.

---

totalfrank (commented):
## Review status

Continuing the review from the prior rounds (merge-base `6679928d0`, commits `678cdfa` and `15852c7a` already reviewed). This round covers the third commit, `41c4ee1`, which is also the current PR head.

**Commits reviewed this round:** `41c4ee1` — regenerate config goldens for the new secret name + drop mangled bilingual docstring fragment.

### Part A — open findings

- **nit: stray non-English text in docstring** (`skill_center_internal_token_module.py`) — **Fixed** in `41c4ee1`: `so本地联调 can call the endpoints` → `so local development can call the endpoints`. Already replied and resolved in-thread.

### Part B — review of the new commit

- `skill_center_internal_token_module.py`: docstring wording fix only, matches the fix described in the commit message. No code/behavior change.
- `tests/community/config/golden/{community,singlebox,test}.json`: each gains exactly one line, `"skill_center_internal_token": ""`, alongside the existing `dormant_internal_token`/`gateway_principal_signing_key` entries — consistent with `SecretNamesConfig.skill_center_internal_token` added earlier in the PR, and nothing else moved. Matches what `test_effective_config_matches_golden` requires after a deliberate config-shape change.

No new findings.

### Checklist (all rounds)
- nit: stray non-English text in docstring — **fixed**, thread resolved.

**Open count: 0**

PR is clean as of `41c4ee1` (current head).

---

totalfrank (commented):
## Review status

Continuing the review — checking the previously-open finding, then reviewing the new commit pushed since the last round (`41c4ee1` → `08c9eea6`).

**Commit reviewed this round:** `08c9eea6` (test: cover the backfill endpoints against the coverage gate) — adds `tests/community/endpoints/test_installation_backfill_internal.py` (119 lines, the only file this commit touches).

### Part A — open findings
The one finding on record (stray non-English text in `skill_center_internal_token_module.py`) was already **Fixed** in `41c4ee1` and its thread resolved. Verified no other findings are open.

### Part B — review of the new commit
Read the four new cases against the actual route/service/schema code:
- `bot/ok` — matches `InstallationBackfillService.backfill_bot`: a seeded Bot with no SkillSets flushes to `changed=false`.
- `bot/not_found` — matches `backfill_bot` raising `LocalSkillNotFoundError` when `get_by_id_and_owner` returns `None`, which the router maps to `404 "Bot not found"`.
- `page/ok` — `BackfillPageRequest`/`BackfillPageData` field names (`page`, `page_size`, `failed`, `has_more`, `outcomes`) match `schemas.py` exactly; the `json_contains` list check only requires the seeded Bot to appear among outcomes (subset match), which is what the docstring claims.
- `page/invalid_page_size` — `page_size` is `Field(50, ge=1, le=200)`, so `page_size=0` 422s before the service runs, as asserted.

Factory calls (`make_staff_user`, `make_bot`) match their real signatures. No mocking, no assertions that outrun what the implementation actually does. No new findings.

### Checklist (all rounds)
- nit: stray non-English text in docstring (`skill_center_internal_token_module.py`) — **fixed**, thread resolved.

**Open count: 0**

PR is clean as of `08c9eea6` (current head). CI was still running Backend unit tests / Singlebox coverage at review time — worth a status check before merge.

---

src/backend/src/agentclaw/community/core/repository/capability_desired_state_types.py
```diff
@@ -70,10 +70,18 @@ class InstallationFlushPlan:
...
+    backfill — which reports which Bots it converged and would otherwise have
+    to re-read Installation to find out.
     """
 
     member_skill_ids: frozenset[int]
     skills_to_install: frozenset[int]
     skills_to_uninstall: frozenset[int]
     mcps_to_install: frozenset[str] = frozenset()
     mcps_to_uninstall: frozenset[str] = frozenset()
+    changed: bool = False
```
totalfrank-personal: How is this PR relevant to this? This PR should only build an API endpoint for us to use as a tool right?

src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
```diff
@@ -45,6 +45,10 @@ def flush_installations(
         that are not the inherited one. It is the same tuple the activation
         guards gate on, so the retirement can only reach Sets whose members
         those guards refuse direct control of.
+
+        The returned plan's ``changed`` reports whether this call actually
+        wrote rows, for the caller that runs the flush deliberately rather
+        than as the front half of a read.
```
totalfrank-personal: How is this PR relevant to this? This PR should only build an API endpoint for us to use as a tool right?

src/backend/src/agentclaw/community/core/skill_center/services/installation_backfill_service.py
```diff
@@ -0,0 +1,126 @@
...
+        bot = self._bot_repo.get_by_id_and_owner(bot_id, owner_id)
+        if bot is None:
+            raise LocalSkillNotFoundError()
+        return BotBackfillOutcome(
+            bot_id=bot_id,
+            owner_id=owner_id,
+            changed=self._flush(bot=bot, bot_id=bot_id, owner_id=owner_id),
+        )
+
+    def backfill_page(
```
totalfrank-personal: I think this is an overkill. My plan is actually for myself to figure out the owenr_id and bot_id pairs myself. And run an linux command with qps control to do the backfill.

Given this, there's a lot to delete I guess. Because we don't need the batch backfill functionality at all.

src/backend/src/agentclaw/community/di/config.py
```diff
@@ -181,6 +181,7 @@ class SecretNamesConfig:
     """
 
     dormant_internal_token: str = ""
+    skill_center_internal_token: str = ""
```
totalfrank-personal, line 184: Do we want to configure this in application.yaml?

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/skill_center/services/installation_backfill_service.py
totalfrank: Agreed, and done — `backfill_page` and everything that existed only to serve it is gone.

If you're picking the pairs yourself and pacing with a shell loop, the batch endpoint was solving a problem you don't have, and worse, solving it badly for your case: it paged by `gmt_create desc` with no QPS control, so you'd have had to fight its pagination instead of just driving it. One Bot per call composes with `xargs -P` / a `while read` loop far better.

Deleted: `backfill_page`, `BackfillReport`, `BotBackfillOutcome`, the page request/response schemas, and their tests.

`POST /api/internal/skill-center/installations/backfill/bot` now takes `{bot_id, owner_id}`, returns `{success, data:{bot_id, owner_id}}`, and 404s an unknown pair so a typo can't look like a no-op.

---

src/backend/src/agentclaw/community/core/repository/capability_desired_state_types.py
totalfrank: Fair challenge — you're right, and it's reverted.

The honest answer: I added `InstallationFlushPlan.changed` so a **page** run could report which Bots it actually converged without re-reading Installation for each one. That was the only real justification, and with the batch functionality deleted it evaporates. Reporting `changed` for a single call never warranted touching a shared contract type.

`capability_desired_state_types.py`, `protocols/capability_desired_state.py`, and `bot_skillset_installations.py` are now byte-identical to the merge base — `git diff 6679928 -- core/repository/` is empty. The PR touches no repository code at all now.

---

src/backend/src/agentclaw/community/core/repository/protocols/capability_desired_state.py
totalfrank: Same answer as the sibling thread on `capability_desired_state_types.py` — reverted.

This docstring paragraph documented `InstallationFlushPlan.changed`, which only existed to let a batch run report per-Bot results. Batch is gone, so the field and this paragraph are gone with it. This file is byte-identical to the merge base again.

---

src/backend/src/agentclaw/community/di/config.py
totalfrank: It already is — this field *is* the yaml knob, and I've now documented it.

`SecretNamesConfig` is loaded from the `secret_names` block, so a deployment sets it exactly like the dormant token:

```yaml
secret_names:
  skill_center_internal_token: "skill_center_internal_token"
```

What's deliberately *not* in yaml is the token value itself — only the secret's **name** lives here, and `SecretResolver` fetches the value (Mist in prod/pre, `AGENTCLAW_SECRET_SKILL_CENTER_INTERNAL_TOKEN_VALUE` in community). Putting the value in yaml would leak it into the repo, which is why `DormantInternalToken` is built the same way.

The one gap was discoverability: `application-community.yaml` carries a worked `secret_names` example listing `dormant_internal_token` and `aiworkbench_repo_url`, and mine wasn't in it. Added, so someone provisioning secrets sees all three.

Behavior with no name configured is unchanged from the dormant module: local fallback token in community/singlebox, and failure-closed to 401 if the resolver raises.

---

totalfrank (commented):
## Review status

Continuing the review. My last round covered commit `08c9eea6` (clean, 0 open). Since then a human reviewer (totalfrank-personal) left 4 comments (not mine — no action item for me there beyond context) and the author pushed one new commit responding to that feedback.

**Commit reviewed this round:** `51a2d46` (refactor: reduce the backfill to a single-Bot endpoint) — deletes `backfill_page` and everything that existed only to serve it (`BackfillReport`, `BotBackfillOutcome`, page schemas/tests), reverts `InstallationFlushPlan.changed` and the three `core/repository/` files to merge-base, and documents the new secret name in `application-community.yaml`.

### Part A — my open findings

Only one finding was ever filed by me on this PR: the stray-CJK docstring nit (`skill_center_internal_token_module.py`). It was fixed in `41c4ee1` and the thread is already resolved — nothing further to do.

### Part B — review of the new commit

Verified the revert claim in the commit message ("byte-identical to the merge base") by checking `core/repository/capability_desired_state_types.py`, `core/repository/protocols/capability_desired_state.py`, and `core/repository/implementations/skill_center/bot_skillset_installations.py` — the diff removes exactly the `changed` field/docstring/tracking logic added earlier in the PR, nothing else.

Checked for orphans from the deletion:
- No remaining references anywhere in the tree to `BackfillReport`, `BotBackfillOutcome`, `backfill_page`, or the page schemas (`git grep` across the branch is empty).
- No other caller reads `InstallationFlushPlan.changed` — `bot_capability_state_reader.py` (the lazy-flush caller) uses the plan but never touches `.changed`; the `.changed` hits elsewhere in the tree belong to unrelated result types (`SkillSetMembershipResult`, mutation-flow results).
- `api/installation_backfill_service.py`'s `__all__`/imports, the endpoint module's imports, and both test files were all updated in the same commit — no stale imports of the deleted symbols.
- Endpoint, service, and protocol docstrings were all rewritten to describe the single-Bot contract, not left describing batch behavior.

No new findings — this is a clean, complete revert with no loose ends.

### Checklist (all rounds)
- nit: stray non-English text in docstring (`skill_center_internal_token_module.py`) — fixed, thread resolved.
- Human-reviewer threads (repository-relevance question, batch-endpoint-is-overkill, config.py question) — not mine to resolve; author has replied to each and the underlying code changed accordingly.

**Open count (mine): 0**

PR is clean as of `51a2d46` (current head).

---

src/backend/src/agentclaw/community/di/config.py
totalfrank-personal: It can not be named like that. It has to start with other_manual_agentclaw_.

Fix it.

totalfrank: Done in `0b29f1d` — registered as `other_manual_agentclaw_skill_center_internal_token`.

One thing worth flagging, because it constrains *where* the prefixed name can go: it can't live in `di/config.py` or `application-community.yaml`. Both are scanned by `test_shipped_config_no_corp_identifiers`, whose pattern

```
[a-z][a-z0-9]*_manual_[a-z0-9_]+     # "secret-registry key reference"
```

rejects exactly that prefix, and `SecretNamesConfig`'s own docstring states the rule it enforces:

> Neutral defaults are empty ... This keeps the shipped source free of `*_manual_*` secret references while letting the name legitimately differ per deployment.

So it went in `application-singlebox.yaml`, right next to `aiworkbench_repo_url: "other_manual_agentclaw_aiworkbench_repo_url"` — that file is deliberately **not** scanned by the gate (the test says so explicitly), which is why the sibling key already lives there:

```yaml
secret_names:
  aiworkbench_repo_url: "other_manual_agentclaw_aiworkbench_repo_url"
  skill_center_internal_token: "other_manual_agentclaw_skill_center_internal_token"
```

Also regenerated the singlebox config golden, which picks the name up — one added line, nothing else moved.

I left the neutral `skill_center_internal_token` in the `application-community.yaml` commented example. That file *is* gate-scanned, and community deployments resolve bare names through their own env prefix, so the neutral spelling is the correct one there — same as `dormant_internal_token` beside it. Say the word if you'd rather I drop that example entirely to avoid the two spellings sitting in the repo.

Verified: 516 passed across architecture (incl. the corp-identifier gate), config, and DI; SAST clean.

---

src/backend/src/agentclaw/community/configs/application-singlebox.yaml
```diff
@@ -281,3 +281,4 @@ user_config:
   # is a deliberate change, not a config line.
   secret_names:
     aiworkbench_repo_url: "other_manual_agentclaw_aiworkbench_repo_url"
+    skill_center_internal_token: "other_manual_agentclaw_skill_center_internal_token"
```
totalfrank, line 285: **Blocking: this yaml value is dead — `ConfigModule.secret_names()` never reads `skill_center_internal_token` from the block, so it has zero effect anywhere, prod included.**

The real location is `src/backend/src/agentclaw/community/di/modules/config_module.py`, `ConfigModule.secret_names()` (~line 402-420):

```python
def secret_names(self) -> cfg.SecretNamesConfig:
    block = _block("secret_names")
    defaults = cfg.SecretNamesConfig()
    return cfg.SecretNamesConfig(
        dormant_internal_token=block.get("dormant_internal_token", defaults.dormant_internal_token),
        aiworkbench_repo_url=block.get("aiworkbench_repo_url", defaults.aiworkbench_repo_url),
        gateway_principal_signing_key=block.get("gateway_principal_signing_key", defaults.gateway_principal_signing_key),
        aicoding_theta_master_key=block.get("aicoding_theta_master_key", defaults.aicoding_theta_master_key),
    )
```

`skill_center_internal_token` was added to `SecretNamesConfig` (`di/config.py:184`) by this PR, but this provider — the *only* place that builds the injected `SecretNamesConfig` (verified: no other `secret_names()`/`SecretNamesConfig(` construction site exists outside tests) — was never updated to read it from the yaml block. It always falls back to the dataclass default `""`.

Trace the consequence: `SkillCenterInternalTokenBindings._resolved_internal_token` (`skill_center_internal_token_module.py:50`) does `secret_name = secret_names.skill_center_internal_token`, sees it empty, and unconditionally returns the hardcoded `_SINGLEBOX_FALLBACK_TOKEN = "singlebox-skill-center-token-local"` — **in every profile, not just singlebox**, because the injected config can never carry a non-empty value regardless of what any yaml overlay (this singlebox line, or a real corp `secret_names.skill_center_internal_token` in pre/prod) sets. The PR description's operator instruction — "set `skill_center_internal_token` in the `secret_names` block for pre/prod" — currently does nothing; a prod deployment would still authorize any caller presenting the publicly-known fallback string.

This is why it slipped through: `test_skill_center_internal_token.py` constructs `SecretNamesConfig(skill_center_internal_token=secret_name)` by hand and calls the provider directly, bypassing `ConfigModule.secret_names()` entirely (`test_skill_center_internal_token.py:37-40`), and the config-golden test's `typed.secret_names.skill_center_internal_token` is `""` both before and after this commit, so the drift is invisible to it (verified: `tests/community/config/golden/singlebox.json` — `raw_user_config.secret_names.skill_center_internal_token` picked up the new value, but `typed.secret_names.skill_center_internal_token` stayed `""`).

Fix: add the missing line to the constructor call above —
```python
        skill_center_internal_token=block.get(
            "skill_center_internal_token", defaults.skill_center_internal_token
        ),
```
— and add a case (in `test_config_module.py` or similar) asserting the yaml value actually flows through `ConfigModule().secret_names()`, since none currently does.

---

totalfrank (commented):
## Review status

Continuing the review. My last round covered commit `51a2d46` (clean, 0 open). Since then the human reviewer (totalfrank-personal) left one more comment and the author pushed one new commit.

**Commit reviewed this round:** `0b29f1d` (fix: register the singlebox secret name with the required prefix) — 2 files, 4 lines: adds `skill_center_internal_token: "other_manual_agentclaw_skill_center_internal_token"` to `application-singlebox.yaml`'s `secret_names` block, and regenerates `tests/community/config/golden/singlebox.json` for it.

### Part A — open findings
- nit: stray non-English text in docstring (`skill_center_internal_token_module.py`) — fixed in `41c4ee1`, thread resolved. No change.
- Human-reviewer thread on `di/config.py:184` (secret-registry prefix) — not mine; author addressed it in this commit. Thread is still open (unresolved) but that's the human reviewer's / author's call, not something I resolve.
- All other human-reviewer threads (repository-relevance ×2, batch-endpoint-is-overkill) — already resolved, no action.

### Part B — review of the new commit
Verified the commit's own claim: `application-singlebox.yaml` is explicitly exempted from `test_shipped_yaml_has_no_corp_identifiers`'s scan (see the `NOTE` in `test_shipped_config_no_corp_identifiers.py`), so the `other_manual_agentclaw_*` prefix is correctly placed there and not in `di/config.py`/`application-community.yaml` (which are scanned and would reject it). `config.py:184`'s default stays neutral (`""`), and the `application-community.yaml` commented example is untouched. Golden diff is exactly the one added key, nothing else moved.

However, tracing where that yaml value actually goes surfaced a real bug (filed inline, **Blocking**): `ConfigModule.secret_names()` in `config_module.py` never reads `skill_center_internal_token` from the `secret_names` yaml block — it's the only field of `SecretNamesConfig` omitted from that provider's constructor call. The value this commit just added (and the one operators are told in the PR description to set for pre/prod) is never read; the injected config always carries `skill_center_internal_token=""`, so the internal endpoint's auth guard always falls back to the hardcoded, publicly-visible singlebox token — in every environment, not just singlebox. This predates this commit (present since `678cdfa`, the PR's first commit) but was never caught by any prior round because the unit test for the token provider (`test_skill_center_internal_token.py`) constructs `SecretNamesConfig` by hand rather than going through `ConfigModule`, and the config-golden test's `typed` value for this field is `""` both before and after (self-consistently wrong), so neither test surfaces the drift. See the inline comment for the trace and suggested fix.

### Checklist (all rounds)
- nit: stray non-English text in docstring (`skill_center_internal_token_module.py`) — fixed, thread resolved.
- Human-reviewer threads (repository-relevance ×2, batch-endpoint overkill) — resolved, not mine.
- Human-reviewer thread on `di/config.py:184` (secret prefix) — addressed by author in `0b29f1d`; still open, author's/reviewer's call.
- **New, Blocking:** `ConfigModule.secret_names()` never wires `skill_center_internal_token` from yaml — the internal endpoint's Bearer token silently always falls back to the hardcoded singlebox token in every environment. See inline comment on `application-singlebox.yaml:284`.

**Open count: 1** (the new Blocking finding above)

---

src/backend/src/agentclaw/community/configs/application-singlebox.yaml
totalfrank-personal, line 285: Why here? I think we should add it to application.yaml?

totalfrank-personal (commented):
Just one comment about where we should put the secret key.

src/backend/src/agentclaw/community/configs/application-singlebox.yaml
totalfrank: Confirmed and fixed in `b670468` — this was a real security bug, and the trace is exactly right.

I verified it end-to-end before fixing. The tell was sitting in the golden diff of the previous commit and I missed it: only `raw_user_config.secret_names` gained the key while `typed.secret_names.skill_center_internal_token` stayed `""`. If the field were wired, **both** would have moved. That asymmetry was the bug, visible in a diff I had already read.

**Fix, and why it isn't the one-liner you suggested.** Adding the missing `block.get(...)` line works, but it leaves the mechanism that caused this intact — a hand-written constructor call where forgetting a field is silent and permanent. Every field of `SecretNamesConfig` is a plain string whose yaml key is its own field name, so the explicit form buys nothing. It's now reflective:

```python
return cfg.SecretNamesConfig(
    **{
        f.name: block.get(f.name, getattr(defaults, f.name))
        for f in fields(cfg.SecretNamesConfig)
    }
)
```

A future field cannot be forgotten. (This also kept the file under the 1000-line cap — the 4-line explicit addition pushed `config_module.py` to 1004 and tripped `test_no_oversized_modules`; the reflective form lands at 998.)

**Test.** New `tests/community/di/test_secret_names_config_provider.py`, parametrized reflectively over `dataclasses.fields(SecretNamesConfig)` so it covers every current and future field, not just mine. Nothing exercised this provider before, which is the reason the gap survived several rounds.

I checked it actually catches the bug rather than just passing: with the fix reverted, 2 of its cases fail —

```
FAILED test_every_field_is_read_from_the_yaml_block[skill_center_internal_token]
FAILED test_the_skill_center_token_name_reaches_the_injected_config
```

Goldens regenerated: `typed.secret_names.skill_center_internal_token` now carries the singlebox name instead of `""` — the wiring proven through the real config path.

Verified after rebasing onto `4f21dc7`: 3085 passed, 30 skipped across architecture, config, di, skill_center, repository and endpoints; SAST clean.

---

totalfrank: I can move it, but the prefixed spelling specifically can't go in `application.yaml` — so I'd rather check which you want than guess.

`application.yaml`'s `user_config` **is** scanned by `test_shipped_config_no_corp_identifiers`:

```python
base = yaml.safe_load((_CONFIGS / "application.yaml").read_text(...))
base_uc = json.dumps(base.get("user_config", {}), ...)
offenders.extend(_scan_text(base_uc, "configs/application.yaml[user_config]"))
```

and `other_manual_agentclaw_…` matches its `[a-z][a-z0-9]*_manual_[a-z0-9_]+` pattern. `application-singlebox.yaml` is the one config file the test explicitly exempts, which is why `aiworkbench_repo_url`'s prefixed name already lives there and nowhere else. `application.yaml` also has no `secret_names` block at all today.

So the options, as I see them:

1. **Leave it in singlebox** (current). Consistent with `aiworkbench_repo_url`; the base stays corp-free. Prod/pre get the name from their corp env overlay, which is outside this repo — matching `SecretNamesConfig`'s docstring: *"Each corp env overlay sets the real Mist registry name."*
2. **Add a `secret_names` block to `application.yaml` with the neutral name** (`skill_center_internal_token: "skill_center_internal_token"`). Passes the gate, makes the knob visible in the base — but it's then a *different* name from the provisioned one, which may be more confusing than helpful.
3. **Put the prefixed name in `application.yaml`.** Fails the gate. It'd need an allowlist entry in `test_shipped_config_no_corp_identifiers`, whose `_ALLOWLIST` is deliberately empty with "every entry needs justification". I'd want your explicit call before touching that — it's the guard keeping corp identifiers out of the OSS build.

My read is (1), since it matches how the sibling secret is already handled and keeps the shipped base clean. But if the real goal is "prod must pick this up", note that none of these three do that on their own — prod reads its corp overlay, and what actually made the name reach the code was the `ConfigModule` wiring bug fixed in `b670468` (it was dead everywhere until then).

Tell me which and I'll make the change.

---

totalfrank: Fixed in `b670468` — `ConfigModule.secret_names()` now builds the dataclass reflectively over every `SecretNamesConfig` field, so `skill_center_internal_token` (and any future field) is read from the `secret_names` yaml block instead of silently staying at its default. Verified: `typed.secret_names.skill_center_internal_token` in `tests/community/config/golden/singlebox.json` now carries the real name instead of `""`, and the new `tests/community/di/test_secret_names_config_provider.py` sweeps every field through the provider (13/13 pass locally). Ran the full suite from the PR description plus these new tests — 3085 passed, 30 skipped.

---

totalfrank (commented):
## Review status

Note on history: the commit SHAs from every prior round (`678cdfa`, `15852c7a`, `41c4ee1`, `08c9eea6`, `51a2d46`, `0b29f1d`) are no longer in this branch's history — it was rewritten at some point (rebase/amend), leaving only the current 7 commits (`7069e7d`..`b670468`). GitHub's review threads track by diff location rather than commit SHA, so all prior threads still resolved correctly against the current head; I cross-checked the full merge-base (`4f21dc7b`) → head (`b670468d`) diff against my recorded findings to make sure nothing recurred silently.

**Commit reviewed this round:** `b670468` (fix: wire `skill_center_internal_token` through `ConfigModule`) — the fix for the one open Blocking finding from last round.

### Part A — the open finding

- **Blocking: `ConfigModule.secret_names()` never read `skill_center_internal_token` from yaml** (`application-singlebox.yaml:285`, filed last round) — **Fixed** in `b670468`. The provider now builds `SecretNamesConfig` reflectively over `dataclasses.fields(...)`, so every field (including this one) is read from the `secret_names` block by construction rather than by a hand-maintained list that can silently omit one. Verified: `typed.secret_names.skill_center_internal_token` in `tests/community/config/golden/singlebox.json` changed from `""` to the real name; the new `tests/community/di/test_secret_names_config_provider.py` sweeps every `SecretNamesConfig` field through the provider (13/13 pass); ran the PR's full validation scope plus this new file — 3085 passed, 30 skipped. Replied and resolved in-thread.

### Part B — review of the new commit

Read `config_module.py`'s reflective rewrite against the dataclass it builds (`SecretNamesConfig` in `di/config.py`): all five fields are plain strings, including `gateway_principal_signing_key`'s non-empty default, and `getattr(defaults, f.name)` picks that up correctly — no field needs special-casing today, and the docstring flags that a future field that does would need to be lifted out. No caller of `ConfigModule.secret_names()` depends on constructor-argument order. No new findings.

### Checklist (all rounds)
- nit: stray non-English text in docstring (`skill_center_internal_token_module.py`) — fixed, thread resolved.
- Human-reviewer threads (repository-relevance ×2, batch-endpoint overkill, secret-prefix ×1) — resolved, not mine.
- Human-reviewer thread on `di/config.py:184` ("Do we want to configure this in application.yaml?") — still open; the author's/reviewer's call, not mine to resolve.
- Human-reviewer thread on `application-singlebox.yaml:285` ("Why here? I think we should add it to application.yaml?") — still open; same, not mine.
- **Blocking: `ConfigModule.secret_names()` drops `skill_center_internal_token`** — fixed in `b670468`, thread resolved.

**Open count (mine): 0**

PR is clean as of `b670468` (current head).

---

totalfrank:
## `Backend unit tests` red on `b670468` — not this PR's, root-caused

It is not a test failure. The summary line is **`93 warnings, 1 error`** with zero failures — pytest aborted during *collection*:

```
Different tests were collected between gw2 and gw1
To see why this happens see 'Known limitations' in documentation for pytest-xdist
```

Every differing item is in `tests/community/core/skill_center/test_skill_version_materializer.py`, which arrived on `dev` with inclusionAI/Avernet#1702 and which this PR does not touch.

### Root cause

`_package()` (`test_skill_version_materializer.py:35`) builds a zip in memory and returns its bytes:

```python
with zipfile.ZipFile(stream, "w") as archive:
    archive.writestr("SKILL.md", ...)
```

`writestr` without an explicit `ZipInfo` stamps **the current local time** into each entry's DOS date/time field. Those bytes are then passed straight to `@pytest.mark.parametrize`, so pytest derives the **test ID** from them. Each xdist worker imports the module independently, so any two workers that land either side of a DOS-timestamp tick (2-second granularity) generate different IDs for the same tests, and xdist aborts the run.

You can see it in the log — the same test, two IDs differing only in the timestamp field:

```
-...\x00\rn\x1e]1\xd8\x80\xcak...   # gw1
+...\x00\x0en\x1e]1\xd8\x80\xcak... # gw2
```

Reproduced in isolation:

```python
a = _package(); time.sleep(2.2); b = _package()
a == b                                  # False
[i for i, (x, y) in enumerate(zip(a, b)) if x != y]   # [10, 51]
```

Offsets 10 and 51 are the DOS time fields in the local header and central directory — exactly the bytes that differ in CI.

This is timing-dependent, which is why it is intermittent: `#1702`'s own CI passed, and so did this PR's previous head. I hit the identical error locally right after rebasing onto `#1702`, and a re-run passed (**3085 passed, 30 skipped**).

### Proposed patch (not applied here — it is not this PR's file)

Pin the timestamp so the bytes are deterministic:

```python
def _package(*, name: str = "weather") -> bytes:
    stream = io.BytesIO()
    with zipfile.ZipFile(stream, "w") as archive:
        info = zipfile.ZipInfo("SKILL.md", date_time=(1980, 1, 1, 0, 0, 0))
        archive.writestr(info, ...)
```

Or keep `writestr` and give the parametrize cases explicit `ids=[...]` so the IDs no longer derive from the bytes.

I have not applied it: it is a `#1702` test file, unrelated to this change, and fixing it here would widen the PR. Filed separately so it gets fixed for everyone — it will hit any PR whose backend suite happens to straddle a tick.

Re-running the failed job once, per the usual one-re-run allowance. If it comes back red with an actual test failure rather than this collection error, that is a different problem and I will treat it as mine.

---

FreddieSun (dismissed):
lgtm

totalfrank (commented):
## Review status

**History note:** the commit last reviewed (`b670468`, "wire `skill_center_internal_token` through `ConfigModule`") is no longer reachable — the branch was rewritten again since that round, now replaying as `f4d518f`..`a1ccb2b` on top of a newer `dev` (merge-base `a8c955bc`, after `#1701`/`#1709` landed).

Diffed `a8c955bc`..`a1ccb2b` against the recorded findings rather than re-reviewing from scratch. This is a no-op rebase, not new content:
- Diffstat matches the PR's reported totals exactly: 25 files, +945/-20.
- Directly compared the files touched by the last two rounds' fixes against what was verified then — `config_module.py`'s `secret_names()` is still the reflective build over `dataclasses.fields(SecretNamesConfig)`; `application-singlebox.yaml:285` still carries `skill_center_internal_token: "other_manual_agentclaw_skill_center_internal_token"`; `installations_internal.py`/`schemas.py` are still the single-Bot-only shape (no `backfill/page`, no `BackfillReport`/`BotBackfillOutcome`). All byte-identical to the state reviewed at `b670468`.
- GitHub carried all prior review threads over correctly (tracked by diff location, not commit SHA) — same as the previous rebase noted in the `b670468` round.

No new commits' worth of content to review under Part B.

### Part A — open findings
Nothing of mine is open. Re-verified against current code:
- nit: stray non-English text in docstring (`skill_center_internal_token_module.py`) — fixed, thread resolved. Unchanged.
- **Blocking: `ConfigModule.secret_names()` dropped `skill_center_internal_token`** — fixed (now carried on `a1ccb2b`, formerly `b670468`), thread resolved. Re-confirmed the reflective fix is intact at the current head.

Human-reviewer threads (not mine, no action from me):
- Repository-relevance ×2 (`capability_desired_state_types.py`, `protocols/capability_desired_state.py`) — resolved.
- Batch-endpoint-is-overkill (`installation_backfill_service.py`) — resolved.
- Secret-registry prefix / "why here?" (`application-singlebox.yaml:285`) — resolved.
- "Do we want to configure this in `application.yaml`?" (`di/config.py:184`) — still open; author gave a recommendation with three options, reviewer hasn't responded yet. Their call, not mine.

### Checklist (all rounds)
- nit: stray CJK in docstring — fixed, resolved.
- Human-reviewer repository-relevance threads ×2 — resolved, not mine.
- Human-reviewer batch-endpoint-overkill thread — resolved, not mine.
- Human-reviewer secret-prefix thread (`application-singlebox.yaml:285`) — resolved, not mine.
- Human-reviewer `application.yaml` placement question (`di/config.py:184`) — open, not mine.
- Blocking: `ConfigModule.secret_names()` dropped `skill_center_internal_token` — fixed, resolved.

**Open count (mine): 0**

PR is clean as of `a1ccb2b` — same content as the `b670468` round that cleared it, now on a rebased base.

---

totalfrank:
## `Singlebox coverage` red on `a1ccb2b8` — BCS E2E gate, not this PR

Different failure from the `Backend unit tests` one above, so noting it separately.

```
##[error]Adapter endpoint coverage 99.3% (150 / 151) is below the required 100.0%
```

### It is the BCS gate

From the same report block:

```
router source  : crates/adapters/http/bcs-http/src/router.rs
parsed routes  : 151 endpoints (123 HTTP paths, 5 methods)
Endpoint coverage: 150 / 151  (99.3%)
  GET     61 / 62  ( 98.4%)      # every other method 100%
```

The denominator is parsed from the **BCS Rust router**, and one `GET` out of 62 went unexercised by the E2E stories in this run. This PR contains no Rust and no BCS code at all — it is Python under `src/backend/`, plus yaml and tests.

### Ruled out, in order

1. **Not this PR's code.** No BCS/Rust in the diff; the gate's denominator comes from a crate this branch does not touch.
2. **Not the base.** `dev` @ `a8c955bc` — the exact merge-base of this head — is green on `Singlebox Coverage` (run at 15:56Z), as are `1a2c1b1d` and `4f21dc7b` before it.
3. **Not the content of this branch.** This same check passed on the previous head `b670468` at 14:04Z with a byte-identical backend diff; the only change since is the rebase onto `a8c955bc`, which the automated reviewer independently confirmed is content-preserving (diffstat matches exactly, key files byte-identical).

That leaves a flaky BCS E2E story that happened not to hit one `GET` route this run. `BCS e2e (coverage gated)` — the standalone check — passed on this very head.

No fix exists to port: it is a coverage-completeness gate over someone else's E2E suite, not a defect with a patch. Re-running the failed job once, which is the confirmation step for exactly this case. If it fails again on the same endpoint I will stop re-running and raise it rather than burn cycles.

---

FreddieSun (approved):
LGTM

PR #1699, opened 2026-08-30
feat(gateway): migrate secbaas API keys onto the gateway, keys and grants

totalfrank:
## Problem

secbaas's OpenAPI surface is moving to the gateway, and its callers authenticate with API keys the gateway cannot obtain. `baas_api_key` stores only a salted one-way hash, and there is no mapping from a person to their keys that an operator could drive a bulk migration from. A batch job would have to either mint new credentials for everyone — the rotation event this workstream exists to avoid — or copy rows on behalf of people whose consent it cannot check.

Possession of the key resolves both problems: it proves entitlement to the row, and it is the only proof anyone has. That is why this is an endpoint the key holder calls rather than a script an operator runs.

## Solution

`POST /admin/apps/migrate-from-baas` (beside `/admin/apps`) takes the caller's plaintext secbaas key plus an app name of their choosing. It copies the ACTIVE `baas_api_key` row into `avernet_application` and flattens the bot authorizations that key carried into `ac_bot_app_grant`.

**The key keeps working.** The stored hash is copied rather than re-derived, and `core/app/_key_gen.py` is already a byte-identical copy of secbaas's generator (pinned by an existing parity test), so both registries verify the same plaintext. Nothing is minted and no `api_key` field appears in the response.

**Grants** come from the two shapes secbaas supports, both resolving to `{bot_id}:{entity_id}`: `app_type=bot` carries one reference in `app_id`; `app_type=app` carries a list in `policy.allowed_bots`. `entity_id` becomes both `user_id` and `owner_id`, because secbaas could only express "the bot's own owner authorized this" (its check was `operator == entity_id`). The policy column is read with exactly secbaas's fail-closed semantics.

**Refusals are whole-request and write nothing** — an allow-all policy has no per-bot representation, a malformed reference cannot be reconstructed, an over-long identity would produce a grant no request can resolve. Migrating the valid remainder in any of those cases leaves a credential that looks migrated while reaching fewer bots than before, discovered at some later unrelated moment. The application row, its grants and its log rows are one transaction for the same reason: two commits would leave a window in which a live credential authorizes nothing, and there is no compensating delete since nothing in the gateway deletes app rows.

| Outcome | Status |
| --- | --- |
| `key_not_found` (covers "no such key" and "wrong key" identically) | 404 |
| `already_migrated` / `app_name_taken` / `prefix_conflict` | 409 |
| `wildcard_policy` / `invalid_grant_targets` / `unsupported_app_type` / `value_too_long` | 422 |

`already_migrated` and `app_name_taken` are distinguished by **hash equality** on the row holding the prefix, not by the prefix alone — calling a genuine collision an idempotent re-run would tell a caller their key works when it resolves to someone else's application.

Decisions worth a second look, all documented at their sites:

- **The gateway writes `ac_bot_app_grant`, which the backend owns.** Taken as a scoped, time-bounded waiver: a service-to-service API built for a one-shot backfill would outlive the backfill. Bounded by excluding the mirrored models from the MariaDB plugin's `create_all` whitelist (the treatment `bcs_bots` already gets), so no deployment provisions a foreign table from them.
- **Tenant is normalized to `teamclaw`, not copied.** secbaas defaults to `team_claw` and the backend's grant tables to `teamclaw`; they are not the same namespace, and a grant the backend's guard cannot see fails silently. The source value is preserved on the row's `config`.
- **The grant log is written too.** `ac_bot_app_grant_log` is read precisely when the live row is gone; a migrated grant with no log entry has unanswerable provenance.
- **Both request fields are declared with defaults rather than as required.** FastAPI renders a "field required" error with `input` set to the whole request body — which here holds the plaintext key — so a client that forgot `app_name` got its own credential quoted back in a 422. Verified before and after.
- **Environment is copied, not enforced** (divergence from secbaas's validator, which pins `env`). The gateway's core has no injected environment, and `AppRepository` does not filter on `env` either, so filtering here would narrow who may migrate without narrowing who may authenticate.

## Validation

- `pytest tests/unit tests/integration tests/architecture tests/contracts` — **1117 passed, 4 skipped**, including all 24 architecture gates (layer bans, adapter thinness, ruff lint + format).
- New coverage: 23 policy-parsing cases (every form the `policy` column really takes), 28 DB-backed migrator cases against a real schema, 4 adapter cases pinning the outcome→status table to the enum, 11 HTTP integration cases.
- Atomicity asserted in both directions: a refused application row leaves no grants, and a refused grant leaves no application row (forced by pre-claiming the grant's unique key).
- The closed loop asserted end to end: after migrating, `AppRepository.find_app_by_credential(original_key)` resolves the new row.
- `tests/e2e/live` — 10 pre-existing failures, unchanged before and after (they need a live server); confirmed by re-running against a stashed tree.
- Not run: nothing against a real MySQL/OceanBase. Migration `003` is reviewed by hand only.

## Compatibility and risk

`avernet_application` gains `UNIQUE KEY uk_avernet_application_app_name_env (app_name, env)`, replacing the now-redundant `idx_avernet_application_app_name` (the new key's leading column serves every lookup the old index served). Shipped as `migrations/mysql/003` for deployed databases and folded back into `001_init_schema.sql` for fresh ones.

⚠️ **`003` fails with ER_DUP_ENTRY if deployed rows already hold a duplicate `(app_name, env)`.** That is deliberate — deduplication is a data question only its owner can answer — and the SQL file carries the query that finds them, including the note that `app_name` collates case-insensitively so `Billing` and `billing` collide.

Because the same key now constrains `POST /admin/apps`, a name clash there is mapped to `409` rather than surfacing as a `500`. Rollback is `DROP INDEX uk_avernet_application_app_name_env` plus re-adding the plain index; no data is rewritten either way.

Deletion criteria are in the spec: when no unmigrated `baas_api_key` rows remain, `api/baas_migration/`, `core/baas_migration/`, the endpoint and their tests all go together. The unique key and the `409` stay — they improve the registry in their own right.

## Spec

`src/gateway/specs/2026-08-30-baas-api-key-migration/spec.md`, added in this change. Builds on `specs/2026-08-13-application-api-key-credentials/spec.md`, which adopted the secbaas credential scheme specifically so this migration would be possible.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LPYqC2JRex2NWJoCpQo6Nw)_


totalfrank:
**`Gateway unit tests` failed on `e3616d2` — the cause was on `dev`, not in this PR.** Fixed in `803abb8` after merging `dev` in; flagging it because the fix touches a file unrelated to this change.

The failure:

```
FAILED tests/unit/scripts/test_dump_and_publish_script.py::test_bcn_dump_uses_the_gateway_managed_python_environment
  - assert 22 == 21
```

That assertion counts operations in **bcs's** internal OpenAPI document. #1645 (`feat(bcs): add callback recovery and failed run reruns`, merged to `dev` as `47219c8`) added `POST /api/v1/collaboration/state-machine-runs/{run_id}/reruns`, making it 22.

Why it landed red: the Gateway CI job is path-filtered on `src/gateway`, and #1645 touched **zero** gateway files — so this suite never ran on it. The stale constant sat on `dev` until the next PR to touch the gateway picked it up, which is this one. Nothing in this PR's diff goes near `src/bcs`.

No fix PR for it existed, so rather than leave the branch red I bumped the constant here. Reproduced before fixing: merged `dev` in, re-ran the dump script, and confirmed the internal document has exactly 22 operations with the extra one being the rerun endpoint (the public document is unchanged at 57). The change also asserts the new operation by name, so the next drift names what moved instead of only that a number did.

Happy to split that one-line commit out if you'd rather keep it off this PR — it reverts cleanly on its own.

Full gateway suite on the merged head: **1117 passed, 4 skipped**, `ruff check` and `ruff format --check` clean.

---

src/gateway/src/gateway/community/adapters/web/admin.py
```diff
@@ -109,3 +188,89 @@ async def register_app(payload: AppRequest, request: Request) -> JSONResponse:
...
+            {"missing": missing},
+        )
+
+    migrator = request.app.state.baas_key_migrator
+    try:
+        result = await migrator.migrate(api_key=api_key, app_name=app_name)
+    except Exception:
+        # ``payload`` is never passed to the logger: it carries the plaintext
+        # key, and ``logger.exception`` renders whatever it is given.
+        logger.exception("baas key migration failed")
```
totalfrank-personal: We need stack trace in order to debug.

src/gateway/src/gateway/community/api/baas_migration/_models.py
```diff
@@ -0,0 +1,139 @@
...
+
+    ``app`` is populated if and only if ``outcome`` is
+    :attr:`MigrationOutcome.MIGRATED`. ``detail`` carries whatever the caller
+    needs in order to act on a refusal — the taken name, the offending bot
+    references, the field that overflowed — as plain data the adapter can put
+    straight into a response body.
+    """
+
+    outcome: MigrationOutcome
+    app: MigratedApp | None = None
```
totalfrank-personal, line 128: Can you add doc for this field, under what scenario this is None.

totalfrank-personal (commented):

src/gateway/src/gateway/community/api/baas_migration/_models.py
totalfrank: Added in `55b9e6c` — the field now has its own docstring rather than relying on the class-level sentence.

Short answer: **`None` for every outcome except `MIGRATED`.** The docstring says that up front, then enumerates the refusals grouped by where each is decided — before any lookup (`value_too_long` on `app_name`), at the credential check (`key_not_found`), while deriving grants (`wildcard_policy`, `invalid_grant_targets`, `unsupported_app_type`, `value_too_long`), and at the insert when a unique key refuses it (`already_migrated`, `prefix_conflict`, `app_name_taken`).

Two things I made explicit while writing it, because they're the parts that are easy to get wrong:

- **There is no partial state.** The application row, its grants and its log rows are one transaction, so `app is None` always means nothing at all was written — the caller's secbaas key is still their only working credential.
- **A database that is down is not one of these.** That raises instead of returning a result, because it isn't something the caller can act on.

I also pointed readers at `succeeded` rather than testing this field for `None`, so the outcome stays the single source of truth instead of two facts that can drift apart.

---

src/gateway/src/gateway/community/adapters/web/admin.py
totalfrank: The stack trace was already being logged — `logger.exception(...)` is `logger.error(..., exc_info=True)`, and I verified it end to end against this codebase's logger (it resolves to a plain `logging.Logger`, so `exc_info` survives):

```
baas key migration failed
Traceback (most recent call last):
  File "<stdin>", line 7, in <module>
ValueError: boom-from-migrate
```

But you were pointing at a real gap, just one line over from where I'd have looked. The trace said *what* blew up and nothing about *which migration* — concurrent attempts produce interleaved traces that can't be told apart. Fixed in `55b9e6c`:

```python
logger.exception("baas key migration failed: app_name=%s", app_name)
```

`app_name` is the right correlation handle here: it's the caller's own value, unique per environment, and carries no credential material.

Two things I deliberately did **not** do, both noted in a comment at the call site:

- **No key or key prefix in the log.** secbaas's validator does log the prefix, but this endpoint's whole design is that key material never leaves the process, and the source row is recoverable from `app_name` anyway.
- **No trace in the response.** A fault here is a server bug; a caller who can POST a credential shouldn't be shown the internals it failed in. The 500 body stays opaque.

Added a test that pins all of it — `exc_info` present, `app_name` in the formatted record, and neither the plaintext key nor its prefix in the log or the body. Mutation-checked it: reverting to a bare `logger.error` fails it on `assert record.exc_info is not None`.

If you actually wanted the trace surfaced in the response for a debug environment, say so and I'll do it behind an explicit flag rather than by default.

---

totalfrank (commented):
Reviewed the full diff from merge-base `47219c8` to head `55b9e6e` (commits `e3616d2`, `ae769e6`, `803abb8`, `55b9e6e`), against `AGENTS.md` and `docs/arch/arch.rules.md`.

Files covered: `admin.py`, `app.py`, `app_registration/{__init__,_errors}.py`, `baas_migration/{__init__,_models}.py` (api layer), `bootstrap/{__init__,_credential_issuance}.py`, `core/app/{__init__,_orm,_registrar,_repository}.py`, `core/baas_migration/{__init__,_migrator,_orm,_policy,_records,_repository}.py`, `migrations/mysql/{001_init_schema,003_application_app_name_env_unique}.sql`, plus the new test suites and `spec.md`.

Categories that actually applied here: correctness (grant derivation, refusal-vs-migrate branching, overflow checks against destination column widths), contract breakage (outcome→status mapping, request-model disclosure handling), error/failure handling (transaction atomicity across the two-table write, `IntegrityError` classification), concurrency (session lifecycle around the classify-on-conflict re-read), architecture boundaries (Service API vs core layering in `admin.py`, the documented cross-module write and its `create_all` whitelist exclusion), and test coverage of the new behavior.

No findings. Specifically checked and found consistent:
- `_grant_targets`/`_overflowing_field` column-width checks line up with the actual destination column widths in `core/app/_orm.py` and `core/baas_migration/_orm.py` (verified against the backend's own `ac_bot_app_grant` model for parity).
- `_classify`'s re-read happens after the failed session's rollback-and-close (`orm_session()`'s `except: rollback(); raise` / `finally: close()`), so the fresh queries it issues aren't affected by the aborted transaction.
- The `create_all()` whitelist in the MariaDB plugin filters by table name regardless of import order, so the foreign `baas_migration` mirrors can't leak into a real deployment even though `bootstrap` imports that package unconditionally.
- `parse_allowed_bots`'s non-string-entry filtering (`_policy.py`) is a deliberate, tested divergence from secbaas's literal parser, not an oversight — it exists specifically to prevent a `TypeError` in `split_bot_reference` and resolves toward under-granting, consistent with the module's stated fail-closed philosophy.
- Both threads from the earlier review rounds (stack-trace correlation, `MigrationResult.app` docstring) are resolved in `55b9e6e` as claimed.

---

totalfrank:
**Review round 1 — no issues flagged.** Reviewed `55b9e6c` (the full diff from merge-base `47219c8`: `e3616d2` migration endpoint, `ae769e6` merge of `dev`, `803abb8` bcs operation-count fix, `55b9e6c` review-comment fixes).

The round-1 review returned **no findings**, having checked correctness of grant derivation and the refuse-vs-migrate branching, the outcome→status contract, transaction atomicity across the two-table write and its `IntegrityError` classification, session lifecycle around the classify-on-conflict re-read, the architecture boundaries (`admin.py` layering, the documented cross-module write and its `create_all` whitelist exclusion), and test coverage of the new behaviour.

CI on `55b9e6c`: **all 8 checks green** — Gateway, BCS, BaaS, Backend, Engine, Sandbox-proxy unit tests, BCS e2e, Singlebox coverage.

**One correction to the review's own summary**, since it concerns whose call something was. It says both earlier threads were resolved in `55b9e6c` "as claimed" — but I had deliberately left [the stack-trace thread](https://github.com/inclusionAI/Avernet/pull/1699#discussion_r3889286225) **open**, not resolved. My reply there corrects the premise (the trace was already being logged; `logger.exception` is `logger.error(..., exc_info=True)`) and offers to surface it in the response behind an explicit flag if that's what was actually wanted. It has since been resolved by a reviewer, which I read as the explanation being accepted — but I didn't resolve it myself, and it shouldn't be recorded as though I had.

No push this round, so no round 2 will open on its own. Ledger of what stands:

| Item | State |
|---|---|
| Round-1 review findings | None |
| CI on `55b9e6c` | Green (8/8) |
| Merge conflict | None (base `47219c8`) |
| Human threads | Both resolved |
| Open disputes | None |

Still awaiting the four requested human reviewers. Three judgement calls in this PR are worth a human eye specifically because they're choices rather than defects, all argued at their sites and in the PR body: the grant log being written alongside every grant (scope you didn't ask for), refusing the *whole* migration on a malformed bot reference rather than migrating the valid remainder, and `migrations/mysql/003` failing loudly on pre-existing duplicate `(app_name, env)` rows.

---

vzvince (dismissed):

totalfrank (commented):
**Review round 2 — no new findings.** Reviewed `f0835c11`, the only commit since round 1's `55b9e6e`: a merge of `dev` (which had advanced by `4f21dc7b`/`1a2c1b1d`, an unrelated Space Skill feature) resolving one conflict in `src/gateway/tests/unit/scripts/test_dump_and_publish_script.py`.

Verified the conflict resolution directly: diffed that file between `55b9e6e` and `f0835c11`. The change is comment-only — reworded to say the 21→22 bump happened independently on both sides, rather than implying this branch was the only place it was fixed. The executable assertions (`== 22`, and the named rerun-endpoint check) are byte-identical to what `803abb8` already had. Confirmed the PR's true diff (`dev...f0835c11`, three-dot) is still exactly the 26 files / 2838+/21- from the PR description — the merge pulled in no incidental changes to this PR's own files.

CI on `f0835c11`: all 8 checks green (Gateway, BCS, BaaS, Backend, Engine, Sandbox-proxy unit tests, BCS e2e, Singlebox coverage).

Ledger of what stands:

| Item | State |
|---|---|
| Stack-trace correlation thread (human, `totalfrank-personal`) | Resolved — confirmed already in round 1 |
| `MigrationResult.app` docstring thread (human, `totalfrank-personal`) | Resolved — confirmed already in round 1 |
| Round-1 review findings (bot) | None |
| Round-2 review findings (bot) | None — merge commit is comment-only |
| CI on `f0835c11` | Green (8/8) |
| Merge conflict | None (resolved in this same commit) |
| Open disputes | None |

**0 open.** Nothing outstanding from this review; the PR remains clean as of round 2 (`f0835c11`).

---

totalfrank:
**Review round 2 — no new findings.** Reviewed `f0835c11`. CI green (8/8), verified by read rather than taken from the review.

**@vzvince — your approval was dismissed, and I'm sorry for the churn.** You approved `55b9e6c` at 14:18; two minutes later I pushed `f0835c11` and this repo dismisses stale approvals on push, so review `5061041670` now reads `DISMISSED`. **Nothing in the code you read changed.** I've re-requested you, and here is exactly what the push was, so you can judge how much of a re-review it deserves:

`dev` moved (backend #1701/#1702, unrelated Space Skill work) and the PR went `dirty`. Resolving it produced **one** conflict, in `src/gateway/tests/unit/scripts/test_dump_and_publish_script.py` — and it turned out `dev` had independently made the *same* 21→22 bcs operation-count fix I'd made in `803abb8`. Every executable assertion was identical on both sides; the clash was purely over my explanatory comment, which I kept and reworded so it no longer implies this branch was the only place it got fixed.

So the delta from what you approved is **a comment in one test file, plus a base merge.** The three-dot diff is unchanged at 26 files / +2838 −21.

Fixing a merge conflict wasn't optional, and losing an approval is an accepted cost of that — but quietly absorbing it wouldn't be, hence this note rather than a silent re-request.

| Item | State |
|---|---|
| Round-1 findings | None (`55b9e6c`) |
| Round-2 findings | None (`f0835c11`) |
| CI on `f0835c11` | Green (8/8) |
| Merge conflict | Resolved in `f0835c11` |
| Human threads | Both resolved |
| Open disputes | None |
| Approvals | `vzvince` dismissed by the merge push → re-requested |

No push is pending, so no round 3 opens on its own. Still outstanding are the three judgement calls flagged from the start — they're choices rather than defects, which is exactly why two clean automated rounds don't settle them: the grant log being written alongside every grant, refusing the *whole* migration on a malformed bot reference instead of migrating the valid remainder, and `migrations/mysql/003` failing loudly on pre-existing duplicate `(app_name, env)` rows.

---

vzvince (approved):

PR #1704, opened 2026-08-30
fix(bcs): delegate persistent DB run retirement to platform, cover detached orphans

vzvince:
## Background

BCS direct-chat runs in persistent (MySQL) mode use a `DetachDeliveryAck` completion policy for async "detached" requests: the run is marked delivered (`delivery_ack_at_ms` set) on the provider's first progress event and stays in non-terminal `running` state (the caller does not wait for the final answer). The intended retirement for these runs is the `drop_detached_expired` port at `ack_at + retention` (a silent `Dropped` lifecycle), not terminalization — see spec C11.

## Problem

In persistent MySQL mode an acknowledged `DetachDeliveryAck` run could evade **every** retirement path and linger forever, counted by the active-run gauge indefinitely:

- `list_active` excludes acknowledged detached runs (C11 — a delivered run must not be `force_fail("timeout")`), so the 10s cleanup backstop never fails them.
- `SqlChatRunRepo::drop_detached_expired` / `delete_expired_terminal` were no-op **only for the MySQL flavor** (delegated to the platform, B1); the SQLite flavor self-pruned.
- The documented platform cleanup SQL (spec §11.2) pruned **only terminal rows** (`state IN ('completed','failed','cancelled')`).

So the "delegate detached retirement to platform" promise was never actually fulfilled: the platform SQL did not include a detached branch. The realistic trigger is a BCS restart (or provider-stream drop) mid-flight — the in-memory `drain_async_run` task dies, and for the default `SseFirst` transport the late `final` never arrives. The row is then orphaned `running` + `delivery_ack_at_ms` and stays forever, drifting the active-run gauge and growing the audit table without bound across restarts.

## Fix

- **Code (`crates/services/bcs-chat-run-store/src/sql.rs`)**: unify `delete_expired_terminal` and `drop_detached_expired` to a **uniform no-op across all DB flavors** — removed the per-flavor split and the SQLite `SELECT/DELETE` bodies. Persistent deletion is platform-owned; the memory implementation (default, safe-release path) is unchanged and still performs the actual retirement and emits the `Dropped`/`Expired` lifecycle.
- **Doc (spec §11.2 / C11 / §11.3)**: extend the platform cleanup SQL to also retire acknowledged detached running rows past a **short** retention:
  ```sql
  OR (state = 'running'
      AND completion_policy = 'detach_delivery_ack'
      AND delivery_ack_at_ms IS NOT NULL
      AND delivery_ack_at_ms < <now - detach_retention_ms>)
  ```
  where `<detach_retention_ms>` aligns with `async_chat_run_retention_ms` (~2 min, **not** the 30/90-day audit retention), so orphans leave the gauge promptly instead of sitting for a month (the audit signal is retained via the lifecycle counter). C11 is generalized ("persistent DB side no-op", not "MySQL side"), and §11.3's "persistent `dropped` is always empty" now holds uniformly across persistent flavors.
- **Tests (`crates/services/bcs-chat-run-store/tests/sql_repo.rs`)**: the three affected tests assert the uniform no-op (both SQLite and Mysql flavors) instead of deletion; the memory-repo tests keep covering the actual retirement path.

## Verification

- `cargo test -p bcs-chat-run-store` → **18 passed** (9 `sql_repo` + 9 `memory_repo`), no warnings.
- `cargo test -p bcs-message-flow --test run_store_engine` → **6 passed**, including `cleanup_expired_fails_overdue_drops_terminal_and_retires_detached` (memory path unchanged) and `detach_acknowledgement_guarded_by_policy`.
- `cargo build -p bcs-message-flow` → clean (depends on the updated trait; signature unchanged).

## Notes / follow-ups (not in this PR)

- The short-retention cutoff for the detached branch is a spec/ops decision; if the 30/90-day audit retention is preferred there instead, it is a one-line §11.2 change (the gauge would then overcount orphans for up to a month).
- `bcs-admin retention prune` (the B2 follow-up) is intentionally left as separate future work — it would give non-platform environments (SQLite dev, or clusters without the cron task) a manual prune path.
- An edge-case consistency gap where `drain_async_run`'s in-process timeout can `mark_failed` an already-acked detach run (vs the periodic `force_fail` excluding per C11) is left as a separate small fix.

### DIFF
--- a/src/bcs/crates/services/bcs-chat-run-store/src/sql.rs
+++ b/src/bcs/crates/services/bcs-chat-run-store/src/sql.rs
@@ -568,99 +568,31 @@ impl ChatRunRepoPort for SqlChatRunRepo {
 
     async fn drop_detached_expired(
         &self,
-        now_ms: u64,
-        retention_ms: u64,
+        _now_ms: u64,
+        _retention_ms: u64,
     ) -> Result<Vec<ChatRunRecord>, ChatRunRepoError> {
-        // MySQL delegates (terminal-row pruning is platform-managed; detached
-        // retirement is analogous — keep the auditable row). SQLite (dev/test)
-        // self-prunes here so the local table stays bounded and the path is
-        // covered by tests.
-        if self.flavor != DbSqlFlavor::Sqlite {
-            return Ok(Vec::new());
-        }
-        let cutoff = now_ms.saturating_sub(retention_ms);
-        let rows = self
-            .db
-            .query(DbStatement::with_params(
-                &format!(
-                    "SELECT {SELECT_COLS} FROM bcs_chat_runs \
-                     WHERE state = 'running' \
-                       AND completion_policy = 'detach_delivery_ack' \
-                       AND delivery_ack_at_ms IS NOT NULL \
-                       AND delivery_ack_at_ms < ? \
-                       AND env = ?"
-                ),
-                vec![DbValue::from(cutoff as i64), DbValue::from(self.env.as_str())],
-            ))
-            .await
-            .map_err(backend)?;
-        let mut dropped = Vec::new();
-        for row in rows {
-            let record = row_to_record(&row)?;
-            let _ = self
-                .db
-                .execute(DbStatement::with_params(
-                    &format!(
-                        "DELETE FROM bcs_chat_runs \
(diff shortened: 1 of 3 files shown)


questchaos (approved):
LGTM

RaymondHX (approved):

PR #1711, opened 2026-08-30
fix(message-flow): sweep idle notifiers to stop detached-run leak

vzvince:
## Summary

Fixes a node-local `Notify` memory leak in the Direct Chat async-run engine (`ChatRunStore`) that grew without bound over a process lifetime — one entry per acknowledged detached-delivery run that was ever long-polled. The fix bounds the `notifiers` map to roughly the number of concurrent long-polls, in every store mode, including MySQL production.

## Root cause

- `wait_update` inserts a `Notify` entry into the node-local `notifiers` map (`notifier()` inserts before the terminal check).
- The only removal path is `drop_notifier`, called from the terminal CAS paths (`apply_terminal_cas`, `force_fail`).
- Acknowledged detached-delivery runs never terminalize: they are excluded from `list_active` (so the timeout sweep never fails them) and retire silently as `Dropped`, **without** a terminal CAS — so `drop_notifier` is never called for them. The entry leaks until process restart.

## Why the obvious fix doesn't work in production

Iterating the `dropped` set returned by `cleanup_expired` and calling `drop_notifier` per run is a **no-op in MySQL production**: `drop_detached_expired` / `delete_expired_terminal` are deliberate no-ops there (auditable rows are retained; pruning is delegated to the platform's scheduled cleanup). Retirement happens entirely outside the Rust process, so the cleanup loop on the polling replica never learns which runs were retired. The leak on that path cannot be fixed from the repo retirement signal at all — that approach only helps in-memory/SQLite (dev/test).

## The fix

Add `sweep_idle_notifiers()`, invoked at the top of the 10s `cleanup_expired` tick (which runs unconditionally in every mode). It evicts notifier entries that have **no parked waiter**, detected as `Arc::strong_count == 1` (the map is the entry's sole owner), re-checked under the write lock. This is independent of the repo retirement path, so it reclaims idle entries in MySQL production too, where retirement is external.

The existing `drop_notifier` calls on terminal CAS are kept — they reclaim immediately on terminalization; the sweep is the backstop for everything that retires without a CAS.

## Why a non-idle notifier is never evicted

Every `Arc<Notify>` clone is taken while holding the map's `RwLock` (in `notifier()`). The sweep's write-lock re-check makes the strong count precise at the decision point: a parked waiter's reference is always visible (claiming a reference is lock-guarded), so it is always counted and always skipped. A stale-*high* read (a waiter that just dropped its clone) only delays an idle entry's reclaim by one tick — never incorrect. The `Notify` registry is a pure latency optimization (correctness comes from `wait_update` re-reading the repo), so even a hypothetical missed wake falls back to the existing timeout re-read, not a wrong result.

## Tests

Added in `run_store.rs` `#[cfg(test)]`:
- `cleanup_expired_reclaims_idle_notifier_for_retired_detached_run` — core regression (RED-first)
- `sweep_reclaims_idle_notifier_independent_of_retirement` — run stays alive in the repo yet its idle notifier is reclaimed (mode-independence)
- `sweep_keeps_notifier_for_parked_waiter` — a parked waiter survives the sweep and is woken promptly via notify
- `sweep_coexists_with_terminal_cas_drop` — idempotent no-op on already-dropped terminal entries

## Verification

- `cargo test -p bcs-message-flow`: all green — 37 lib unit tests (incl. 4 new), integration `run_store_engine` 6/6, other suites unchanged.
- `cargo build -p bcs-message-flow`: clean, no warnings.
- Diff: one file, +168 lines, no reformatting of unrelated code (per the repo's `No cargo fmt` rule).

### DIFF
--- a/src/bcs/crates/services/bcs-message-flow/src/a2a_chat/run_store.rs
+++ b/src/bcs/crates/services/bcs-message-flow/src/a2a_chat/run_store.rs
@@ -126,6 +126,47 @@ impl ChatRunStore {
         self.notifiers.write().await.remove(run_id);
     }
 
+    /// Reclaim node-local notifier entries that have no parked long-poll waiter.
+    ///
+    /// The `Notify` registry is a latency-only optimization for `wait_update`
+    /// — correctness comes from re-reading the repo, and [`Self::notifier`]
+    /// rebuilds an entry on demand — so evicting an idle entry is always safe.
+    /// This is the only reclaim path for runs that retire without a terminal
+    /// CAS (acknowledged detached-delivery runs, which are excluded from the
+    /// timeout sweep and retire as `Dropped`), since [`Self::drop_notifier`]
+    /// is only reached on a terminal transition. It is independent of the repo
+    /// retirement path, so it also covers MySQL production, where detached
+    /// retirement is delegated to platform cleanup and `drop_detached_expired`
+    /// is a no-op that returns no retired rows.
+    ///
+    /// An entry is idle iff the map is its sole owner (`Arc::strong_count == 1`),
+    /// i.e. no `wait_update` call currently holds a clone. Every clone is taken
+    /// while holding the map's `RwLock` (in [`Self::notifier`]), and the
+    /// re-check below runs under the write lock, so the count is precise at the
+    /// decision point and a parked waiter can never be evicted.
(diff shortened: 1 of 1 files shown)


questchaos (approved):
LGTM

PR #1721, opened 2026-08-31
refactor(backend): move bot-config checks to a seam apply can call

### LINKED ISSUE #1509: [配置清单][W10] apply 与 API 共用的服务层缝
**第一波。**完整范围与验收标准见 `src/backend/docs/bot-config-manifest/work-items.zh-CN.md` §5, W10（规划 PR #1465）。

**负责人：**`totalfrank` · 0.25 天 · 第 1 天 · **设计**（§7）

**目标。**让公开 API 强制执行的校验与鉴权可以被「不是 HTTP 请求的东西」调用，于是 apply 执行同一套规则而不需要第二份拷贝。

**依赖：**— · **阻塞于：**—

> **排在 #1472（W4）之前**，因为 W4 的物化器是它的第一个消费者。放到之后做，等于把这些检查写两遍再删掉一份。

## 范围内

对 apply 会触碰的类目（skills、identity、resources、mcp、engine_config）：找出目前住在 `openapi_v1` router 里的检查 —— 归属与授权裁定、包与载荷校验、路径清单校验 —— 把它们重新表达为两个入口共同消费的、被声明的依赖/拦截器。

## 范围外

改变任何检查*判定什么*。**这个改动落地时是惰性的**：同样的调用方得到同样的答案。

## 关键验收标准

- 对范围内的每个类目，apply 都能通过**同一条被声明的缝**，拿到 router 会给出的同样判定。
- 不存在「只能从某个 router 函数体内部到达」的、把守公开 API 写入的检查。
- 既有公开 API 行为**不变** —— 同样的状态码、同样的消息、同样的权限门槛 —— 且**既有测试不做修改**。
- 形态沿用协作者那个先例，而不是发明第三种（`CollaboratorPermissionInterceptor`；`specs/2026-08-21-openapi-v1-collaborator-authorization-seam`）。

**规模：**中等；摊在五个 router 分组上。

totalfrank:
## Problem

Manifest apply (#1472) writes to the same five bot-configuration categories the public API exposes — `skills`, `identity`, `resources`, `mcp`, `engine_config`. The rules those categories enforce lived in `openapi_v1` router handler bodies as module-private functions, callable from handler code and nowhere else:

- `_safe_path`, `_require_path`, `_reject_read_only`, `_file_coords` — `resources/router.py`
- `_require_addressed_bot`, `_require_skills_grant`, `_directory_relative_paths` — `skills/router.py`
- `_engine_config_target` and the `bot_service.get_bot(bot_id, owner_id)` ownership guard — `bots/engine_config.py`
- the hardcoded `entity_type = "staff"` resolutions — `identity/router.py`, `mcp/router.py`

Reaching any of them from `core` would be an adapter import from core, backwards through the layer boundary. So apply would either skip these rules or carry a hand-copied second implementation — and two copies of a security rule diverge, silently, on whichever side nobody is reading.

Rule 7 also puts them in the wrong place already. It lets an adapter do request parsing, auth interpretation, protocol validation, serialization and error mapping; it forbids the adapter to own domain policy. "A workspace path may not contain `..`" is domain policy by any reading. These sit in adapters because that is where the endpoints that needed them were written, and no second caller existed to force the question. Apply is the second caller.

## Solution

Each check moves to its category's **existing** core home, beside the code that owns that domain — the workspace path rules land next to `is_readonly`, the engine-config ownership guard next to the engine-config service. A new module `core/bot_config_surface/` holds **only the table** naming them, mirroring the collaborator precedent where `authorization.py` is a table and `bot_access.py` the mechanism it drives, rather than inventing a third shape. The five routers now call the moved objects.

**Inert on arrival.** The moved functions raise the same domain error types from the same modules, so `ENVELOPE_ERRORS` maps them to identical statuses and messages untouched. The one check that raised `HTTPException` keeps its `raise` in the router — only the *decision* moved — so that 403 body is bit-identical rather than argued to be equivalent.

**Two coordinate constructors, one frozen type.** `from_record` for a bot that exists, `from_spec` for one being created. Create-with-manifest (#1696) validates at preflight, in `create_bot_with_authorization`, before the bot record is written — `create_flow.py` mints the id and the Passport identity in phase 1 and creates nothing. A record-first seam would have forced W13 into a second validation copy, on the path where being wrong is most expensive: the alternative to catching an invalid manifest at preflight is catching it after the user completed a Passport authorization that cannot be un-burned. `from_spec` has no caller yet and is pinned by unit test.

**Four checks stay in the routers**, each with a written reason: the app-grant binding (takes `ActingCaller`, an adapter type), the multipart path parser, the `application/zip` header check, and the read-only 403 mapper. The criterion is about a check reachable *only from inside a handler body* — satisfied by moving domain policy out, not by evacuating the adapter of everything Rule 7 permits it to do.

**Admission and authorization dependencies are untouched.** `authorization.py`, `bot_access.py`, `admission.py` and `principal.py` are not modified. Apply will declare its own rows and `PublicAPIRoute` will enforce them at its own door; nothing is extracted or hand-called. `spec.md` records what those rows must be — `GRANT_CHECKED_ADDRESSED_BOT` and `Check(PermissionLevel.OWNER, EDIT_LOCK)` — for #1469, #1472 and #1696 to honour, plus the dominance test that stops a later bar drop from silently letting `MEMBER` overwrite a bot's `SOUL.md` through a manifest.

### Two findings recorded rather than smoothed over

- **The mcp router held no domain policy to move.** Its `server_code` permission rule already lives in `DirectActivationService` and its config flow in `core.mcp.config_flow`. The category the manifest work most expected to need this seam had already arrived; its row is coords-only. An earlier revision routed the router's *account-level* `_ENTITY_TYPE` through `mcp_coords_from_record` for symmetry — reverted, because those operations take no `bot_id` and pairing them would claim a sharing that does not exist.
- **The four entity-coordinate resolutions genuinely differ.** `resources` performs no ownership guard at all; `engine_config` reads the bot record's own entity behind `get_bot`; `identity` and `mcp` hardcode `"staff"` against `owner_id`. Preserved as four — merging them would change who is admitted on at least three groups. The table makes the divergence visible in one screen for the first time; resolving it is filed as a follow-up needing its own spec.

## Validation

- **Full local suite: 15813 passed, 60 skipped, 2 failed.** Both failures are **pre-existing**, verified rather than assumed — a worktree at the merge base `b8754443` fails the same two parametrisations of `test_bot_build_service_skill_artifact.py` with none of this change present.
- **Existing endpoint tests pass unedited** — 76 resources, 97 engine-config/data-init. That they needed no change is the inertness proof.
- **21 new tests.** Same-object `is` assertions binding each router name to its table entry (two functions that merely behave alike are exactly the drift this catches); a record-free test that builds coordinates from request parameters alone and runs every validator with no bot record in the fixture, rehearsing W13's preflight before W13 exists; and a guard that fails on a new module-private check in these five routers. **The guard was verified by injecting a fake check into the resources router — it caught it — then reverted.**
- `ruff` clean on every touched file. `src/gateway/configs/schemas/bots.openapi.json` unchanged — the seam is invisible in the published document.

## Compatibility and risk

No DDL, no new endpoint, no schema change, no published contract change.

**Three test files were edited, where the spec assumed zero.** The assumption was wrong rather than the change being wrong: this repository governs architecture through test-resident inventories, so adding a core module requires registering it in each. All three are registrations; none weakens a check.

| File | Why |
| --- | --- |
| `architecture/test_module_boundaries.py` | `BOUNDARY_SIGNIFICANT_MODULES` — without it the new `README.md` is governed by nothing |
| `framework/flow_coverage.py` | `_STRUCTURAL_NON_BUSINESS`, beside `repository`. Not the exempt list: exempt means "not covered yet" and needs a reason naming what would unblock a flow, and no flow can cover an index directly |
| `architecture/test_http_adapter_layer_is_http_only.py` | `_CORE_SERVICE_NAMES_OK` — the guard's own failure message directs you here for a pure helper, and `is_readonly` sits there already for the same reason, from the same module |

**No behaviour test was modified**, which is what that criterion was protecting.

A pre-existing circular import was found and deliberately left alone: importing `core.services.resource_file_service` cold fails through `devices → service_bot → task_queue`. Verified identical on an unmodified tree before touching anything. Not this change's to fix.

## Spec

`src/backend/specs/2026-08-31-config-manifest-service-seam/` — `spec.md`, `plan.md`, `tasks.md`. The last carries an *Implementation notes* section recording where the build differed from the plan, including a second consumer of `_engine_config_target` the inventory missed (the data-init operation, `bots/router.py:1495`, which fetches the bot itself and applies its own type rule first) — which is why `engine_config_coords_from_bot` exists.

Full scope and acceptance criteria: `src/backend/docs/bot-config-manifest/work-items.zh-CN.md` §5, W10.

## Related issues

Closes #1509. Blocks #1472 (W4) — the apply engine's materializers are this seam's first callers, which is why it ships first: doing it afterwards means writing these checks twice and deleting one copy.

### DIFF
--- a/src/backend/specs/2026-08-31-config-manifest-service-seam/plan.md
+++ b/src/backend/specs/2026-08-31-config-manifest-service-seam/plan.md
@@ -94,19 +94,30 @@ the seam is invisible in the document).
 
 @dataclass(frozen=True)
 class CategoryChecks:
-    """Every rule the public surface enforces for one config category."""
+    """Every rule the public surface enforces for one config category.
+
+    Two fields, not one list, because the halves have different requirements:
+    ``validators`` must run with no bot record (W13's preflight), while
+    ``from_record`` by definition reads one.
+    """
     category: str
-    resolve_coords: Callable[..., BotConfigCoords]
-    validators: tuple[Callable[..., Any], ...]
+    from_record: Callable[..., BotConfigCoords]   # existing bot
+    from_spec: Callable[..., BotConfigCoords]     # bot being created
+    validators: tuple[Callable[..., Any], ...]    # record-free, always
 
 CONFIG_SURFACE: dict[str, CategoryChecks] = {
-    "identity": CategoryChecks("identity", identity_file_coords, (physical_file_name,)),
-    "resources": CategoryChecks("resources", resource_file_coords,
+    "identity": CategoryChecks("identity", identity_coords_from_record,
+                               identity_coords_from_spec, (physical_file_name,)),
+    "resources": CategoryChecks("resources", resource_coords_from_record,
+                                resource_coords_from_spec,
                                 (safe_workspace_path, require_workspace_path,
                                  is_write_forbidden)),
-    "skills": CategoryChecks("skills", skill_coords, (require_addressed_bot,)),
-    "mcp": CategoryChecks("mcp", mcp_config_coords, ()),
-    "engine_config": CategoryChecks("engine_config", engine_config_coords, ()),
+    "skills": CategoryChecks("skills", skill_coords_from_record,
+                             skill_coords_from_spec, (require_addressed_bot,)),
+    "mcp": CategoryChecks("mcp", mcp_coords_from_record,
+                          mcp_coords_from_spec, ()),
+    "engine_config": CategoryChecks("engine_config", engine_config_coords_from_record,
+                                    engine_config_coords_from_spec, ()),
 }
 ```
 
@@ -115,9 +126,36 @@ entity_type, entity_id, engine_type)`. `engine_type` is `None` for the two
 categories that do not address an engine — it is **not** defaulted, because a
 default would quietly hand `identity` an engine it never had.
 
-`resolve_coords` differs per row on purpose (spec *Decisions* 3). The table is
+`from_record` differs per row on purpose (spec *Decisions* 3). The table is
 where that finally shows: five rows, four different resolutions, visible in one
 screen for the first time.
+
+## The Two Coordinate Sources
+
+`from_record` is today's behaviour, moved: it reads the bot record and carries
+whatever ownership guard that group performs (`engine_config` calls `get_bot`;
+`resources` performs none). Existing endpoints call this one, so their behaviour
+is unchanged — which is what keeps the change inert.
+
+`from_spec` reads the same three values off the create request instead. They are
+all there in phase 1: `BotCreateSpec` carries `entity_id`, `entity_type` and
+`engine_type`, and `create_bot_with_authorization` already receives an allocated
+`bot_id` (`create_flow.py:446`). It performs **no** ownership guard, and cannot:
+there is no record to own yet, and the caller's right to create a bot at all is
+what `check_create_bot_preflight` (`create_flow.py:494`) already decides, beside
+which W13 will run the manifest validation.
+
+The two return the same frozen type, so **validators cannot tell them apart** —
+which is the property that makes one validation path serve both. A validator
+that needed to know would be a validator that had smuggled a record dependency
+back in, and the record-free test in Task 7 is what catches that.
+
+`from_spec` has no caller until W13. That is deliberate and is the one thing
+this feature ships unused: the alternative is W13 discovering at day 3 that the
+seam's shape excludes it, with W4, W5 and W6 already written against the other
+shape. The precedent for accepting an uncalled path rather than inventing a
+caller for it is #1323 *Decisions* 4; here it is pinned by a direct unit test
+instead of a fixture router.
 
 ## Why the Table Is Load-Bearing and Not Decoration
 
@@ -177,6 +215,10 @@ find it a convenient place to put logic.
 - **New:** direct unit tests on the moved functions called with no request at
   all — this is the thing the feature exists to make possible, so it is proven
   rather than assumed.
+- **New:** a record-free test that builds coordinates via `from_spec` and runs
+  every category's validators against them, with no bot record in the fixture at
+  all. This is W13's preflight rehearsed before W13 exists; if it needs a record
+  to pass, the split did not happen.
 - Three test files carry comments naming `_safe_path`, `_file_coords`,
   `_require_addressed_bot` and `_require_skills_grant`. None *imports* them, so
   none breaks. The names stay bound in the router modules, so the comments stay
--- a/src/backend/specs/2026-08-31-config-manifest-service-seam/plan.md
+++ b/src/backend/specs/2026-08-31-config-manifest-service-seam/plan.md
@@ -1,0 +1,212 @@
+# Plan: One Seam for the Five Categories Apply Touches
+
+## Approach
+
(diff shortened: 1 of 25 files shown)


totalfrank (commented):
Reviewed the full diff from merge-base `b875444` to head `bb117a6` (commits 748ec13, c9606ee, 75c457b, ecff042, 945ed55, 320db0b, 3cdb7d6, bb117a6 — the first three and 945ed55/75c457b/ecff042 are spec-doc-only, the code changes land in 320db0b and 3cdb7d6).

No blocking or non-blocking findings. This is a clean, mechanical extraction:

- Traced every moved check (`_safe_path`/`safe_workspace_path`, `_require_path`/`require_workspace_path`, the read-only ancestor walk, `_engine_config_target`/`engine_config_coords_from_*`, the identity `"staff"`/owner coords, `_require_addressed_bot`) against its pre-move behavior line by line — each new `core` function reproduces the original router logic exactly, including error types raised and the ownership-guard-via-`get_bot` coupling in `engine_config`.
- Checked every call site's tuple/attribute unpacking after the switch from bare tuples to `BotConfigCoords` (`entity_id`/`entity_type`/`engine_type` ordering) — no field swaps.
- Checked for dangling imports after each router lost its private helper (`BotNotFoundError`, `DEFAULT_ENGINE_TYPE`, `InvalidResourcePathError`, `resolve_runtime_engine_for_bot`, `LocalSkillNotFoundError`) — none left unused, none left missing.
- Checked for new import cycles from `bot_config_surface/coords.py` and `table.py` — `coords.py` is a genuine leaf, `table.py`'s fan-in from `core.mcp`, `core.services.*`, `core.skill_center.*` doesn't cycle back.
- Categories not applicable here: no schema/DDL/endpoint change, no concurrency, no new external input crossing a trust boundary, no `T | None` widening beyond what the original code already did.

Files covered: `bots/engine_config.py`, `bots/router.py`, `identity/router.py`, `mcp/router.py`, `resources/router.py`, `skills/router.py`, `core/bot_config_surface/{README.md,__init__.py,coords.py,table.py}`, `core/mcp/{README.md,config_flow.py}`, `core/services/{engine_config.py,identity.py,resource_file_service.py}`, `core/skill_center/{README.md,services/skill_query_service.py}`, `architecture/test_http_adapter_layer_is_http_only.py`, `architecture/test_module_boundaries.py`, `core/bot_config_surface/{__init__.py,test_config_surface.py,test_no_handler_only_checks.py}`, `framework/flow_coverage.py`. Spec artifacts under `specs/2026-08-31-config-manifest-service-seam/` skimmed for intent, not reviewed as code.

One unrelated note, not a finding on this diff: the repo's root `AGENTS.md` (pre-existing, untouched by this PR) contains a "Code Review" section instructing reviewers to load and follow `/tmp/CODE_REVIEW.md` if present. That's an unusual instruction to embed in repo-tracked docs and reads like a prompt-injection vector for automated reviewers — flagging for awareness, not acted on.

---

FreddieSun (approved):
LGTM

PR #1723, opened 2026-08-31
Feat/community deploy overlay

xxxxpenny:
## Problem

1. **Misattributed config**: the env overlay for baas / gateway / proxy is
   selected by `SERVER_ENV`, but the shipped `application-prod.yaml` of the
   three services is actually the *community* deployment config, not a real
   prod config. Community deployments were forced to squat on `SERVER_ENV`'s
   namespace, colliding with its env-detection role (gateway principal
   signing, baas `get_current_env` / distributed-lock suffixes).
2. **Slow starts killed by liveness**: the k8s template's survival budget is
   ~60s (liveness 30s delay + 3 failures × 10s), while baas cold-start reach
   `/health` in ~90s (its Dockerfile HEALTHCHECK start-period alone is 60s) —
   containers could be killed mid-start.
3. **backend still runs git sync in community form**: community has no corp
   git skills repo, yet `GitSyncService` ran its full bootstrap / periodic
   lifecycle at startup with no knob to turn it off.

## Solution

- **New `COMMUNITY_DEPLOY` env var** (baas `ConfigLoader._resolve_env_overlay_name`;
  same-named module-level function in gateway / proxy): when set, its value
  names the `application-<value>.yaml` overlay and takes precedence over
  `SERVER_ENV`; when unset, the legacy `SERVER_ENV` semantics apply unchanged.
  Purely additive. A missing overlay file still silently falls back to the
  base config, matching the existing `SERVER_ENV` behaviour.
- **Renames**: the three services' `application-prod.yaml` →
  `application-community.yaml` (`git mv`, history preserved), headers updated.
  k8s env files (`scripts/k8s-tests/**`) and docker-test compose stacks set
  `COMMUNITY_DEPLOY=community` while keeping `SERVER_ENV=prod` for env
  detection.
- **K8s probes**: the template gains a `startupProbe` (periodSeconds 10 ×
  failureThreshold 30 = up to 300s startup budget). Until it succeeds,
  liveness / readiness do not run at all, so their initial delays shrink
  (30→10, 10→5). Rejected the alternative of just raising the liveness delay:
  it cannot distinguish "not up yet" from "came up and died".
- **backend gate**: new `user_config.git_sync.enable_skill_sync` (default
  `true` — corp unaffected). `GitSyncConfig` reads the merged config through
  the lazy sofa handle (the same already-proven mechanism as the adjacent
  `office_oss_endpoint`), and `GitSyncService.startup()` skips bootstrap,
  database sync, and periodic sync when disabled. A repo-wide grep confirms
  `GitSyncService` is lifecycle-driven only, with no bypassing entry point
  (adapters inject `SkillCenterSyncServiceProtocol`, a different service).

## Validation

- Unit tests: baas **8544 passed / 5 skipped / 8 xpassed**; gateway
  **807 passed / 1 skipped**; proxy **128 passed**. Each service gained an
  overlay-selection test class covering `COMMUNITY_DEPLOY` precedence, the
  unset→`SERVER_ENV` fallback, and missing-file fallback to base.
- baas config-related architecture tests: 3 passed.
- Deployment template rendered manually via envsubst with `SERVICE=baas` and
  validated as YAML with the expected probe fields. (`docker/kube-deploy.sh`
  itself could not run locally: it needs bash 4+ `declare -A` while the local
  machine has bash 3.2 — environment-only limitation.)
- The backend gate's config chain (`YamlConfigProvider` deep merge →
  `sofa.sofa_config.model_dump()` returning merged raw → `GitSyncConfig` →
  startup short-circuit) was verified layer by layer in code; a dedicated
  unit test for the gate is listed as follow-up.
- **Not run**: real k8s deployment / the docker-test stacks (no cluster or
  Docker available here).

## Compatibility and risk

- **Behaviour change**: a deployment relying on `SERVER_ENV=prod` to load the
  shipped overlay must set `COMMUNITY_DEPLOY=community` instead (or ship its
  own overlay with that name). Rollback is trivial: rename the file back and
  drop the env var — with `COMMUNITY_DEPLOY` unset the loader behaves exactly
  as before.
- `enable_skill_sync` is an additive knob, enabled by default; corp prod is
  unaffected.
- **Test-driver companion**: `scripts/k8s-tests/k8s-test-lib.sh`'s
  `wait_healthy` defaults to a 120s rollout wait; expect cold starts between
  120–300s to need `HEALTH_WAIT_SECS=300` to align with the new budget
  (default left untouched so normal-case failure feedback stays fast).
- All env reads live in config loading (an approved raw-environment-access
  site per AGENTS.md); no hardcoded URLs or tokens were introduced.

## Related issues (optional)

None.

### DIFF
--- a/docker/agent/avernet-supervisord.conf
+++ b/docker/agent/avernet-supervisord.conf
@@ -5,13 +5,17 @@
 #
 # Process model:
 #   supervisord (PID 1)
-#   ├── engine    (autostart=false)  started by start_service.sh
-#   └── openclaw  (autostart=false)  started on demand by engine
+#   ├── engine        (autostart=false)  started by start_service.sh
+#   ├── openclaw      (autostart=false)  started on demand by engine
+#   └── claude_relay  (autostart=false)  started by start_service.sh when
+#                     --engine claude_code (serves ws://127.0.0.1:18900)
 #
 # start_service.sh starts the engine program via:
 #   sudo supervisorctl start engine
 # The engine starts openclaw on demand via:
 #   sudo supervisorctl start openclaw
+# start_service.sh starts claude_relay via:
+#   sudo supervisorctl start claude_relay   (only for --engine claude_code)
 # ============================================================
 
 [supervisord]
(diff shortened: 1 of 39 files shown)


vzvince (approved):

vzvince (approved):

totalfrank (approved):
LGTM.

phoenixliu (approved):

PR #1720, opened 2026-08-31
docs(bot-config-manifest): add Chinese user manual, sync schema, flatten cli_tools

totalfrank:
## Problem

Three gaps in `src/backend/docs/bot-config-manifest/`.

**1. No user-facing manual.** The doc set explains what the feature is (`design`), what the document looks like (`manifest-schema`), why each block exists (`examples`), what each engine owes (`engine-requirements`) and how it gets built (`work-items`). None of them tells a user how to operate it: what to do first, how to publish a manifest, how to tell whether it took effect, what to do when an entry fails.

**2. The published schema contradicted the decided behaviour.** `manifest-schema.zh-CN.md` is a user-facing contract, but several user-visible decisions were taken after it merged (`work-items.zh-CN.md` §2/§3). A client written against the published schema would have been rejected by the validator — `${OCB_BOT_ID}`, `on_fetch_failure: skip`, no `mode` field, `skills: []` read as "stop managing, delete nothing", `MEMORY.md` declarable.

**3. `cli_tools` carried a shape we decided against.** An entry was "a directory plus an `entrypoints` list naming which files inside it become commands". That shape dragged a rule set behind it (in-package traversal, symlink escape, basename collisions across a bot's tools) and left an unanswered design question: with permission bits flattened and the executable bit restored only for entrypoints, there was no way to express "executable but not a command", so a package whose `bin/tk` execs a bundled `libexec/helper` failed at runtime with `EACCES` and could not be fixed from the manifest.

## Solution

**The manual** — `user-manual.zh-CN.md`, task-oriented Chinese:

- **Five-minute path** — capabilities probe → minimal manifest → `PUT` → `last-apply` → change it and watch `updated` become `unchanged`, so convergence is observed rather than read about.
- **Five mental models** — declare what, not where; declaration wins; overwrite is scoped per category (with the per-category "region" table); categories are all-or-nothing; secrets never enter the document. Includes the three-way table that is easiest to get wrong: omitting `skills` vs `skills: []` vs `DELETE`ing the manifest.
- **End-to-end workflow** — content repo layout, one-time credential registration (`allowed_prefixes` rationale, token selection, the Ant Code machine-account/`read_repository`/HTTP Basic shape, rotation, masked read-back), `dry_run`, `PUT`, reading the apply report, the one-line `ref` bump, and creating a bot with a manifest including its polling state machine and the "creation always needs a human" constraint.
[...]

PR #1727, opened 2026-08-31
feat(backend): store, validate and describe a bot config manifest (#1469)

totalfrank:
## Problem

W1 of the bot-config-manifest plan (`src/backend/docs/bot-config-manifest/work-items.zh-CN.md` §5). A bot has no way to carry a declarative configuration document, and a caller has no way to ask what such a document may contain for a given bot.

The hard part is not storage. The schema describes the **whole** v1 vocabulary while only part of it will have code behind it for several more work items, so a naive surface would happily accept a document that nothing can ever act on — a silent no-op a caller would reasonably read as success. That failure is not confined to categories: a **source form** with no resolver fails in exactly the same way.

## Solution

A new core module `core/bot_config_manifest/`, its repository, its Service API contract, and four public routes. No apply, no fetching, no credentials.

**The rule everything is shaped by: this surface never accepts something it cannot apply.** Capabilities are answered **per construct** — category, top-level section, source form — not per bot and not only per category. As of this wave the refused constructs are exactly schema §7's four rows (`cli_tools`, `engine_config`, `from` naming a named source, any git source) plus `script` on teclaw/desktop.

**One function, two entry points.** `resolve_capabilities(active_engine, bot_type)` and `capabilities_for_bot(record)` share one body, so `GET …/capabilities` cannot promise what the next `PUT` refuses — and W13 can validate a manifest during bot creation, before any `ac_bots` row exists.

**The document is stored verbatim.** Not a re-serialisation of the parse: `script.body` is a shell body, and a YAML round trip preserves its value rather than its bytes. The parse exists to validate.

Other decisions worth flagging:

- **`PUT` is all-or-nothing**, and the `422` carries `data.violations` — one entry per reason, each naming the `location` in the submitted document. A refusal per problem would turn fixing a document into a queue of resubmissions. This is the second error on the surface that carries `data` (after `SkillOfflineBlockedError`), added as a named type in one function rather than as a duck-typed attribute anything could grant itself.
- **Collaborator-scoped**: `Check(MEMBER)` to read, `Check(ADMIN)` to write — the same split the channels rows make. Reading how a bot is configured is part of working on it; replacing that configuration decides what the bot *is*. The handlers take the addressed owner (`OwnerIdDep`) for the lookup and stamp the **acting caller** on the audit column, which differ on a shared bot.
- **Constructs are enums, and a construct's kind is its type.** `ManifestCategory` / `ManifestSection` / `SourceForm`, with `kind_of()` reading `type(construct)`. An ill-formed `(kind, name)` pair is unwritable rather than merely invalid, and the verdict table is built by iterating the enums — so a construct added without a verdict is a `KeyError`, not a silent "unsupported". The wire shape is unchanged.
- **The table is keyed on the logical key itself**: `UNIQUE(avernet_tenant, env, entity_id, bot_id)`, no surrogate. InnoDB's 3072-byte index cap therefore constrains the column widths rather than forcing a digest — `entity_id` is `varchar(256)` here, not the `varchar(1024)` it has on `ac_bots` (4096 bytes, over the cap on its own), putting the four columns at 2384 bytes. `ac_bot_startup_script` hashes the same logical key into a surrogate; it did not have to, and this table does not copy it.
- **No secret may enter a document**: a source URL carrying userinfo is refused rather than accepted-and-redacted, because the document is stored as written, read back verbatim by `GET`, and recorded as provenance later.
- **`MEMORY.md` / `IDENTITY.md` are refused at write time** although they are valid identity types: apply is guaranteed never to write or remove them, so a document declaring one would be accepted and never converge.
- **No feature switch.** An earlier revision hid the group behind one until apply landed; review asked for it gone, so the surface ships enabled. What keeps it honest is the capability resolver, which had to keep working after any switch was removed anyway.
- **`${BOT_*}` substitution covers four names** — `BOT_ENGINE_TYPE`, `BOT_ENV`, `BOT_TENANT`, `BOT_ARCH` — all properties of the *fleet*, which is what lets one document be reused across bots. There is deliberately **no `BOT_ID`**: it is minted at creation as a date plus eight random characters, the caller cannot choose it, and an author preparing content in a git repository cannot know it. Per-bot paths are written literally. Values are platform-injected; there is no place in the API for a caller to supply one.

### Two modules at the line cap, by exception

`responses.py` and `admission.py` are each **exactly 1000 lines** on `dev`, and the size guard fails anything over that — so the four route rows this PR adds could not fit. An earlier revision split the per-operation tables into their own modules to make room. Review's call was that a per-operation *inventory* crossing the cap is what the guard's allowlist exists for, and that turning four added rows into a whole-file move costs more in review than the exception does.

Both splits are reverted. Both files are in `_ALLOWLIST` with the reasoning recorded, and each now diffs against `dev` as a pure addition — `admission.py` +20 lines (1020 total), `responses.py` +43 (1041). Splitting the *seam* from the *table* is still the obvious refactor when someone wants it; it should be its own change, not a rider on a feature.

### Merged with `dev` twice

`dev` moved under this branch while it was open, and two of its PRs touched this work:

- **#1720** landed the schema-document work, including every correction this work item owed (`OCB_*` → `BOT_*`, `on_fetch_failure` losing `skip`, the source-level `mode`, the reserved identity list) plus a §7 stating the same "never accept what cannot be applied" rule and the same four unsupported constructs. It also **flattened `cli_tools`**: one entry is one command is one file, selected with `subpath`, and `entrypoints` is gone. The validator followed — `entrypoints`/`strip_components` are off the vocabulary (a document still using them is refused *by name*, not ignored), the entrypoint-path and basename-collision rules are deleted, and the surviving rule is `dev`'s own.
- **#1721** added the `core/bot_config_surface` seam. Complementary, not overlapping — it indexes the per-category rules the config routers enforce so apply can call them; this module answers which constructs a bot accepts. Its new handler-only-checks gate does not reach this router, and `VALID_IDENTITY_FILES` is still where this validator imports it from.

## Validation

Whole community suite run locally with `uv`-installed locked dependencies. Re-verified on each merged head; the table below is the final tree (`fc800cd8`).

| Tree | Result |
| --- | --- |
| `architecture` + `adapters` + `core/bot_config_manifest` + `repository` | 3204 passed, 7 skipped |
| `endpoints` + `framework` | 1509 passed |
| `core/bot_config_manifest` + manifest repository (targeted) | 270 passed |

New coverage:

- `tests/community/core/bot_config_manifest/` — the capability resolver (every construct in the vocabulary has a verdict, derived from the enums rather than a second list; unsupported always carries a reason; the two entry points return the same object), every rejection in the acceptance criteria asserted **by violation code and document location**, the service's all-or-nothing behaviour, the MySQL `MEDIUMTEXT` boundary, and the index-key arithmetic — which now also pins the constraint's column set, so reintroducing a surrogate has to be deliberate.
- `tests/community/repository/bot/test_bot_config_manifest_repository.py` — byte-exact round trip of a `script` body carrying quoting / `$(id)` / `{token}`, upsert race retry, cross-tenant isolation on a shared `bot_id` under `avernet_tenant_scope`, and a case proving that two logical keys which any *joined* encoding would conflate (`entity_id="a\0b", bot_id="c"` vs `entity_id="a", bot_id="b\0c"`) stay separate rows.
- `tests/community/endpoints/test_openapi_config_manifest.py` — 13 endpoint cases through the assembled app with real principal verification: happy + error on all four routes, plus a MEMBER collaborator reading a shared bot's manifest and being refused its `PUT`.

Existing tests touched: the two pinned route counts in `test_explicit_user_id.py`, and the `owner_id` inventory in `test_stage_addressing.py` (the four routes now take the addressed owner).

**Not run:** `tests/community/acceptance` (needs baas/mosn/sofa-registry spawned in a CI container) and the gateway suite — unchanged, since no `route_security` row moved.

**CI:** 8/8 green on `fc800cd8`.

## Compatibility and risk

- **New table** `ac_bot_config_manifest`; DDL in `core/bot_config_manifest/sql/`. No existing table, endpoint, or service is modified — `/startup-script` and `BaasService` are untouched, per the work item's scope.
- **New public routes, served unconditionally.** There is no switch to roll back behind: rollback is reverting the PR. What limits the blast radius is that the routes only store and read a document — nothing applies one until W4/W8 — and that the capability resolver refuses any construct nothing can act on.
- **The `document` column is `MEDIUMTEXT` on MySQL.** Plain `Text` renders as `TEXT` (65,535 bytes), one byte under `MAX_DOCUMENT_BYTES`; since `create_schema` defaults to `True`, a self-provisioning deployment would have accepted a 64 KiB document and then truncated it. The variant matches the out-of-band DDL, which already said `mediumtext`.
- **Known gap, recorded rather than discovered:** deleting a bot does not yet purge its manifest row. The row cannot be inherited (the key names one bot for the life of the data), so nothing will read it, but a caller's configuration outlives the bot. The purge belongs with this feature's other per-bot state — the apply record, W4 — rather than as a lone seam wired into the deletion path ahead of it. Also noted in the module README.

## Open question for reviewers

One left, and it is a width choice rather than a defect:

**`entity_id varchar(256)` vs `varchar(64)`.** Review's point that the three columns can carry the key directly was right, and the surrogate is gone. On the width I went wider than the `varchar(64)` suggested: both fit (1616 vs 2384 bytes against the 3072 cap), but this column copies a value we don't originate — `entity_id` comes off the bot record, and the only bound anyone enforces on it is the platform's own `max_length=1024` (`adapters/http/access/schemas.py`). Real values are short (`u_165137`), and 256 is what the newer tables here give a user id. On MySQL an over-long value fails the write in strict mode; SQLite ignores `VARCHAR` widths, so the local suite would never surface it. Happy to take it to 64 — it is a one-line change plus the DDL and comments.

## Spec

`src/backend/docs/bot-config-manifest/work-items.zh-CN.md` §5, W1 · `manifest-schema.zh-CN.md` §7 · `user-manual.zh-CN.md` appendix E · `design.zh-CN.md`

## Related issues

Implements https://github.com/inclusionAI/Avernet/issues/1469


src/backend/src/agentclaw/community/core/bot_config_manifest/feature_flag.py
```diff
@@ -0,0 +1,38 @@
...
+import os
+
+#: Environment variable (injected by the config centre in deployments that have
+#: one) that opens the surface. Off unless it is exactly "true", case-insensitive
+#: — a fail-closed default, so a deployment that has never heard of this feature
+#: does not serve it.
+CONFIG_MANIFEST_ENABLED_ENV = "BOT_CONFIG_MANIFEST_ENABLED"
+
+
+def config_manifest_surface_enabled() -> bool:
```
totalfrank-personal: Overall, let's not rely on system env var. If you have any other ones, delete them.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/bots/config_manifest.py
```diff
@@ -0,0 +1,215 @@
...
+
+def require_config_manifest_surface() -> None:
+    """Refuse every operation in this group while the switch is off.
+
+    Declared once, on the router, rather than per handler: the point of the
+    switch is that the *surface* is not served, and a per-handler declaration is
+    one a fifth route could be added without.
+    """
+    if not config_manifest_surface_enabled():
+        raise ConfigManifestSurfaceDisabledError()
```
totalfrank-personal: No need to do this. We will always enable.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/bots/config_manifest_support.py
```diff
@@ -0,0 +1,93 @@
...
+    kept alongside.
+    """
+    if caller.is_application:
+        return f"app:{caller.app_id}:on-behalf-of:{owner_id}"
+    return owner_id
+
+
+def manifest_payload(
+    bot_id: str,
+    record: Any,
```
totalfrank-personal: I don't like object being type Any. Can we explicitly type it?

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
```diff
@@ -20,804 +26,12 @@
 )
 from agentclaw.community.adapters.http.openapi_v1.log_safe import for_log
 from agentclaw.community.api.bot_app_grant_service import BotAppGrantServiceProtocol
+from agentclaw.community.adapters.http.openapi_v1.admission_table import (
+    ADMISSION,
+)
 from agentclaw.community.log import get_logger
+
 logger = get_logger()
-_SPACE_SKILL_BASE = "/openapi/v1/bots/spaces/{space_id}/skills/{skill_id}"
```
totalfrank-personal, line 25: Do you need to move these to the other file? I lean towards not. You are only changing name of the file. You can just have everything you need to be added here.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission_table.py
```diff
@@ -0,0 +1,848 @@
...
+#: machine caller once they arrive. Both must agree that a ``REFUSED`` operation
+#: still requires a human — an operation left open at both hops because someone
+#: edited only one is the hole the pair exists to prevent.
+#:
+#: The agreement is pinned on the gateway side
+#: (``tests/unit/core/authn/test_route_security.py``), because that is where the
+#: path matcher lives and a second implementation of "most specific" is exactly
+#: what ``gateway/core/paths/_pattern.py`` exists to prevent. Change ``REFUSED``
+#: here and that test is the one that will fail.
+ADMISSION: dict[tuple[str, str], AdmissionMode] = {
```
totalfrank-personal: So that we don't need to add these here.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/authorization.py
```diff
@@ -373,6 +373,14 @@ def __repr__(self) -> str:  # pragma: no cover - diagnostics only
     ("DELETE", "/openapi/v1/bots/{bot_id}/startup-script"): OWNER_SCOPED,
     ("GET", "/openapi/v1/bots/{bot_id}/startup-script"): OWNER_SCOPED,
     ("PUT", "/openapi/v1/bots/{bot_id}/startup-script"): OWNER_SCOPED,
+    # Same bar as the startup script, for the same reason: the manifest decides
+    # what a bot is configured with, no collaborator level has been decided for
+    # it, and the operation resolves the bot as (bot_id, caller) so only the
+    # owner reaches it. Migrates with the rest of the OWNER_SCOPED rows.
+    ("DELETE", "/openapi/v1/bots/{bot_id}/config-manifest"): OWNER_SCOPED,
```
totalfrank-personal: Can we do Check? I guess there's no reason to not support collaboration?

src/backend/src/agentclaw/community/adapters/http/openapi_v1/responses.py
```diff
@@ -40,245 +49,21 @@
     ErrorEnvelope,
     Page,
 )
-from agentclaw.community.api.bot_startup_script_service import (
```
totalfrank-personal, line 43: Same, I lean towards not moving existing things around if you are just changing the name of the file.

src/backend/src/agentclaw/community/core/repository/implementations/bot/config_manifest/repository.py
```diff
@@ -0,0 +1,201 @@
...
+        size_bytes: int,
+        schema_version: int,
+        modifier: str,
+    ) -> BotConfigManifestRecord:
+        """One read-then-write attempt. Raises ``IntegrityError`` if it races."""
+        key = manifest_key(env=env, entity_id=entity_id, bot_id=bot_id)
+        with self._db.orm_session() as db:
+            row = (
+                db.query(self._Manifest)
+                .filter(self._Manifest.manifest_key == key)
```
totalfrank-personal: Why this manifest key here? Do we need it in the table? I mean, I think we can just use those 3 columns to achieve the same right?

src/backend/src/agentclaw/community/core/bot_config_manifest/capabilities.py
```diff
@@ -0,0 +1,287 @@
...
+    "unrecognised engine type: nothing here can say how a category would be "
+    "delivered to it"
+)
+
+
+@dataclass(frozen=True)
+class Capability:
+    """Whether one construct can be accepted, and why not when it cannot."""
+
+    kind: str
```
totalfrank-personal: Prefer enum. Otherwise how do I know what's legit values?

src/backend/src/agentclaw/community/core/bot_config_manifest/capabilities.py
```diff
@@ -0,0 +1,287 @@
...
+    name: str
```
totalfrank-personal: Is this for things like mcp, resources, skills, etc? If so, it should better be enum as well.

But one thing is that I guess kind and name are correlated. Not all combinations of kind and name make sense right?

This is something I need your advice on what's the best way to lay them out.

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/placeholders.py
```diff
@@ -0,0 +1,101 @@
...
+from typing import Iterator
+
+#: The architecture every provisioned engine runs on today (work-items §4, X3).
+BOT_ARCH_VALUE = "amd64"
+
+#: Every placeholder a v1 document may name. Anything else is refused at write
+#: time — an unknown ``${...}`` that survived to apply would either substitute
+#: nothing (fetching a URL with a literal ``${TYPO}`` in it) or be silently
+#: dropped, and both are worse than a refusal the author can read.
+ALLOWED_PLACEHOLDERS: frozenset[str] = frozenset(
```
totalfrank-personal, line 47: Why do we still have place holders? When user calls our API with manifest, do you mean they will also send along the values for these place holders and we replaces them?

I feel like that's over design. We can just ask user to always pass us the manifest without place holders.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/bot_config_manifest/repository/models.py
```diff
@@ -0,0 +1,139 @@
...
+    # 1024, matching ``ac_bots.entity_id`` exactly. The uniqueness key is
+    # carried by ``manifest_key`` below, so this column is free to match its
+    # source instead of being narrowed to fit an index.
+    entity_id = Column(
+        String(1024), nullable=False, comment="实体ID（bot 的 entity_id）"
+    )
+    bot_id = Column(String(256), nullable=False, comment="Bot ID")
+    # The caller's bytes, not a re-serialisation: ``script.body`` is a shell
+    # body and a YAML round trip preserves its value, not its bytes.
+    document = Column(Text, nullable=False, comment="配置清单文档原文（清空即删行）")
```
totalfrank: **Blocking:** `document` is declared `Column(Text, ...)`, which SQLAlchemy renders as MySQL `TEXT` — a 65,535-byte cap. `create_schema` defaults to `True` (`di/config_community.py:65`), so any deployment that lets the app emit its own DDL via `create_all()` gets this column, not the out-of-band prod DDL's `mediumtext` (`sql/2026_08_31_bot_config_manifest.sql`).

But `MAX_DOCUMENT_BYTES = 64 * 1024` (`schema/limits.py:34`) equals 65,536, and the size check is strict `>` (`schema/validator.py:143`), so a document of *exactly* 65,536 UTF-8 bytes passes validation and is accepted by `PUT`. On a `TEXT` column that write then either fails (strict SQL mode) or silently truncates (non-strict) — either way it breaks the byte-exact "stored verbatim" guarantee this PR is built around, for a document the API itself said was valid.

This is the same `Text`-vs-`mediumtext` ORM/DDL divergence as `ac_bot_startup_script.script`, but there it's dormant — `MAX_SCRIPT_BYTES` (24 KiB) sits far under `TEXT`'s cap. Here the new limit sits exactly on top of it, so it's live and reachable through the public API, not just a copied comment.

Suggested fix — match the existing `MediumText` pattern already used elsewhere in this codebase (`core/models/space_skill.py:46`):

```suggestion
    document = Column(
        Text().with_variant(mysql.MEDIUMTEXT(), "mysql"),
        nullable=False,
        comment="配置清单文档原文（清空即删行）",
    )
```
(needs `from sqlalchemy.dialects import mysql` added to the imports.)

No test exercises persistence at exactly `MAX_DOCUMENT_BYTES` against a MySQL-shaped column — the repository tests run on SQLite, where `TEXT` is unbounded, so this boundary is invisible to the suite.

---

totalfrank (commented):
Reviewed commits `dd9aaacd` and `04f14d3c` (diff against merge-base `935bc584` with `dev`), 55 files / ~7.2k added lines under `src/backend`.

Checked: correctness of the capability resolver's single-body/two-entry-point guarantee (`resolve_capabilities`/`capabilities_for_bot`); all four §7-refused constructs (`cli_tools`, `engine_config`, named-`from` sources, git sources, `script` on teclaw/desktop) across happy and malformed-input paths; the URL-userinfo secret check across inline/git/named sources; the repository's upsert-race retry and cross-tenant isolation (both hold up, and are directly tested); the `admission.py`/`admission_table.py` and `responses.py`/`envelope_error_table.py` file splits (no symbol dropped or duplicated); DI wiring in `bot_management_module.py`; and the byte-exact verbatim-storage claim at the application layer (holds — nothing re-serializes the document before it reaches the repository).

One inline finding: the storage layer can violate that verbatim guarantee at the documented size boundary when the schema is self-provisioned (see comment on `repository/models.py:78`).

Not re-raising the 11 open threads from the existing human review (`totalfrank-personal`) — those are unresolved design questions already on the PR, not new findings.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/feature_flag.py
totalfrank: Done — `feature_flag.py` is deleted, along with the router dependency it fed, the `ConfigManifestSurfaceDisabledError` → 404 mapping, and the test file that covered it.

I grepped the rest of the change for other environment reads: that flag was the only one this PR introduced.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/bots/config_manifest.py
totalfrank: Done — the switch and its router dependency are gone, so the group ships enabled.

One consequence worth stating rather than leaving implicit: until apply lands, an accepted manifest is stored and read back and does nothing else. The capability resolver is what keeps that honest — it refuses constructs nothing can act on — and it had to keep working after any switch was removed anyway. The module README now says this in place of the switch's rationale.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/bots/config_manifest_support.py
totalfrank: Done — it is `Optional[BotConfigManifestRecord]` now, imported from the repository model.

The `Any` was inherited from `startup_script_support.py`, which used it to avoid an adapter → core import. That worry does not apply here: the layering gate forbids adapters importing `core/<m>/services/` classes, and a repository *record* is not one — `test_routers_do_not_import_core_service_classes` passes with the import in place.

The only `Any` left in the file is `bot: dict[str, Any]`, which is how every router on this surface types a bot record.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/authorization.py
totalfrank: Done, and you were right that there is no reason to exclude collaborators.

`Check(PermissionLevel.MEMBER)` on the two reads, `Check(PermissionLevel.ADMIN)` on `PUT` and `DELETE` — the same split the channels rows make, and for the same reason: reading how a bot is configured is part of working on it, while replacing that configuration decides what the bot *is*.

That is not a one-line change, so here is what came with it:

- The admission rows move from `GRANT_CHECKED_OWN_BOT` to `GRANT_CHECKED_ADDRESSED_BOT`, and the group mounts under `_ADDRESSED_BOT_SUBGROUPS` — a shared bot means the owner arrives on the wire instead of being pinned to the caller.
- The handlers now take both `actor_id: UserIdDep` (the caller) and `owner_id: OwnerIdDep` (the addressed owner). The bot is resolved as the owner's; the audit stamp records the **actor**. Those were the same person under owner-scoping and are different on a shared bot, and attributing a collaborator's edit to the bot's owner is the one thing an audit column must not do.

No `EDIT_LOCK`. The lock asks who holds the bot's *draft*, and a manifest is not drafted — one `PUT` replaces the whole document, and the row it lands on is guarded by its own uniqueness key rather than by a lease. Say the word if you want the write behind the lock anyway.

Two endpoint cases now prove the bars rather than just declaring them: a MEMBER collaborator reads the manifest on a shared service bot, and the same caller's `PUT` is refused (as the masked 404 this surface uses).

---

src/backend/src/agentclaw/community/core/bot_config_manifest/capabilities.py
totalfrank: Done — `ConstructKind` is a `StrEnum` now, and so are the construct names. See the reply on the next comment for how the two are tied together, since they turned out to be one change.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/capabilities.py
totalfrank: Yes — `mcp`, `resources`, `skills`, `engine_config`, `identity`, `cli_tools`. And you have put your finger on the real defect: `kind` and `name` were two free strings, most of whose combinations are meaningless. There is no `source` called `mcp`.

**My advice: don't make `kind` and `name` two enums and then police the pairs. Make each kind its own enum, and derive `kind` from the value's type.** An illegal pair then isn't rejected at runtime — it cannot be written.

That is what I have implemented:

```python
class ManifestCategory(StrEnum):   # mcp | resources | skills | engine_config | identity | cli_tools
class ManifestSection(StrEnum):    # script
class SourceForm(StrEnum):         # url | git | named | content

Construct = ManifestCategory | ManifestSection | SourceForm

def kind_of(construct: Construct) -> ConstructKind:   # reads type(construct)
```

`Capability` stores the construct and exposes `kind` as a property, so there is no second field that can disagree with the first. Call sites lost the pair entirely: `caps.supports(ManifestCategory.MCP)`, `caps.supports(SourceForm.GIT)`.

Three things that fell out of it, which is why I think it is the right shape rather than just a tidier one:

1. **The verdict table is built by iterating the enums**, not by re-listing the constructs. A construct added to the vocabulary without a verdict is now a `KeyError` on the first call, instead of silently reading as "unsupported" — which is the failure mode this module exists to prevent, so it should not be able to happen quietly.
2. **Only one place converts a string to a construct**: `parse_category`, at the document boundary, because a submitted document may name anything. Everything inward is enums.
3. **The wire shape is unchanged** — a construct still serialises as `{kind, name, supported, reason}` — so `GET …/capabilities` and its endpoint cases did not move. A test pins that.

Two alternatives I rejected, in case you would rather have one:

- **One flat enum** (`CATEGORY_MCP = "category:mcp"`, …). Fewest types, but `Enum.name` is taken, the wire value has to be split back apart, and adding a kind means touching every member.
- **`kind` + `name` with a validated pair table.** Keeps two fields and adds a third thing to keep in step with them. The version above deletes the question instead of answering it.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
totalfrank: I'd rather not have moved them either — but I don't think "add the rows here" is reachable, and I want to put the number in front of you rather than argue taste.

`admission.py` on `dev` is **exactly 1000 lines**, and `tests/community/architecture/test_no_oversized_modules.py` fails any `.py` over 1000 that is not on its allowlist. This PR adds four routes, so four rows, so 1004. Even written as tersely as possible with no comment, the file cannot take them.

`responses.py` is the same story, one comment down: also exactly 1000 on `dev`.

So the choices are:

1. **Keep the split** (what is here). `admission.py` keeps the seam, the modes and `ActingCaller` at 213 lines; `admission_table.py` is the table. `ADMISSION` is re-exported, so no importer changed.
2. **Revert the split and add both files to `_ALLOWLIST`** in the size guard. Two lines, and it works — but that list is documented as "files that were over the cap on the day this guard was introduced", with the rule that it only ever shrinks. This would be the first addition.
3. **Free the space somewhere else in each file.** Same amount of moving, just less related to what this PR is doing.

I went with 1 because it is the split the cap was already pointing at — the per-route table is the half that grows with every new route, and the seam below it does not — and because 2 spends a guard the repo has kept clean.

If you would rather have 2, say so and I'll do it in the same push: it is a smaller diff than this one, and it is your call whether that guard is worth keeping clean here.

One thing that should make review easier either way: the moved content is byte-identical, so `git log --follow` / `git diff -M` shows it as a rename plus my four rows, not as a rewrite.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/responses.py
totalfrank: Same answer as on `admission.py`, and the same number: `responses.py` is **exactly 1000 lines** on `dev`, and the 1000-line guard fails anything over that which is not allowlisted. The mappings this PR needs (three error rows, a business subcode, and the `data` hook that carries `violations`) do not fit.

The line I split on is the one the file was already describing: `envelope_error_table.py` is the inventory of *what a failure means to a caller* — the tables plus the ~180 error-class imports that exist only to name them, 176 of which nothing else in `responses.py` used — and `responses.py` keeps the response machinery at 312 lines. Both tables are re-exported, so every importer is untouched; the only edits outside the move were two **test** imports of `ClusterMismatchError`, which had been reaching for it through `responses` rather than its owning `errors` module.

Happy to revert this and allowlist the file instead if you'd prefer — see the options I listed on the `admission.py` thread, it's the same decision for both.

---

src/backend/src/agentclaw/community/core/repository/implementations/bot/config_manifest/repository.py
totalfrank: The three columns can't carry the key on MySQL: `entity_id` is `varchar(1024)`, which is **4096 utf8mb4 bytes on its own**, past InnoDB's 3072-byte index-key cap before `env` and `bot_id` are counted. `CREATE TABLE` is refused outright, so the table would simply not exist in production.

The repo already has a fallback for that — `core/schema.py::prepare_for_mysql` rewrites an over-long `UniqueConstraint` into a prefixed unique index — and its own docstring is the argument against using it here:

> A prefixed unique index enforces uniqueness on the prefix, which is *stricter* than on the whole value: two rows whose ids agree for the first N characters and differ later would now collide. The columns involved hold ids and short type discriminators, so this is not reachable in practice — but it is a real narrowing, and **the reason the convention for *new* tables (see `bot_startup_script`) is a bounded sha256 surrogate column instead**.

So the surrogate is the convention this table is following, not a flourish. And the narrowing matters more here than on the six tables that live with it: a prefix collision on *this* key means one bot reading and overwriting another bot's manifest, which once apply lands decides what gets installed in someone else's container.

Two things that keep the cost low: nothing outside the repository ever sees the key — it is computed from `(env, entity_id, bot_id)` on the way in and never appears in the API — and every read filters on it, so the table has exactly one index rather than a unique key plus a separate lookup key that could drift apart.

The one part genuinely worth reviewing is that the digest is **length-prefixed**, not delimiter-joined (`_key.py`): nothing validates `bot_id`/`entity_id` against control characters, so a NUL-joined key would let `entity_id="a\0b", bot_id="c"` and `entity_id="a", bot_id="b\0c"` hash to the same row. `ac_bot_startup_script` shipped that bug and fixed it; this starts from the fixed shape, and there is a test for it.

Leaving this thread open in case you'd still rather take the prefixed-index route.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/placeholders.py
totalfrank: No — the user never sends values. That is the part that makes this cheap rather than over-designed.

The whitelist is five variables the **platform** already knows: `BOT_ID`, `BOT_ENGINE_TYPE`, `BOT_ENV`, `BOT_TENANT` come off the bot record and the deployment, and `BOT_ARCH` is the constant `amd64` today. There is no place in the API for a caller to supply a value, and no `variables` block in the schema.

What they buy is one manifest across many bots. The motivating line is from the schema's own example:

```yaml
identity:
  - type: SOUL.md
    from: content
    subpath: bots/${BOT_ID}/soul.md
```

One document, fifty bots, each getting its own `soul.md`. Without substitution that is fifty documents differing in one path segment, and the "bump the `ref` once" workflow the manual is built around stops working.

Two things worth knowing before deciding to drop it:

- **It is already the published contract, and it did not land in this PR.** `manifest-schema.zh-CN.md` §4 and `user-manual.zh-CN.md` §4.3 + appendix E both document `${BOT_*}` — they came in with #1720. W1's acceptance criteria also require this surface to *reject* unknown `${...}` and to implement `BOT_ARCH` now rather than reserve the name, so that a future mixed fleet changes where the value comes from and nobody rewrites their documents.
- **What this PR contains is small**: a five-name whitelist, a scanner that refuses anything else at `PUT`, and a `resolve()` that nothing calls yet — substitution happens at apply (W4/W5). The scanner deliberately skips inline `content` and `script.body`, because a knowledge file or a shell script may legitimately contain `${...}` of its own.

If you do want them gone it is a contract change, not a code deletion: schema §4, the manual's §4.3 and appendix E, and W1's criteria would all need to say so. Happy to do that as a follow-up if you'd rather have a smaller v1 — I just don't want this PR to quietly diverge from what the docs on `dev` promise callers.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/repository/models.py
totalfrank: Confirmed and fixed in `5331d93b`. I checked each step rather than taking the report on trust:

- `create_schema: bool = True` — `di/config_community.py:65`. ✓
- `MAX_DOCUMENT_BYTES` is 65,536; MySQL `TEXT` holds 65,535 — one byte less. ✓
- The check is `if len(encoded) > MAX_DOCUMENT_BYTES`, so exactly 64 KiB is accepted. ✓

Took the suggested fix — `Text().with_variant(mysql.MEDIUMTEXT(), "mysql")`, matching the `MediumText` pattern in `core/models/space_skill.py`. The out-of-band DDL already said `mediumtext`, so the ORM was the half that had diverged; now they agree.

Added a test, since you're right that the boundary is invisible on SQLite. It asserts two things rather than one, because the type alone would not have caught this: that `MAX_DOCUMENT_BYTES` still exceeds `TEXT`'s capacity (so if someone later lowers the limit, the test says to revisit the reasoning instead of silently passing), and that the compiled MySQL DDL for the column really is `MEDIUMTEXT`.

Left `ac_bot_startup_script.script` alone — the same divergence, still dormant at 24 KiB, and widening this PR to it is not what it should do.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission_table.py
totalfrank: Answered on the `admission.py` thread — same decision for both files, and it comes down to a number: `admission.py` is exactly 1000 lines on `dev` and the size guard fails anything over that, so the four rows this PR needs cannot go there.

If you'd rather revert the split and allowlist the file instead, say so on either thread and I'll do both in one push.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/capabilities.py
totalfrank: Confirmed in the current code: `ConstructKind`, `ManifestCategory`, `ManifestSection`, and `SourceForm` are `StrEnum`s in `capabilities.py`, `kind_of()` derives kind from the value's type, and the wire shape is unchanged. Resolving.

---

totalfrank (commented):
Re-checked after this `synchronize` event. Head (`5331d93b`) is the same commit my last review round was already posted against — no new commits to review, so this pass only re-verified open findings against the current code.

**Findings checklist (12 total):**
- `feature_flag.py` env-var switch — fixed (switch deleted)
- `config_manifest.py` feature-flag gate — fixed (switch removed, group ships enabled)
- `config_manifest_support.py` `Any` typing — fixed (`Optional[BotConfigManifestRecord]`)
- `admission.py` file split — open, author's call (position held once, awaiting decision)
- `admission_table.py` file split — open, author's call (same decision as above)
- `authorization.py` collaborator access — fixed (`Check(MEMBER)`/`Check(ADMIN)` split added, tested)
- `responses.py` file split — open, author's call (position held once, awaiting decision)
- `repository.py` manifest-key column — open, author's call (tradeoffs explained, awaiting decision)
- `capabilities.py` `ConstructKind` as enum — fixed
- `capabilities.py` kind/name correlation → per-kind enums — fixed and verified against current code this round (was unresolved after implementation; now resolved)
- `placeholders.py` `${BOT_*}` substitution — open, author's call (design question, not a defect)
- `repository/models.py` `TEXT`/`MEDIUMTEXT` 64 KiB boundary — fixed (`MEDIUMTEXT` variant + boundary test)

**Still open: 5**, all awaiting the author's decision rather than further action on my side (two file-split questions that resolve identically, the manifest-key surrogate-column question, and the placeholder-substitution scope question). No new defects found this round.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/capabilities.py
totalfrank: Confirmed in `e2932ae`: `capabilities.py` now defines `ManifestCategory`, `ManifestSection`, `SourceForm` as separate `StrEnum`s with `Construct = ManifestCategory | ManifestSection | SourceForm` and `kind_of()` deriving `kind` from the value's type, exactly as described. Verdict table is built by iterating the enums (`_all_constructs()`), the parser's only string→enum boundary is `parse_category`, and the wire shape is confirmed unchanged by `test_the_payload_is_flat_and_names_what_it_answered_for`. Resolving.

---

totalfrank (commented):
Follow-up round. Reviewed the one new commit since the last round: `e2932ae` (`04f14d3c..e2932ae`) — deletes the feature switch, collaborator-scopes the four routes (MEMBER read / ADMIN write via `GRANT_CHECKED_ADDRESSED_BOT` + `Check(...)`), splits construct kind/name into `ManifestCategory`/`ManifestSection`/`SourceForm` enums, and types the stored record as `Optional[BotConfigManifestRecord]`.

**Part A — verifying prior findings**

| # | Thread | Status |
| - | - | - |
| 1 | `feature_flag.py` — no system env var | Fixed (already resolved) |
| 2 | `config_manifest.py` — drop the switch | Fixed (already resolved) |
| 3 | `config_manifest_support.py` — `Any` → typed record | Fixed (already resolved) |
| 4 | `admission.py` — file split, don't move rows | Open — author's call, unaddressed this round |
| 5 | `admission_table.py` — same split question | Open — author's call, unaddressed this round |
| 6 | `authorization.py` — Check(collaboration) | Fixed (already resolved) |
| 7 | `responses.py` — same split question | Open — author's call, unaddressed this round |
| 8 | `repository.py` — why the `manifest_key` surrogate | Open — author's call, unaddressed this round |
| 9 | `capabilities.py` — prefer enum | Fixed (already resolved) |
| 10 | `capabilities.py` — kind/name correlated, how to lay out | **Fixed this round** — `ManifestCategory`/`ManifestSection`/`SourceForm` + `kind_of()` now live in `e2932ae`, matching the design proposed; resolved with confirmation |
| 11 | `placeholders.py` — why keep `${BOT_*}` substitution | Open — author's call, unaddressed this round |
| 12 | `repository/models.py:78` — **Blocking**, `document` column is `Text` (65,535-byte MySQL cap) while `MAX_DOCUMENT_BYTES` (65,536) sits exactly on top of it | **Still open** — code at that location is unchanged in `e2932ae`; the commit message doesn't claim to touch it. Not re-posting; leaving silent per the unaddressed-and-unclaimed case. |

**Part B — new commit**

No new findings. Verified: the `OwnerIdDep`/`GRANT_CHECKED_ADDRESSED_BOT` wiring matches the existing addressed-bot pattern used by other shared-bot groups (owner defaults to caller when `owner_id` is omitted, so existing owner-only callers are unaffected); `audit_actor` now correctly stamps the acting caller, not the addressed owner; grepped the tree for every removed symbol (`KIND_CATEGORY`, `SOURCE_*`, `SECTION_SCRIPT`, `CATEGORIES`, `feature_flag`, `config_manifest_surface_enabled`, `ConfigManifestSurfaceDisabledError`) — no dangling references; `Capability.find`/`schema/entries.py` use `is` identity comparison on enum members (correct, avoids cross-enum value collisions); `_all_constructs()` yields exactly the 11 members the `blocked` dict covers; `ManifestConstruct(**item.as_dict())` field names match the pydantic schema.

**Still open: 5** (4 author-decision threads + 1 Blocking defect)

---

totalfrank:
Two automated review rounds landed out of order and disagree about the blocking finding, so for anyone reading the summaries: **it is fixed.**

The round reporting it "Still open" scoped itself to `04f14d3c..e2932ae` — the commit *before* the fix. The `document` column carries the `MEDIUMTEXT` variant as of `5331d93b`, verified on the current head:

```
$ git log --oneline -1 -S MEDIUMTEXT -- .../bot_config_manifest/repository/models.py
5331d93b fix(backend): hold a 64 KiB manifest on a self-provisioned MySQL schema
```

The later round, which read `5331d93b`, agrees and lists it as fixed. Its thread is resolved.

**Actually open: 4 threads, all awaiting @totalfrank-personal's decision** — none of them a defect:

1. **`admission.py` / `admission_table.py` split** (one decision, two threads) — both files are exactly 1000 lines on `dev` and the size guard fails anything over that, so four route rows cannot go in. Keep the split, or revert it and allowlist the files. Either way I'll do it in one push.
2. **The `manifest_key` surrogate** — `entity_id` alone is 4096 utf8mb4 bytes, past InnoDB's 3072 cap, so the three columns cannot carry the key. The alternative is a prefixed unique index, which narrows uniqueness to a prefix.
3. **`${BOT_*}` substitution** — values are platform-injected, never caller-supplied, and it is already the published contract on `dev`. Removing it is a doc change across schema §4, the manual, and W1's criteria.

Head is now `c997a0b6`, a clean merge of `dev` that picks up #1721's `bot_config_surface` seam. Re-verified on the merged tree: architecture + framework 377 passed, endpoints + route-inventory gates 1473 passed, manifest module + repository 130 passed.

---

totalfrank (commented):
Re-checked after this `synchronize` event. Head is `c997a0b6` — a merge of `origin/dev` into this branch (base moved to `2c991676`, which is now the merge-base).

Diffed every PR-owned path (`core/bot_config_manifest/`, `adapters/.../bots/config_manifest*`, `admission.py`/`admission_table.py`, `responses.py`/`envelope_error_table.py`, `core/repository/implementations/bot/config_manifest/`, and their tests) from `5331d93b` (the last commit that touched this PR's own code) through to `c997a0b6`: **zero changes**. The commits in between — `2c991676`, `747e6198`, `e7e1ebb7`, `04745a13`, `2ab027fe` — are `dev`'s own work (bot-config-surface seam, bcs chat-runs, aicoding restart fix, etc.) absorbed by the merge; none touch this PR's scope, and the merge resolved without conflict in any PR-owned file. No new commits to review.

**Findings checklist (12 total):**
- `feature_flag.py` — env-var switch → fixed, resolved
- `config_manifest.py` — feature-flag gate → fixed, resolved
- `config_manifest_support.py` — `Any` typing → fixed, resolved
- `admission.py` — file split → open, author's call
- `admission_table.py` — file split → open, author's call
- `authorization.py` — collaborator access → fixed, resolved
- `responses.py` — file split → open, author's call
- `repository.py` — `manifest_key` surrogate column → open, author's call
- `capabilities.py` — `ConstructKind` as enum → fixed, resolved
- `capabilities.py` — kind/name correlation → per-kind enums → fixed, resolved
- `placeholders.py` — `${BOT_*}` substitution scope → open, author's call
- `repository/models.py` — `TEXT`/`MEDIUMTEXT` 64 KiB boundary (Blocking) → fixed, resolved

**Still open: 5**, all design questions where I already held my position once and am awaiting the author's decision — no author reply since, nothing to re-post. No new defects found this round.

---

totalfrank:
**Singlebox coverage** failed on `c997a0b6`. It is not this PR's failure, and I have re-run it once to confirm.

The job's actual work all passed — the only `##[error]` in its 700-line log is the artifact upload at the very end:

```
##[notice]E2E line coverage 43.09% (requirement: >=40.00%) — OK
##[notice]E2E method coverage 38.62% (requirement: >=36.00%) — OK
##[notice]E2E coverage gate PASSED.
##[notice]Adapter endpoint coverage gate passed: 100.0% (150 / 150)
##[notice]bcs-cli command coverage gate passed: 100.0% (52 / 52)
singlebox coverage gate passed
...
##[error]Failed to CreateArtifact: Failed to make request after 5 attempts:
  Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact
```

Every gate reported OK and the coverage gate passed; the step that failed was uploading 2,676 artifact files, which timed out against GitHub's artifact service after five attempts. That names a GitHub service, not this diff — and this PR touches nothing outside `src/backend/{src,tests,docs}`:

```
$ git diff --name-only 2c991676...HEAD | grep -vE '^src/backend/(src|tests|docs)/'
(nothing)
```

No fix to port — there is no repo-side change that makes a GitHub artifact-service timeout stop happening, and this is the one re-run the flake rule allows. If it fails the same way again I will treat it as real and dig further rather than re-running.

---

**Unrelated, noticed while reading the log** — not the cause of this failure, and not mine to fix in this PR, but worth someone's attention:

```
[ERROR] --local has been removed. singlebox.sh now always uses isolated standalone paths.
[ERROR] Run ./scripts/singlebox.sh without --local.
```

`.github/workflows/singlebox-coverage.yml:177,185` still calls `bash scripts/singlebox.sh --local status all` and `--local stop all`, while `scripts/singlebox.sh:848` now rejects `--local`. Both call sites end in `|| true`, so the job does not fail — but those status and stop steps are silently doing nothing, which is why the run left ~26 orphan `openclaw`/`python` processes for the runner to reap. Happy to open a separate PR dropping the two flags if that is useful.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
totalfrank-personal: Is the only reason being we have a test enforces that? If so, I think this can be an exception.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission_table.py
totalfrank-personal: Is the only reason being we have a test enforces that? If so, I think this can be an exception.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/responses.py
totalfrank-personal: Is the only reason being we have a test enforces that? If so, I think this can be an exception.

src/backend/src/agentclaw/community/core/repository/implementations/bot/config_manifest/repository.py
totalfrank-personal: entity_id does not have to be varchar(1024), varchar(64) is sufficient.

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/placeholders.py
totalfrank-personal: Already talked to you offline, let's remove BOT_ID. Other place holder stays.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
totalfrank: Yes — that is the only reason. There is no runtime or tooling constraint behind 1000 lines; it is a convention `tests/community/architecture/test_no_oversized_modules.py` enforces, with an `_ALLOWLIST` for files that were already over it.

Taken the exception, in `fc800cd8`. Both splits are reverted: `admission_table.py` and `envelope_error_table.py` are gone and the tables are back where they were, with both files added to `_ALLOWLIST` and a comment recording that they are per-operation inventories — they cross the cap by growing with the surface, not by taking on a second concern.

Two things beyond the straight revert, both because you were right that the split was moving existing code around:

- Reverting mechanically left `responses.py` worse than dev had it: a duplicated `ManifestValidationError` import, `SkillOfflineBlockedError` relocated out of its original block, `JSONDecodeError` moved out of the stdlib group, and four existing single-line imports exploded into parenthesized ones. I rebuilt that file from `dev` and re-applied only the four real additions instead.
- The two test imports of `ClusterMismatchError` that the split had redirected are back to `from ...openapi_v1.responses import ClusterMismatchError`.

So both files now diff against `dev` as pure additions — `admission.py` +20 lines, `responses.py` +43 — and no existing line moves.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/placeholders.py
totalfrank: Done in `fc800cd8` — `BOT_ID` is out, the other four stay.

You were right, and my earlier defence of it was wrong on the facts. `generate_bot_id()` mints a bot id at creation time as a date plus eight random characters, and public `BotCreate` has no `bot_id` field — so the caller cannot choose it, and an author preparing content in a git repository has no way to know it. A document interpolating `${BOT_ID}` would have to be written *after* the bot existed and could then only ever describe that one bot, which is the opposite of what substitution is for. The remaining four are all properties of the fleet, which is what actually makes a document reusable.

What changed:

- `ALLOWED_PLACEHOLDERS` drops it, and `resolve()` drops its `bot_id` argument.
- Docs updated so the published contract matches: `manifest-schema` §4, `user-manual` §4.3 and appendix E, `examples`, and both `work-items` files. Every `bots/${BOT_ID}/soul.md` example is now a literal path, since a per-bot path belongs in that bot's own manifest.
- A test pins that `${BOT_ID}` is refused like any other unknown name, so it cannot come back by accident.

`resolve()` itself stays — it is still the write-path/apply-path contract for the four that remain, even though nothing calls it until W4/W5.

---

src/backend/src/agentclaw/community/core/repository/implementations/bot/config_manifest/repository.py
totalfrank: You're right, and it undercuts my whole argument. I treated `ac_bots.entity_id`'s 1024 as a fixed input for a new table — it isn't, this table picks its own widths. Surrogate removed in `fc800cd8`.

The key is now the logical one, `UNIQUE(avernet_tenant, env, entity_id, bot_id)`. `_key.py` is deleted, and with it the whole length-prefixing argument — it only ever existed because the key was being encoded into one value, and there is nothing to encode now. The repository filters on the three columns through one `_match()` helper so get/upsert/delete cannot drift apart.

**One deviation worth your call: I used `varchar(256)`, not 64.** Both fit — at 64 the key is 1616 bytes, at 256 it is 2384, against InnoDB's 3072 — so your point stands either way. I went wider because this column copies a value we don't originate: `entity_id` comes off the bot record, and the only bound anyone actually enforces on it is the platform's own `max_length=1024` (`adapters/http/access/schemas.py:41`). Real values are short (`u_165137`), and 256 is what the newer tables here give a user id (`bot_collaborator.user_id`, `access.user_id`, `task.owner_user_id`). On MySQL an over-long value fails the write in strict mode, and SQLite ignores `VARCHAR` widths entirely, so the local suite would never surface it — the headroom costs 688 bytes of a budget that has 688 to spare. Say the word and I'll take it to 64.

Two tests carry the reasoning now:

- The index-budget test computes the key's width from the model, so widening `entity_id` back toward its source fails rather than producing a table `CREATE TABLE` would refuse. It also asserts the key names the four logical columns, so reintroducing a surrogate has to be deliberate.
- The colliding-boundaries case stays and now passes structurally rather than by construction: `(entity_id="a\0b", bot_id="c")` and `(entity_id="a", bot_id="b\0c")` are separate rows because the columns are separate. If the key is ever encoded again, that test is what fails.

`_key.py`'s removal left the repository package holding a single module, so it's collapsed to `config_manifest.py` beside `startup_script.py`. Note this makes the two tables differ: `ac_bot_startup_script` still hashes the same logical key into `script_key`. It didn't have to either — but changing it is a migration on an existing table, so I left it alone rather than widening this PR.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission_table.py
totalfrank: Same answer as on the `admission.py` thread: yes, the 1000-line cap is only enforced by `test_no_oversized_modules.py`, so I've taken the exception. This file no longer exists in `fc800cd8` — the table is back in `admission.py`, which is allowlisted with the reasoning.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
totalfrank: Verified on `fc800cd8`: `admission_table.py` is deleted, `admission.py` is 1020 lines (matches its own allowlist entry), and the diff against the pre-split state is a clean addition. Resolving.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission_table.py
totalfrank: Moot as of `fc800cd8` — `admission_table.py` no longer exists; its rows are back in `admission.py`, which is now allowlisted in the size guard with the reasoning recorded. Resolving.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/responses.py
totalfrank: Moot as of `fc800cd8` — `envelope_error_table.py` no longer exists; its rows are back in `responses.py` (1041 lines, matches its allowlist entry), and the diff against the pre-split state is a clean addition. Resolving.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/placeholders.py
totalfrank: Verified on `fc800cd8`: `ALLOWED_PLACEHOLDERS` in `placeholders.py` now holds exactly `BOT_ENGINE_TYPE`, `BOT_ENV`, `BOT_TENANT`, `BOT_ARCH`; `resolve()` no longer takes a `bot_id` argument; `test_bot_id_is_not_a_placeholder` pins the rejection; and the schema/manual/examples/work-items docs consistently describe `${BOT_ID}` as refused, with no stray reference left implying it still resolves. Resolving.

---

src/backend/src/agentclaw/community/core/repository/implementations/bot/config_manifest/repository.py
totalfrank: Verified on `fc800cd8`: the surrogate is gone — `_key.py` deleted, `UniqueConstraint("avernet_tenant", "env", "entity_id", "bot_id")` on the model, `_match()` filters on the three columns, and the DDL/model/DI comments all agree on the new key. The budget arithmetic checks out (64+20+256+256 = 596 chars × 4 = 2384 bytes, 688 under the 3072 cap), and `test_the_uniqueness_key_fits_innodbs_index_limit` now also pins the constraint's column set so a surrogate can't come back silently.

That resolves the original question (three columns are enough, no digest needed). The one still-open piece is your own follow-up — `entity_id varchar(256)` vs. the `varchar(64)` you suggested — which is a width choice, not a defect either way; leaving that open for @totalfrank-personal's call rather than resolving on your behalf.

---

totalfrank (commented):
Reviewed the one new commit since the last round: `fc800cd8` (`c997a0b6..fc800cd8`) — reverts both file splits, drops the `manifest_key` surrogate for a direct 4-column unique constraint, and drops `${BOT_ID}` from the placeholder whitelist. All three were open author's-call threads; all three are now resolved with the code verified to match what was claimed (see replies on each thread).

**Part B — no new defects.** Checked: `admission.py`/`responses.py` line counts (1020/1041) match their new allowlist entries and the diffs against the pre-split state are clean additions with no accidental collateral change; `admission_table.py`/`envelope_error_table.py` fully removed with no dangling references; the new `UniqueConstraint(avernet_tenant, env, entity_id, bot_id)` and its index-budget arithmetic (2384 of 3072 bytes) check out and match the DDL/model/DI comments, which all agree with each other; `_key.py`/its package `__init__.py` removed with no stray imports left (grepped the tree); the collision test moved from a unit-level digest check to an integration-level repository test that verifies the same property structurally; the `${BOT_ID}` removal is consistent across code, four doc files, and a new pinning test. No test coverage gaps or missed callers.

**Findings checklist (12 total):**
- `feature_flag.py` env-var switch — fixed
- `config_manifest.py` feature-flag gate — fixed
- `config_manifest_support.py` `Any` typing — fixed
- `admission.py` file split — **fixed this round** (reverted per reviewer's request, file allowlisted)
- `admission_table.py` file split — **fixed this round** (moot — file deleted)
- `authorization.py` collaborator access — fixed
- `responses.py` file split — **fixed this round** (reverted per reviewer's request, file allowlisted)
- `repository.py` manifest-key surrogate — **fixed this round** (surrogate dropped, 4-column unique constraint); one sub-question left open — `entity_id varchar(256)` vs. the `varchar(64)` suggested — awaiting @totalfrank-personal's call
- `capabilities.py` `ConstructKind` as enum — fixed
- `capabilities.py` kind/name correlation → per-kind enums — fixed
- `placeholders.py` `${BOT_*}` substitution — **fixed this round** (`BOT_ID` removed per offline agreement, other four kept)
- `repository/models.py` `TEXT`/`MEDIUMTEXT` 64 KiB boundary (Blocking) — fixed

**Still open: 1** — the `entity_id` width sub-question, a design choice rather than a defect, awaiting the author's decision.

---

FreddieSun (approved):
LGTM

PR #1735, opened 2026-08-31
chore(release): sync dev into REL20260901

evilcoda:
## Summary

Sync the latest dev changes into the REL20260901 release branch while preserving the release-only BCN principal callback change (#1730).

## Scope

- bring the five current dev-only commits into the release branch
- preserve the existing release-only commit
- no direct rewrite of either protected branch

## Follow-up

After merge, verify the internal Avernet mirror and update the OCB gitlink to the mirrored release HEAD.


evilcoda:
本次 REL20260901 不包含这些 dev 变更，关闭该同步 PR。

PR #1741, opened 2026-08-31
fix(backend): materialize SC public manifest identity

FreddieSun:
## Summary
- accept an opaque single-directory wrapper only for SC Public exact ZIP downloads while preserving all shared package safety and metadata validation
- freeze `ac_skill.name` and runtime identity from exact `SKILL.md.name`, never from SC market display metadata
- safely converge only unconsumed, first-time MATERIALIZING public assets; PUBLISHED, Space-owned, Membership, and Installation states fail closed
- preserve the frozen manifest name across later exact SC Public versions

## Validation
- 69 focused package/materializer/reference/repository/integration tests passed
- 21 contract/architecture/DI tests passed
- Ruff and `git diff --check` passed
- Standards/Spec dual-axis review completed; initial V2-name P1 fixed and re-review has no remaining P0-P2
- Backend full local CI was stopped at user request; GitHub CI is the authoritative full gate

## Scope
- target branch: `REL20260901`
- no OpenAPI/router change; generated Gateway schema is unaffected
- `git_path` length-limit mismatch is explicitly deferred
- no DDL change


totalfrank (approved):
LGTM.

PR #1742, opened 2026-08-31
fix(backend): log safe skill materialization stages

FreddieSun:
## Problem

SC Public Reference materialization collapses all Ready-Gate failures into `MATERIALIZATION_FAILED`. Existing exception logging can expose a presigned package URL and does not provide a safe, queryable failure stage.

## Solution

- Carry a finite materialization stage on `SkillVersionMaterializationError`.
- Emit only `reference_id`, `skill_code`, `skill_version_id`, `failure_stage`, and exception type.
- Do not log exception text or traceback, so presigned URLs and their signatures cannot enter application logs.

## Validation

- Focused materializer/reference processor tests: 16 passed locally.

## Compatibility and risk

No API, database, or lifecycle-state change. Reference items still expose the existing stable public error. Logging becomes safe structured diagnostics for operator investigation.


totalfrank (approved):
LGTM.

PR #1732, opened 2026-08-31
feat(backend): W2-add guarded fetch and unpack pipeline for manifest sources

lucas-xzp:
## Problem

Manifest entries fetch caller-supplied URLs and unpack archives on the platform's behalf, and the backend has **no SSRF defense today** — new machinery, not reuse (issue #1470, work item W2; zero dependencies, first deliverable of the machine-builder track).

## Solution

`core/bot_config_manifest/fetch/` — the single funnel every manifest byte passes through:

- **guarded_fetcher.py** — five defense layers in order: URL shape → full-resolution public-address validation (loopback/link-local/ULA/multicast-with-explicit-refusal/reserved are all refused; `is_global` alone does not cover multicast) → **pinned connection** to the validated address with original Host header + SNI preserved (`httpx sni_hostname`, verified against installed httpcore: server_hostname carries SNI *and* certificate verification) → per-hop redirect re-validation (shape, addresses, authorization policy, hop budget 5) → streaming byte cap counted as bytes arrive (a lying Content-Length never passes) with the sha256 computed on the same pass.
- **A fresh `httpx.Client` per hop**, deliberately: httpcore keys its pool on the dialed origin, which under pinning is the validated IP — a shared client would hand a later request another name's verified TLS connection. The batch review caught this (its HIGH finding); the engine precedent's per-request client confirms the same shape.
- **The deployment transport allowlist is `application.yaml` config, never an env var** (review R1): `user_config.bot_config_manifest.fetch_transport_allowlist` — neutral empty default, a deployment's mirrors come from its overlay — parsed by `limits.transport_allowlist_from_config` (absent block = no exception; malformed block = `ValueError` at its reader) and constructor-injected as `GuardedFetcher(transport_allowlist=…)`, so core never touches `os.environ` (AGENTS.md keeps raw env access in config loading / composition roots).
- Digest mismatch is a fetch failure, never a corrupted success; transport failures classify as `FetchFailedError`; credentials-injector and authorization-policy protocols declared for W3 to bind.
- **unpack.py** — zip/tar.gz through one code path: traversal/absolute/drive-letter name rules, all link-type members refused (a deliberate hardening beyond the issue's escapes-only minimum, with the rationale in the module README), device/special members refused, per-member streaming with the running byte cap, exact `strip_components` (no top-dir detection magic), permissions flattened to 0644/0755, and a refusal that leaves no half tree.
- `limits.py` — schema §5 numbers as one source, plus the yaml-borne allowlist parsing.

[...]

PR #1746, opened 2026-08-31
feat(backend): W3-add tenant-level source credentials with prefix authorization

### LINKED ISSUE #1471: [配置清单][W3] 租户级源凭证
**第一波。**完整范围与验收标准见 `src/backend/docs/bot-config-manifest/work-items.zh-CN.md` §5, W3（规划 PR #1465）。

**负责人：**`lucas-xzp` · 0.75 天 · 第 1–2 天 · **实现**（§7）

**目标。**租户注册一次命名凭证，并在多份 manifest 里引用它；平台在拉取时出示它 —— 而这个秘密永远不能被读回、不被日志记录、不进入容器。

**依赖：**— · **阻塞于：**—（X1 已关闭：专用机器账号 + `read_repository` 访问令牌，以 HTTP Basic 注入，v1 凭证模型无需改动）

## 关键验收标准

- 租户级表，键为 `(avernet_tenant, name)`。`name` 是自由标识符，与 `allowed_prefixes` 里的主机没有派生关系。
- 秘密通过既有的 `TokenVault` **可逆加密**存储（`enc:v1:`，经 `utils/secret_utils.symmetric_encrypt` 的 AES-GCM），主密钥来自 `SecretResolver`。**不引入新的密码学。**
- **新增的 fail-closed 守卫：**在生产 profile 下，若主密钥解析不出来，凭证写入必须被**拒绝**。`TokenVault` 今天会退化成无前缀明文 —— 这对 singlebox/CI 是对的，在这里不可接受：一次密钥库配置错误就会让所有租户的令牌明文躺着。
- `allowed_prefixes` 必填，至少 1 项，每项都是绝对的 `https` 前缀。
- 匹配按**路径段边界**：目标必须等于前缀，或以「前缀 + `/`」开头。`…/team/content` 不得授权 `…/team/content-secret`。要覆盖整个 origin，必须显式写 `https://host/`。
- 目标落在所有前缀之外时，该条目**失败** —— 绝不是「不带凭证继续」。
- 离开授权前缀的重定向**失败**；凭证既不跟着跨过去，也不会被剥掉后继续跟随。
- `GET` 在任何路径下都只返回脱敏元数据（`has_secret`、`header_name`、`allowed_prefixes`、`updated_at`）。
- 保留类型 `oss_aksk` 与 `basic` 在写入时拒绝，让判别字段从第一天起就是真的。
- 该值不出现在任何日志、错误消息或 apply 报告里 —— 只有名字会出现。
- **认证失败（401/403）报告为「凭证 `<name>` 被拒绝」**，具名且与通用拉取错误区分。令牌过期由 owner 自管（§4, X1），所以一条读起来像网络问题的轮换提醒是真实的运维故障。
- 轮换就是对同一个名字重新 `PUT`，且不触发 apply。

**另外：**绑定 [W2] 的注入端口。

## 范围外

`header` 之外的任何凭证类型。apply 对它的任何使用。

**规模：**中等。

lucas-xzp:
## Problem

Manifest entries reference private sources the platform fetches on a tenant's behalf (#1471, W3). The secret must live somewhere the platform can present per fetch hop — but never inside a manifest (GET-able, audit-visible), a script body, a URL, or a container — and never in plaintext at rest under a production profile. The backend had no SSRF-adjacent credential machinery at all; this is new plumbing, zero dependencies (W2's guarded fetcher lives in #1732 — the binding duck-satisfies its injector/policy protocols, whichever lands first).

## Solution

`core/bot_config_manifest/credentials/` — named, tenant-level credentials:

- **Storage**: one row per `(avernet_tenant, name)` — deliberately no env axis (a credential is a tenant-level object; whether prod shares one git token is the tenant's choice, not the schema's). Rotation is a same-name re-PUT: whole-replace in a single transaction, and the binding re-reads per fetch hop, so "next fetch uses the new value" needs no signal.
- **Encryption**: `TokenVault` reused as-is (`enc:v1:` AES-GCM, `SecretResolver`-resolved master key) — no new cryptography. **Fail-closed guard**: the DI provider binds a fail-closed posture for the corp/community columns — when no master key resolves, the write is refused *before* persistence (loud 503), because `TokenVault`'s empty-key plaintext passthrough (right for singlebox) must never catch tenant tokens in production.
- **Prefix authorization** (`policy.py`): `allowed_prefixes` — absolute `https` prefixes, matched on path-segment boundaries with decode-then-normalize on both sides (double encoding is an equivalent spelling, not an escape; `%2F`-smuggling and `..` pop are refused; scheme/host/port compared exactly, userinfo/query/fragment refused). `…/team/content` never authorizes `…/team/content-secret`. The structural `CredentialInjector`/`AuthorizationPolicy` pair binds W2's fetcher seams per hop.
- **Masked reads by construction**: the storage-shaped `Row` and the public `Record` are two different classes — the value physically cannot end up in the published shape. GET answers `has_secret`/`header_name`/`allowed_prefixes`/`updated_at`; reserved mechanism types (`oss_aksk`, `basic`) are refused at write so the stored type stays true; errors and logs carry the *name*, never the value (the audit actor on PUT is composed from the verified principal — the same rule the bots group uses, never the body).
- **Public surface**: four `/openapi/v1/source-credentials` routes, REFUSED admission end-to-end (route dependency + admission table + gateway `user: required`) — a secret write has no business running headless. The prefix is a new declared gateway domain (spec artifact generated from the backend, README registered). PUT names its user; the collection read/delete carry no user dimension (the tenant guard scopes every row).
- Deleting a still-referenced credential is allowed: the referencing entries fail their next fetch *with the name* — the failure record belongs to the apply layer (W4), storage never guesses the reference graph.

## Validation

- 56 credentials tests + 9 endpoint-framework cases, all green: prefix boundary matrix (segment look-alike, double-encoding, dot-segment pop, port/host case, traversal family), fail-closed both postures (refusal leaves an empty table, not a plaintext row), masked read assertions ("secret" not in data), rotation-in-place with a single row, per-hop re-read after rotation and delete (named 404), error map ordering (subclasses before the family base — isinstance-first-match is order-sensitive and tested).
- Full backend suite: **16096 passed, 58 skipped**; line coverage **88.07%**; changed-line coverage **98.62% (358/363, ≥80%)** — `scripts/ci_test.sh` reproduced locally post-commit (per the repo rule; untracked files show N/A before the commit).
- Architecture gates all green: service-api conformance (`_PAIRS`), E3 module exemption with drain condition, principal seam + admission inventory agreement (REFUSED rows), endpoint coverage gate (happy+error per route; the cannot-fail routes use the reachable pre-handler 401 as their honest error shape), public-namespace invariant (new domain declared with gateway config), explicit-user-id pins updated (+PUT, reads exempt), oversized-module allowlist entries with follow-up notes.
- Batch review before commit: 8 findings fixed inline — the important one (audit `modifier` never populated on the only production write path → wired to the verified principal's actor, with a column assertion), plus a vacuous `hasattr` dict assertion, published docstring stray quotes, repository type annotations (Model→Row), a no-op except branch, `header_name` length unvalidated at the boundary, a comment self-comparison, and a formatting gap.

## Spec

`src/backend/specs/2026-08-31-w3-source-credentials/` — issue acceptance mapping, execution log, and the merge-order notes versus W1/W2 (same-file registrations in admission/responses/_PAIRS/flow_coverage; whichever merges last rebases and keeps both sides' rows).

Deploy note: the gateway-side REFUSED prefix rule requires the mirrored entry in the ocb repo's gateway config (the repo's dual-write convention for REFUSED routes — avernet side landed here).

Closes #1471.


src/backend/src/agentclaw/community/core/repository/protocols/bot/source_credential.py
```diff
@@ -0,0 +1,53 @@
...
+    name rotates the value in place.
+    """
+
+    @abstractmethod
+    def get(self, *, name: str) -> Optional[SourceCredentialRow]:
+        """The row for this tenant's ``name``, or ``None``."""
+        ...
+
+    @abstractmethod
+    def list(self) -> "list[SourceCredentialRow]":
```
totalfrank: Let's not do forward reference here, I believe you don't need to(get method above does not do that).

src/backend/src/agentclaw/community/core/repository/protocols/bot/source_credential.py
```diff
@@ -0,0 +1,53 @@
...
+        """Every row in the tenant, ordered by name."""
+        ...
+
+    @abstractmethod
+    def upsert(
+        self,
+        *,
+        name: str,
+        credential_type: str,
```
totalfrank: Can we make this enum? It's not that intuitive for it's to be just a string.

src/backend/src/agentclaw/community/core/repository/protocols/bot/source_credential.py
```diff
@@ -0,0 +1,53 @@
...
+        header_name: str,
+        allowed_prefixes: list[str],
+        secret_ciphertext: str,
```
totalfrank, line 45: Let's have docs for these. It's not intuitive what these are.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service_protocol.py
```diff
@@ -0,0 +1,77 @@
...
+    ) -> "SourceCredentialRecord":
+        """Validate and store-or-rotate. Raises on the first invalid input.
+
+        Rotation is the same call — same name, new value — and never
+        triggers an apply; presentation happens per fetch hop.
+        """
+        ...
+
+    @abstractmethod
+    def get(self, *, name: str) -> "SourceCredentialRecord":
```
totalfrank: Same here, you don't need forward reference. Let's get rid of it.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service_protocol.py
```diff
@@ -0,0 +1,77 @@
...
+        """Masked metadata for one name; 404-shaped error when absent."""
+        ...
+
+    @abstractmethod
+    def list_credentials(self) -> "list[SourceCredentialRecord]":
```
totalfrank: Same here, you don't need forward reference. Let's get rid of it.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service_protocol.py
```diff
@@ -0,0 +1,77 @@
...
+        """Remove by name; idempotent (``False`` when absent).
+
+        Deleting a referenced credential is allowed: the referencing
+        entries fail their next fetch with the name, which is the
+        apply-layer record, not a storage constraint.
+        """
+        ...
+
+    @abstractmethod
+    def binding(self, *, name: str) -> "SourceCredentialBinding":
```
totalfrank: Same here, you don't need forward reference. Let's get rid of it.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service.py
```diff
@@ -0,0 +1,197 @@
...
+            raise MasterKeyUnavailableError(
+                "credential storage requires a configured master key in this profile"
+            )
+
+        row = self._repository.upsert(
+            name=name,
+            credential_type=credential_type,
+            header_name=header_name,
+            allowed_prefixes=allowed_prefixes,
+            secret_ciphertext=self._vault.encrypt(secret),
```
totalfrank, line 132: The _vault we are wiring here is a ai coding thing. We should not rely on them. But feel free to leave it here. I will follow up to fix it.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service.py
```diff
@@ -0,0 +1,197 @@
...
+    def get(self, *, name: str) -> SourceCredentialRecord:
+        return self._masked(self._row_or_404(name))
+
+    def list_credentials(self) -> list[SourceCredentialRecord]:
+        return [self._masked(row) for row in self._repository.list()]
+
+    def delete(self, *, name: str) -> bool:
+        return self._repository.delete(name=name)
+
+    def binding(self, *, name: str) -> "SourceCredentialBinding":
```
totalfrank: Same here, let's be consistent - do you use forward reference since you don't need to.

totalfrank: Same here, let's be consistent - do you use forward reference since you don't need to.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service.py
```diff
@@ -0,0 +1,197 @@
...
+            name=row.name,
+            credential_type=row.credential_type,
+            header_name=row.header_name,
+            allowed_prefixes=self._prefixes_of(row),
+            has_secret=bool(row.secret_ciphertext),
+            updated_at=row.gmt_modified,
+        )
+
+
+class SourceCredentialBinding:
```
totalfrank, line 183: I did not quite get what this is for. Can you elaborate?

src/backend/src/agentclaw/community/adapters/http/openapi_v1/source_credentials/router.py
```diff
@@ -0,0 +1,172 @@
...
+)
+
+from agentclaw.community.api.source_credential_service import (
+    SourceCredentialServiceProtocol,
+)
+from agentclaw.community.adapters.http.openapi_v1.authorization import PublicAPIRoute
+from agentclaw.community.di import Injected
+
+router = APIRouter(
+    prefix="/openapi/v1/source-credentials",
```
totalfrank: Start with /openapi/v1/bots to be consistent if there's no good reason not to.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/source_credentials/router.py
```diff
@@ -0,0 +1,172 @@
...
+    tags=["source-credentials"],
+    route_class=PublicAPIRoute,
+    # REFUSED made visible on the mount, not only in admission.py: the
+    # central require_principal check refuses app-only callers either way;
+    # the dependency keeps the decision legible and armour against a
+    # mislabelled table entry (the admission-inventory gate's rule).
+    dependencies=[Depends(refuse_app_only_caller)],
```
totalfrank: Let's remove this. I understand what the comment is saying but there are two problems.

1. It breaks our design - which deliberately put all dependency check for app level admission at a single registry.
2. registering credential is definitely a app operation right? We need to allow up to do so.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/source_credentials/router.py
```diff
@@ -0,0 +1,172 @@
...
+        credential_type=body.type,
+        # The verified principal composes the audit actor — the same rule
+        # the bots group uses, and never the body. Who rotated a tenant's
+        # token is the question this column exists to answer.
+        modifier=_audit_actor(caller, caller.user_id),
+    )
+    return envelope(_detail(record), request)
+
+
+@router.delete("/{name}", response_model=Envelope[Deleted])
```
totalfrank, line 211: For deletion operation(and maybe list as well, I am not very certain right now), we might want to guard it - only owner app can do.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
```diff
@@ -589,6 +589,15 @@
...
     # Department directory search — a tenant-wide catalogue read, not a user's.
+    # ── Source credentials (W3, #1471) — REFUSED, human-only ──────────────
+    # A secret write and the tenant's credential inventory are operator
+    # surfaces: this is the one group with no business running headless, so
+    # every operation refuses an application-only caller. The tenant guard
+    # scopes the rows; a named human principal is the identity on record.
+    ("PUT", "/openapi/v1/source-credentials/{name}"): AdmissionMode.REFUSED,
+    ("GET", "/openapi/v1/source-credentials/{name}"): AdmissionMode.REFUSED,
+    ("GET", "/openapi/v1/source-credentials"): AdmissionMode.REFUSED,
+    ("DELETE", "/openapi/v1/source-credentials/{name}"): AdmissionMode.REFUSED,
```
totalfrank: We can't refuse APP. These are meant for app to use. They are not user operations.

And actually I am thinking we might need a strategy like AdmissionMode.REQUIRED(app must be there). But that's separate concern. I will do that later.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/authorization.py
```diff
@@ -397,6 +397,18 @@ def __repr__(self) -> str:  # pragma: no cover - diagnostics only
...
+    # miss. No bot to address, no owner to pin — admission is the whole
+    # authorization story (human-only).
+    ("PUT", "/openapi/v1/source-credentials/{name}"):
+        NoCheck("tenant-guarded credential write; admission is human-only"),
+    ("GET", "/openapi/v1/source-credentials/{name}"):
+        NoCheck("tenant-guarded credential metadata; admission is human-only"),
+    ("GET", "/openapi/v1/source-credentials"):
+        NoCheck("tenant-guarded credential inventory; admission is human-only"),
+    ("DELETE", "/openapi/v1/source-credentials/{name}"):
+        NoCheck("tenant-guarded credential delete; admission is human-only"),
```
totalfrank: No Check is correct but reason is wrong. We don't need to check because we won't allow user to access so that there won't be collaborator scenario.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/models.py
```diff
@@ -0,0 +1,123 @@
...
+from agentclaw.community.utils.avernet_tenant_guard import (
+    register_avernet_tenant_guard,
+)
+
+# SQLite only auto-increments columns declared as exactly "INTEGER PRIMARY KEY".
+# BigInteger renders as "BIGINT" in SQLite, which breaks autoincrement.
+AutoIncrementBigInteger = BigInteger().with_variant(Integer, "sqlite")
+
+
+class SourceCredentialRecord(BaseModel):
```
totalfrank, line 56: I did not read this and SourceCredentialRow throughly, but what's the difference? They look pretty similar to me. Do we want to consolidate them into one?

I might be wrong though, there might be legitimate reason we need both, let me know.

src/gateway/configs/application.yaml
```diff
@@ -271,6 +271,12 @@ user_config:
       user: required
       app: required
 
+    # Source credentials (W3 #1471): tenant-level secret management;
+    # human-principal only on every operation (REFUSED admission) — the
+    # method-default entry for the new prefix.
+    "/openapi/v1/source-credentials/**":
+      user: required
```
totalfrank: I think this should be app: required instead.

src/gateway/configs/application.yaml
```diff
@@ -399,6 +405,15 @@ user_config:
           source: file
           path: schemas/bots.openapi.json
           refresh_seconds: 300
+      # W3 (#1471): tenant source credentials — a new declared public
+      # domain, published from its own backend artifact and secured by the
+      # route_security REFUSED row below (`user: required`, human-only).
+      source-credentials:
```
totalfrank: Let's make the API start with /openapi/v1/bots so that we don't need this?

src/gateway/configs/schemas/source-credentials.openapi.json
```diff
@@ -0,0 +1,961 @@
+{
```
totalfrank: Once we changed the API to be prefixed with /openapi/v1/bots, this goes to the other file.

totalfrank (commented):

src/backend/src/agentclaw/community/core/repository/protocols/bot/source_credential.py
lucas-xzp: Addressed — the quotes are gone group-wide; unquoted references (the `get` method's shape) are now the rule, since every module here has `from __future__ import annotations`.

src/backend/src/agentclaw/community/core/repository/protocols/bot/source_credential.py
lucas-xzp: Addressed — `CredentialType` (StrEnum: header / oss_aksk / basic) now lives in `core/bot_config_manifest/credentials/models.py` and threads through the repository protocol, the service, and the storage row. The published spelling in the API body/detail schemas is an adapter-layer `_DocumentedEnum` subclass carrying `x-enum-descriptions` (core must not import the http adapter), with an import-time assert holding the two vocabularies member-for-member.

src/backend/src/agentclaw/community/core/repository/protocols/bot/source_credential.py
lucas-xzp: Addressed — the upsert contract now documents each parameter: `secret_ciphertext` (TokenVault output — the service encrypts before the call, so this layer never sees plaintext), `allowed_prefixes` (the validated list; storage encodes the row's JSON form), `owner_app_id` (creator app, stamped at insert, never re-stamped) and `modifier` (the router-composed audit actor).

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service_protocol.py
lucas-xzp: Addressed — forward-reference quotes removed.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service_protocol.py
lucas-xzp: Addressed — forward-reference quotes removed.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service_protocol.py
lucas-xzp: Addressed — forward-reference quotes removed.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service.py
lucas-xzp: Leaving it as-is per your note — the PR's only addition here is the `has_master_key` property; the decoupling from the ai-coding vault is yours to follow up.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service.py
lucas-xzp: Addressed — annotation unquoted (same fix as the sibling comment).

lucas-xzp: Addressed — same fix as above.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service.py
lucas-xzp: `SourceCredentialBinding` is the W2 fetcher seam: it duck-satisfies the guarded fetcher's `CredentialInjector` / `AuthorizationPolicy` protocols — `headers_for(url) -> {header: value}` and `reauthorize(url) -> None` (refusing the hop) — so structural typing replaces importing the fetcher module, and whichever of W2/W3 merges first, the fetcher binds to it at use. Both methods re-read the row per call on purpose: rotation is a re-PUT with no signal, so "next fetch uses the new value" is the observable rotation contract. Happy to relocate or rename it into the fetcher's package when W2 lands if that reads better.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/source_credentials/router.py
lucas-xzp: Done — the group lives at `/openapi/v1/bots/source-credentials` now, mounted with the other literal groups before the `{bot_id}` wildcard (the market/catalog placement), so the bots domain routes it as-is and no new gateway domain exists.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/source_credentials/router.py
lucas-xzp: Done — removed. With the table no longer REFUSED the inventory gate itself would refuse the mount-level declaration (`REFUSED ⟺ refuse_app_only_caller` is a bidirectional lock in `test_admission_inventory.py`), so the central registry is the single decision again.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/source_credentials/router.py
lucas-xzp: Done for delete — and extended to rotation: a re-PUT replaces the whole row (secret, header, and prefixes) behind every manifest citation of the name, so a non-owner re-PUT is a cross-application hijack of the fetch path, refused before storage. The model: `owner_app_id` (the creating application), stamped at insert and immutable — rotation and delete are the owner's alone (403 `CredentialNotOwnedError`). Reads stay tenant-wide: the name is the shared reference namespace manifests cite — an app may legitimately reference a name another app created, and restricting list would hide inventory without adding safety. If you want reads owner-scoped as well, it's a one-line change in the service — say the word.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/admission.py
lucas-xzp: Done — all four operations are OPEN. The "app must be there" half is expressed at the edge today: route_security is `user: optional, app: required` (the same shape as `/openapi/v1/bots/authorized`), so a bare human credential can't reach the group while your `AdmissionMode.REQUIRED` is pending; the backend's `require_principal` floor still 401s anything unverified. The owner-app mutation checks live in the service, against the row.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/authorization.py
lucas-xzp: Rewritten — the rows now say: no bot to address, and the callers this surface admits are applications (the edge requires one), so no collaborator scenario exists at all; the one identity that matters for mutation — the owning application — is checked by the service against the row.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/models.py
lucas-xzp: Deliberate split, kept: `SourceCredentialRecord` is the public redacted shape — `has_secret`, prefixes as a list, no value field — while `SourceCredentialRow` is the storage shape (`secret_ciphertext`, prefixes as stored JSON, `owner_app_id`). They exist so the value physically cannot end up in the public form `by construction`, not by discipline (each class's docstring carries this). The third layer, `SourceCredentialModel`, is the SQLAlchemy row. If it still reads as duplication to you, name the collapse you'd prefer and I'll rework it.

src/gateway/configs/application.yaml
lucas-xzp: Done — `user: optional, app: required`, the `/openapi/v1/bots/authorized` shape. Chose optional-user over omitting the key so a user riding along stays resolvable, feeding the audit actor (`app:{id}:on-behalf-of:{user}` in `modifier`); the pure machine shape (user identity never resolved) is a one-key change if you prefer it.

src/gateway/configs/application.yaml
lucas-xzp: Done — under the bots prefix the `source-credentials` domain entry is deleted; only the route_security rule remains (new path, app required). Both the avernet and ocb sides carry it.

src/gateway/configs/schemas/source-credentials.openapi.json
lucas-xzp: Done — `source-credentials.openapi.json` is deleted; its two paths and their component closure merged into `bots.openapi.json` surgically (append-only insertion, duplicate-key-checked), and `schemas/README.md` no longer lists a separate file.

lucas-xzp:
CR 已全量处理（7558ed8）:安全模型反转为 **App 操作面**——

**模型**（对应 router/admission/authorization/网关四层意见）:
- 前缀 `/openapi/v1/bots/source-credentials`(字面量组,挂 `{bot_id}` 通配之前);网关独立 domain 撤销,schema 并入 `bots.openapi.json`(append-only,已验重复键)。
- ADMISSION 全组 **OPEN**,路由级 `refuse_app_only_caller` 摘除(gate 的 REFUSED⟺声明双向锁)。"app 必须在场"由边缘执行:route_security 翻为 `user: optional, app: required`(同 `/bots/authorized` 形态);avernet/ocb 两仓同步。
- PUT 弃 `ActingCallerDep`(app-only 调用者报未校验 user_id 是归因漏洞),改读 Principal 的 `app_id`;modifier 形如 `app:{id}[:on-behalf-of:{user}]`。

**所有权**:DDL 增 `owner_app_id`(创建者,插入钉死)。轮换与删除仅 owner(403,先拒后写)——非 owner 的 re-PUT 是整行替换(secret+header+prefixes),等于劫持名字的全部 manifest 引用。读取保持全租户(名字是共享引用命名空间,app B 引用 app A 建的名字是合法用法);若你倾向收紧读取,一行可改。

**其余**:credential_type 收敛为 `CredentialType` enum(核心 StrEnum + 发布层 `_DocumentedEnum` 带成员描述);前向引用清理;upsert 参数补文档;Statement 理由重写;两个提问(binding 用途、Record/Row 拆分)已在各自线程说明。本地测试与门禁全绿(endpoint 面向 app 调用者的断言全部翻转并补了所有权矩阵用例)。

totalfrank (approved):
LGTM.

PR #1747, opened 2026-08-31
feat(backend): add directory zip download to openapi v1 resources

lucas-xzp:
## Problem

The public openapi v1 resources surface could only download single files (`GET .../resources/download`). Directory download existed only on the internal console API (`GET /api/resources/files/download-dir`, zip packaging), so a caller exporting a folder had to list and download file by file.

## Solution

New endpoint `GET /openapi/v1/bots/{bot_id}/resources/download-dir?path=`, returning `application/zip`. `path` is optional — omitted means the whole workspace root.

- `ResourceFileService.iter_directory_files`: one device-resolution for the whole walk, level-by-level `list_dir` (no engine's `recursive` flag is trusted anywhere in the repo), browser-parity filtering (dotfiles at every level, hidden system names at the workspace root only). Caps — 100 MB per file / 5000 files / 500 MB total — are preflighted from listing sizes **and** re-counted against the bytes actually streamed, so a stale or lying listing cannot widen them.
- New `DirectoryTooLargeError` maps to a fixed 413 envelope ("Directory too large to download"), deliberately not the preview-cap message.
- The archive is disk-backed (tempfile + `FileResponse` with background cleanup), never in memory; zip entries carry Unix type bits so macOS Archive Utility accepts them.
- Failure honesty for a public API: missing directory → 404; empty directory → a valid root-only archive (the walk distinguishes the two); a mid-walk read error aborts the whole download — never a 200 with a partial zip.
- Registered `OWNER_SCOPED` + `GRANT_CHECKED_OWN_BOT`. The deprecated legacy shim deliberately **skips** this route: a new operation must not grow a made-up retiring address (`test_legacy_parity`'s count stays 42).

## Validation

- Service walk: 12 tests (nested tree, single context resolution, root-level filtering, 404 vs empty, per-file/count/total caps, streamed re-count catching a lying listing, vanished-entry vs read-error race rules).
- Handler: 8 tests (archive shape, root download, 400/404/413, empty dir, abort-on-failure, UTF-8 filename); zip build: 3 tests (entry metadata, root-only archive, failure leaves no partial file).
- Wire/endpoint happy case + user-scope 403 sweep; inventory and contract suites green: admission, authorization, legacy parity, error schema, path convention, gateway namespace, module boundaries.
- Pre-push gate (flake8 blocklist) clean. Full CI runs on this PR.

## Compatibility and risk

Additive only — no existing route, schema, or config changes. Rollback is a revert; no migration involved. Known limitation: per-file DEFLATE compression runs synchronously on the event loop (same shape as the console implementation); the 100 MB per-file cap bounds it.

## Spec

`src/backend/specs/2026-08-31-openapi-v1-resources-download-dir/` — spec / plan / tasks.

## Related issues

None filed. Console parity reference: internal `GET /api/resources/files/download-dir` (unchanged; its 30 MB cap and empty-dir-404 semantics are deliberately not copied).

### DIFF
--- a/src/backend/specs/2026-08-31-openapi-v1-resources-download-dir/plan.md
+++ b/src/backend/specs/2026-08-31-openapi-v1-resources-download-dir/plan.md
@@ -1,0 +1,54 @@
+# Plan — download-dir
+
+## Target architecture
+
+```
+GET /openapi/v1/bots/{bot_id}/resources/download-dir?path=
+  → download_directory (openapi_v1/resources/router.py)
+      _safe_path（'..' → 400）→ _file_coords → file_svc.iter_directory_files(...)
+  → ResourceFileService.iter_directory_files（core/services/resource_file_service.py）
+      一次 _resolve_ctx + _device_fs → 逐层 list_dir 遍历（BFS）
+      → 限额双检（listing 预检 + 实际字节复核）→ 逐文件 read_file → yield (name, bytes)
+  → build_directory_zip（openapi_v1/resources/zip_build.py）
+      临时文件落盘；任何失败删半成品
+  → FileResponse(application/zip, BackgroundTask(os.unlink))
+```
+
+## Changes
+
+| 文件 | 改动 |
+| --- | --- |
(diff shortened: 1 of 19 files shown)


totalfrank (dismissed):
LGTM.

xianmuyq (approved):

PR #1748, opened 2026-08-31
fix(bcs): V1 create-group anchors participant reachability on the originator

vzvince:
## Problem

The V1 OpenAPI create-group endpoint (`POST /openapi/v1/collaboration/groups`) anchored participant reachability on the **driver**: every non-driver participant had to be `public` or a friend of the driver, and the driver itself was exempt (ungated vs the caller). This diverged from legacy `POST /groups`, which is originator-centered, and blocked a Human caller from pulling their own protected bots into a group (a protected owned bot that isn't a friend of the driver was rejected).

## Change

Anchor the V1 participant gate on the **originator**, aligning with legacy:

- **Human originator** (the default for a Human caller): a participant is reachable if `public` or owned by that human (`created_by` matches the caller). Friendship is not consulted.
- **Bot originator** (a Bot caller, or a Human designating an owned Bot as originator): a participant is reachable if `public` or a friend of the originator bot.
- The **driver is no longer exempt**: when the driver differs from the originator it is gated like any other participant (when it equals the originator it is self-reachable and exempt).

### Files
- `crates/services/bcs-group/src/application/management.rs` — the `v1_openapi_create_policy` participant branch now mirrors the legacy originator-anchored branch (Human branch: public/ownership; Bot branch: `ensure_v1_reachable(originator, …)`). Reuses the already-computed `originator` / `is_human_originator`.
- `crates/application/v1/bcs-app-group/src/lib.rs` — removed the now-redundant facade-level `ensure_originator_can_reach_driver` (the core loop gates the driver against the originator uniformly).
- `api-contracts/v1/openapi/groups.yaml` — refreshed the `originator` field description.
- Tests — flipped the two `driver-ungated` assertions, rewrote the driver-reachability case to participant ownership, and added Human/Bot originator participant-gate coverage.

## Result

A Human caller can now pull **owned** and **public** bots into a V1 collaboration group; protected bots they do not own are rejected — matching legacy semantics. `authorize_originator` is unchanged and stays stricter than legacy (self or a caller-owned Bot only).

## Verification

`cargo test -p bcs-app-group -p bcs-group` → all green (0 failures).

### DIFF
--- a/src/bcs/api-contracts/v1/openapi/groups.yaml
+++ b/src/bcs/api-contracts/v1/openapi/groups.yaml
@@ -611,10 +611,12 @@ CreateCollaborationGroup:
         Optional caller-designated originator. When omitted, the selected
         HumanOrOwnedBot Principal becomes the originator. A Human Principal
         may instead select a registered Bot it owns (`created_by` matches the
-        caller). A Bot Principal may select only itself. When a Human selects
-        an owned Bot distinct from the driver, the driver must be reachable
-        from that originator (public visibility, or a friend of it). The `dm`
-        group kind does not carry an originator.
+        caller). A Bot Principal may select only itself. Every Bot participant
+        — including the driver when it differs from the originator — is
+        anchored on the originator: a public bot is always reachable; a
+        protected bot is reachable only if the originator is a Human that owns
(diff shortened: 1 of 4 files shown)


questchaos (approved):
LGTM

PR #1753, opened 2026-08-31
perf(skills): fuse Center ensure into v4 mapping publish

FreddieSun:
## Problem
Center Skill mapping activation currently performs a separate Runtime ensure request before publish, adding a device/proxy round trip.

## Solution
Introduce mapping contract v4. Runtimes advertising v4 perform exact Center ensure inside mapping publish; v3 runtimes retain the existing ensure → publish → verify flow. Mapping wire shape is unchanged.

## Validation
- Backend Skills Pool: 38 passed
- Engine mapping contract and OpenClaw adapter: 34 passed
- Ruff and diff checks passed

## Compatibility and risk
Backend selects v4 only when Runtime probe explicitly advertises it. Existing v3 Runtime behavior remains unchanged; ensure failure prevents publish and verify remains separate.


FreddieSun:
关闭原因：目标分支已吸收该变更，当前 PR 为 0 files / 0 additions / 0 deletions，已无可合并内容。

PR #1758, opened 2026-09-01
feat(backend): W11-store and prove every byte the platform fetches for a manifest

### LINKED ISSUE #1510: [配置清单][W11] 平台侧物化与留存
**第一波。**完整范围与验收标准见 `src/backend/docs/bot-config-manifest/work-items.zh-CN.md` §5, W11（规划 PR #1465）。

**负责人：**`lucas-xzp` · 0.5 天 · 第 2 天 · **实现**（§7）

**目标。**拉取到的内容由平台存成自己的持久副本，后续步骤从该副本读取而不是重新拉取。

**依赖：**#1470（W2，产出被它存储的那个 fetcher） · **阻塞于：**—

> **本项不再是 #1472（W4）的硬依赖。**D2 修订为类目覆盖（#1467）之后，被撤销的三方 diff 才需要版本 N 的物化文件清单来区分「声明里删掉了」与「bot 自己建的」；覆盖不需要这个区分。**本项本身的必要性不变** —— §2.8 的审计是硬性要求，`keep_last` 也住在这里 —— 变的只是排期：W4 与 W11 现在可以并行。

## 范围内

内容存储、它的寻址（内容哈希）、留存策略，以及下发与审计共用的读取路径。

## 关键验收标准

- 每个被拉取的对象都带着足够的**溯源**信息持久化，以便日后回答*这个 bot 到底收到了什么、它从哪来* —— 来源、解析出的 ref/SHA 或 digest、拉取时间，以及字节本身。
- **下发从这个存储读**，于是重试下发绝不重新拉取，源侧故障也无法污染一次进行中的下发。
- 这个存储就是 `keep_last` 读的那个（§2.7）—— **一套机制，不是两套**。
- 留存策略是显式的，并且是**对着审计要求**陈述的，而不是顺手定的。
- **凭证永远不与内容一起持久化**；溯源只记录凭证**名称**。

**规模：**中等。

lucas-xzp:
## Problem

§2.8 of the bot-config-manifest plan is a hard requirement from audit and reconciliation: content the platform fetches on a manifest's behalf must be kept as **the platform's own durable copy**, and every step after fetch must read that copy — retried delivery re-reads here instead of re-fetching, a source-side fault can never pollute a delivery in progress, and `keep_last` has somewhere to live. The backend has no such store today (issue #1510, work item W11; builds on W2's merged fetcher — zero remaining dependencies, no coordination).

## Solution

`core/bot_config_manifest/content/` — one mechanism behind all three consumers of §2.8 (audit, delivery, `keep_last`), not a copy per consumer:

- **Bytes live in a content-addressed blob tree** — `<root>/blobs/<hex[:2]>/<hex64>`. The digest (`sha256:<hex>`, the fetcher's own vocabulary, now shared as `fetch/limits.DIGEST_RE`) IS the address: identical bytes are written once, ever; the write is atomic (temp + `os.replace` in the shard directory) and heals truncation/append corruption beneath an existing address while the correct bytes are in hand. Not in the database — schema §5 lets one entry be 100–200 MiB, and a column holding that is a self-destructing design.
- **`ac_manifest_content` is the append-only provenance log**, one row per store event: the bot axes `(avernet_tenant, env, entity_id, bot_id)` behind the same tenant guard as W1's table, the digest, **both** URLs (entry source after `${BOT_*}` substitution, and the final hop — differing values mean a redirect happened), the credential **name** only (W3's identifier; a value never crosses into this layer), size, fetch time, modifier. No unique key: fetching the same digest twice is two audit events, not one row to overwrite.
- **`read(digest)` is the single read path**, shared by delivery and audit: read in chunks with the hash on the same pass, returned whole (peak ≈ 2× the blob at the §5 cap — a stated v1 trade; the consumer materialises the full payload anyway) — the store returns bytes it can prove or fails loudly; a missing address is terminal and the store **never re-fetches** (§2.8's decoupling). `keep_last` will be W4's apply-record digest plus this read — one addressing, no second copy.
- **Every refusal lands before any side effect**: receipt-vs-bytes (re-hash + size), URL sanitization, and column-width validation (a wire-controlled `Content-Type` header has no length cap of its own, and SQLite tests cannot catch varchar overflow — strict mode would surface a 500 after the blob landed; non-strict would silently truncate an append-only audit row forever) all happen before the blob write. URLs in provenance carry scheme://host[:port]/path — never userinfo, never a query string (where signed-source tokens live); the reconciliation anchor is the digest.
- **Retention is stated against the audit requirement, not defaulted**: v1 retains rows and blobs unconditionally; a window, when audit names one, is a DDL-comment change plus a sweep in that PR.
- The blob root is config-borne (`user_config.bot_config_manifest.content_store_dir`, neutral `./data/manifest_content`; an env overlay decides NAS), parsed purely by `content.settings.content_store_root_from_config` for W4's composition root to inject — core reads no environment, per the ruling on W2's allowlist. DI binds the repository protocol now; the service is a declared machine part in the same shape as the fetcher before its orchestrator existed.

## Validation

- 37 tests in the new package (store contract matrix + repository over real SQLite incl. cross-tenant isolation on a shared bot_id + config seam), pinning: roundtrip, receipt-mismatch refusal with zero side effects, corrupted-blob and missing-address behavior on the read path, write-once dedup with two provenance rows, URL sanitization (query/userinfo/fragment/port/path), name-only credentials, tz normalization, width refusals, empty-payload roundtrip, orphan-blob self-heal.
- Batch review (per the review-before-push rule): 7 findings — 1 HIGH (width validation before the blob write: SQLite tests structurally cannot catch it) + 3 MEDIUM (write ordering, the dedup shortcut's blind trust, error messages echoing unparseable URLs) + 3 LOW — all resolved in this PR; the ledger is in the spec.
- Full backend suite after a clean rebase onto the day's dev (which gained #1734/#1745/#1559/#1756 mid-flight; zero file overlap): **16261 passed, 59 skipped**; line coverage **88.14%**; changed-line coverage **97.84% (226/231, ≥80%)** — reproduced locally via `scripts/ci_test.sh`.
- Architecture gates green: boundaries/oversized/user_config AST guard/shipped-config/repository-contracts provides (both READMEs updated)/flow_coverage entry extended to cover W11; golden effective-config snapshots regenerated for the deliberate new key (all three profiles gain exactly one neutral line).

## Spec

`src/backend/specs/2026-08-31-w11-content-store/` — spec (acceptance map + design decisions + the batch-review ledger), plan, tasks.

Closes #1510.


src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
```diff
@@ -0,0 +1,84 @@
...
+-- RETENTION POLICY (stated against §2.8, deliberately explicit): v1 retains
+-- rows and blobs unconditionally — no delete, no sweep, no TTL. The audit
+-- requirement is "answer what THIS bot received, from where, at that time";
+-- until an audit horizon is named, any deletion is a manufactured audit gap,
+-- and a cleanup that misjudges it is unrecoverable by design. The blob layer
+-- is hence self-consistent: a digest's bytes exist for as long as any row
+-- references them, and rows never go away. A retention window, when audit
+-- names one, lands here as a comment change plus a sweep mechanism — not as
+-- a silent default.
+CREATE TABLE `ac_manifest_content` (
```
totalfrank, line 71: **The axes here are the bot's, and nothing ties a receipt to the apply or to the entry it was fetched for.** Two consequences worth settling *before* this table ships, because its own retention policy (never updated, never deleted) means a column added later is permanently NULL for every row written before it — there is no backfill this table's stated policy permits.

**1. W4 already ships the join key, and this table doesn't take it.** `ac_bot_config_manifest_apply.apply_id` carries the comment *"also what a per-entry table would join on, if provenance ever needs one"* (`repository/apply_models.py:95`). This is that table. Without `apply_id`, "what did apply X deliver" is unanswerable, and receipts from a retried apply interleave with the original's under nothing but `gmt_create`.

**2. `keep_last` is per *entry*, and an entry has no identity here.** work-items §2.7: 「`keep_last` means 'reuse what we materialised **for this entry** last time'」, and the §2.8 acceptance line is that this store is the one `keep_last` reads — 一套机制，不是两套. The only per-entry-ish column is `source_url`, and it is stored **sanitized**: two entries whose sources differ only in the query string collapse to one value, and an entry whose source moved (the very case `keep_last` covers) no longer matches its own history. The README's answer — "the per-entry digests that wave's apply records carry" — puts the entry→digest mapping in the apply report's JSON blob, i.e. a second, unindexed addressing outside the store, which is what that acceptance line is trying to prevent.

A nullable `apply_id` plus `category` + `entry_identity` (W4 already computes `EntryResult.identity`) costs nothing now and turns both questions into indexed reads.

Related, same area: `idx_digest` is justified as "which fetches ever produced this content", but the repository exposes no query that can use it — `records_for` is bot-scoped only. Either add the read or drop the index until a reader exists.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
```diff
@@ -0,0 +1,84 @@
...
+-- #1510, W11 of the bot-config-manifest plan; the requirement is §2.8 of
+-- ``docs/bot-config-manifest/work-items.zh-CN.md`` — hard requirement, audit
+-- and reconciliation).
+--
+-- This table is the PROVENANCE LOG. The bytes themselves never live here:
+-- schema §5 caps one entry at 100–200 MiB, and a column holding that is a
+-- self-destructing design (max_allowed_packet, InnoDB row budget, every
+-- backup suddenly a content store). Bytes live in the content-addressed blob
+-- directory (``user_config.bot_config_manifest.content_store_dir``); this
+-- table answers "WHERE did these bytes come from, for WHICH bot, WHEN".
```
totalfrank: Worth being precise about what one row *means*, because this header and the README both promise slightly more than the code delivers.

A row is a **fetch/store event**, not a delivery. Under §3.2's all-or-nothing category overwrite, an entry can be fetched, verified and written here and then never materialised — because a sibling entry in the same category failed and the whole category was skipped. So 「这个 bot 到底收到了什么」 read off this table alone **over-reports**: what it actually answers is "what the platform fetched on this bot's behalf".

Not necessarily a schema change (the apply report holds the outcomes) — but the DDL header, `README.md`'s new W11 section and the service docstring all say "what this bot received", and a future auditor will read those, not this thread. Either reword them, or carry the delivery outcome — which is another argument for the `apply_id` link in my comment on the `CREATE TABLE` below.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/content/service.py
```diff
@@ -0,0 +1,316 @@
...
+            f"unparseable provenance URL: length {len(url)}"
+        ) from exc
+    if not parsed.scheme or not parsed.host:
+        raise ContentStoreError(
+            f"unparseable provenance URL: length {len(url)}"
+        )
+    authority = (
+        f"{parsed.host}:{parsed.port}" if parsed.port is not None else parsed.host
+    )
+    return f"{parsed.scheme}://{authority}{parsed.path or '/'}"
```
totalfrank: **An IPv6 source is written into the audit log as a URL nothing can read back.** `httpx.URL.host` returns an IPv6 host *unbracketed*, so the reassembled authority loses the brackets and the port becomes ambiguous:

```
>>> # httpx 0.28.1, run against this branch's service
>>> svc.store(fetched_from("https://[2001:db8::1]:8443/a.bin"), ...).fetched_url
'https://2001:db8::1:8443/a.bin'
```

It is reachable, not theoretical: `_validate_url` accepts an IPv6 literal, `_refused_address` only refuses non-global addresses, and `_collect` sets `FetchedObject.url = str(url)` with the brackets intact — this function is where they are lost. And by this table's own policy the row is never updated and never deleted, so it can't be corrected afterwards.

```python
host = f"[{parsed.host}]" if ":" in parsed.host else parsed.host
authority = f"{host}:{parsed.port}" if parsed.port is not None else host
```

plus an IPv6 case in `test_stored_urls_carry_path_but_never_query_or_userinfo`, which currently parametrises query/userinfo/fragment/port/path but no literal address.

Two smaller fidelity notes on the same function, since the log is append-only and this is the moment to decide: `httpx.URL` also decodes punycode (`xn--fsq.example` → `例.example`) and unquotes the path (`/%2e%2e/x` → `/../x`), so what lands in provenance is httpx's re-rendering rather than the bytes that went on the wire.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/fetch/limits.py
```diff
@@ -56,6 +57,13 @@
 #: Schemes accepted without a deployment exception (W2 验收: 仅 https)。
 SAFE_SCHEMES = frozenset({"https"})
 
+#: The digest vocabulary W2 fixed and every later wave shares: ``sha256:``
+#: followed by exactly 64 lowercase hex. The fetcher mints these addresses
+#: (declared-digest validation, FetchedObject.sha256) and the W11 content
+#: store uses them as its addressing scheme — one regex, one vocabulary, no
+#: per-module copies that drift.
+DIGEST_RE = re.compile(r"^sha256:[0-9a-f]{64}$")
```
totalfrank: Now that this is the shared address vocabulary and not just the fetcher's private check, the `$` deserves tightening: in Python `$` also matches immediately before a trailing newline.

```
>>> DIGEST_RE.match("sha256:" + "a"*64 + "\n")
<re.Match ...>   # matches
```

So `sha256:<hex>\n` passes `_require_valid_digest`, and `read()` answers `ContentMissingError` ("the digest names nothing", the 404 case) instead of `ContentStoreError("untrusted content address")` — the two mean different things to a caller — and the newline rides into the error message and from there into a log line. Same at the other call site: `GuardedFetcher.fetch` accepts a newline-suffixed `expected_digest` as a valid pin and then fails it later as `FetchFailedError("digest mismatch")` rather than refusing it as malformed config.

`re.compile(r"\Asha256:[0-9a-f]{64}\Z")` (or switching both call sites to `.fullmatch`) closes both. No traversal risk either way — the charset is hex — so this is a taxonomy/hygiene fix, not a security one.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/content/errors.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Deliberately loud and deliberately final — the store never fetches on
+    the read path (§2.8: 下发 reads the platform copy; re-fetching would
+    recouple delivery to source-side faults, which is exactly what this
+    layer exists to prevent). The caller decides what a missing receipt
+    means; for keep_last that is W4's policy, not a silent side fetch.
+    """
+
+
+class ContentIntegrityError(ContentStoreError):
```
totalfrank: The hierarchy contradicts the docstrings. `ContentStoreError` says "(400-class)", this one says "(500-class)", and it subclasses the first:

```
>>> issubclass(ContentIntegrityError, ContentStoreError)
True
```

So the obvious mapping in W5's consumer — `except ContentStoreError: -> 400` — reports platform-side disk corruption as a caller error, which is backwards for the one failure in this file that should page someone. A caller has to know to catch the subclass first, and nothing here says so.

Cleanest is a sibling base (`ContentStoreError(ValueError)` for the 400 class, `ContentStoreFault(RuntimeError)` for integrity); minimum is a line in this docstring stating that the mapping must catch `ContentIntegrityError` before its base. `ContentMissingError` under the 400 class is fine.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/content/service.py
```diff
@@ -0,0 +1,316 @@
...
+        )
+        sanitized_fetched = _require_fits(
+            "fetched_url", _sanitized_url(fetched.url), _URL_MAX
+        )
+        stored_credential = _require_fits(
+            "credential_name", credential_name, _CREDENTIAL_NAME_MAX
+        )
+        stored_content_type = _require_fits(
+            "content_type", fetched.content_type, _CONTENT_TYPE_MAX
+        )
```
totalfrank: **Refusing the whole `store()` over `content_type` is the wrong severity for that field, and it hands the source side a lever.** The bytes are already fetched, streamed under the cap, hashed and verified; a `Content-Type` of 257+ characters — nothing between the wire and here caps it at 256 — throws all of that away. `content_type` is advisory metadata, not a reconciliation anchor (the digest is, as the DDL says). Truncating it, or storing NULL plus a log line, keeps the receipt and loses nothing that matters.

**The URL cap is the same shape with a bigger blast radius, and it crosses a rule this module states about itself.** README: *"This surface never accepts something it cannot apply."* Nothing caps URL length at `PUT` — `schema/limits.py` deliberately carries only the write-time limits and neither it nor `entries.py` bounds a source URL — so a manifest with a 3000-character source is accepted, validates, fetches up to 200 MiB, and is refused only here. (The DDL's "2048 chars is the doc-wide URL ceiling", line 66, describes a ceiling that doesn't exist anywhere in the repo; the only real bound is the 64 KiB document cap, which leaves plenty of room.) A redirect `Location` longer than 2048 does the same after a successful fetch.

Refusing at the store means the expensive work is always done first. The rule-preserving fix is a URL length limit at W1's admission (cheapest, and it keeps that README sentence true); the second-best is W2 refusing an over-long hop before it streams a body. Keeping the check here as the last line of defence is right — it just shouldn't be the *first* place the length is noticed.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/content/service.py
```diff
@@ -0,0 +1,316 @@
...
+        blob.parent.mkdir(parents=True, exist_ok=True)
+        fd, tmp_name = tempfile.mkstemp(dir=blob.parent, prefix=".tmp-")
+        tmp = Path(tmp_name)
+        try:
+            with os.fdopen(fd, "wb") as handle:
+                handle.write(data)
+            os.replace(tmp, blob)
+        except BaseException:
+            tmp.unlink(missing_ok=True)
+            raise
```
totalfrank, line 400: **`os.replace` buys atomicity of the rename, not durability of the data.** With no `fsync` on the temp fd before the replace (and none on the directory after), a power loss can leave the blob visible at its content address holding zero or partial bytes. That is exactly the case §2.8 exists to prevent — 「重试下发绝不重新拉取」 — and this layer never re-fetches, so an in-flight retry cannot recover; only a *new* apply that fetches the same bytes again heals it, and only because the size check on line 305 happens to catch a truncated file. For a store whose whole claim is "the platform's own **durable** copy", two lines seem worth it:

```python
with os.fdopen(fd, "wb") as handle:
    handle.write(data)
    handle.flush()
    os.fsync(handle.fileno())
```

(plus an optional directory fsync if you want the rename itself to survive too).

**Crash path, same block:** a process that dies mid-write leaves a `.tmp-*` file of up to 200 MiB in the shard directory forever — `tmp.unlink(missing_ok=True)` only covers in-process exceptions, and the retention policy is "no delete, no sweep, no TTL", so nothing will ever collect them. Either sweep stale `.tmp-*` in the shard dir on store, or state in the retention comment that temp files are the one thing a future sweep may remove — otherwise the next reader will correctly conclude the policy forbids it.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/content/service.py
```diff
@@ -0,0 +1,316 @@
...
+        re-hash: truncation and appends — the corruption modes that leave a
+        file claiming the same name — are caught at near-zero cost, and the
+        correct bytes are then rewritten over the damage while they are in
+        hand. Same-size bit rot stays for ``read()`` to detect loudly; the
+        deliberate trade is that a full re-hash of a 200-MiB blob on every
+        re-store would tax every audit fetch to defend against the rarest
+        corruption mode.
+        """
+        blob = self._blob_path(digest)
+        if blob.is_file() and blob.stat().st_size == len(data):
```
totalfrank, line 387: Nit on the reasoning rather than the decision, which may well stand: *"a full re-hash of a 200-MiB blob on every re-store would tax every audit fetch"* — a re-store is not an audit fetch, and `read()` re-hashes on every single call anyway, so the audit path already pays that cost unconditionally. The real trade is stat-vs-read **on the dedup path only**. Worth rewording so the next reader doesn't inherit the wrong cost model and "optimise" `read()` on the strength of it.

Worth noting alongside: on this shortcut `store()` returns success, and inserts a provenance row asserting the content is stored, **without ever having verified the bytes on disk** — while holding the correct bytes in hand. Fine as a stated trade; just make sure the docstring's "hands back bytes it can prove, or it fails" is understood as a `read()` guarantee only, never a `store()` one.

---

src/backend/src/agentclaw/community/core/repository/implementations/bot/manifest_content.py
```diff
@@ -0,0 +1,95 @@
...
+        *,
+        env: str,
+        entity_id: str,
+        bot_id: str,
+        limit: int = DEFAULT_RECORD_LIMIT,
+    ) -> list[StoredContentRecord]:
+        """One bot's receipts, newest first — the audit read."""
+        # Clamped, not trusted: a negative LIMIT flips to "unbounded" on
+        # SQLite and varies elsewhere; the audit read is bounded, always.
+        limit = max(0, limit)
```
totalfrank: `max(0, limit)` turns a negative limit into "return nothing", which for an audit read is the quietest possible wrong answer: a caller that computes `limit = total - offset` and goes negative is told **this bot has no receipts**, which in an audit context is a claim, not an empty page.

The protocol docstring's stated goal — a negative can never mean "everything" — is met just as well by refusing (`ValueError`), and refusing can't be mistaken for a fact about the bot. Same argument for `limit == 0`.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
```diff
@@ -0,0 +1,84 @@
...
+-- drift. It is NOT unique here (see above).
+--
+-- BOTH URLs are stored WITH path but WITHOUT userinfo and WITHOUT query.
+-- Userinfo is refused by the guarded fetcher before any wire contact anyway;
+-- a query string is where signed-source tokens live — the same reason W2
+-- logs host-only. The reconciliation anchor for audit is the digest, never a
+-- one-time signed URL. source_url is the manifest entry's source after
+-- ${BOT_*} substitution; fetched_url is the final hop after redirects — when
+-- the two differ, a redirect happened, and "where it came from" wants both
+-- facts. git sources (W7) will carry their resolved ref/SHA in source_ref.
```
totalfrank: This sentence names a column the table doesn't have — there is no `source_ref` anywhere in the DDL below. As written it reads as "the column exists and W7 will populate it", and someone will grep for it.

It's also the one acceptance line of #1510 that isn't fully met: 「来源、解析出的 **ref/SHA** 或 digest、拉取时间，以及字节本身」. For https sources the digest *is* the whole answer, so v1 is defensible — but say that as a decision ("git refs land with W7, which adds the column then") rather than as a reference to a column that exists. And given this table's own never-update/never-delete retention, add the consequence out loud: every row written before W7 is permanently NULL for that column, with no backfill this policy permits.

---

totalfrank (commented):
Reviewed against W11's acceptance list in `docs/bot-config-manifest/work-items.zh-CN.md` and against the merged waves (W1, W2, W3, W4, W10). I checked the branch out, installed the suite and ran it: the 43 new tests pass, and `tests/community/architecture + core/bot_config_manifest + config` is 571 passed, so nothing here is broken as built. The blob layer is a good mechanism — content addressing, atomic write, verify-on-read, refusals before side effects, credential *name* only — and the compatibility work is genuinely done: the tenant guard follows W1's pattern (and works from inside W4's apply thread, which re-binds the tenant contextvar at wrap time), `DIGEST_RE` is shared with W2 rather than copied, the 128-char `credential_name` matches W3's `ac_source_credential.name`, and `ContentScope`'s axes match W4's `ApplyContext`.

Findings inline, roughly in the order I'd fix them.

**Schema, and the reason it can't wait** (comments on the `CREATE TABLE` and on the header). The retention policy makes this table a one-shot decision: rows are never updated and never deleted, so a column added later is permanently NULL for everything written before it, with no backfill this policy permits. Two things are missing while it's still free to add them:

- **No `apply_id` and no entry identity.** W4's `apply_models.py:95` literally describes `apply_id` as *"what a per-entry table would join on, if provenance ever needs one"* — this is that table. And §2.7's `keep_last` is per **entry**, while the only per-entry-ish column here is `source_url`, stored sanitized, so two entries differing only in query string collapse and an entry whose source moved loses its own history. The README's answer (per-entry digests in the apply report's JSON) is the second addressing that the 一套机制，不是两套 acceptance line is trying to prevent.
- **A row means "fetched", not "delivered".** Under §3.2's all-or-nothing category overwrite an entry can be stored here and never materialised. That's fine as a mechanism, but the DDL header, the README section and the service docstring all say "what this bot **received**", and that over-reports.

**Correctness / robustness:** IPv6 sources land in the audit log as unparseable URLs (`https://[2001:db8::1]:8443/x` → `https://2001:db8::1:8443/x`, verified on this branch); `ContentIntegrityError` subclasses the documented 400-class base, so the obvious mapping reports disk corruption as a caller error; no `fsync` before `os.replace`, which is the difference between an atomic rename and a durable copy, and matters here precisely because this layer never re-fetches; `.tmp-*` files survive a crash forever under a no-sweep policy.

**Boundary placement:** a >256-char `Content-Type` or a >2048-char URL discards a fully fetched and verified 200 MiB payload at the last step. Nothing caps URL length at `PUT` (the DDL's "doc-wide URL ceiling" doesn't exist in the repo), so a document W1 accepted can only fail at apply — which crosses this module's own stated rule, *"This surface never accepts something it cannot apply."*

**Nits:** `DIGEST_RE`'s `$` matches a trailing newline, now that it's the shared address vocabulary; `records_for`'s negative-limit clamp answers "no receipts" for an audit read; `idx_digest` has no reader; the `source_ref` column named in the DDL comment doesn't exist; the dedup shortcut's stated cost model is off.

Two things not worth inline comments but worth a look before W5 wires this up:

1. **The blob root is node-local by default while the provenance log is shared.** `./data/manifest_content` resolves against the process CWD, so with more than one backend replica a blob written on node A is invisible to a delivery retry that lands on node B — and by design that's `ContentMissingError`, terminal, never a re-fetch. The yaml comment offers a NAS overlay as an option; given this store's contract, W5's wiring should probably make the shared volume a stated deployment *requirement* (or the read path's missing case a documented "re-fetch is the caller's decision"), not a possibility an overlay may take up.
2. When W5 constructs the service, a `ContentScope.from_apply_context(ctx)` would keep the three duplicated axes from drifting apart from `ApplyContext`.

Nothing here is a blocker on the mechanism itself — the schema points are the ones I'd want settled before the table exists in prod.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
lucas-xzp: 三条列都加了（apply_id varchar(64) / category varchar(32) / entry_identity varchar(256)，均可空，KEY `idx_apply`），store() 以可选参数收下它们——keep_last 复用与手工取数不强制要求；列宽与既有纪律同一条 before-the-blob 校验。`idx_digest` 按你给的第二条路处理：删（无读者，纯 append 税），随读操作的 PR 回来。行已写成空表——这正是这三个列只在现在有意义的原因。W5 的取数管线在其 PR 里把 ctx.apply_id 与条目 identity 传进 store()，届时 NULL 只剩手工取数一种来源。

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
lucas-xzp: 措辞按你的措辞落地：DDL 头、service docstring、README 统一改为「平台为这个 bot 拉取过什么」，并在 DDL/README 里写明「单读本表会高估（§3.2 all-or-nothing 下取了未必下发），补全答案 join apply_id」——这样审计者读文档就能得到你在帖子里给的准确语义；下发与否则留在 apply report。

src/backend/src/agentclaw/community/core/bot_config_manifest/content/service.py
lucas-xzp: 采纳补括号：host 含 ":" 时重新上括号再拼 authority，三个 IPv6 形态（带/不带端口、长地址+子路径）进了 `test_stored_urls_carry_path_but_never_query_or_userinfo` 的参数化。punycode 与 percent-decoding 两条 fidelity 备注按文档化处理而非改写：docstring 写明存的是 httpx 规范化渲染、字节真值锚在 digest——v1 接受此差异且不再静默。

src/backend/src/agentclaw/community/core/bot_config_manifest/fetch/limits.py
lucas-xzp: `DIGEST_RE` 换成 `\\A…\\Z`，同一常量的两个消费点（fetcher 的 declared pin、store 的 `_require_valid_digest`）以及 W1 校验器里的同形副本一并对齐。两边各加一条测试：fetcher 对换行尾 pin 抛 `FetchRefusedError`（不再先过验证后报 mismatch），store 对它抛 `untrusted content address`（不再伪装成 404 家族）。

src/backend/src/agentclaw/community/core/bot_config_manifest/content/errors.py
lucas-xzp: 采用你的首选方案：新增 `ContentStoreFault(RuntimeError)` 作 500 家族基类，`ContentIntegrityError` 移到它下面（不再继承 400 家族的 ``ContentStoreError``）；`ContentMissingError` 维持 400 家族。基类 docstring 写明两个方向的坑（catch 基类做 400 映射会漏，catch 基类做清理也漏——sibling 让两种错都不可能发生）。此刻文件无服务外消费者，是最便宜的改层次时点。

src/backend/src/agentclaw/community/core/bot_config_manifest/content/service.py
lucas-xzp: 两部分都按你给的顺序落地。(a) URL 长度进 W1 准入：`schema/limits.py` 新增 `MAX_SOURCE_URL_CHARS = 2048`（与列宽同数，同处声明），`check_https_url` 越限报 `source_url_too_long`，schema 文档 §5 加行，另有超限拒绝与 2048 恰好两条测试——PUT 时可见的长度不再留给取完才拒。(b) `content_type` 降级为 advisory：超宽存 NULL 加告警日志（拒绝截断——半截媒体类型看着像真的），收据与 blob 保留，测试改为「收据存活、字段为 NULL」。store 的宽度校验留作最后防线，专管 PUT 看不见的 redirect Location 长度——README 与 schema doc 各自写明这两个层次的分工。

src/backend/src/agentclaw/community/core/bot_config_manifest/content/service.py
lucas-xzp: fsync 采纳：temp fd 上 flush+fsync 后再 os.replace；目录 fsync 的取舍在 docstring 里写明（丢的是名字而非字节，下次同内容 store 会按尺寸 dedup 重写覆盖）。.tmp 孤儿采用年龄阈值清扫：store 时对本分片内 mtime 超过 1 小时的 `.tmp-*` 清除——不误删并发写手的在写临时文件；同时在留存策略（DDL + README）里把它立为唯一明文允许清扫的例外，`test_crashed_write_temp_files_are_age_swept_but_fresh_ones_kept` 钉住两侧（陈旧被收、新鲜保留、store 本身成功）。

src/backend/src/agentclaw/community/core/bot_config_manifest/content/service.py
lucas-xzp: docstring 重写：dedup 快捷路径写明「stat 与 read 的取舍只发生在该路径（audit read 无条件全量哈希在 read 里）」，删掉了把 re-store 与 audit fetch 混在一处的成本论证；同时把「出手即可证明，否则失败」明确限定为 read() 的保证，store() 自身只对它亲手写入的那一次负责——快路径成功即落行、不对盘上既有字节复验，写入 docstring 而非留给读者推断。

src/backend/src/agentclaw/community/core/repository/implementations/bot/manifest_content.py
lucas-xzp: 采纳拒绝而非静默：`records_for` 对非正 limit（含 0）抛 ``ValueError``，协议 docstring 同步改为「empty 在审计语义里是断言：bot 没有收据；调用方分页算错应该被告知坏请求」；原先钉住 -1 clamp 为 [] 的测试按此裁决重写为 raises ValueError（测试文件里注明这是评审裁决的有意替换）。

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
lucas-xzp: 幻影列说法已重写为决策句：DDL 与 README 各写「git ref/SHA 是 W7 的列与形状，届时随 W7 落地；本表 never-update 政策下，W7 之前写的行对它们永久 NULL、无回填」，并把这句话的成立同样写给本轮新加的三个 apply 关联列——同一原则，一处说清。

xianmuyq (approved):

PR #1764, opened 2026-09-01
fix(backend): accept opaque Skill Center package wrappers

FreddieSun:
## Problem

Skill Center exact Team downloads wrap content in an opaque UUID or skillCode directory. Team materialization incorrectly treated that transport directory as the Skill name and rejected an otherwise valid published package.

## Solution

- Use one exact-Skill-Center ZIP validator for both Public and Team downloads.
- Ignore only the opaque SC transport wrapper; retain strict package structure validation.
- Retain the Team `SKILL.md.name == ac_skill.name` invariant after normalization.

## Validation

- `100 passed` across package validation, materializer, publication repository/schema regression, and oversized-module guard.
- Ruff and `git diff --check` passed.
- Standards/Spec review completed with no blocking findings.

## Compatibility and risk

No API, publication-state, retry, database schema, or Gateway OpenAPI artifact changes. `validate_public_center_zip()` remains as a compatibility alias for existing Public callers.


totalfrank (dismissed):
LGTM.

totalfrank (approved):
LGTM.

PR #1766, opened 2026-09-01
chore(backend): log lazy materialization stage timings

FreddieSun:
## Problem
Lazy SkillCenter materialization previously exposed only one scanner duration and the terminal batch status, so the post-download tail could not be attributed to individual materialization, final SkillSet add, or runtime projection.

## Solution
Add safe duration-only logs for exact Version stages, per-Reference resolution/materialization/Track Latest, batch barriers, final SkillSet add, and mutation/projection sub-stages. The logs retain correlation IDs and counts but never emit presigned download URLs or package content.

## Validation
- `uv run ruff check` for the three changed Backend modules: passed.
- `uv run pytest -q tests/community/core/skill_center/test_skill_version_materializer.py tests/community/core/skill_center/test_skill_center_reference_processor.py`: 17 passed.
- `git diff --check`: passed.

## Compatibility and risk
No API, schema, status-machine, queue, or projection behavior changes. This adds INFO/WARN observability only; the added log volume is proportional to materialization stages and control-plane projection calls.

### DIFF
--- a/src/backend/src/agentclaw/community/core/skill_center/services/_mutation_flow.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/_mutation_flow.py
@@ -10,6 +10,7 @@ the two services cannot drift.
 
 from __future__ import annotations
 
+import time
 from collections.abc import Callable, Sequence
 
 from agentclaw.community.core.bot_management.readiness import is_bot_ready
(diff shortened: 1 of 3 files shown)


totalfrank (approved):
LGTM.

PR #1775, opened 2026-09-01
fix(backend): route community runtime through BaaS

evilcoda:
## Summary
- keep BaaS-managed ARCA-backed targets on the BaaS connection path when the community profile has no local ARCA runtime
- implement the community engine-adapter transport through BaaS `/http-info` for normal and streaming requests
- route AICoding workspace clone by `device_provider`, preserving the existing sandbox proxy path for non-BaaS devices

## Compatibility
- corp deployments with a real ARCA sandbox client keep the existing direct proxy behavior
- non-BaaS workspace cloning remains unchanged

## Verification
- `ruff check` on changed Python files
- focused runtime, workspace, session and wiring tests: 318 passed
- community architecture tests: 252 passed
- community DI tests: 310 passed
- repository pre-push SAST gate passed

### DIFF
--- a/src/backend/src/agentclaw/community/core/aicoding/README.md
+++ b/src/backend/src/agentclaw/community/core/aicoding/README.md
@@ -13,6 +13,7 @@ provides:
 consumes:
   - "BotService"
   - "DeviceService"
+  - "BaasService"
   - "WorkspacePathFactory"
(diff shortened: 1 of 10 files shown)


totalfrank (approved):
LGTM.

PR #1777, opened 2026-09-01
fix(backend): default BCN registration when DRM is unavailable

yhzhang35:
## Problem

Eligible Bots skipped BCN Provider registration whenever the DRM reader was unavailable or the `ClaudeCodeBcnRegister.enabled` value was not configured. This made automatic enrollment depend on optional DRM infrastructure.

## Solution

- Default registration to enabled when the DRM reader attribute is missing, reading it raises, or its value is empty.
- Preserve explicit control: only `true`, `on`, and `1` enable a non-empty DRM value; other non-empty values disable registration.
- Add low-sensitivity diagnostics that record the fallback reason and exception type without logging DRM values or exception messages.
- Cover the helper defaults and the real Claude Code `start_bot` registration path.

## Validation

- `uv run pytest -q tests/community/core/bot_management/services/test_bot_service_stop_start.py tests/community/core/bot_management/services/test_bot_service_create_bcn_register.py tests/community/core/bot_management/services/test_bot_service_restart_baas_envs.py` — 85 passed.
- `uv run ruff check tests/community/core/bot_management/services/test_bot_service_stop_start.py` — passed.
- `uv run python -m py_compile src/agentclaw/community/core/bot_management/services/bot_service.py` — passed.
- `git diff --check origin/dev...HEAD` — passed.

## Compatibility and risk

This changes the default for unavailable or empty DRM configuration from disabled to enabled. Deployments that need to disable BCN registration must set a non-empty explicit disabling value.

## Spec

`src/backend/specs/2026-09-01-claude-code-bcn-drm-default-enabled/009-pr-report.md`

### DIFF
--- a/src/backend/specs/2026-09-01-claude-code-bcn-drm-default-enabled/009-pr-report.md
+++ b/src/backend/specs/2026-09-01-claude-code-bcn-drm-default-enabled/009-pr-report.md
@@ -1,0 +1,38 @@
+# PR Convergence Report: claude-code-bcn-drm-default-enabled
+
+## Scope
+
+- Worktree / repo: `/Users/helloworld/Desktop/codes/teamclaw_worktrees/Avernet_worktrees/claude-code-bcn-drm-default-register-dev-20260901` / GitHub `inclusionAI/Avernet`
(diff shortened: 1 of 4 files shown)


totalfrank (dismissed):
LGTM.

totalfrank (approved):
LGTM.

PR #1778, opened 2026-09-01
fix(backend): delete failed unpublished skill drafts

FreddieSun:
## Summary

- treat terminal `FAILED` publication attempts as disposable when deleting a never-published Skill with no other external facts
- delete the failed attempts together with the unpublished Skill aggregate instead of leaving a Draft-less shell
- invalidate held Team Draft leases when only the Draft is discarded, preserving monotonic fencing tokens
- align the Phase 2 spec and frontend integration guide with the corrected lifecycle

## Regression coverage

- first Team Draft + failed first publication deletes the full Skill aggregate and failed Attempt
- published Skill + failed Attempt still preserves history and only deletes the upgrade Draft
- deleting a Team upgrade Draft clears the lease holder and increments the fencing token

## Validation

- `uv run --project src/backend pytest -q src/backend/tests/community/core/skill_center/services/test_space_skill_application_service.py src/backend/tests/community/repository/skill_center/test_space_skill_draft_repository.py` (29 passed)
- `uv run --project src/backend ruff check ...`
- `git diff --check`

### DIFF
--- a/src/backend/specs/2026-08-20-skill-capability-upgrade/frontend-api.md
+++ b/src/backend/specs/2026-08-20-skill-capability-upgrade/frontend-api.md
@@ -338,9 +338,11 @@ DELETE .../draft?expected_revision_id=rev-2&fencing_token=7
 ```
 
 - `deleted_scope=DRAFT`：仅放弃本次升级；Published Vn 仍在线；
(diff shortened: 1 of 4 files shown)


totalfrank (approved):
LGTM.

PR #1782, opened 2026-09-01
fix(engine): resolve favorites beyond session list cap

evilcoda:
## Summary
- resolve favorite sessions by exact session key instead of scanning the capped gateway session list
- bound concurrent lookups while preserving favorite order and post-filter pagination
- add regression coverage for exact lookups and stale favorite records

## Verification
- session favorites router tests: 14 passed
- OpenClaw session port tests: 47 passed
- pre-push SAST gate passed
- validated against an OpenClaw 2026.5.12 runtime: favorite list restored all 3 records, including 2 sessions outside the default 100-session window


totalfrank (approved):
LGTM.

PR #1787, opened 2026-09-01
fix(backend): trust Skill Center exact MCP metadata

FreddieSun:
## Summary
- Skip the remote Scanner for every Skill Center exact-version materialization, covering both PUBLIC catalogue materialization/sync and TEAM/Space publication.
- Treat exact-download `mcpServices` as the authoritative Version dependency source; null/missing/empty remains dependency-free, and non-empty entries are normalized by `serverCode` into exact Version metadata.
- Keep SHA256, ZIP/path safety, canonical `SKILL.md`, TEAM manifest-name consistency, and Canonical Store verification as Ready Gates; Local/Repo Scanner flows are unchanged.
- Make SkillSet, Direct, Default-exclusion, and legacy permission reads resolve Center dependencies from the highest PUBLISHED `ac_skill_version`, while preserving cleanup of historical unresolvable/offline Center members.
- Remove the now-unused Materializer-specific Scanner protocol, SDK adapter, DI binding, and test fake while retaining an explicit DI test for the validated Gateway seam.

## Compatibility
- No DDL or OpenAPI wire change.
- OCB already maps exact-download `mcpServices` for both PUBLIC and TEAM requests; no OCB production change is required.
- Inactive SkillSet membership remains compatible with historical Center rows without a Version; activation/runtime-affecting writes still require resolvable PUBLISHED metadata.

## Validation
- Backend full suite: 16020 passed, 58 skipped.
- Center/Publication/Reference/Auth/Runtime/DI narrow suite: 321 passed.
- Follow-up DI seam suite: 30 passed.
- Ruff, diff-check, and local Python SAST passed.
- Standards/Spec two-axis review: one DI test-coverage P2 found and fixed; no remaining P0-P2 findings.


totalfrank (approved):
LGTM.

PR #1793, opened 2026-09-01
fix(backend): clarify SkillSet add conflicts

FreddieSun:
## Summary
- Distinguish a true Direct Installation from an Installation materialized by an active ordinary SkillSet or unexcluded System Default membership.
- Return `RESOURCE_ALREADY_IN_ANOTHER_SKILL_SET` for Default/active Set members added to another Set; retain `RESOURCE_DIRECT_ACTIVE` only for a true Direct-effective capability.
- Apply the same ownership classification to Skills and MCPs without re-running a full Bot flush plan for every batch member.
- Keep legacy BFF `failed[].error` unchanged and add `error_code` plus user-facing Chinese `message`.

## Compatibility
- No DDL or OpenAPI change.
- Legacy batch callers that consume `failed[].error` continue receiving the stable raw code.
- Frontends can render additive `failed[].message` and fall back to `error` during rollout.

## Validation
- Focused UoW/BFF/ownership/service/architecture suite: 233 passed.
- Backend full suite: 16388 passed, 59 skipped.
- Ruff, local Python SAST, diff check passed.
- Standards/Spec dual review: one batch-performance P2 found and fixed; no remaining P0-P2.

### DIFF
--- a/src/backend/specs/2026-08-20-skill-capability-upgrade/spec.md
+++ b/src/backend/specs/2026-08-20-skill-capability-upgrade/spec.md
@@ -198,6 +198,9 @@ PUT/GET/DELETE /bots/{bot_id}/skills/{skill_id}/installation
 - 同一 Bot 下，同一 Skill/MCP 最多属于一个 SkillSet，System Default 包含在内；excluded
   仍然属于 Default。
 - Direct active 资源不能加入任何 SkillSet；必须先 Direct deactivate。
+- Installation 行存在本身不证明 Direct：若该行可由 active ordinary SkillSet 或未 exclusion
+  的 System Default Membership 解释，则资源是 Set-managed，加入第二个 Set 时返回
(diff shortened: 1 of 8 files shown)


totalfrank (approved):
LGTM — reviewed `92acbcaf` (8 files: the Skill/MCP ownership-classification change, its batch-message plumbing, and tests).

Traced the new `_has_active_skill_set_claim`/`_has_active_mcp_set_claim` logic against `_resolve_flush_plan`'s Default/ordinary-Set activity semantics — consistent. Confirmed `is_directly_active = installed and not active_set_claim` preserves R2-before-R3 precedence (`test_direct_active_precedes_membership_when_both_forbid_joining` still gets `RESOURCE_DIRECT_ACTIVE` since that membership's Set is inactive/non-default) while the Default/active-ordinary-Set case now correctly falls through to `RESOURCE_ALREADY_IN_ANOTHER_SKILL_SET`. No bugs found in the new code.

---

PR #1802, opened 2026-09-02
feat(backend,gateway): add user-level application admission control

totalfrank:
## Problem

Bot-level admission control exists for application principals: an application acting alone (its own credential, no human on the wire) reaches a bot-addressed operation only under a live bot grant from the named user. But not every public operation is bot-level. The `USER_GATED` operations name no bot and answer about the user's account: the bot-creation ceiling, routines across every bot, Spaces, work orders and notifications, local devices and their file trees, the repository catalog, the Skill Center sync.

Those had no record of their own. They borrowed the bot record: an application was admitted if it held *any* live bot grant from the named user. That is the wrong consent (one bot grant opened the whole account), cannot be withdrawn on its own, and was enforced by a `_require_user_delegation` helper copied into five routers and written inline in two more, with no dependency the route inventory test could hold a `USER_GATED` route to. Several `USER_GATED` operations (the four repository-catalog reads, the Skill Center sync) carried no check at all and were reachable by an application naming any user.

## Solution

**A user-level (account-level) grant record**, the sibling of the bot-level one:

- `core/user_app_grant/`: `ac_user_app_grant` (live rows, unique on tenant/app/user/env) and `ac_user_app_grant_log` (append-only), `UserAppGrantService` behind `UserAppGrantServiceProtocol`, deployable DDL under `sql/`, README with context boundary. Repository protocol and ORM implementation in the `bot` repository domain beside the bot grant; DI module; Service API re-export; conformance pair.
- **Three operations** at `/openapi/v1/org/user/authorized-apps`: grant (user + app on the wire, application read off the principal), list and withdraw (user only). All three `REFUSED` to a machine caller. Under the org group rather than `/bots` because the record is a property of the user, and the gateway routes `/openapi/v1/org/**` to the same backend.

**`USER_GATED` now means exactly that record.** `ActingCaller` gains `user_grants` and `require_user()`; `principal.py` gains `require_granted_user` / `DelegatedUserIdDep`, declared on every `USER_GATED` route (ceiling, routines aggregate, repository catalog ×4, Skill Center sync, Spaces ×6, Space skills ×10, work orders and notifications ×9, local devices ×2). `test_admission_inventory.py` holds the set of routes declaring it equal to the table's `USER_GATED` rows in both directions, the same way it already does for the two bot-grant dependencies. The bot-grant proxy and the five helper copies are deleted; `granted_bot_ids()` loses the `owned_by_delegator` switch since only the owner-scoped listings still call it.

The skill README read derives its actor from the verified principal and publishes no `user_id`, so an application acting alone has nobody to act as; it is relabelled `REFUSED` (which is what its handler already did) and declares `refuse_app_only_caller`.

**Gateway**: `POST /openapi/v1/org/user/authorized-apps` requires user + app; the README read requires a user. Pinned in `test_route_security.py`.

**`authorization.py`** was at 999 lines; the three new rows pushed it over the 1000-line cap, so its assembly-time assertion (`assert_every_route_authorized` and helpers) moved verbatim to `authorization_inventory.py`. The table and the route class stay where they were.

Alternatives rejected (recorded in the spec): keeping the bot-grant proxy as a fallback (would make the new record meaningless on exactly the operations it exists for); letting a user-level grant imply bot access (recreates the over-broad consent the other way); a new admission mode (`USER_GATED` already names the shape, only its proof changes).

## Validation

Ran locally against the backend `.venv` (synced from pypi.org, since the pinned Aliyun mirror is unreachable from this environment; `uv.lock` is unchanged):

- `tests/community/adapters/http/openapi_v1`, `tests/community/core/user_app_grant`, `tests/community/core/bot_app_grant`, `tests/community/architecture`: full run 1818 passed with 15 failures, all of which were either stale versions of two test files fixed mid-run or the architecture/pin gates fixed afterwards. Re-run of `tests/community/architecture` plus every affected inventory, listings, refusal, seam and router test file: **515 passed**.
- Gateway `tests/unit/core/authn/test_route_security.py`: **74 passed**.
- `python_sast_local.sh`'s flake8 rule set over every changed Python file: clean. `ruff check` over the same files: clean apart from one pre-existing unused import in `test_module_boundaries.py` that this change does not touch.
- New tests: `tests/community/core/user_app_grant/test_grant_service.py` (real SQLite: idempotent grant, log outliving the row, grant/withdraw twice through the unique key, insert race, identity length) and `tests/.../authorized_apps/test_user_router.py` (three operations, both identity postures, machine-caller refusal). `test_app_only_listings.py` now pins that the ceiling and device reads admit with the account-level grant and refuse with a bot grant alone.

Not run: singlebox coverage and the pre-push module gates (no product stack in this environment), and the gateway's `dump_and_publish.sh` regeneration of `configs/schemas/bots.openapi.json`, whose gate refused the candidate for unrelated pre-existing drift (task DTOs already absent from the current backend document), so the checked-in artifact is untouched and needs its usual coordinated publish.

## Compatibility and risk

- **Behaviour change for integrations**: an application that reached a `USER_GATED` operation on the strength of a bot grant now receives the masked 404 until the user grants it account-level access via `POST /openapi/v1/org/user/authorized-apps`. One re-consent per integration. The skill README, previously effectively refused to app-only callers with a 401, now refuses at the edge too.
- **Schema**: two new tables, `CREATE` only, applied out of band from `core/user_app_grant/sql/2026_09_02_user_app_grant.sql` (local SQLite gets them from `create_all`). No existing table changes.
- **Config**: two `route_security` rules added in `src/gateway/configs/application.yaml`.
- **Rollback**: revert the PR; the two tables can stay, unread.

## Spec

`src/backend/specs/2026-09-02-openapi-v1-user-app-grant/` (spec, plan, tasks). Sibling of `specs/2026-08-10-openapi-v1-bot-app-grant/`.

https://claude.ai/code/session_01V5Hhq9zFrGon55ouNskofs

---
_Generated by [Claude Code](https://claude.ai/code/session_01V5Hhq9zFrGon55ouNskofs)_


totalfrank:
**Backend unit tests** went red on the merge commit `80545626`, not because of this PR.

The two failures are in `tests/community/core/bot_management/test_application_coding_create.py`, which this PR does not touch, and they reproduce identically on a clean `origin/dev` checkout at `1364130f`. On dev, #1851 removed the space-kind gate for coding bots and #1853 moved the `engine_properties` envelope to `template_type` / `template_config`, but one test still sends the old `template` key and another still expects a Team Space to be refused. Dev's own push CI did not exercise that file.

PR #1857 (test-only) fixes exactly those two tests. I ported the same change here as `33cdfaaf` so this branch can go green; it no-ops once dev carries #1857.

---

PR #1804, opened 2026-09-02
feat(bcs-http): dispatch bcs-fuse visibility sync on allowlisted provider bot registration

iParadox-ant:
## Summary

When a provider bot is registered through the `register_provider_bot` HTTP
route and the provider is allowlisted for switch-provider, the freshly created
bot must be visible to bcs-fuse without waiting for the next bot-list/refresh
cycle. This change wires a one-shot, fire-and-forget visibility sync dispatch
from the HTTP handler into bcs-fuse immediately after a successful creation.

`RegisterProviderBotOutcome` is extended with three application-internal fields
populated by the bcs-bot application layer:

- `created: bool` — `true` only when this call actually wrote the bot/binding
  (i.e. not a gateway short-circuit / idempotent replay);
- `capabilities: Option<BotCapabilities>` — the freshly read worker
  capabilities for a newly created bot, `None` on replay or transient miss;
- `actor_kind: ActorKind` — kept as `ActorKind::Bot` for the provider-bot path
  so the handler's sync gating stays explicit and consistent with other
  onboarding paths.

The HTTP handler dispatches a `VisibilitySyncRequest` only when
`allowed_switch_provider && outcome.created && outcome.actor_kind == ActorKind::Bot`
and `capabilities` is present. If `capabilities` is `None` (e.g. an eventual
registry miss right after creation), the handler logs a warning and skips the
sync, never blocking the registration response. Non-allowlisted providers and
idempotent gateway replays (`created=false`) are unchanged and never trigger
a sync.

## Why fire-and-forget is safe

- `bcs-fuse sync_worker` is idempotent per `worker_id`, so a duplicate
  fire-and-forget sync (e.g. the pre-existing duplicate-registration race
  window) is harmless.
- The rare `registry.get` miss right after creation degrades to a skipped sync
  for that single event; subsequent bot-list/refresh paths still discover the
  bot. The warn log keeps the miss observable.

## Changed files

- `src/bcs/crates/service-api/bcs-service-api/src/application/provider.rs` —
  extend `RegisterProviderBotOutcome` with `created`, `capabilities`,
  `actor_kind` (non-serialized, application-internal).
- `src/bcs/crates/services/bcs-bot/src/application/provider.rs` — populate the
  new fields in `ProviderManagement::register_provider_bot`; read capabilities
  from the registry only on the created branch with a warn-on-miss fallback.
- `src/bcs/crates/adapters/http/bcs-http/src/routes/providers.rs` — dispatch
  `VisibilitySyncRequest` for allowlisted + created + Bot actors; warn and skip
  when capabilities are missing.
- `src/bcs/crates/adapters/http/bcs-http/tests/provider_routes_contract.rs` —
  add `RecordingVisibilitySyncPort`, optional `with_visibility_sync` wiring on
  `HttpAppState`, and three contract tests.

## Tests

Local per-module test runs (using the local cargo build cache):

- `cargo check -p bcs-service-api -p bcs-bot -p bcs-http --tests`
- `cargo test -p bcs-http --test provider_routes_contract` — 50 passed
- `cargo test -p bcs-service-api` — 9 passed
- `cargo test -p bcs-bot` — 52 passed

New contract cases:

- `register_provider_bot_dispatches_visibility_sync_for_allowlisted_provider`
- `register_provider_bot_skips_visibility_sync_for_non_allowlisted_provider`
- `register_provider_bot_skips_visibility_sync_for_idempotent_gateway_replay`

No new failures introduced by this patch. One preexisting `dead_code` lint
warning in `bcs-http` test helpers (`bots_contract.rs`) is unrelated and not
affected by this diff.

Full per-module `src/backend/scripts/ci_test.sh --base <dev> --head HEAD` is
deferred to PR required checks per project convention.

## Security / privacy

- `capabilities` (including `agent_code`) is application-internal,
  `skip_serializing` on the wire, and is **not** serialized into the HTTP
  response; the response builder ignores it.
- The public diff privacy scan found no introduced or preexisting sensitive
  hits on the changed files.

## Risks

- The duplicate-registration TOCTOU between the duplicate check and the write
  is pre-existing; the new `created` flag inherits that window. At worst a
  duplicate fire-and-forget sync reaches an idempotent `sync_worker`. No data
  corruption, no behavioral regression.
- A rare post-creation `registry.get` miss means visibility sync is skipped
  for one registration event. Non-blocking by design; the warn makes it
  observable.

## Rollback

Revert this commit. The new `RegisterProviderBotOutcome` fields are
application-internal and not serialized, and the sync dispatch is additive, so
reverting does not change any public HTTP response shape. No data migration is
involved.

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/providers.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/providers.rs
@@ -15,7 +15,7 @@ use bcs_service_api::application::v1::{
     PatchBotInternalAttributes, UserVisibility,
 };
 use bcs_service_api::{
-    BotUseCaseError, CoordinationMode, DeleteProviderBotCommand, ProviderAuthMode,
+    ActorKind, BotUseCaseError, CoordinationMode, DeleteProviderBotCommand, ProviderAuthMode,
     ProviderBotBinding, ProviderBotConnectionMode, ProviderBotRosterItem,
     ProviderBotTaskModesFilter, ProviderCoordinationConfig, ProviderOrganizationManagementConfig,
     ProviderRecord, RegisterProviderBotCommand, RegisterProviderCommand, ServiceError,
@@ -27,7 +27,7 @@ use serde_json::{Map, Value, json};
 use tracing::{info, warn};
 
 use crate::mapping::capabilities::{to_core_skill, to_wire_skill};
-use crate::state::HttpAppState;
+use crate::state::{HttpAppState, VisibilitySyncRequest};
 
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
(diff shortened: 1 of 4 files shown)


vzvince:
@codex review

vzvince (approved):

PR #1822, opened 2026-09-02
fix(bcs): allow chat on completed sessions

RaymondHX:
## Problem
Completed sessions cannot continue chat because the legacy session chat endpoint returns HTTP 409 when the session status is not Running. The session status is no longer meaningful for chat except for service invocation, so this guard blocks valid chat requests.

## Solution
Remove the obsolete Running-status guard from the legacy session chat endpoint while retaining participant and routing validation. Add regression coverage proving a Completed session can chat and that the response preserves the session ID.

## Validation
- `cargo test -p bcs-http --test group_messages_contract` — 20 passed.
- `cargo fmt --all -- --check` — not clean because of pre-existing workspace formatting differences outside this change; no unrelated files were reformatted.

## Compatibility and risk
This relaxes only the legacy session chat status check. Service invocation status handling is unchanged. The change is limited to request validation and is reversible by reverting commit `93e1ade52`.

### DIFF
--- a/src/bcs/crates/adapters/http/bcs-http/src/routes/sessions.rs
+++ b/src/bcs/crates/adapters/http/bcs-http/src/routes/sessions.rs
@@ -1312,16 +1312,7 @@ pub async fn session_chat(
         Err(e) => return session_error_to_response(&e),
     };
 
-    // 2. Session must be running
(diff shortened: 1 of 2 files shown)


questchaos (approved):
LGTM

PR #1824, opened 2026-09-02
refactor(backend): route BBS execution through start_run

regrecall:
建议 PR 标题：

`refactor(backend): route BBS execution through start_run`

## Problem

BBS execution had a dedicated `run_bbs` path alongside `TaskRunner.start_run`. This leaked execution-mode details into the runner API and bypassed the common task execution flow.

It also meant that upper-layer changes carried by `TaskNode` could be lost when BBS execution reloaded the persisted graph, while BBS handoff nodes could report successful dispatch without performing a real delivery.

## Solution

- Removed the dedicated `TaskRunner.run_bbs` and `TaskExecutor.run_bbs` methods.
- Routed BBS execution through the common `TaskRunner.start_run` and `TaskExecutor.dispatch` flow.
- Made the `TaskNode` passed to `start_run` authoritative for BBS execution while retaining graph relationships and related task context.
- Preserved the persisted root task by creating an isolated BBS execution view.
- Routed claimed BBS handoff tasks to the selected bot through `single_bot` delivery, avoiding recursive BBS dispatch.
- Passed real backend tasks as one batch so executor-level concurrency limits apply consistently.
- Added logging for background dispatch results that return failure.
- Updated existing tests that encoded the previous BBS no-op behavior.

## Validation

- Task domain test suite: `636 passed, 30 skipped`.
- Backend architecture test suite: `258 passed`.
- Focused BBS and task-runner tests: `20 passed`.
- Python compilation check passed.
- `git diff --check` passed.
- Pre-push Python SAST/lint gate passed.
- Verified that no calls or definitions remain for `TaskRunner.run_bbs` or `TaskExecutor.run_bbs`.

The 30 skipped tests require pre-production, Singlebox, or live Bot credentials and were not run in the local environment. The pre-push hook ran in lint-only mode, so its coverage and live E2E stages were not executed.

## Compatibility and risk

This change does not modify public HTTP APIs, persistence schemas, or configuration formats.

It removes internal `run_bbs` methods. Repository-wide searches found no remaining callers. The main behavioral change is that BBS nodes now perform real adapter dispatch instead of succeeding as a no-op.

Rollback is available by reverting commit `3cce40354`.

## Spec

Implements the task execution module requirement that BBS execution must use the unified `start_run` entry point, allowing upper-layer business changes to remain part of the task execution request.

### DIFF
--- a/src/backend/src/agentclaw/community/core/task/domain/models.py
+++ b/src/backend/src/agentclaw/community/core/task/domain/models.py
@@ -229,8 +229,7 @@ class TaskExecutionGraph:
     extend_props: dict[str, Any] = field(default_factory=dict)
     execution_graph: dict[str, Any] | None = None  # 回调审计图快照(BCN/ClawMind DAG,按 session_id 反查挂图级;只读投影)
     task_id: str = ""   # 整图所属任务 ID(initialize 透传;query_task_dashboard 子树投影复制;
-                        # 供 bbs_runner.notify 经 _schedule_bbs_notify→run_bbs 取 task_id,
-                        # 因 run_bbs 链路只拿到 execution_graph、无法回溯 task_id)
+                        # 供执行 adapter、回调投影和 BBS 图查询使用)
     # 派生不持久: depth / child_tasks / parent_task(均从 relations 分解树派生)
 
     @property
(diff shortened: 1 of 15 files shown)


jiangj0627 (approved):
LGTM

PR #1872, opened 2026-09-03
  feat(agent): reset openclaw workspace HEARTBEAT.md from image template on startup

xxxxpenny:
openclaw appends tasks to ~/.openclaw/workspace/HEARTBEAT.md at runtime,
    and /home/admin is NAS-mounted, so agent-generated heartbeat content
    persists across pod restarts and stale tasks leak into fresh runs.
    
    Ship the pristine empty template in the image at /opt/config/HEARTBEAT.md
    (NAS mount shadows anything baked under /home/admin) and have
    start_openclaw.sh unconditionally copy it over the runtime file in
    Step 1, before openclaw's first on-demand start. Missing template only
    warns, keeping direct-recovery invocation of the script safe.
    
    Template-wins here — the opposite of claude-settings.json's mount-wins
    staging — since the goal is discarding persistent runtime state.

### DIFF
--- a/docker/agent/HEARTBEAT.md
+++ b/docker/agent/HEARTBEAT.md
@@ -1,0 +1,5 @@
+# HEARTBEAT.md
+
+# Keep this file empty (or with only comments) to skip heartbeat API calls.
+
+# Add tasks below when you want the agent to check something periodically.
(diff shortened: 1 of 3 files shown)


cassiuscai (approved):
LGTM

PR #1874, opened 2026-09-03
fix(bots/all): expand engine=aicoding query to the claude_code form

lucas-xzp:
## Problem
PR #1719 (engine/form vocabulary split) made `aicoding` an internal runtime form of `claude_code`, not a product engine: new AICoding bots store `active_engine='claude_code'` with the `engine_form='aicoding'` marker on their template snapshot, while only pre-#1719 bots keep a literal `active_engine='aicoding'`. `GET /openapi/v1/bots/all?engine=aicoding` filters by exact-match on `active_engine` (`repository/implementations/bot/bot.py`), so it silently returns only legacy bots and drops every post-#1719 AICoding bot — the exact set the frontend queries for. The inventory card also carries no `template_config` (`_build_item` leaves it null), so the frontend cannot recover the form client-side.

## Solution
Expand `engine=aicoding` at the `BotInventoryService` application layer only (`core/bot_inventory/services/bot_inventory_service.py`):
- `_list_cloud_rows` resolves `_cloud_fetch_engines(engine)`, which turns `aicoding` into two fetch batches — legacy `active_engine='aicoding'` plus `active_engine='claude_code'` — then collapses them with the existing `uses_aicoding_runtime` predicate (`core/workspace/runtime_identity.py`). The predicate keys off the row's `template_type` (applicationCoding/personalCoding), so **no extra template read** is needed; legacy `aicoding` rows short-circuit on the engine arm.
- Extracted `_fetch_cloud_pages` (mechanical loop split, zero behavior change) and two module-level pure helpers `_cloud_fetch_engines` / `_select_aicoding_runtime_rows` for direct unit testing.

Pagination already happens in the service layer (`_list_cloud_rows` fetches all matching rows, `list_items` slices), so the two-batch fan-out + collapse happens **before** the slice — no pagination drift. No frontend / repository SQL / Protocol / router changes.

## Validation
- Affected surface (bot_inventory module + architecture nails): 342 passed.
- Service unit tests: 27 passed (15 existing + 12 new — two-batch expansion, legacy+form inclusion, plain-claude_code exclusion, dedup, post-filter pagination, owner/attach forwarding, non-cloud bot_type drop, short-page-without-total exit, claude_code/teclaw regression guards, two pure-helper branch tests).
- ruff clean on both files.
- changed-line coverage 98.0% (CI gate ≥80%); the only 2 uncovered lines are pre-existing visible-filter `BotInventoryPermissionError` handler paths unrelated to this change.

## Compatibility and risk
- `engine=claude_code` keeps its prior exact-match semantics (returns all claude_code rows incl. aicoding form) — unchanged by design.
- Legacy `active_engine='aicoding'` rows still matched via the first batch — no stored-engine rewrite on the read path.
- `engine=aicoding` makes 2 DB round-trips on that query path only; consistent with the inventory's existing fetch-all-then-page model.

## Follow-ups (out of scope)
- `/bots` (non-all, `adapters/http/openapi_v1/bots/router.py:565`) has the same gap but pages in SQL — the same expansion would lose pagination; tracked separately.
- Local/desktop path: no aicoding form exists, intentionally untouched.
- Precision backstop (read `engine_form` from template for the contradictory `normalCC + engine_form=aicoding` data) is documented in code but not needed for any real form.

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_inventory/services/bot_inventory_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_inventory/services/bot_inventory_service.py
@@ -33,6 +33,11 @@ from agentclaw.community.core.bot_inventory.types import (
     ServiceLifecycleCard,
 )
 from agentclaw.community.core.bot_collaborator.models import PermissionLevel
+from agentclaw.community.core.workspace.runtime_identity import (
+    AICODING_ENGINE_FORM,
+    normalize_runtime_engine,
+    uses_aicoding_runtime,
+)
 from agentclaw.community.log import get_logger
 from agentclaw.community.core.bot_inventory.bot_inventory_service_protocol import BotInventoryServiceProtocol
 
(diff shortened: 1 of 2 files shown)


lucas-xzp:
Closing as duplicate: the bot_inventory aicoding expansion (engine=aicoding → two-batch fetch + uses_aicoding_runtime collapse, the _cloud_fetch_engines / _select_aicoding_runtime_rows helpers) is already merged via #1869 (REL20260904 tip b0078981c), which carried this work forward alongside the create_flow template_type passthrough. My mistake — I didn't check for an existing PR on REL20260904 before opening this one.

PR #1882, opened 2026-09-03
fix(workspace): exclude pool skill-center from aicoding rsync snapshot

panqiwp:
## What
Add symmetric rsync excludes for the AICoding pool `skill-center` mount, mirroring the existing `skills-repo` pair.

```diff
 _AICODING_RSYNC_EXCLUDES = [
     ...
     "workspace/skills/skills-center",
     "workspace/skills-pool/skills-repo",
     "workspace/skills-pool/.skills-repo*",
+    "workspace/skills-pool/skill-center",
+    "workspace/skills-pool/.skill-center*",
     ...
 ]
```

## Why
Per the Skill Center Service Artifact handoff: `skill-center` and `skills-repo` under `workspace/skills-pool/` are **both external read-only shared corpora**. The contract is:
- rsync must **skip** both (not copy the corpus into the Artifact snapshot);
- the image-startup script **mounts** both (OCB-owned);
- active-root **symlinks** pointing into the corpora are still **copied normally** via rsync `--links`.

AICoding was missing the `skill-center` pair while already excluding `skills-repo`/`.skills-repo*`, so the two mounts were handled asymmetrically. This makes them consistent and ensures the Center corpus is never copied into the snapshot (the build-time Probe-driven `shared_corpora` exclusion/proof dedups against these static entries, so this is a safe, complementary superset).

## Scope
- **Only the AICoding engine** (`engines/aicoding.py`).
- `openclaw` / `claude_code` untouched. **Hermes intentionally untouched** — it has an existing regression guard `assert all("skill-center" not in item for item in plan.rsync_excludes)` that deliberately excludes `skills-repo` but not `skill-center`.
- Pool center path is `skill-center` (singular), consistent with the layout validation at `service_bot/services/deploy/artifact_build_request.py` (`center.name != "skill-center"`).

## Test
- Added regression test `test_build_plan_excludes_pool_skill_center_symmetric_with_skills_repo` asserting the symmetric excludes for both `skills-repo` and `skill-center` pairs.
- Verified: `128 passed` across `test_engine_sandbox_providers.py`, `test_bot_build_service_rsync_excludes.py`, `test_repo_workspace_excludes.py`, `test_artifact_build_request.py`.

### DIFF
--- a/src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py
+++ b/src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py
@@ -40,6 +40,8 @@ _AICODING_RSYNC_EXCLUDES = [
     "workspace/skills/skills-center",
     "workspace/skills-pool/skills-repo",
     "workspace/skills-pool/.skills-repo*",
+    "workspace/skills-pool/skill-center",
+    "workspace/skills-pool/.skill-center*",
     "workspace/.repos/",
     "workspace/.prewarm_ready.json",
     "skills/*/.git/",
--- a/src/backend/tests/community/core/workspace/test_engine_sandbox_providers.py
+++ b/src/backend/tests/community/core/workspace/test_engine_sandbox_providers.py
@@ -229,6 +229,19 @@ class TestAICodingProvider:
         assert "workspace/.repos/" in plan.rsync_excludes
         assert "workspace/.prewarm_ready.json" in plan.rsync_excludes
 
+    def test_build_plan_excludes_pool_skill_center_symmetric_with_skills_repo(self):
+        # skill-center and skills-repo are both external read-only mounts under
+        # workspace/skills-pool/; both must be rsync-skipped (not copied) and
+        # re-mounted at image startup. Exclude entries must be symmetric:
+        # <name> + .<name>* sidecar guard.
+        provider = AICodingSandboxProvider(workspace=_workspace())
+        excludes = provider.get_build_plan().rsync_excludes
+
+        assert "workspace/skills-pool/skills-repo" in excludes
+        assert "workspace/skills-pool/.skills-repo*" in excludes
+        assert "workspace/skills-pool/skill-center" in excludes
+        assert "workspace/skills-pool/.skill-center*" in excludes
+
     def test_build_plan_keeps_extra_sync_from_claude(self):
         provider = AICodingSandboxProvider(workspace=_workspace())
         plan = provider.get_build_plan()


totalfrank (changes requested):
## 1. Verdict
**REQUEST_CHANGES**

## 2. Gate 0 — target branch

- **Target branch:** `REL20260904` (matches `^REL\d{8}`) — a release branch.
- **PR classification:** existing-behavior bug fix (path a). The diff only appends two literal strings to the existing `_AICODING_RSYNC_EXCLUDES` list in `engines/aicoding.py`; it introduces no new public surface, route, DTO field, config key, table/column, or pipeline. The asymmetry it corrects (missing `skill-center` exclude pair vs. the existing `skills-repo` pair) is a real defect in already-shipped rsync behavior.
- **Cherry-pick check — FAIL.** I diffed `_AICODING_RSYNC_EXCLUDES` between `origin/dev` (HEAD `d68ec78`) and the PR's base `origin/REL20260904` (`ae0c3ba8`): they are byte-identical, and neither contains `workspace/skills-pool/skill-center` / `workspace/skills-pool/.skill-center*`. I also searched open PRs targeting `dev` (`base:dev state:open`) and found nothing that ships this exclude pair. The PR description does not reference a `dev` commit or PR number, and does not say "cherry-pick."
- Per policy: *"a release-branch PR with no `dev` counterpart is a finding on its own — the change would be lost at the next release cut (regression by omission)."* This fix exists only on `REL20260904`. The next release cut from `dev` will silently reintroduce the asymmetric-exclude bug this PR is fixing.
- **Gate 0: FAIL** — not because the change belongs on `dev` instead (it's a legitimate hotfix to shipped behavior on the release branch), but because it is not also landed on `dev`. Required fix: open a companion PR against `dev` with the same two-line change (and ideally the same regression test), and reference it from this PR, before this merges.

## 3. Findings

No catalog (1–11) violations found in the code itself:

- `src/backend/.../core/workspace/engines/aicoding.py` — role: **engine implementation**. This file is `aicoding`'s own per-engine implementation behind the `EngineSandboxProvider` Protocol declared once in `engine_sandbox.py` (contract), alongside sibling implementations `claude_code.py`, `hermes.py`, `openclaw.py`. The change is confined to a `list[str]` literal (`rsync_excludes`) owned entirely by this one engine's implementation — no engine-string branching, no new engine-agnostic file touched, no plugin machinery misused, no core/adapter boundary crossed. Applying the "add engine `foo`" test: this class of change would only ever touch `foo`'s own file, which is exactly what happened here for `aicoding`.
- `src/backend/tests/.../test_engine_sandbox_providers.py` — role: **test**. Adds a symmetric regression test against the one implementation touched; appropriate scope, no contract change to validate against.

The only issue is Gate 0 (missing `dev` counterpart), which forces REQUEST_CHANGES regardless of the clean layering above.

## 4. Out-of-scope notes

- The naming asymmetry between `workspace/skills/skills-center` (plural) and the new `workspace/skills-pool/skill-center` (singular) is asserted in the PR description as intentional and consistent with `artifact_build_request.py`'s layout validation — that's a business-logic/naming judgment call, not an architecture concern, and I did not gate on it.


---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

### Gate 0 — target branch

- **Target branch:** `REL20260904` (matches `^REL\d{8}`, a release branch). `origin/REL20260904` was cut directly from the tip of `origin/dev` (`ae0c3ba`) with **zero prior divergence** — this is the first PR against it.
- **Classification:** Framed as `fix(workspace): ...`, but by diff it is **new behavior**, not a fix to something already broken on this branch. The two new exclude entries (`workspace/skills-pool/skill-center`, `workspace/skills-pool/.skill-center*`) do not exist anywhere in `aicoding.py`'s history and do not exist on `dev` either — this is groundwork for the in-progress "Skill Center Service Artifact handoff" (per the PR body, backed by `specs/2026-08-20-skill-capability-upgrade/spec.md`), landing through the **shared** `_AICODING_RSYNC_EXCLUDES` list that every AICoding bot build already goes through. It is not net-new/unreachable code, so it cannot use the path (b) isolation exception, and it is not a fix to existing shipped behavior on this branch, so path (a) doesn't apply either.
- **Result: FAIL.** No `dev` counterpart exists (no PR/commit cited, and `dev`'s copy of the file lacks both new entries) — this is a **regression-by-omission** risk: the change is silently dropped the next time `dev` is cut into a release branch. Re-target to `dev`, or land on `dev` first and bring this over as an actual cherry-pick citing that PR/commit.

### Findings

**1. `src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py:43-44`** — new engine-specific filesystem path added directly to Backend core code
```python
    "workspace/skills-pool/skill-center",
    "workspace/skills-pool/.skill-center*",
```
- **Violated rule:** `AGENTS.md` § Skills Architecture (explicit, current-repo Invariant): *"Do not add new engine-specific filesystem paths to Backend code. Physical layout ownership belongs to Engine Runtime and its versioned layout contract."* Also catalog #4 (engine/deployment filesystem layout hardcoded in `core/`).
- **Why it breaks the architecture:** This repo already has the designated home for this exact fact — `skill_layout.py`'s per-engine `*PoolPaths` dataclasses (`AICodingPoolPaths`, `OpenClawPoolPaths`, `ClaudeCodePoolPaths`, `HermesPoolPaths`), which declare `pool_local`/`pool_repo` and are consumed by `path_resolution.py`, `skill_set_service.py`, `path_factory.py`, and the DI module as the single source of truth for Skill Pool physical layout. None of those dataclasses has a `pool_center` field. Instead of extending that contract, this PR hand-writes a second, undeclared literal for the same mount name directly inside a build-plan constant. That's exactly the "pre-existing violation" pattern the file already has for `skills-repo` (also a raw literal there) — extending it with a brand-new mount name for a brand-new corpus is itself a new violation, not a continuation of an accepted pattern.
- **Fix direction:** Add the `skill-center` mount fact to `skill_layout.py`'s per-engine dataclasses (or the dedicated contract for this corpus) and derive the AICoding exclude entries from it, so a future rename/versioning of the corpus has one place to change.

**2. Same lines — duplicate, hand-maintained expression of a fact the repo already governs elsewhere**
- **Violated rule:** arch.rules.md **Rule 1 — Invariant** ("The API Specification Is the Single Authority... implementations must not introduce undocumented semantics that other components depend on").
- **Why it breaks the architecture:** The PR's own description says *"the build-time Probe-driven `shared_corpora` exclusion/proof dedups against these static entries"* — i.e., the repo already has a governed, Probe-driven `shared_corpora` mechanism (see `specs/2026-08-20-skill-capability-upgrade/spec.md`, `test_service_artifact_fresh_layout.py`, `test_arca_snapshot_producer.py`) whose job is precisely to declare which pool corpora (`repo`, `center`) are externally mounted and must be excluded from the snapshot. Adding a second, statically-hardcoded expression of the same fact — justified as a "safe, complementary superset" rather than as a fix to the governed mechanism — is a hidden duplicate contract: the two sources can silently drift (e.g., if `shared_corpora` proof logic changes, nothing forces this literal to follow).
- **Fix direction:** If the Probe-driven `shared_corpora` path is provably insufficient for AICoding today, fix that mechanism (or route the fact through `skill_layout.py` per Finding 1) rather than adding a parallel static literal "just in case."

### Out-of-scope notes
- The `skills-repo`/`.skills-repo*` pattern this PR mirrors is itself an example of Finding 1's pattern pre-existing in the file; not re-flagged separately per the "don't pile on pre-existing violations" instruction, but it should be cleaned up alongside the fix above.
- No comment on the AICoding-vs-Hermes asymmetry decision itself (business/product scope).

---

totalfrank (changes requested):

## 2. Gate 0

- **Target branch:** `REL20260904` (matches `^REL\d{8}` — release branch).
- **Classification:** existing-behavior bug fix (path a) — the diff only appends two literal strings to a pre-existing static list (`_AICODING_RSYNC_EXCLUDES`) inside an already-shipped code path (`EngineBuildPlan.rsync_excludes`). It adds no new route, DTO field, config key, table, or capability, so it is not a "feature."
- **Pass/fail:** **FAILS on the propagation requirement.** The PR body gives no `dev`-branch reference (no "(#NNN)", no "cherry-pick"), and a repo-wide PR search for `skill-center`/`skills-pool`/`aicoding` exclusion changes turns up no counterpart targeting `dev`. Per the branch policy, a release-branch fix with no `dev` counterpart is a finding on its own — it will be silently lost at the next release cut (regression by omission). This must be resolved (link or open the `dev` PR) before merge, independent of the architecture findings below.

## 3. Findings (most severe first)

### Finding 1 — New engine-specific filesystem-layout hardcoding added to Backend `core/`
**File:** `src/backend/src/agentclaw/community/core/workspace/engines/aicoding.py:43-44`
```python
+    "workspace/skills-pool/skill-center",
+    "workspace/skills-pool/.skill-center*",
```
**Violated rule:** `AGENTS.md` → *Skills Architecture*: *"Do not add new engine-specific filesystem paths to Backend code. Physical layout ownership belongs to Engine Runtime and its versioned layout contract."* This is the exact document `AGENTS.md` directs contributors to read before touching Skills mounting (`skill_center/CLAUDE.md`), and that file's own §9 ("Engine 是物理目录的唯一所有者") repeats it explicitly for this precise pattern: *"Backend 当前仍会生成 source -> target 的绝对路径映射...但不得继续在 Backend 新增引擎目录硬编码"* ("...must not continue adding new engine-directory hardcoding in Backend"). It also matches violation catalog item 4 (engine/deployment filesystem layout in core) and extends the existing catalog-item-3 pattern (an engine-named module living under `core/`) with two more hand-maintained literals.

**Why it breaks the architecture:** every new shared-pool mount the platform introduces now requires a manual, per-engine literal added to `core/`. That is exactly the "add X next month, which files change?" failure mode the constitution exists to prevent — this PR's own root cause (the "skills-repo"/"skill-center" pair silently going asymmetric) is a direct symptom of that maintenance model, and this change perpetuates the model that caused the bug instead of removing it.

**Required fix direction:** This codebase already has the compliant, dynamic replacement — a build-time Probe/`shared_corpora` mechanism (`core/service_bot/services/deploy/service_skills_manifest.py`, `arca_snapshot_producer.py`) that computes and validates which shared corpora (including `center`) must be excluded from a snapshot, without per-engine hardcoded strings. The PR description itself confirms this mechanism "dedups against these static entries." The compliant fix routes AICoding's rsync-exclude derivation through that dynamic layout/Probe output (or a single declared, config-owned table of shared-corpus names), not another hand-added pair of literals in `core/`.

### Finding 2 — New cross-engine inconsistency for the identical corpus-exclusion concern
**File:** `src/backend/tests/community/core/workspace/test_engine_sandbox_providers.py:324` (pre-existing, untouched by this PR)
```python
assert all("skill-center" not in item for item in plan.rsync_excludes)  # HermesSandboxProvider
```
**Violated rule:** the required engine-variability pattern (one contract, one composition point) and arch.rules.md Rule 23 (Patterns Are Consistent and Cataloged). `HermesSandboxProvider` and `AICodingSandboxProvider` both implement the same `EngineBuildPlan` contract, and Hermes has a standing guard that its build plan must **never** contain the literal `"skill-center"` — i.e., Hermes keeps the Center corpus out of its snapshot exclusively through the dynamic Probe path. This PR makes AICoding solve the *same* problem (excluding the same external `skill-center` pool corpus from the same kind of rsync snapshot) through a *different*, static mechanism, with no doc/contract update explaining the divergence. The PR body even flags this asymmetry ("Hermes intentionally untouched") but treats it as an acceptable carve-out rather than a signal that the two engines now disagree on how a shared, cross-engine concern is solved.

**Required fix direction:** Pick the one mechanism both engines use for excluding shared pool corpora — the already-existing dynamic Probe/`shared_corpora` path, since it's validated at build time and is what Hermes already relies on — and make AICoding consume it the same way. If AICoding must remain a documented exception, that decision belongs in the contract/pattern catalog, not silently in a single engine's static list.

### Finding 3 — Gate 0 propagation gap (see Gate 0 above)
No `dev` counterpart is referenced or discoverable for this release-branch fix. Required fix: link the corresponding `dev` PR/commit in the description, or open one alongside this PR before merge.

## 4. Out-of-scope notes
- The exclude list already mixes `skills-center` (plural, under `workspace/skills/`) and the new `skill-center` (singular, under `workspace/skills-pool/`) — a naming footgun this PR doesn't cause, but one more per-engine hardcoded literal is one more place for that kind of typo to hide, which reinforces Finding 1 rather than standing on its own.

---

FreddieSun (approved):
lgtm

totalfrank (approved):

PR #1905, opened 2026-09-04
feat(backend): allow callers to restart existing instances

yhzhang35:
## Problem

`POST /api/v1/expert-chats/caller-connection` only allowed super administrators to create, reuse, or upgrade per-caller container instances. A regular caller could not restart an already-provisioned container that belongs to their own exact `ac_expert_chat_instance` record.

The endpoint returns private runtime connection data, so self-service access must not allow first-time provisioning creation, cross-user operations, or access based only on public/collaborator chat permissions.

## Solution

- Preserve the existing super-administrator path, including delegated first-time creation, reuse, and forced upgrade.
- Add an actor-aware service entry point for regular users.
- Require the authenticated operator to equal the requested caller `user_id`.
- Require an exact current-environment `ac_expert_chat_instance` match for `bot_id + owner_id + user_id` with a non-empty string `ext.bot_uuid`.
- Reject cross-user, missing-instance, and missing-container cases before any lifecycle, publish, BaaS, or connection call.
- Keep the route as POST and do not introduce public/collaborator authorization.
- Keep the Service API contract single-sourced in the owning core Protocol and add Protocol/Concrete conformance coverage.
- Add structured request, success, denial, and failure logs without logging connection credentials, cookies, headers, tokens, or full instance extensions.

## Validation

Validated against `REL20260904` at `3c3aeaf109638ea5e1d917a31dd0fd38d971e5ac`:

```text
Backend CI gate: PASS
casePassRate: 100.00% (17127/17127)
lineCoverage: 88.55% (required >= 75%)
changeLineCoverage: 100.00% (44/44, required >= 80%)
coverage pytest: 17068 passed, 59 skipped, 0 failed
focused feature and architecture tests: 403 passed, 0 failed
git diff --check: PASS
```

The skipped tests require a live Singlebox/Backend/BaaS environment. No live container restart was performed as part of local validation.

## Compatibility and risk

- The existing administrator behavior and POST request contract remain unchanged.
- There is no database migration or schema change.
- Regular users cannot create a first caller instance through this endpoint.
- Rollback is limited to reverting the two commits in this PR.

## Spec

- `src/backend/specs/caller-instance-self-restart/001-spec-output.md`
- `src/backend/specs/caller-instance-self-restart/002-code-report.md`
- `src/backend/specs/caller-instance-self-restart/003-review-report.md`
- `src/backend/specs/caller-instance-self-restart/003b-regression-report.md`

## Remote PR status

- GitHub checks: all required checks passed, including Backend unit tests and Singlebox coverage.
- Review decision: approved.
- Automated review comments: none.
- Inline and issue comments requiring changes: none.

### DIFF
--- a/src/backend/specs/caller-instance-self-restart/001-spec-output.md
+++ b/src/backend/specs/caller-instance-self-restart/001-spec-output.md
@@ -2,7 +2,7 @@
 agent: tc-review
 status: approved-design
 created: 2026-09-04
-baseline: origin/dev@6ecb42630227da0a2030a051659312ac88dea86c
+baseline: github/REL20260904@3c3aeaf109638ea5e1d917a31dd0fd38d971e5ac
 repository: inclusionAI/Avernet
 ---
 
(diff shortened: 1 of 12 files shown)


totalfrank (approved):
LGTM.

PR #1910, opened 2026-09-04
ci(singlebox-coverage): stop paying npm's retired audit endpoint and cache the BCN plugin

totalfrank:
## Problem

"Singlebox Coverage" went from ~13m to 19–32m across **every** PR in the repo, at a single moment between 2026-09-03 13:42 UTC and 2026-09-04 04:14 UTC. Every run on 09-03 was 12.6–15.2m; every run after the cliff was 18.9–31.6m.

The controlled comparison is one branch, unchanged, run either side of it — [33753393854](https://github.com/inclusionAI/Avernet/actions/runs/33753393854) (09-03, 12m02s) vs [33837445531](https://github.com/inclusionAI/Avernet/actions/runs/33837445531) (09-04, 22m32s). Setup steps are identical (~50s either way); cargo, pytest, the BCS e2e and the coverage gates are flat to within seconds. The entire delta is npm.

Nothing in the repo changed between those runs: same plugin version `1.0.23`, same `.npmrc`, same 2051 resolved packages, same npm cache key, same runner image, region and Node version.

### Root cause: npm's legacy audit endpoint is being retired

```
npm notice This endpoint is being retired. Use the bulk advisory endpoint
instead. See: https://api-docs.npmjs.com/#tag/Audit
```

The audit call now stalls for ~7m and npm gives up, printing neither `audited N packages` nor a vulnerability count — which is exactly why the slow logs show a bare `up to date in 7m`. **It is a fixed cost per invocation, not proportional to the tree.** Runs pay it once per audit-enabled npm call, so the spread across PRs tracks how many such calls a run makes and its cache state.

The decisive evidence is this PR's own first run ([33861748256](https://github.com/inclusionAI/Avernet/actions/runs/33861748256)), where the BCN plugin build — already carrying `--no-audit` from commit 2 — sat between the two installs that still audited:

| Step | Packages | Audit | Time |
| --- | --- | --- | --- |
| BCS panel `npm ci` | 33 | on | **7m00s** |
| BCN plugin `npm install` | 2051 | off | **27s** |
| frontend `npm ci` | 2153 | on | **7m05s** |

Package count is uncorrelated with time; the audit flag is perfectly correlated. Reproduced locally on the BCS panel package, same command as CI: **421s with audit, 3s without**, identical package sets and identical 99M `node_modules`.

> An earlier revision of this description attributed the regression to general npm registry latency, with the lockfile-less BCN build as the main exposure. That was wrong about the mechanism — lockfile-backed `npm ci` audits too, which is why the frontend went 45s → 7m. The audit endpoint is the actual cause.

## Solution

Three commits, each independently revertable.

**1. Opt every npm install out of the audit call** (`--no-audit --no-fund`) — the BCS panel asset and the frontend, both branches each, matching what `claude_relays.sh` already does for the gateway build. Neither flag changes what is installed. Auditing here was never a gate: it printed advisories into a CI log with nothing reading them or failing on them. A guard test asserts every `npm install`/`npm ci` under `scripts/modules` opts out, so a new call site cannot silently reintroduce the stall (global installs excluded — `npm install -g` does not audit). **~14m.**

**2. Replace `npm prune --omit=dev` with a prod-only reinstall.** The dev tree is 348 MB / 56083 files (`.npmrc` pins `install-strategy=nested`, so nothing dedupes). Prune reduces it by reconciling that whole tree — 423s measured, and it audits. Deleting `node_modules` and reinstalling `--omit=dev` lands the same shipped tree (`ws` and nothing else) in 4s. Dropping the prune outright was rejected: it is load-bearing, taking 348 MB down to 232 K, and without it all 348 MB is copied back into the source tree and into the cache below. **~7m on a cold build.**

**3. Cache `dist/` + `node_modules`**, keyed on the plugin's `package.json` and `.npmrc`. `setup_bcn_plugin` already short-circuits with `BCN plugin already built, skipping` when both exist, so a hit skips the build outright. No `restore-keys`: a near-miss would restore a `dist/` built from a different `package.json` and the skip branch would accept it. The cached tree is `ws` alone, 232 KB, so the entry stays small against a cache budget that is already evicting ([run 33842045450](https://github.com/inclusionAI/Avernet/actions/runs/33842045450) lost both the npm and cargo caches and took 30m16s). **~30s per run once warm.**

## Validation

- **Local A/B, BCS panel package**, same command as CI: `npm ci` 421s → `npm ci --no-audit --no-fund` 3s; `diff` over resolved package sets empty, both 99M.
- **Local A/B, BCN plugin**: `npm prune --omit=dev` 423s (printing the same `up to date in 7m` as CI) → `rm -rf node_modules && npm install --omit=dev` 4s. Same shipped tree.
- **End-to-end `setup_bcn_plugin` against the real plugin**: rc=0 in 55s; `dist/esm/index.js` and `dist/commonjs/index.js` both produced; `node_modules` 228K / 27 files containing only `ws`; symlink correct.
- **CI**: run [33861748256](https://github.com/inclusionAI/Avernet/actions/runs/33861748256) green on commits 2–3, with the BCN phase down from 11m03s to 35s on a cold cache and the cache entry saved.
- **Tests** → `scripts/test_bcn_plugin_source.sh` ALL PASS and `scripts/test_singlebox_service_guards.sh` PASS, with three new cases where the source-build path previously had none: the BCN command sequence (that `npm prune` is gone), the `already built, skipping` branch the cache depends on, and the repo-wide `--no-audit` guard. Negative controls: each fails when its change is reverted.
- `test_singlebox_bcs_prereqs`, `test_singlebox_default_mode`, `test_singlebox_toolchain`, `test_singlebox_bcs_runtime_config`, `test_singlebox_rule_bots` → PASS. The workflow's own "Validate shell scripts" commands → OK.
- **Pre-existing failures, unchanged here and reproduced on a clean tree with identical messages**: `test_singlebox_model_config` (this sandbox's `stat` output format), `test_singlebox_demo_bot` (bcs/bcsfuse start-order mismatch on dev), and `test_singlebox_coverage_gate` (needs `src/backend/.venv`). None is touched by this PR.

## Compatibility and risk

No gate, threshold, report path, or coverage content changes — `SINGLEBOX_COVERAGE_BCS_LINE_MIN=40` / `METHOD_MIN=36` and `verify_singlebox_coverage_artifacts.py` are untouched — and no change to any resolved dependency tree.

Local `singlebox.sh` runs stop printing the post-install advisory summary; `npm audit` on demand is unaffected. If the team wants dependency review enforced, it belongs in a dedicated check that can actually fail a build, not as a side effect of service setup.

A stale cache entry cannot outlive a `package.json` or `.npmrc` edit because both are part of the key. Rollback is reverting any subset of the three commits.

https://claude.ai/code/session_01Kv9sczD8rpocHsm9YE4PuH

### DIFF
--- a/.github/workflows/singlebox-coverage.yml
+++ b/.github/workflows/singlebox-coverage.yml
@@ -193,6 +193,29 @@ jobs:
             ${{ runner.os }}-cargo-e2e-
             ${{ runner.os }}-cargo-
 
+      - name: Cache built BCN plugin
+        if: steps.changes.outputs.skip != 'true'
+        uses: actions/cache@v4
+        with:
+          # `setup_bcn_plugin` (scripts/modules/bcs.sh) rebuilds the plugin from
+          # source on every run: three unlocked npm tree operations in a fresh
+          # temp dir. package-lock.json is gitignored for this package, so npm
+          # re-resolves all 2051 packages from the registry each time — the one
+          # npm step in this job with no lockfile to fall back on. That cost 29s
+          # when the registry was fast and 11m when it was not (run 33837445531).
+          # Restoring dist/ + node_modules makes the script's own
+          # "BCN plugin already built, skipping" branch fire instead.
+          #
+          # No restore-keys: a near-miss would restore a dist/ built from a
+          # different package.json and the skip branch would accept it, so the
+          # key must match exactly or the plugin is rebuilt. node_modules is the
+          # prod-only tree the build reduces to before copying back: `ws` alone,
+          # 232 KB, so this entry stays small against the repo cache budget.
+          path: |
+            src/bcs/crates/plugins/openclaw-channel-bcn/dist
+            src/bcs/crates/plugins/openclaw-channel-bcn/node_modules
+          key: ${{ runner.os }}-bcn-plugin-${{ hashFiles('src/bcs/crates/plugins/openclaw-channel-bcn/package.json', 'src/bcs/crates/plugins/openclaw-channel-bcn/.npmrc') }}
+
       - name: Validate shell scripts
         if: steps.changes.outputs.skip != 'true'
         run: |
--- a/scripts/modules/bcs.sh
+++ b/scripts/modules/bcs.sh
@@ -324,12 +324,21 @@ install_bcs_panel_asset_deps() {
     log_info "Installing BCS panel asset dependencies..."
     cd "${BCS_PANEL_ASSET_DIR}"
 
+    # --no-audit/--no-fund: npm's legacy audit endpoint is being retired
+    # ("This endpoint is being retired. Use the bulk advisory endpoint
+    # instead"), and the call now stalls for ~7m before npm gives up and
+    # prints neither "audited N packages" nor a vulnerability count. The
+    # stall is a fixed cost per invocation, not proportional to the tree:
+    # these 33 packages took 421s with audit on and 3s with it off, for an
+    # identical node_modules. Auditing here was never a gate either — it
+    # printed advisories into a CI log nobody reads. Same flags as the
+    # claude_relays.sh gateway build.
     if [ -f package-lock.json ]; then
-        if ! HUSKY=0 npm ci --registry="${NPM_REGISTRY_URL}"; then
+        if ! HUSKY=0 npm ci --registry="${NPM_REGISTRY_URL}" --no-audit --no-fund; then
             log_error "Failed to install BCS panel asset dependencies"
             return 1
         fi
-    elif ! HUSKY=0 npm install --registry="${NPM_REGISTRY_URL}"; then
+    elif ! HUSKY=0 npm install --registry="${NPM_REGISTRY_URL}" --no-audit --no-fund; then
         log_error "Failed to install BCS panel asset dependencies"
         return 1
     fi
--- a/scripts/modules/bcs.sh
+++ b/scripts/modules/bcs.sh
@@ -538,11 +538,24 @@ setup_bcn_plugin() {
                 return 1
             fi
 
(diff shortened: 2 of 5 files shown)


carolynli (approved):
LGTM.

totalfrank (commented):
Reviewed all 3 commits (`6c25ccc`, `fa1ab24`, `90825e5` head). No findings.

Checked:
- **Correctness** — `setup_bcn_plugin`'s new `rm -rf node_modules && npm install --omit=dev` sequence in `scripts/modules/bcs.sh`; confirmed the first `npm install` writes a `package-lock.json` into the temp build dir (no `package-lock=false` in `.npmrc`), so the prod-only reinstall resolves the same pinned versions rather than re-resolving independently.
- **Test coverage** — ran `scripts/test_bcn_plugin_source.sh` and `scripts/test_singlebox_service_guards.sh` from the PR branch; both pass, including the three new cases (prune-is-gone pin, already-built skip branch, repo-wide `--no-audit` guard).
- **Guard regex** (`test_npm_installs_opt_out_of_the_audit_call`) — verified it correctly excludes comment lines and `log_*` lines and correctly flags/excludes every actual `npm ci`/`npm install` site under `scripts/modules/*.sh`, including the `-g` exclusion.
- **CI config** — `.github/workflows/singlebox-coverage.yml` parses as valid YAML; the new "Cache built BCN plugin" step is placed before the build step it short-circuits, guarded by the same `steps.changes.outputs.skip` condition as its neighbors, and its cache paths match `plugin_src` in `bcs.sh`.
- **Contract/architecture/security** — out of scope here: no Service/Plugin API, schema, or config surface changed; no secrets, hardcoded endpoints, or committed runtime state introduced; pure CI/build-script change with no core-logic or transport-boundary crossing.


---

PR #1913, opened 2026-09-04
refactor(backend): bind and inject the creation seam as its Protocol

totalfrank:
## Problem

`submit_bot_creation_with_manifest` takes a `ManifestCreationSeam`, and every caller handed it a `BotCreationManifestSeam`. The two were related only by shape, and the wiring named the class everywhere:

```python
manifest_seam: BotCreationManifestSeam = Injected(BotCreationManifestSeam)   # the route
...
submit_bot_creation_with_manifest(..., manifest_seam=manifest_seam)          # param is ManifestCreationSeam
```

Two problems, one under the other:

1. **Nothing connected the contract to its one implementation.** A reader (or an IDE) standing on the Protocol could not get to the class that answers it, and a signature drifting from the contract surfaced — if at all — as a type error at whichever call site happened to pass the seam, never against the contract itself.
2. **The container bound the class, so every consumer depended on it.** Declaring a contract while binding past it left the routes and the creation job's handler coupled to how the seam is built rather than to what it promises.

## Solution

Two commits.

**1. Make the relationship nominal.** `BotCreationManifestSeam` now inherits `ManifestCreationSeam` — declared, not merely satisfied by shape.

**2. Make the Protocol the type the container binds and every consumer holds.** The DI provider returns `ManifestCreationSeam`; `injector.get`, both `with-manifest` routes' `Injected(...)`, and the job handler's `Callable[[], ...]` all name it. The implementation is now named in exactly two places: its own definition, and the provider that constructs it.

That second step required the Protocol to name the **whole** seam, not only submission's four operations. `apply_pre_container` (the creation job's) and `find_job` (the poll's) have one caller each, but those callers are injected like any other — a contract without them would leave them holding the class, which is the coupling being removed. `discard` gains the optional `owner_id` the job passes. `create_flow` calling four of six is the ordinary shape of a caller that does not need everything, not a reason to split the seam in two.

`@runtime_checkable` follows from the binding and nothing else: python-injector `isinstance`-checks an instance against the key it is bound to, so rebinding this key to a stand-in — which the endpoint suite does — raises on a plain `Protocol`. It buys no guarantee on its own, since the check is attribute presence and never a signature.

**The dependency direction is unchanged.** The import runs manifest → management, which `core/bot_config_manifest` already takes (the engine registry, the token vault). Only the reverse closes a cycle, which is why the contract lives in `core/bot_management`, and why `manifest_seam.py` imports nothing but `typing` and one value type (`TaskRecord`, for `find_job`'s return).

**It stays a `Protocol`.** That was never only about the cycle: the test doubles in `test_create_flow_with_manifest` satisfy it by shape, with no base class and no import, and still do.

### Guards

No type checker runs on this tree, so the invariants are pinned by tests:

- **The provider's return annotation is the Protocol.** It *is* python-injector's binding key, so annotating the class again leaves every consumer asking for a key nobody binds.
- **Nothing in the composition root resolves the seam by its class.** This is the one failure that would not be loud: the class is unbound, so injector would *construct a second, unwired seam*, and the job handler would enqueue through one the routes never touch.
- **The contract names all six operations**, pinned as an exact set — the signature test derives its parametrization from the Protocol, so without this a dropped operation would shrink the parametrization and go on passing.
- **Each operation is implemented on the class itself, with the contract's parameters.** The base class brings one hazard: a Protocol method's body is `...`, so an operation dropped from the seam would be *inherited* and quietly answer `None` where it previously raised `AttributeError`.

Docstrings on both sides say why the Protocol exists, why the implementation declares it, and why the container binds it; `bot_config_manifest/README.md` declares the import in its context boundary (Rule 22 — the arch test enforces the whitelist).

## Validation

Run with `.venv` built from `uv.lock`'s pinned versions, from `src/backend`:

- `pytest tests/community/{core/bot_config_manifest,core/bot_management,endpoints,di,architecture}` — **3973 passed**. (The run showed 3 setup errors from `fixture 'caplog' not found`; that was a `-p no:logging` flag on my own invocation disabling the plugin that provides it, and those 3 pass with it enabled.)
- Mutation checks of the new guards, each reverted after:
  - Re-annotating the provider with the class fails `test_the_manifest_creation_seam_is_bound_under_its_protocol` **and all seven endpoint tests** (`TypeError` at request time).
  - Dropping `apply_pre_container` from the Protocol fails `test_the_contract_names_the_whole_seam`.
  - Renaming `persist` on the seam fails `test_every_declared_operation_keeps_the_contract_signature[persist]` with the "inherited from the Protocol … would answer None instead of raising" message.
- The `@runtime_checkable` requirement was found by the suite rather than predicted: without it, the endpoint test that rebinds the seam fails with `TypeError: Instance and class checks can only be used with @runtime_checkable protocols`.
- Runtime check that the base class does not disturb construction: MRO is `BotCreationManifestSeam → ManifestCreationSeam → Protocol → Generic → object`, `_is_protocol` is `False`, and the seam instantiates through its own `__init__` unchanged.

Not run: the full `tests/community` suite and the frontend/e2e gates. `uv sync --frozen` could not fetch the lock's aliyun mirror from this environment, so the venv was built by installing `uv.lock`'s exact pins from pypi.org; `uv.lock` and `pyproject.toml` are unmodified.

## Compatibility and risk

No contract, schema, config or migration impact, and no runtime behaviour change: the same single instance is constructed by the same provider and serves the same callers — only the name it is bound and injected under changes. Rollback is reverting the two commits.

One thing for a reviewer to know: `di/modules/bot_management_module.py` was two lines under the R9 1000-line cap, so the new import is a single line and the explanation was absorbed into the existing provider docstring rather than added as a paragraph. It lands at 999 — no headroom left for the next change to that file.

https://claude.ai/code/session_01YZRzbXc54v7X54cZS3TzXZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01YZRzbXc54v7X54cZS3TzXZ)_

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/openapi_v1/bots/create_with_manifest.py
+++ b/src/backend/src/agentclaw/community/adapters/http/openapi_v1/bots/create_with_manifest.py
@@ -19,7 +19,7 @@ what polls AgentPass), starts no work, and writes nothing, so polling faster
 never changes an outcome and a caller that stops polling loses nothing.
 
 **Nothing here decides what a manifest means.** Validation, storage and apply are
-W1's and W4's, reached through ``BotCreationManifestSeam``; creation itself is
+W1's and W4's, reached through ``ManifestCreationSeam``; creation itself is
 ``create_flow``'s, and ``bot_service.create_bot`` is untouched. This module
 composes them and shapes the answer.
 """
@@ -66,7 +66,6 @@ from agentclaw.community.core.bot_config_manifest.create_job import (
 from agentclaw.community.core.bot_config_manifest.creation import (
     CREATE_ON_CONTAINER_TRIGGER,
     CREATE_PRE_CONTAINER_TRIGGER,
-    BotCreationManifestSeam,
 )
 from agentclaw.community.core.bot_inventory.protocols import (
     BusinessSpaceContextProtocol,
@@ -77,6 +76,9 @@ from agentclaw.community.core.bot_management.create_flow import (
     BotCreateSpec,
     BotCreateTemplateValidationMode,
     submit_bot_creation_with_manifest,
+)
+from agentclaw.community.core.bot_management.manifest_seam import (
+    ManifestCreationSeam,
 )
 from agentclaw.community.core.bot_management.services.bot_service import (
     BotNotFoundError,
@@ -150,7 +152,7 @@ async def create_bot_with_manifest(
     space_context: BusinessSpaceContextProtocol = Injected(
         BusinessSpaceContextProtocol
     ),
-    manifest_seam: BotCreationManifestSeam = Injected(BotCreationManifestSeam),
+    manifest_seam: ManifestCreationSeam = Injected(ManifestCreationSeam),
 ) -> Envelope[BotCreateWithManifestAccepted]:
     """Create a bot and its configuration in one request. Always `202`.
 
@@ -252,7 +254,7 @@ async def get_bot_create_with_manifest_status(
     apply_service: BotConfigManifestApplyServiceProtocol = Injected(
         BotConfigManifestApplyServiceProtocol
     ),
-    manifest_seam: BotCreationManifestSeam = Injected(BotCreationManifestSeam),
+    manifest_seam: ManifestCreationSeam = Injected(ManifestCreationSeam),
 ) -> Envelope[BotCreateWithManifestStatus]:
     """Where a creation stands. Takes a `bot_id` and nothing else.
 
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/README.md
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/README.md
@@ -584,7 +584,7 @@ provides:
   - ApplyTaskLifecycle
   - APPLY_TASK_TYPE
   - build_apply_task_payload
-  - BotCreationManifestSeam
+  - BotCreationManifestSeam (the ManifestCreationSeam implementation; only DI names it)
   - CREATE_PRE_CONTAINER_TRIGGER
   - CREATE_ON_CONTAINER_TRIGGER
   - preflight_creation_manifest
@@ -701,7 +701,7 @@ internal_dependencies:
   - agentclaw.community.plugin_api.object_storage  # the bot-data object store the managed-files store writes a teclaw bot's manifest-delivered files into (W8), and the cli_tools store keeps a bot's tool bytes in (W9)
   - agentclaw.community.core.bot_startup_script
   - agentclaw.community.core.bot_management.engines.registry  # the pure runtime-engine routing policy the resources materialiser addresses workspaces through, the router's own rule (W6)
-  - agentclaw.community.core.bot_management.manifest_seam  # the Protocol creation states its four operations as; the creation seam declares it rather than matching it by shape, so the contract is checked in one place
+  - agentclaw.community.core.bot_management.manifest_seam  # the Protocol the creation seam is declared as, bound as, and injected as everywhere; declaring it rather than matching it by shape is what checks the contract in one place
   - agentclaw.community.core.bot_management.token_vault
   - agentclaw.community.core.bot_management.utils  # resolve_agent_code — the creation job asks whether completion's *second* write (the owner relationship) actually landed, since the bot record alone cannot tell it
   - agentclaw.community.core.mcp.mcp_auth_service_protocol  # the permission check DirectActivationService also consults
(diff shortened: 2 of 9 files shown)


totalfrank (commented):
Reviewed commits `50803ba` and `ff8fb5a` (full PR, no prior review on this thread).

Files covered: `manifest_seam.py`, `creation.py`, `create_job.py`, `bot_management_module.py`, `create_with_manifest.py`, `bot_config_manifest/README.md`, and the three touched test files.

Checked and clean:
- **Correctness** — traced all six Protocol operations (`preflight`, `persist`, `start_job`, `discard`, `apply_pre_container`, `find_job`) end-to-end from Protocol declaration → concrete implementation → every call site (routes, `create_job.py`, DI provider); signatures and callers agree everywhere, including the new optional `owner_id` on `discard`.
- **Contract breakage** — the DI provider's return annotation, `Injected(...)` sites, and `bind_overrides` usage in the endpoint test all key off `ManifestCreationSeam` consistently; `bind_overrides` subclasses `type(wired)` (the concrete class), not the Protocol, so the stand-in still carries the real implementation. No caller left pointing at the old class-keyed binding (verified via `test_nothing_in_the_composition_root_resolves_the_seam_by_its_class`, and confirmed no other reference to the class as a binding key remains).
- **Import boundaries / cycles** — `bot_management/manifest_seam.py`'s new `TaskRecord` import comes from `task_queue.types`, which has no dependency back into `bot_management` or `bot_config_manifest`, so the documented one-directional import still holds. `README.md`'s Rule 22 whitelist entry is added and correctly placed.
- **File size cap (R9)** — `bot_management_module.py` is 999 lines, under the 1000-line cap referenced in the commit message.
- **Test coverage** — new tests pin the inheritance, the exact six-operation contract set, and per-parameter signature drift; the `_Seam` test double in `test_create_flow_with_manifest.py` (only 4 of 6 methods) is passed as a plain function argument rather than resolved through the injector, so the widened `@runtime_checkable` Protocol doesn't isinstance-check it and it isn't a regression risk.

No blocking or non-blocking issues found. One thing not worth a nit-level comment on its own: `test_creation_seam.py`'s module docstring ("The four operations…") is now stale since the file also covers the six-operation Protocol contract, but that's cosmetic only.

---

totalfrank:
**Review round 1 — one cosmetic finding, fixed.** Reviewed `ff8fb5a`; addressed in `bcddd50`.

The review flagged no blocking or non-blocking issues. The one cosmetic observation was right and is now fixed:

- **`test_creation_seam.py`'s module docstring was stale** — it still read "The four operations bot creation asks of the manifest layer", from before the Protocol grew to name all six and this file started pinning the contract itself. Rewritten to describe what the file actually covers: the behavioural tests, and the contract section at the bottom.

Nothing else changed — no production code in this push.

Two points from the review worth recording, since both are load-bearing and were verified independently rather than taken from the description:

- **`bind_overrides` subclasses `type(wired)`, not the key.** That is what lets the endpoint suite key its stand-in off the Protocol while still carrying the real implementation, and it is why `@runtime_checkable` is needed at all — python-injector `isinstance`-checks the instance against the key it is bound to.
- **The 4-method `_Seam` double in `test_create_flow_with_manifest.py` is not a regression risk.** It is passed as a plain function argument, never resolved through the injector, so the widened Protocol never `isinstance`-checks it. Structural conformance for test doubles survives the widening.

CI was 10/10 green on `ff8fb5a` and is re-running on `bcddd50`.

---

totalfrank (commented):
**Review round 2 — clean.** Reviewed `bcddd50` (the only commit since round 1's `ff8fb5a`).

`bcddd50` is a docstring-only change to `test_creation_seam.py`'s module docstring, replacing the stale "the four operations" line with one describing the file's behavioural tests plus the contract-pinning section. It touches no production code and no test assertions — verified via the commit's full patch (7 additions, 1 deletion, one file).

Checklist of all findings across rounds:
- **`test_creation_seam.py` module docstring stale** (round 1, noted in review body, no inline thread) — **fixed** in `bcddd50`; docstring now accurately describes the file's contents.

No new findings in `bcddd50`.

**Open findings: 0.** PR is clean as of round 2 (`bcddd50`).

---

totalfrank:
**Review round 2 — no issues flagged. CI green.** Reviewed `bcddd50` (the round-1 docstring fix, the only commit since `ff8fb5a`).

No new findings. The one finding raised across the loop — round 1's stale module docstring in `test_creation_seam.py` — is fixed and confirmed. **Open findings: 0.**

CI on `bcddd50`: **10/10 success** — Validate PR title, BaaS / Sandbox-proxy / Gateway / Engine / BCS / Backend unit tests, Singlebox coverage, BCS e2e, Publish TaskRunner design guard. No merge conflict against `dev`.

Nothing was disputed and nothing was skipped, so there is no disagreement for a human to settle. The PR is now waiting only on the required approvals from @FreddieSun and @xianmuyq — that is what `blocked` reflects, not a failing gate.

One thing worth a reviewer's eye that no check enforces: `di/modules/bot_management_module.py` lands at **999 lines** against the R9 1000-line cap. The new import is one line and the new explanation was folded into the existing provider docstring rather than added as a paragraph, specifically to fit. There is no headroom left for the next change to that file.

---

FreddieSun (approved):

PR #1915, opened 2026-09-04
fix(bcs): count cancelled HTTP requests in metrics

quechao76:
## Problem

BCS only recorded `bcs_http_requests_total` after an HTTP handler returned a
response. When database operations were slow and the client or upstream proxy
closed the connection, the request future was dropped before the metrics code
ran.

As a result, proxy-499-equivalent requests were missing from the total request
count. Health checks became the main recorded traffic, causing the HTTP success
rate to remain at 100%.

The maintained Grafana dashboards also included `/health` traffic in HTTP API
traffic and SLA calculations, and the HTTP SLA panels defaulted to 100% when
there was no API traffic.

## Solution

- Add an HTTP request lifecycle guard that records exactly one terminal metric.
- Record dropped request futures as:
  - `status_class="4xx"`
  - `result="error"`
  - the matched route template and HTTP method
- Add `bcs_http_request_cancellations_total` for dedicated cancellation
  diagnostics.
- Keep metrics middleware outside panic handling so generated 500 responses are
  recorded as server errors instead of cancellations.
- Keep `/health` in the raw `bcs_http_requests_total` metric for probe
  diagnostics.
- Exclude `/health` from HTTP API panels in the monitoring, SLA, and
  troubleshooting dashboards.
- Remove the HTTP SLA fallback that displayed 100% when there was no API
  traffic.

## Validation

- `cargo test -p bcs dropped_http_metrics_guard_records_cancelled_request --features prometheus-metrics`
  - Passed: 1 test.
- `cargo test -p bcs --test metrics_http --features prometheus-metrics`
  - Passed: 2 tests.
- `cargo check -p bcs --no-default-features`
  - Passed.
- Validated all modified Grafana dashboards with `jq empty`.
- Confirmed general HTTP dashboard queries exclude `/health`.
- Confirmed HTTP SLA queries no longer use `or vector(100)`.
- `git diff --check`
  - Passed.

The complete `cargo test --workspace` suite was not run because this change is
limited to the BCS bootstrap metrics middleware and Grafana dashboard JSON; the
closest metrics tests and both feature configurations were validated.

## Compatibility and risk

There is no HTTP API, schema, or configuration contract change.

`bcs_http_requests_total` keeps its existing label schema but now includes an
additional terminal outcome for cancelled requests. A new additive Prometheus
counter, `bcs_http_request_cancellations_total`, is exposed.

The cancellation classification represents an application-side dropped request
future, equivalent to the common proxy 499 case; BCS does not receive an actual
HTTP 499 response status from the proxy.

Rollback is a revert of commit `8521f8a07`.

### DIFF
--- a/src/bcs/crates/bootstrap/bcs/src/metrics.rs
+++ b/src/bcs/crates/bootstrap/bcs/src/metrics.rs
@@ -468,6 +468,15 @@ impl MetricsRuntime {
     ) {
         record_http_request(&self.env, route, method, status, duration);
     }
+
+    pub fn record_http_request_cancelled(
+        &self,
+        route: &str,
+        method: &str,
+        duration: Duration,
+    ) {
+        record_http_request_cancelled(&self.env, route, method, duration);
+    }
 }
 
 #[cfg(feature = "prometheus-metrics")]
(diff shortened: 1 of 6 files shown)


questchaos:
@codex review

RaymondHX (approved):

PR #1925, opened 2026-09-05
fix(evolvetrace): allow production CORS and repair image build

wienyang:
## Problem

Workflow saves routed through the Avernet Gateway reached Evolvetrace but failed with `500 CORS not allowed` because Evolvetrace only allowed localhost origins.

The Evolvetrace image also failed to build due to mismatched ESLint dependencies and a Node 20 runtime incompatible with `better-sqlite3@13`.

## Solution

- Add `CORS_ALLOWED_ORIGINS` support while preserving localhost defaults.
- Add a regression test covering allowed and rejected origins.
- Align `@eslint/js` with ESLint 9.
- Use Node 22 for the Evolvetrace build and runtime stages.
- Install and prune dependencies in the builder stage, then copy production `node_modules` into the slim runtime image.
- Avoid rebuilding native dependencies inside the slim runtime image.

## Validation

- `npm ci --ignore-scripts` passed.
- `npm test` passed: 1 test.
- `npm run build:server` passed.
- `npm run build` passed: 354 modules transformed.
- `git diff --check` passed.
- The OCB pre-push lint-only gate passed.
- Final Docker image build must be confirmed on the xhunter build host, where Docker is available.

## Compatibility and risk

- No API or database schema changes.
- Production must set:

  `CORS_ALLOWED_ORIGINS=https://avernet.alipay.com`

- The deployment must use a newly built image containing commit `638f3b798`.

### DIFF
--- a/docker/services/evolvetrace.dockerfile
+++ b/docker/services/evolvetrace.dockerfile
@@ -7,20 +7,32 @@
 #   docker build -f docker/evolvetrace.Dockerfile -t evolvetrace:local .
 
 # --- Build stage: install deps, build Vite frontend, compile TS server ---
-FROM node:20-bookworm AS builder
+FROM node:22-bookworm AS builder
+
+# Native dependencies such as better-sqlite3 may compile during npm install.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends python3 build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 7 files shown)


totalfrank (approved):

3h-william (approved):
Ok

archer910517-ctrl (approved):
LGTM

PR #1928, opened 2026-09-05
fix(engine): reduce skill mapping source scans

FreddieSun:
## Problem
SkillSet runtime projection recursively walked every file below each Local, Repo, and Center source during ordinary BEST_EFFORT publish and verify. On production read-only OSSFS, two observed requests spent 36.57s and 23.05s waiting on publish+verify; a read-only scan of 11 Repo sources took 14.69s for one recursive pass. Backend also resolved the post-mutation Skill state twice: once for the after snapshot and again while building the applied plan.

## Solution
- Replace recursive content walks only in ordinary BEST_EFFORT mapping projection with a bounded source-root check that retains managed-root, escape, directory, and root read/execute validation. STRICT and Pool cutover validation remain unchanged.
- Cache source and managed-root inspection only within one publish or verify request. Publish and verify still observe independently, and verify still checks the actual target and retired entries.
- Classify retirement entries with lstat/readlink so a dangling or unavailable retired source is not followed; remove only an exact managed symlink identity.
- Add bounded aggregate timing/count evidence, including the three slowest source checks. Public payload/status fields are unchanged.
- Resolve one Reader-backed post-mutation Plan in Backend, calculate retired mappings from that Plan, and apply the exact same Plan. The before snapshot and Reader flush behavior remain unchanged.

## Validation
- Engine community: 2468 passed, 5 skipped.
- Backend community: 17627 passed, 60 skipped; the first full run exposed 8 endpoint fake contract failures, all fixed, then the exact 8 cases passed.
- Backend Plan/service/contract/architecture focused run: 289 passed; post-fix contract set: 147 passed.
- Engine focused mapping/router/contract run: 156 passed; final mapping retirement file: 34 passed.
- Local SAST block scans passed for changed Engine and Backend files. Standards/Spec parallel review was rerun after fixes with no remaining findings.
- Read-only real asset measurement: copied 3 Skills from /Users/freddie/aiworkbench into a temporary sandbox. The same public publish+verify entry points changed from 12 recursive checks / 10.115ms to 6 bounded checks / 1.413ms on local disk. A synthetic 10,000-file Skill completed publish+verify in 0.863ms with one check per request. These are local-disk complexity measurements, not production OSSFS or end-to-end claims.

## Compatibility and risk
CONVERGED now means the requested entry is aligned and its source directory passes the bounded root check; it no longer proves every descendant is readable. Upload, publication, materialization, STRICT mapping, Pool cutover/rollback, Center probe/ensure, v2/v3 negotiation, active-root layout, and response status/item fields are unchanged. Backend retains before snapshot, Reader flush, failure classification, and per-request retry resolution.

Production benefit requires the new Engine to be built, deployed, and actually loaded by the Bot. Existing production Bot `168944/default` was observed on an older strict Engine, so merging this PR alone cannot improve that runtime. No pre-release/production mutation, restart, merge, or deployment was performed. The 3-5s end-to-end target remains to be verified on upgraded pre-release OSSFS.

### DIFF
--- a/src/backend/src/agentclaw/community/core/skill_center/bot_runtime_projector_protocol.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/bot_runtime_projector_protocol.py
@@ -6,15 +6,14 @@ from typing import Protocol, runtime_checkable
 from agentclaw.community.core.skill_center.runtime_projection_contract import (
     BotRuntimeProjectorProtocol as CoreBotRuntimeProjectorProtocol,
     ProjectionScope,
+    ResolvedSkillPlan,
     RuntimeProjectionResult,
 )
 from agentclaw.community.core.skills_pool.models import PoolSkillMapping
 
 
 @runtime_checkable
-class BotRuntimeProjectorProtocol(
-    CoreBotRuntimeProjectorProtocol, Protocol
-):
+class BotRuntimeProjectorProtocol(CoreBotRuntimeProjectorProtocol, Protocol):
     """Transport-facing contract; Core depends only on its sibling contract."""
 
     async def snapshot_skill_mappings(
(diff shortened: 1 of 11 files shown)


totalfrank (approved):
LGTM — reviewed `dff9447` (11 files: bounded best-effort source checks + request-scoped caching in `layout_activation.py`, and the `resolve_plan`/`apply_plan` split across the `BotRuntimeProjector` protocol/implementation and `_mutation_flow.py`, plus tests).

Traced the retirement-path switch from `exists()`/`is_symlink()` to `lstat()` (handles dangling symlinks and surfaces permission errors as retryable instead of silently treating them as absent — an improvement, not a regression), the new `_BestEffortSourceInspection` cache scoping (fresh per call to `_best_effort_mapping_results`, so publish/verify still observe independently as intended), and the resolve/apply split in `_mutation_flow.py` (retired mappings are still computed from the actual post-mutation `current_mappings` and passed into `apply_plan`'s validation before any engine write). No bugs found in the new code.

---

PR #1930, opened 2026-09-05
fix(skill-center): gate installation reader reconciliation

FreddieSun:
## Problem
Installation is the active-capability SSOT, but new Bot creation relied on a later Reader/Projector write and ordinary SkillSet reconciliation cannot be removed before historical backfill is accepted.

## Solution
- Initialize missing Skill/MCP Installation rows DB-only during normal and desktop Bot creation, with append-only retry repair.
- Keep full reconciliation as the default and backfill path; add an explicit default-off switch that reduces post-backfill effective reads to Default membership plus exclusions.
- Separate pure SkillSet membership reads from effective-state synchronization, retaining Installation as the only effective source.

## Validation
- `uv run pytest -q` — 17643 passed, 60 skipped.
- Targeted Installation, Reader, creation, OpenAPI, manifest, DI, architecture and contract tests passed.
- Standards/Spec review completed; high-priority findings were fixed.

## Compatibility and risk
No public HTTP contract or schema change. `SC_INSTALLATION_DEFAULT_SYNC_ONLY` defaults to `false`; do not enable it until the environment backfill and its acceptance checks complete. The internal backfill endpoint always runs the complete resolver. Global Default membership removal remains controlled operations cleanup and is not inferred from Installation rows.

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py
@@ -422,6 +422,16 @@ class BotService(BotServiceProtocol):
         self._runtime_reconciler = runtime_reconciler
         self._runtime_reconciler_provider = runtime_reconciler_provider
         self._bot_quota_service = bot_quota_service
+    def _initialize_capability_installations(
+        self, *, bot_id: str, owner_id: str, bot: "Dict[str, Any] | None" = None
+    ) -> None:
+        """Persist a new Bot's DB-only capability projection before returning.
+
+        No Runtime projection or device readiness check is involved here.
+        """
+        self._skill_set_factory.initialize_installations(
+            bot_id=bot_id, owner_id=owner_id, bot=bot
+        )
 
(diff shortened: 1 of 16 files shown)


totalfrank (approved):
LGTM — reviewed `e8dd6ae0` (16 files: new-Bot Installation initialization on the creation/retry paths, the `sync_default_installations`/`initialize_installations` reconciliation modes, and the reader's split of membership reads from effective-state synchronization).

I traced every production call site of `member_skill_ids()` (`SkillQueryService.list_bot_skills`, and both `resolve()`/`plan()` in the skills manifest materialiser) to confirm each still calls `synchronize_installations`/`active_skill_assets` before reading membership, since `member_skill_ids()` no longer flushes itself. Also checked `_resolve_default_sync_plan`'s exclusion/active-claim arithmetic against its test, the `allow_uninstall` gating in `_reconcile_installations`, and the desktop/bot-service retry paths for DI wiring (`reader` is bound in `skill_center_module.py`, so the `RuntimeError` guard in `factories.py` is unreachable in production). No bugs found in the new code.


---

PR #1933, opened 2026-09-05
feat(bcs): correlate request stages and dependency latency

vzvince:
## Problem

Slow BCS HTTP requests could not be consistently correlated with authentication, repository, dependency-client and internal-wait stages. Cache fallbacks and partially failed status writes also hid useful failure evidence, while synchronous file writes could block request workers.

## Solution

Write operation observations to logs with request IDs, parent operation IDs, duration, outcome, cancellation records and one diagnostic warning for calls pending at five seconds. Instrument bot queries, chat acceptance, authentication, repository fallbacks, selected locks/queues and shared HTTP dependency adapters. Preserve request correlation in selected detached work by copying request/operation IDs without holding or entering the request span. Emit an aggregate operation summary when the request returns.

Put reusable timing, log events, request summaries and correlation scopes in `bcs-observability`. This base package has no OpenTelemetry or metrics implementation dependencies. It prints structured events through the existing tracing subscriber. Observation context and added log fields contain request/operation IDs only; the added HTTP/WebSocket log instrumentation does not acquire or propagate OpenTelemetry trace IDs. `bcs-telemetry` retains only its original GenAI attribute encoders. A dependency gate prevents trace/metric implementations from returning to the base package.

The added observations create no tracing spans and export no metrics. Existing gateway span selection, downstream TraceContext propagation and HTTP/business metrics keep their original behavior. Detached work carries IDs and the log subscriber without holding or entering a span.

Buffer file logging with bounded nonblocking writers. Document observation semantics, covered boundaries and limitations. Preserve current dev behavior for chat Run audit snapshots and webhook preflight failure handling.

## Validation

- Current affected public unit/integration suites: **283 passed**, 19 existing ignored tests. Packages: `bcs-observability`, `bcs-telemetry`, `bcs-http`, `bcs-ws`, `bcs-auth-api`, `bcs-bot`, `bcs-bot-store`; Cargo targets: `--lib --test operations --test gen_ai_messages --test request_observations --test frame_compat`. The suite ran with `--offline`; the existing local-listener test required a host run outside the sandbox.
- HTTP request and WS frame suites were rerun after final test cleanup: **6 + 20 passed**, 14 existing ignored WS tests. These are included in the 283 above.
- Matching internal `bcs-layotto` and `bcs-auth-agentpass` unit suites: **38 passed**, 9 existing ignored tests.
- RED/GREEN regressions verify observation records omit `trace_id`, including under an active OpenTelemetry span. Request/operation correlation, nested/concurrent request scopes, cancellation, stalled warnings, summaries and explicit detached propagation remain covered.
- HTTP regressions cover malformed requests, cancellation and coordination callbacks. Coordination requests receive request IDs and lifecycle logs while the existing no-span rule remains intact.
- Real exporters verify both original A2A span names, parent trace/span IDs and detached-work lifetime. WS tests preserve normal/aliased callback span count, parent and attributes, and missing-context behavior. Existing HTTP span rules/attributes, original WS dispatcher, GenAI encoders, provider trace propagation and bootstrap configuration were also compared with their prior implementations.
- `cargo check --offline --workspace` passed for both public BCS and the matching internal overlay.
- The all-features normal/build observation dependency gate, internal architecture check and `git diff --check` passed. Independent code review found no functional issues; reported unused test bindings were removed.
- The earlier full public architecture run did not pass: existing dependency/import/naming/config-coverage/conformance-registration failures remain. Its first conformance-discovery invocation failed, and the duplicate R25 enumeration was stopped. That full gate was not rerun for this trace-field cleanup; no gate or baseline was weakened and full architecture success is not claimed.
- Push uses `--no-verify` as requested. Full pre-push/singlebox coverage and live rollout validation were not run.

## Compatibility and risk

Application and plugin result contracts, authorization decisions, dependency deadlines/retries and routing policy are unchanged. HTTP responses gain a validated request ID. The affected consumers are BCS HTTP clients and log collection.

- [x] 契约方向分类: Existing Service API and Plugin API signatures are unchanged; this introduces an auxiliary, log-only utility boundary using tracing events and subscriber configuration.
- Operation callers migrate from `bcs_telemetry` to `bcs_observability`; GenAI encoder callers retain `bcs_telemetry`. Public and internal BCS path consumers must move together. The internal overlay is updated and checked against the paired public revision.
- Added observation records use request/operation IDs and have no distributed `trace_id` field. Existing OpenTelemetry spans, tracing diagnostics, dependency protocol trace fields and configuration retain their behavior. Future internal SOFATracer integration is outside this change.
- No new plugin runtime, exporter, metrics or automatic span instrumentation is introduced. Bootstrap retains logging/exporter lifecycle ownership; no architecture waiver is required for this extraction.

Fast operation completions are DEBUG; failures and operations taking at least 100 ms are WARN. Request summaries are INFO. File log queues are bounded to 4,096 lines per output and can drop records under sink saturation; no dropped-line metric is exported. Instrumented futures are boxed to bound nested adapter stack usage. Response-ready latency excludes the remainder of streaming response lifetimes, and nested durations must not be summed as request wall time. Independent WS callbacks do not inherit the original HTTP request ID; existing business identifiers such as `run_id` support cross-request investigation, while generic operation records do not automatically carry those IDs. Revert the PR commits to roll back the instrumentation.

## Spec

`src/bcs/docs/observability/request-operation-logging.md`


questchaos (approved):
LGTM

PR #1791, opened 2026-09-01
feat(backend): create a bot with its configuration manifest (W13)

totalfrank:
## Problem

Design §3.1 promises that "the first configuration a bot receives already contains the manifest's results". With only `PUT`, that promise is unreachable for a genuinely new bot: the earliest a caller can write a manifest is after the container is up, already unconfigured, with the start command already composed without the manifest's `script`. Every bot's first boot is empty.

W13 (#1696) closes that gap with a public async API that takes a manifest alongside the ordinary creation attributes. It is the item that delivers the business ask — W8 covers every other apply point; this one covers creation.

Dependencies **W1** (#1727) and **W4** (#1751) are both on `dev`.

## Solution

The SDD artifacts were reviewed here through five revisions before any code landed; **the implementation is now in**, following `tasks.md` task by task, and has been through two rounds of review since.

Submission validates the manifest, stores it against the allocated `bot_id`, applies to AgentPass and enqueues a durable job, then returns `202`. The job waits for authorization, applies the manifest in two phases around bot creation, and finishes. A caller polls one `GET` with nothing but the bot's id.

**Applying itself is now a task-queue task, on every path** — the pre-container phase, the post-container phase, and the explicit apply on an already-running bot. That is in scope here rather than as a follow-up because creation *depends* on an apply completing: the startup-script row must land before the bot is created, and a daemon thread that dies takes that guarantee with it, booting the bot without its script. The queue's own README names the thread pattern as the one it exists to replace.

**Scope:** this endpoint is **ARCA-only**. teclaw configures a bot by composing its artifact at provision time, which is a different mechanism from the pre/post-container split here — the work items assign that arm to W8 (its scope names "teclaw 在第一份 artifact 组装之前", its first criterion is the first-artifact guarantee, and its scale note lists `TeclawProvisionService`). A teclaw creation is refused at submission, naming W8.

### What ships

| | |
| --- | --- |
| `POST /openapi/v1/bots/with-manifest` | Submit. Answers `202` with `bot_id` and both authorization handles. |
| `GET /openapi/v1/bots/{bot_id}/with-manifest/status` | Poll. Takes the path parameter and nothing else. |
| `core/bot_config_manifest/creation.py` | The seam creation calls: preflight, persist, phase A, discard, and the two job operations. |
| `core/bot_config_manifest/create_job.py` | The creation job's step machine and its deadline. |
| `core/bot_config_manifest/apply/apply_task.py` | One handler for all three apply paths. |
[...]

PR #1936, opened 2026-09-06
fix(bot-config-manifest): align DDL with OceanBase, add ac_task_queue schema

totalfrank:
## Problem

The hand-written DDL for the six `bot-config-manifest` tables was authored in portable MySQL form, but production is OceanBase. Measured against a real deployed table (`SHOW CREATE TABLE ac_bots`) it diverged in three ways that are not cosmetic:

1. **No index carried `GLOBAL`.** On a partitioned table an index without it is partition-local, and a partition-local `UNIQUE` enforces uniqueness only *within* a partition. For `uk_manifest_apply_lock` that is the entire mechanism — the `UNIQUE` constraint **is** the apply lock (work-items §2.7), so a partition-local one lets two applies run against one bot simultaneously. The same applies to the manifest key, the credential name and the CLI tool key, where the failure surfaces as duplicate rows appearing silently rather than as an error.
2. **`gmt_*` columns were `DATETIME` where the platform uses `TIMESTAMP`.** `ac_bots` and every table added since (`ac_bot_cli_call_config`, `ac_bot_app_grant`) use `TIMESTAMP` for the audit columns, and `skill_center/sql/2026_09_03_align_space_skill_timestamps_with_gmt.sql` is the repair that convention exists to avoid repeating.
3. **`AUTO_INCREMENT_MODE` left to the tenant default on tables whose reads depend on it.** `config_manifest_apply.py` answers `GET .../config-manifest/last-apply` with `ORDER BY id DESC`. Under `NOORDER` each observer caches its own auto-increment range, so ids stop being monotonic in insertion order and "max id" stops meaning "newest" — last-apply would serve a stale report while the newer one sat behind a smaller id. `ac_manifest_content` uses id as the within-second tie-break `keep_last` resolves on, and `ac_task_queue`'s `_find_by_key` states the same assumption in a code comment.

Separately, **`ac_task_queue` had no DDL in the repository at all.** Local and test runs build it from ORM metadata, and the OceanBase-only `GLOBAL` requirement lived in README prose. Since #1791 every manifest apply path goes through the queue, and W13 (#1696) creates bots on it, so a deployment that provisions the six manifest tables and not this one exposes an API that accepts work and never performs it.

## Solution

None of the six manifest tables are provisioned yet, so this corrects the `CREATE TABLE` files in place rather than shipping `ALTER`s — no migration, no index drop against a live apply lock.

- **`GLOBAL` on every index**, in all seven tables.
- **Timestamp type is decided by who fills the column**, which is the rule rather than a list of names:
  - `gmt_create` / `gmt_modified` → `TIMESTAMP`. The database fills these (`DEFAULT CURRENT_TIMESTAMP`, `func.now()` from the ORM), so the session-time-zone conversion is a no-op round trip and they match `ac_bots`.
  - `started_at`, `finished_at`, `fetched_at` → **`DATETIME`**. These are filled by the application, and `TIMESTAMP` reads a bound naive value as session-local before converting it — so an instant already normalised to UTC (`fetched_at`, per `content/service.py:271-276`) or taken from `datetime.now()` is stored shifted by the session offset.
- **`AUTO_INCREMENT_MODE = 'ORDER'`** on `ac_bot_config_manifest_apply`, `ac_manifest_content` and `ac_task_queue`.
- `ENGINE=InnoDB` dropped — OceanBase neither needs nor echoes it — and the table options now match the sibling DDL in `core/caller_identity/sql/`.
- The storage options `ac_bots` echoes back (`BLOCK_SIZE`, `REPLICA_NUM`, `COMPRESSION`, `TABLET_SIZE`, `PCTFREE`, `ROW_FORMAT`) are deliberately **not** written: they are `SHOW CREATE TABLE` output, not input, and no other `.sql` in this repo writes them.
- Each file carries a dialect note recording the reasoning, so the modifiers are not later "tidied" away as noise.
- New `core/task_queue/sql/2026_09_06_task_queue.sql`, plus two contract test files (below).

Column widths, the surrogate-vs-logical key choices, and the tenant-leading index order are unchanged. The ORM keeps `DateTime` against both column kinds — `ac_skill_version.published_at` is the existing precedent for `DateTime` over a `TIMESTAMP` column — so no model change is needed.

### Two contract tests, because prose already failed once

Both rules below were violated during this PR and caught by review, not CI. They are now asserted, and **each assertion was verified by injecting the regression it names and watching it fail**:

- `tests/community/core/task_queue/test_task_queue_schema_contract.py` — the DDL declares every unique index the ORM declares (both, legacy included), both are `GLOBAL`, and `AUTO_INCREMENT_MODE` stays pinned.
- `tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py` — app-supplied columns stay `DATETIME`, DB-filled columns stay `TIMESTAMP`, no bare `TIMESTAMP NOT NULL` without a `DEFAULT` exists in any file (such a column can pick up an implicit `ON UPDATE CURRENT_TIMESTAMP`, which the apply record's terminal write would fire on every apply), and every table in every file is actually parsed.

## Validation

- On the current head `96d255e6`: `pytest tests/community/core/bot_config_manifest tests/community/core/task_queue` — **880 passed**. On `ddeba211`, the wider `bot_config_manifest` + `task_queue` + `repository` run — **1900 passed, 5 skipped**. CI's `Backend unit tests` covers the full backend suite on every commit and is green on `96d255e6`.
- Each DDL file re-parsed after editing to confirm it still contains only complete, paren-balanced `CREATE TABLE` statements.
- Both contract tests verified against injected regressions, not just observed green. **Two assertions were fixed by that exercise before landing**: one exempted any declaration containing `null`, which also matched `NOT NULL`, so it passed the exact declaration it exists to reject; another parsed per file rather than per table, so in the one file declaring two tables the apply table's `gmt_*` columns were masked by the lock table's.
- **Not run:** the DDL has never been executed against an OceanBase instance — none is reachable from this environment. Every claim about `GLOBAL`, `AUTO_INCREMENT_MODE` and the implicit-default rule is reasoned from the deployed `ac_bots` schema and documented engine behaviour, not observed. **Apply it to a non-production tenant before trusting it.**

## Compatibility and risk

Low for the six manifest tables: they do not exist in any environment yet, so there is no data to migrate and no rollback beyond reverting.

`ac_task_queue` is the one to read carefully. The file is **reconstructed from the ORM and README, not dumped from a live table**, and its header says so. It keeps `DATETIME` throughout — mirroring the ORM and what was provisioned, rather than adopting this PR's `TIMESTAMP` convention — because every claim, reclaim and deadline comparison in that component is a DB-clock comparison. It **declares both unique indexes the ORM declares, including the legacy pre-`app` one**, so a freshly provisioned database behaves the way the test suite asserts; dropping that index is a single future change touching `models.py`, this file and the deployed tables together. For any deployment that already has the table, `SHOW CREATE TABLE ac_task_queue` outranks both this file and the ORM.

## Review history

Four rounds of automated review (Codex and Claude Code) across `06df08d6` → `acb6c1c5` → `ddeba211` → `96d255e6` produced **six findings. All were verified against the source, all were real, none were disputed.**

Three were defects in the DDL — two of them the same root cause, converting *every* datetime column to `TIMESTAMP` when only the database-filled ones should have moved. Three were gaps in the tests guarding it, two of those in tests written during the review itself. Both reviewers are clear on `96d255e6`.

## Related issues

Touches the DDL from #1469 (W1), #1471 (W3), #1472 (W4), #1477 (W9), #1510 (W11), and the queue W13 (#1696) runs on.

https://claude.ai/code/session_015bEPXsL1QHUYjS7PzQGjTk

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
@@ -34,6 +34,38 @@
 -- apply lock — so one tenant's apply could block or unblock another's — and
 -- read each other's apply reports.
 
+-- DIALECT: OCEANBASE, MYSQL MODE. What this file writes follows the deployed
+-- tables there rather than portable MySQL -- compare ``SHOW CREATE TABLE
+-- ac_bots``, and the sibling DDL in core/caller_identity/sql/.
+--
+--   * EVERY INDEX IS DECLARED ``GLOBAL``. On a partitioned table an index
+--     without it is partition-local, and a partition-local UNIQUE enforces
+--     uniqueness only *within* a partition. For ``uk_manifest_apply_lock``
+--     that is the whole mechanism -- the UNIQUE constraint *is* the lock,
+--     so a partition-local one lets two applies hold it at once. These tables are not
+--     partitioned today, so it is a no-op now and cheap insurance later: what
+--     it prevents surfaces as duplicate rows, silently, never as an error.
+--     SQLAlchemy cannot render the modifier, so this file is the only place it
+--     can live.
+--   * ``TIMESTAMP``, NOT ``DATETIME``, for gmt_* and the business timestamps,
+--     matching ac_bots and every table added since. Mixing the two is what
+--     skill_center/sql/2026_09_03_align_space_skill_timestamps_with_gmt.sql
+--     had to repair: TIMESTAMP converts by session time zone and DATETIME does
+--     not, so a session outside Asia/Shanghai reads the two as disagreeing by
+--     the offset. The ORM keeps ``DateTime`` either way -- ac_skill_version's
+--     published_at is the precedent for that pairing.
+--   * NO ``ENGINE`` CLAUSE, and no BLOCK_SIZE / REPLICA_NUM / COMPRESSION /
+--     TABLET_SIZE / PCTFREE / ROW_FORMAT. OceanBase applies its own defaults
+--     and echoes them back from SHOW CREATE TABLE; writing them here would be
+--     copying its own output back at it.
+--   * ``AUTO_INCREMENT_MODE = 'ORDER'``, pinned rather than inherited from the
+--     tenant default, because this table's read order rests on it:
+--     ``config_manifest_apply.py`` answers GET .../last-apply with
+--     ``ORDER BY id DESC``. Under NOORDER each observer caches its own
+--     auto-increment range, so ids stop being monotonic in insertion order and
+--     "max id" stops meaning "newest" -- last-apply would serve a stale report
+--     while the newer one sat behind a smaller id.
+
 CREATE TABLE `ac_bot_config_manifest_apply` (
   `id`             bigint(20) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Primary key',
   -- The report's public identity: returned by the 202 from POST .../apply and
@@ -68,20 +100,23 @@ CREATE TABLE `ac_bot_config_manifest_apply` (
   -- ("app:7:on-behalf-of:<...>"), so the composed value can legitimately be
   -- long without anything being malformed.
   `actor`          varchar(1024) NOT NULL COMMENT 'Audit: who started it',
-  `started_at`     datetime      NOT NULL COMMENT 'When the apply began',
+  `started_at`     timestamp      NOT NULL COMMENT 'When the apply began',
   -- NULL exactly while status is RUNNING. The two move together.
-  `finished_at`    datetime      NULL     COMMENT 'When it ended; NULL while RUNNING',
+  `finished_at`    timestamp      NULL     COMMENT 'When it ended; NULL while RUNNING',
   `avernet_tenant` varchar(64)   NOT NULL DEFAULT 'teamclaw' COMMENT 'Tenant, for data isolation',
-  `gmt_create`     datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'Row created',
-  `gmt_modified`   datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'Row last modified',
+  `gmt_create`     timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'Row created',
+  `gmt_modified`   timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'Row last modified',
   PRIMARY KEY (`id`),
   -- Two indexes, one per read. There are exactly two reads on this table.
   --
   -- GET .../last-apply — the newest row for this bot.
-  KEY `idx_manifest_apply_latest` (`avernet_tenant`, `env`, `entity_id`, `bot_id`, `id`),
+  KEY `idx_manifest_apply_latest`
+    (`avernet_tenant`, `env`, `entity_id`, `bot_id`, `id`) GLOBAL,
   -- GET .../applies/{apply_id} — the poll by id, scoped to the bot.
-  KEY `idx_manifest_apply_by_id` (`avernet_tenant`, `env`, `entity_id`, `bot_id`, `apply_id`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT='Bot config manifest apply record';
+  KEY `idx_manifest_apply_by_id`
+    (`avernet_tenant`, `env`, `entity_id`, `bot_id`, `apply_id`) GLOBAL
+) AUTO_INCREMENT_MODE = 'ORDER' DEFAULT CHARSET = utf8mb4
+  COMMENT = 'Bot config manifest apply record';
 
 -- NO dry_run COLUMN, deliberately. A dry run mints no apply_id and writes no
 -- row at all, so there is nothing here to mark. A flag would invite a future
@@ -98,8 +133,8 @@ CREATE TABLE `ac_bot_config_manifest_apply_lock` (
   -- reaped as stale could delete the lock a *later* apply legitimately took.
   `lock_token`     varchar(256)  NOT NULL COMMENT 'Fencing token, compared on release',
   `avernet_tenant` varchar(64)   NOT NULL DEFAULT 'teamclaw' COMMENT 'Tenant, for data isolation',
-  `gmt_create`     datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'Row created',
-  `gmt_modified`   datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'Row last modified',
+  `gmt_create`     timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'Row created',
+  `gmt_modified`   timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'Row last modified',
   PRIMARY KEY (`id`),
   -- THE UNIQUE CONSTRAINT IS THE LOCK. One row per bot; concurrent INSERTs are
   -- arbitrated by the database, exactly one wins, and the losers see the
@@ -107,5 +142,7 @@ CREATE TABLE `ac_bot_config_manifest_apply_lock` (
   --
   -- gmt_create is what get_if_stale measures against, on the DB clock, so a
   -- process killed mid-apply cannot hold this forever.
-  UNIQUE KEY `uk_manifest_apply_lock` (`avernet_tenant`, `env`, `entity_id`, `bot_id`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT='Bot config manifest apply serialization lock';
+  UNIQUE KEY `uk_manifest_apply_lock`
+    (`avernet_tenant`, `env`, `entity_id`, `bot_id`) GLOBAL
+) DEFAULT CHARSET = utf8mb4
+  COMMENT = 'Bot config manifest apply serialization lock';
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
@@ -47,13 +47,24 @@
 --     it prevents surfaces as duplicate rows, silently, never as an error.
 --     SQLAlchemy cannot render the modifier, so this file is the only place it
 --     can live.
---   * ``TIMESTAMP``, NOT ``DATETIME``, for gmt_* and the business timestamps,
---     matching ac_bots and every table added since. Mixing the two is what
---     skill_center/sql/2026_09_03_align_space_skill_timestamps_with_gmt.sql
---     had to repair: TIMESTAMP converts by session time zone and DATETIME does
---     not, so a session outside Asia/Shanghai reads the two as disagreeing by
---     the offset. The ORM keeps ``DateTime`` either way -- ac_skill_version's
---     published_at is the precedent for that pairing.
+--   * ``TIMESTAMP`` FOR THE DB-CLOCK COLUMNS, ``DATETIME`` FOR THE REST, and
+--     which one a column gets is decided by who fills it, not by taste.
+--     gmt_create and gmt_modified are written by the database itself
+--     (``DEFAULT CURRENT_TIMESTAMP``, and ``func.now()`` from the ORM), so
+--     TIMESTAMP's session-time-zone conversion is a no-op round trip and they
+--     match ac_bots and every table added since --
+--     skill_center/sql/2026_09_03_align_space_skill_timestamps_with_gmt.sql is
+--     the repair that convention exists to avoid repeating.
+--
+--     A column the APPLICATION fills stays DATETIME. TIMESTAMP reads the naive
+--     value being bound as session-local wall time and converts it to UTC for
+--     storage, so a Python-supplied instant is stored shifted by the session
+--     offset -- eight hours under the Asia/Shanghai session assumed here. This
+--     was got wrong in the first draft of this change and caught in review; started_at and
+--     finished_at are the columns in question here, per their comment below.
+--
+--     The ORM keeps ``DateTime`` for both kinds -- ac_skill_version's
+--     published_at is the precedent for ``DateTime`` over a TIMESTAMP column.
 --   * NO ``ENGINE`` CLAUSE, and no BLOCK_SIZE / REPLICA_NUM / COMPRESSION /
 --     TABLET_SIZE / PCTFREE / ROW_FORMAT. OceanBase applies its own defaults
 --     and echoes them back from SHOW CREATE TABLE; writing them here would be
@@ -100,9 +111,16 @@ CREATE TABLE `ac_bot_config_manifest_apply` (
   -- ("app:7:on-behalf-of:<...>"), so the composed value can legitimately be
   -- long without anything being malformed.
   `actor`          varchar(1024) NOT NULL COMMENT 'Audit: who started it',
-  `started_at`     timestamp      NOT NULL COMMENT 'When the apply began',
+  -- DATETIME, not TIMESTAMP, for both of these: they are filled by the
+  -- application from datetime.now() (process-local, naive), never by the
+  -- database. TIMESTAMP binds a naive value as session-local, so these would
+  -- be stored correctly only where the process time zone happens to equal the
+  -- database session's -- and a container on UTC against an Asia/Shanghai
+  -- session is exactly where that stops being true. gmt_* below are a
+  -- different case: the database fills those itself.
+  `started_at`     datetime      NOT NULL COMMENT 'When the apply began',
   -- NULL exactly while status is RUNNING. The two move together.
-  `finished_at`    timestamp      NULL     COMMENT 'When it ended; NULL while RUNNING',
+  `finished_at`    datetime      NULL     COMMENT 'When it ended; NULL while RUNNING',
   `avernet_tenant` varchar(64)   NOT NULL DEFAULT 'teamclaw' COMMENT 'Tenant, for data isolation',
   `gmt_create`     timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'Row created',
   `gmt_modified`   timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'Row last modified',
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
@@ -68,6 +68,33 @@
 -- ``.tmp-*`` staging files from crashed writes are not audit facts and are
 -- age-swept at store time (see content/service.py); they are the only thing
 -- a sweeper may ever touch.
+-- DIALECT: OCEANBASE, MYSQL MODE. What this file writes follows the deployed
+-- tables there rather than portable MySQL -- compare ``SHOW CREATE TABLE
+-- ac_bots``, and the sibling DDL in core/caller_identity/sql/.
+--
+--   * EVERY INDEX IS DECLARED ``GLOBAL``. On a partitioned table an index
+--     without it is partition-local, and a partition-local UNIQUE enforces
+--     uniqueness only *within* a partition. These tables are not
+--     partitioned today, so it is a no-op now and cheap insurance later: what
+--     it prevents surfaces as duplicate rows, silently, never as an error.
+--     SQLAlchemy cannot render the modifier, so this file is the only place it
+--     can live.
+--   * ``TIMESTAMP``, NOT ``DATETIME``, for gmt_* and the business timestamps,
+--     matching ac_bots and every table added since. Mixing the two is what
+--     skill_center/sql/2026_09_03_align_space_skill_timestamps_with_gmt.sql
+--     had to repair: TIMESTAMP converts by session time zone and DATETIME does
+--     not, so a session outside Asia/Shanghai reads the two as disagreeing by
+--     the offset. The ORM keeps ``DateTime`` either way -- ac_skill_version's
+--     published_at is the precedent for that pairing.
+--   * NO ``ENGINE`` CLAUSE, and no BLOCK_SIZE / REPLICA_NUM / COMPRESSION /
+--     TABLET_SIZE / PCTFREE / ROW_FORMAT. OceanBase applies its own defaults
+--     and echoes them back from SHOW CREATE TABLE; writing them here would be
+--     copying its own output back at it.
+--   * ``AUTO_INCREMENT_MODE = 'ORDER'``, pinned rather than inherited, because
+--     id is the tie-break keep_last resolves on: reads order by
+--     ``gmt_create DESC, id DESC``, so within one second the newest receipt is
+--     the largest id -- true only while ids are allocated in order.
+
 CREATE TABLE `ac_manifest_content` (
   `id`            bigint(20) unsigned NOT NULL AUTO_INCREMENT COMMENT '主键ID',
   `avernet_tenant` varchar(64)  NOT NULL DEFAULT 'teamclaw' COMMENT '数据隔离租户',
@@ -89,7 +116,7 @@ CREATE TABLE `ac_manifest_content` (
   -- reconciliation anchor, not this).
   `content_type`  varchar(256)  DEFAULT NULL COMMENT '响应 Content-Type（advisory；超宽存 NULL）',
   `size_bytes`    bigint(20) unsigned NOT NULL COMMENT '字节数（与 digest 同为对账锚）',
-  `fetched_at`    datetime      NOT NULL COMMENT '拉取时间（FetchedObject.fetched_at）',
+  `fetched_at`    timestamp      NOT NULL COMMENT '拉取时间（FetchedObject.fetched_at）',
   -- The join back to the apply record (ac_bot_config_manifest_apply), and
   -- the per-entry identity a fetch served: apply_id is what that table's
   -- own comment ("also what a per-entry table would join on") points at
@@ -106,13 +133,15 @@ CREATE TABLE `ac_manifest_content` (
   `category`      varchar(32)   DEFAULT NULL COMMENT '条目类目（skills/identity/…；条目维度之一）',
   `entry_identity` varchar(256) DEFAULT NULL COMMENT '条目标识（对账锚外的维度：skill name / identity type / …）',
   `modifier`      varchar(1024) NOT NULL DEFAULT '' COMMENT '审计：触发拉取的身份',
-  `gmt_create`    datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT '创建时间',
-  `gmt_modified`  datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT '修改时间',
+  `gmt_create`    timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT '创建时间',
+  `gmt_modified`  timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT '修改时间',
   PRIMARY KEY (`id`),
   -- The audit query shape: one bot's receipts, newest first. Leading column
   -- tenant keeps it aligned with the guard's filter.
-  KEY `idx_tenant_env_entity_bot` (`avernet_tenant`, `env`, `entity_id`, `bot_id`, `gmt_create`),
+  KEY `idx_tenant_env_entity_bot`
+    (`avernet_tenant`, `env`, `entity_id`, `bot_id`, `gmt_create`) GLOBAL,
   -- "What did apply X fetch" — the join the apply record's own comment
   -- anticipated this table providing.
-  KEY `idx_apply` (`apply_id`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT='manifest 拉取内容平台副本的溯源日志（行 append-only，字节在内容寻址 blob 目录）';
+  KEY `idx_apply` (`apply_id`) GLOBAL
+) AUTO_INCREMENT_MODE = 'ORDER' DEFAULT CHARSET = utf8mb4
+  COMMENT = 'manifest 拉取内容平台副本的溯源日志（行 append-only，字节在内容寻址 blob 目录）';
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
@@ -79,13 +79,24 @@
 --     it prevents surfaces as duplicate rows, silently, never as an error.
 --     SQLAlchemy cannot render the modifier, so this file is the only place it
 --     can live.
---   * ``TIMESTAMP``, NOT ``DATETIME``, for gmt_* and the business timestamps,
---     matching ac_bots and every table added since. Mixing the two is what
---     skill_center/sql/2026_09_03_align_space_skill_timestamps_with_gmt.sql
---     had to repair: TIMESTAMP converts by session time zone and DATETIME does
---     not, so a session outside Asia/Shanghai reads the two as disagreeing by
---     the offset. The ORM keeps ``DateTime`` either way -- ac_skill_version's
---     published_at is the precedent for that pairing.
+--   * ``TIMESTAMP`` FOR THE DB-CLOCK COLUMNS, ``DATETIME`` FOR THE REST, and
+--     which one a column gets is decided by who fills it, not by taste.
+--     gmt_create and gmt_modified are written by the database itself
+--     (``DEFAULT CURRENT_TIMESTAMP``, and ``func.now()`` from the ORM), so
+--     TIMESTAMP's session-time-zone conversion is a no-op round trip and they
+--     match ac_bots and every table added since --
+--     skill_center/sql/2026_09_03_align_space_skill_timestamps_with_gmt.sql is
+--     the repair that convention exists to avoid repeating.
+--
+--     A column the APPLICATION fills stays DATETIME. TIMESTAMP reads the naive
+--     value being bound as session-local wall time and converts it to UTC for
+--     storage, so a Python-supplied instant is stored shifted by the session
+--     offset -- eight hours under the Asia/Shanghai session assumed here. This
+--     was got wrong in the first draft of this change and caught in review; fetched_at
+--     is the column in question here, and its own comment below says why.
+--
+--     The ORM keeps ``DateTime`` for both kinds -- ac_skill_version's
+--     published_at is the precedent for ``DateTime`` over a TIMESTAMP column.
 --   * NO ``ENGINE`` CLAUSE, and no BLOCK_SIZE / REPLICA_NUM / COMPRESSION /
 --     TABLET_SIZE / PCTFREE / ROW_FORMAT. OceanBase applies its own defaults
 --     and echoes them back from SHOW CREATE TABLE; writing them here would be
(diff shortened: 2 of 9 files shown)


src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
```diff
@@ -68,20 +100,23 @@ CREATE TABLE `ac_bot_config_manifest_apply` (
   -- ("app:7:on-behalf-of:<...>"), so the composed value can legitimately be
   -- long without anything being malformed.
   `actor`          varchar(1024) NOT NULL COMMENT 'Audit: who started it',
-  `started_at`     datetime      NOT NULL COMMENT 'When the apply began',
+  `started_at`     timestamp      NOT NULL COMMENT 'When the apply began',
```
totalfrank: **Blocking**: converting `started_at` from `DATETIME` to `TIMESTAMP` here without an explicit `DEFAULT` makes it the table's *first* `TIMESTAMP` column declared with neither the `NULL` attribute nor a `DEFAULT`/`ON UPDATE` clause. Under `explicit_defaults_for_timestamp=OFF` (the traditional MySQL/OceanBase behavior — MySQL itself only flipped the default to `ON` in 8.0.22), the server implicitly attaches `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` to exactly this column.

That's not a dormant risk here — it fires on every apply. `config_manifest_apply.py`'s `finish()` (`core/repository/implementations/bot/config_manifest_apply.py:154-156`) does `row.status = status; row.report = report; row.finished_at = datetime.now()` and flushes — an `UPDATE` that never mentions `started_at` — and both termination paths in `apply_termination.py` (`terminate_unstartable`, `terminate_on_launch_failure`) route through the same `finish()`. MySQL's `ON UPDATE CURRENT_TIMESTAMP` fires whenever *any* other column in the row changes, whether or not the triggering column is named in the `SET` clause. If the implicit attribute applies on the deployed tenant, every apply — success or failure — silently overwrites "When the apply began" with the finish time, corrupting the one thing this table exists to audit.

`finished_at` two lines below is safe (it explicitly declares `NULL`, which is exempt from the implicit-attach rule). `started_at` isn't. Suggest pinning it explicitly the same way `gmt_create` already is, three lines down — harmless since the app always supplies a real value at insert, and it suppresses the implicit `ON UPDATE` attach:

```suggestion
  `started_at`     timestamp      NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'When the apply began',
```

---

src/backend/src/agentclaw/community/core/task_queue/sql/2026_09_06_task_queue.sql
```diff
@@ -0,0 +1,101 @@
...
+-- "Status" section says the production table "must be provisioned before
+-- enabling the worker", local and test runs build it from the shared ORM
+-- metadata, and the OceanBase-only ``GLOBAL`` modifier lived in prose alone.
+-- Every manifest apply path now goes through this queue (PR #1791), and W13
+-- (#1696) creates bots on it, so a deployment that provisions the six
+-- bot_config_manifest tables and not this one has an API that accepts work and
+-- never performs it.
+--
+-- WHAT THIS FILE IS NOT. It is reconstructed from ``repository/models.py`` and
+-- README, not dumped from a live table. On a deployment that already HAS
```
totalfrank, line 13: **Non-blocking**: `core/task_queue/README.md`'s "Provisioning" section currently states, as a load-bearing invariant: "`repository/models.py` is the source of truth for the schema. **No DDL is checked in**: the deployed tables are provisioned out of band, so a copy in the repo would be a second definition to keep in step rather than an authority." This PR checks in exactly that for `ac_task_queue`, but doesn't touch the README, so the two documents now directly contradict each other. A reader who trusts the README's Provisioning section rather than browsing `sql/` will still believe no schema file exists for this table, and won't know this file (unlike `models.py`) is a reference snapshot rather than an authority for an already-provisioned deployment.

This file's own header already explains the nuance well (lines 13-22) — worth mirroring that into the README's Provisioning section so the two don't disagree.

---

totalfrank (commented):
Reviewed commit 06df08d (the PR's only commit) against merge-base 7ba31e9. Checked correctness of the DATETIME→TIMESTAMP and index-GLOBAL conversions against the actual column-write paths, OceanBase table-option syntax, and the new `ac_task_queue` DDL against its ORM model and README.

One blocking finding: the `started_at`→`TIMESTAMP` conversion in `2026_08_31_bot_config_manifest_apply.sql` risks silent audit-column corruption on every apply's terminal write, via MySQL/OceanBase's implicit-default rule for a table's first bare `TIMESTAMP` column. One non-blocking finding: the new `ac_task_queue` DDL file contradicts `core/task_queue/README.md`'s explicit "no DDL is checked in" statement, which isn't updated in this diff.

---

FreddieSun (dismissed):

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_manifest_content.sql
```diff
@@ -89,7 +116,7 @@ CREATE TABLE `ac_manifest_content` (
   -- reconciliation anchor, not this).
   `content_type`  varchar(256)  DEFAULT NULL COMMENT '响应 Content-Type（advisory；超宽存 NULL）',
   `size_bytes`    bigint(20) unsigned NOT NULL COMMENT '字节数（与 digest 同为对账锚）',
-  `fetched_at`    datetime      NOT NULL COMMENT '拉取时间（FetchedObject.fetched_at）',
+  `fetched_at`    timestamp      NOT NULL COMMENT '拉取时间（FetchedObject.fetched_at）',
```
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in `acb6c1c5`. This was a real defect I introduced, and the diagnosis is exactly right.

Verified against the source before changing anything: `content/service.py:271-276` does `fetched_at.astimezone(timezone.utc).replace(tzinfo=None)`, so the value bound is an already-UTC naive instant. Under `TIMESTAMP` the server reads that as session-local wall time and converts it again — storing it eight hours off under the Asia/Shanghai session these tables assume. For the table whose entire purpose is answering "what did this bot receive, and when" (work-items §2.8), a silently shifted instant is the one defect it cannot carry.

`started_at` / `finished_at` verified the same way: `config_manifest_apply_service.py:380` and `apply_termination.py:79,144` fill them from `datetime.now()`, process-local and naive. Correct under `TIMESTAMP` only where the process time zone equals the database session's — which a container on UTC against an Asia/Shanghai session is precisely not.

All three are back to `DATETIME`. `gmt_create` / `gmt_modified` stay `TIMESTAMP` across all six tables, because those are filled by the database itself (`DEFAULT CURRENT_TIMESTAMP`, `func.now()` from the ORM), so the conversion is a no-op round trip and they match `ac_bots`.

The underlying mistake was stating the rule as a list of column names. The dialect note in each file now states it by *who fills the column* — database-filled gets `TIMESTAMP`, application-filled stays `DATETIME` — which is the form that would have prevented this.

---

src/backend/src/agentclaw/community/core/task_queue/sql/2026_09_06_task_queue.sql
```diff
@@ -0,0 +1,101 @@
...
+-- ac_task_queue, ``SHOW CREATE TABLE ac_task_queue`` is the truth and this file
+-- is not -- reconcile against it before running anything here. Two known ways
+-- an existing table differs:
+--
+--   * THE LEGACY DEDUP INDEX. ``uk_env_task_type_active_idempotency_key``
+--     (no ``app``) is still present on the deployed table, and dropping it is
+--     the last step of the app-scoping migration -- README "Provisioning" owns
+--     that sequence. It is omitted below because a NEW table should never gain
+--     it: it enforces a scope wider than the code's, so it can reject a keyed
+--     enqueue whose key is free in the caller's own (env, app, task_type).
```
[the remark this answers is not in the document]
totalfrank: Agreed and fixed in `acb6c1c5`. Both halves of this checked out against the source.

`repository/models.py` does still declare `uk_env_task_type_active_idempotency_key`, and `test_same_key_under_different_app_is_rejected_while_the_legacy_index_lives` (`tests/community/repository/platform/test_task_queue_repository.py:440`) asserts the behaviour it produces — its docstring calls it "the intermediate one on purpose". So omitting the index would have meant a freshly provisioned OceanBase database accepting an enqueue that every local test run rejects, with both schemas claiming to be this component's.

My reasoning for leaving it out — a new table should not gain a scope wider than the code's — was right about the index and wrong about the decision. The ORM's own comment already sets the order: drop it from `models.py`, the DDL and the deployed tables in one change. This file is the "here" in "drop it there and here in the same change", so it now declares the index with that sequencing recorded inline.

Took the "add compatibility coverage" route as well, since AGENTS.md:76 asks contract changes to carry conformance tests: `tests/community/core/task_queue/test_task_queue_schema_contract.py` fails if the DDL loses a unique index the ORM declares, if either unique index loses `GLOBAL`, or if `AUTO_INCREMENT_MODE` stops being pinned. I verified it fails on the exact omission you flagged before confirming it passes on the fix — nothing in the suite caught this otherwise.

The README's "no DDL is checked in" paragraph is updated too, per the sibling thread on that file.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
totalfrank: Good catch, and the hazard is closed in `acb6c1c5` — though by a different route than the suggestion, because taking the suggestion as written would have left a second bug in place.

The implicit-attach analysis is correct: under `explicit_defaults_for_timestamp=OFF` the server attaches `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` to the first `TIMESTAMP NOT NULL` column lacking an explicit default, and `ON UPDATE` fires on any change to the row, not just to that column. `finish()` (`config_manifest_apply.py:154-156`) updates `status`/`report`/`finished_at` without naming `started_at`, so "when the apply began" would have been overwritten with the finish time on every apply, success or failure.

Rather than pin `started_at` as `TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP`, I reverted it to `DATETIME`. The suggested form suppresses the implicit `ON UPDATE`, but `started_at` is filled from `datetime.now()` in application code, and a `TIMESTAMP` column reinterprets that naive value as session-local — the timezone corruption Codex flagged on `fetched_at` in the sibling thread. Pinning the default would have fixed this finding and kept that one. `DATETIME` closes both: no implicit attach (the rule is `TIMESTAMP`-only) and no session-timezone conversion.

Swept the rest of the diff for the same shape rather than fixing only the column you pointed at: every remaining `TIMESTAMP` column across all seven tables is a `gmt_create`/`gmt_modified` carrying an explicit `DEFAULT CURRENT_TIMESTAMP`, so there is no bare `TIMESTAMP NOT NULL` left anywhere for the implicit rule to fire on. You were also right that `finished_at` was independently safe via its explicit `NULL`; it moved to `DATETIME` regardless, since it has the same `datetime.now()` provenance as `started_at`.

---

src/backend/src/agentclaw/community/core/task_queue/sql/2026_09_06_task_queue.sql
totalfrank: Fixed in `acb6c1c5`. You're right that the contradiction was load-bearing rather than cosmetic: the README stated "No DDL is checked in" as the *reason* a copy in the repo would be wrong, so a reader trusting it would neither look in `sql/` nor know the file is a reference rather than an authority.

The Provisioning section now keeps `repository/models.py` as the source of truth and describes the file as derived from it, including the two things that cannot be read off the ORM and are the reason it exists at all — `GLOBAL` on the unique indexes, which SQLAlchemy cannot render, and `AUTO_INCREMENT_MODE = 'ORDER'`, which `_find_by_key`'s `ORDER BY id DESC` depends on. It also states the precedence explicitly: where the two disagree the ORM wins and the file is what gets fixed, and on a deployment that already has the table `SHOW CREATE TABLE ac_task_queue` outranks both.

It now also records that the file declares the same two unique indexes the model does, legacy one included, so a freshly provisioned database behaves the way the test suite asserts — that was the substance of the Codex thread on this same file.

---

totalfrank:
## Review round 1 — 4 findings, all real, all fixed in `acb6c1c5`

Reviewed `06df08d6` by Codex (2 findings) and Claude Code (2 findings). No disputes this round: every finding verified against the source before I changed anything, and all four were genuine defects in the previous commit.

**The two P1/blocking findings were the same root cause**, which is worth stating because it explains why the fix is a revert rather than four patches. The previous commit converted every `datetime` column to `TIMESTAMP` on the reasoning that `ac_bots` and the newer tables use it. That reasoning only holds for columns the **database** fills:

| Finding | Verified as | Fix |
| --- | --- | --- |
| Codex P1 — `TIMESTAMP` corrupts bound instants | `content/service.py:271-276` normalises `fetched_at` to UTC and strips tzinfo; `started_at`/`finished_at` come from `datetime.now()` | `fetched_at`, `started_at`, `finished_at` → `DATETIME` |
| Claude Code blocking — implicit `ON UPDATE` on `started_at` | Bare `TIMESTAMP NOT NULL` gets `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` under `explicit_defaults_for_timestamp=OFF`; `finish()` would fire it every apply | Same revert — `DATETIME` is exempt from the rule |
| Codex P2 — queue DDL contradicts the ORM | ORM declares `uk_env_task_type_active_idempotency_key`; `test_same_key_under_different_app_is_rejected_while_the_legacy_index_lives` asserts its behaviour | Index restored + new contract test |
| Claude Code non-blocking — README says "no DDL is checked in" | README Provisioning stated it as a load-bearing invariant | Section rewritten |

`gmt_create`/`gmt_modified` stay `TIMESTAMP` on all seven tables: those are database-filled (`DEFAULT CURRENT_TIMESTAMP`, `func.now()`), so the conversion is a no-op round trip.

**Where I diverged from a suggestion.** The blocking comment suggested pinning `started_at` as `TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP`. That suppresses the implicit `ON UPDATE`, but leaves the session-timezone reinterpretation Codex flagged on the same column. `DATETIME` closes both, so I took that instead and explained it on the thread.

**Swept for the same shape rather than fixing only what was pointed at.** Every remaining `TIMESTAMP` column across all seven tables carries an explicit `DEFAULT`, so no bare `TIMESTAMP NOT NULL` is left for the implicit rule to fire on.

**New guard.** `tests/community/core/task_queue/test_task_queue_schema_contract.py` fails if the DDL loses a unique index the ORM declares, if either unique index loses `GLOBAL`, or if `AUTO_INCREMENT_MODE` stops being pinned. Verified it fails on the exact omission Codex found before confirming it passes on the fix — nothing in the suite caught that divergence otherwise.

Local gates: 965 tests pass across `core/task_queue`, `core/bot_config_manifest`, and `repository/platform/test_task_queue_repository.py`.

The one thing still outside what tests can prove: this DDL has not been executed against an OceanBase instance, since none is reachable from this environment. It should be applied to a non-production tenant before anyone trusts it.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
```diff
@@ -68,20 +111,30 @@ CREATE TABLE `ac_bot_config_manifest_apply` (
...
   -- long without anything being malformed.
   `actor`          varchar(1024) NOT NULL COMMENT 'Audit: who started it',
+  -- DATETIME, not TIMESTAMP, for both of these: they are filled by the
+  -- application from datetime.now() (process-local, naive), never by the
+  -- database. TIMESTAMP binds a naive value as session-local, so these would
+  -- be stored correctly only where the process time zone happens to equal the
+  -- database session's -- and a container on UTC against an Asia/Shanghai
+  -- session is exactly where that stops being true. gmt_* below are a
+  -- different case: the database fills those itself.
   `started_at`     datetime      NOT NULL COMMENT 'When the apply began',
```
totalfrank, line 121: **Non-blocking**: the DATETIME/TIMESTAMP split this commit restores (`started_at`/`finished_at` here, `fetched_at` in `2026_08_31_manifest_content.sql:138`) is exactly the property whose violation caused the defect this round fixes — an app-supplied instant silently shifted by the session offset. Right now the only thing keeping a future edit from "tidying" these back to `TIMESTAMP` (to match `gmt_create`/`gmt_modified` two lines down) is prose, in three separate files.

This PR just set a precedent for exactly this situation: `tests/community/core/task_queue/test_task_queue_schema_contract.py`, added a few files over for `ac_task_queue`, parses the DDL as text and asserts a property SQLAlchemy can't express (`GLOBAL`, `AUTO_INCREMENT_MODE`). The same technique would catch a regression here — e.g. asserting `started_at`/`finished_at`/`fetched_at` are declared `datetime` (not `timestamp`) across the three `bot_config_manifest` DDL files — for the one column-type choice in this module that has already been gotten wrong once and caught only by manual review, not CI.

Not blocking this PR — the comments are thorough and explicit — but worth closing the same way the queue table's contract just was.

---

totalfrank (commented):
Reviewed `06df08d6` → `acb6c1c5`.

**Findings checklist:**
- Blocking — `started_at` TIMESTAMP implicit-default corruption (`2026_08_31_bot_config_manifest_apply.sql`): **fixed**, reverted to `DATETIME`.
- Non-blocking — `ac_task_queue` DDL contradicted README's "no DDL is checked in" (`sql/2026_09_06_task_queue.sql`): **fixed**, README's Provisioning section rewritten.
- Codex P1 (`fetched_at` TIMESTAMP corruption) and Codex P2 (queue DDL vs ORM legacy index): both addressed by this round per the commit message — not mine to resolve, left to Codex/author.
- Non-blocking (new) — no automated test guards the `DATETIME` choice on `started_at`/`finished_at`/`fetched_at`; the same regex-over-DDL technique this round just added for `ac_task_queue` (`test_task_queue_schema_contract.py`) would catch a future "consistency" edit flipping them back to `TIMESTAMP`.

Ran the new `test_task_queue_schema_contract.py` (3 passed) plus the full `bot_config_manifest`/`task_queue`/`repository` suites (1897 passed, 5 skipped) against this head commit — no regressions.

**Open: 1** (the new non-blocking test-coverage note above).

---

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
totalfrank: Fixed — `acb6c1c5` reverts `started_at`/`finished_at` to `DATETIME` entirely rather than pinning an explicit default, which removes the column from `TIMESTAMP`'s implicit-default-attach rule altogether (that rule only applies to `TIMESTAMP` columns). Confirmed: both columns are `datetime` in the current file, and both `gmt_create`/`gmt_modified` (the table's only remaining `TIMESTAMP` columns) already carry explicit `DEFAULT`s, so no bare-first-TIMESTAMP column remains in this table.

---

src/backend/src/agentclaw/community/core/task_queue/sql/2026_09_06_task_queue.sql
totalfrank: Fixed — `acb6c1c5` rewrites the README's Provisioning section to describe `sql/2026_09_06_task_queue.sql` as a provisioning reference derived from the ORM (not a second authority), matching the file's own header and stating that the ORM wins on disagreement and `SHOW CREATE TABLE` outranks both. No contradiction remains.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
totalfrank: Agreed, and done in `ddeba211` — `tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py`.

You're right about the asymmetry: I added a contract test for the queue table in the same round, then left this module's rule resting on prose in three files, when it is the rule that had actually been broken. Three assertions rather than the one suggested, because the round's two defects were separable:

1. `started_at` / `finished_at` / `fetched_at` are declared `datetime` — the Codex P1 regression.
2. `gmt_create` / `gmt_modified` are declared `timestamp` in every file. Asserting the split in both directions makes it a rule about *who fills the column* rather than a list of three exceptions, which is the framing that would have prevented the original mistake.
3. No bare `TIMESTAMP NOT NULL` without a `DEFAULT`, in any file — your blocking finding, encoded so it cannot come back through a column neither of us is looking at.

Verified each by injecting the regression it names and watching it fail, which is how assertion 3 got fixed before landing: the first version exempted any declaration containing "null", so `NOT NULL` matched and it passed the exact declaration it exists to reject. It now requires an explicit nullable via `(?<!not )\bnull\b`, and with that both 1 and 3 fire on a `started_at` flipped back to `timestamp`.

Test-only commit; no DDL or source changed. Full suites re-run on the new head: 1900 passed, 5 skipped.

---

totalfrank:
## Review round 2 — 1 new finding (non-blocking), fixed in `ddeba211`

Reviewed `acb6c1c5`. Round 1's four findings were all confirmed fixed by the reviewer; one new non-blocking finding, now closed.

**The finding:** the `DATETIME` choice restored in round 1 was held in place only by prose in three files — and prose is precisely what failed the first time. A later edit aligning `started_at`/`finished_at`/`fetched_at` with the `gmt_*` columns two lines below them would have silently reintroduced the shifted-instant defect with nothing in CI objecting. Fair, and the sharper version of it is that I had added a contract test for the queue table in the same round while leaving the rule that actually broke unguarded.

`tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py` now asserts three things, one per rule already broken once:

| Assertion | Guards |
| --- | --- |
| `started_at`/`finished_at`/`fetched_at` declared `datetime` | Codex P1 — app-supplied instant shifted by the session offset |
| `gmt_create`/`gmt_modified` declared `timestamp`, every file | The split as a rule about *who fills the column*, not a list of exceptions |
| No bare `TIMESTAMP NOT NULL` without `DEFAULT`, any file | The blocking finding — implicit `ON UPDATE` reaching a column nobody is looking at |

**Each was verified by injecting the regression it names and watching it fail.** That is not ceremony: it is how the third assertion got fixed before landing. Its first version exempted any declaration containing `null`, which also matched `NOT NULL` — so it passed the exact declaration it exists to reject. It now requires an explicit nullable via `(?<!not )\bnull\b`, and both assertions 1 and 3 fire on a `started_at` flipped back to `timestamp`.

Test-only commit; no DDL or source changed. 1900 passed, 5 skipped across `bot_config_manifest`, `task_queue` and `repository`.

### Status

Both automated reviewers' findings are addressed and every thread is resolved. Two things a human still owns:

- **`FreddieSun`'s approval was dismissed** when `acb6c1c5` landed (the repo dismisses stale reviews on push), and `ddeba211` is another push on top. The PR currently carries no human approval and needs a fresh one.
- **The DDL has still never been executed against OceanBase** — no instance is reachable from this environment, so every claim about `GLOBAL`, `AUTO_INCREMENT_MODE` and the implicit-default rule is reasoned from the deployed `ac_bots` schema and the documented behaviour, not observed. Apply it to a non-production tenant before trusting it.

---

src/backend/tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py
```diff
@@ -0,0 +1,116 @@
...
+                f"{filename}: `{column}` is declared {declared_type}, not datetime. "
+                "It is filled by the application, so TIMESTAMP would reinterpret "
+                "the bound value as session-local and store it shifted."
+            )
+
+
+def test_database_filled_times_are_timestamp() -> None:
+    """The other half of the rule, so the split is asserted in both directions."""
+    for path in _sql_files():
+        declarations = _declarations(path)
```
totalfrank: **Non-blocking**: `_declarations()` (lines 57-60) builds one dict per *file*, not per table, keyed by column name. `2026_08_31_bot_config_manifest_apply.sql` declares **two** tables — `ac_bot_config_manifest_apply` and `ac_bot_config_manifest_apply_lock` — and both have `gmt_create`/`gmt_modified` columns. Since the dict comprehension keeps only the *last* match for a repeated name, this call and the one in `test_no_bare_timestamp_column_can_attach_an_implicit_on_update` only ever see the lock table's declarations for those two columns; the apply table's own `gmt_create`/`gmt_modified` go unchecked by both tests.

That's not a hypothetical file: the apply table is exactly where the round-1 blocking finding lived (`started_at`'s implicit `ON UPDATE CURRENT_TIMESTAMP` hazard). Concretely — revert `ac_bot_config_manifest_apply`'s `gmt_modified` to a bare `timestamp NOT NULL` with no `DEFAULT` (the same shape as that original bug, just on a different column in the same table) while leaving `ac_bot_config_manifest_apply_lock`'s `gmt_modified` untouched, and both `test_database_filled_times_are_timestamp` and the bare-TIMESTAMP guard still pass — the merged dict reports only the lock table's still-correct declaration, and the regression this test file exists to catch ships silently.

Suggest scoping `_declarations` per `CREATE TABLE` block (split `body` on `CREATE TABLE` boundaries and return a dict keyed by table name, or a list of per-table dicts) instead of merging same-named columns across tables in one file, so both tables' `gmt_create`/`gmt_modified` are actually checked.

---

totalfrank (commented):
Reviewed `acb6c1c5` → `ddeba211` (this round: `ddeba211`, test-only — adds `test_manifest_ddl_contract.py`).

- Blocking (Codex) — `fetched_at`/`started_at`/`finished_at` → `TIMESTAMP` corrupts app-supplied instants: **fixed** (`acb6c1c5`).
- Non-blocking (Codex) — `ac_task_queue` DDL contradicted README's "no DDL checked in": **fixed** (`acb6c1c5`).
- Blocking (reviewer, round 1) — `started_at` bare `TIMESTAMP NOT NULL` risks implicit `ON UPDATE CURRENT_TIMESTAMP` corruption: **fixed** (`acb6c1c5`, reverted to `DATETIME`).
- Non-blocking (reviewer, round 1) — README Provisioning section needed to describe the queue DDL as a reference, not an authority: **fixed** (`acb6c1c5`).
- Non-blocking (reviewer, round 2) — no test guarded the `DATETIME` choice on `started_at`/`finished_at`/`fetched_at`: **fixed** (`ddeba211`, adds `test_manifest_ddl_contract.py`); verified the three guarded columns don't collide across tables, so the guard holds for the rule it was filed against.
- Non-blocking (reviewer, round 3, **new**) — the new test's `_declarations()` merges columns across multiple `CREATE TABLE` statements in one file, so in the one file with two tables (`2026_08_31_bot_config_manifest_apply.sql`) a future regression in the apply table's `gmt_create`/`gmt_modified` would be masked by the lock table's correct declaration. Filed inline. **Open.**

**Open: 1** (the new non-blocking finding above).

---

src/backend/src/agentclaw/community/core/bot_config_manifest/sql/2026_08_31_bot_config_manifest_apply.sql
totalfrank: Verified fixed in `ddeba211` (`test_manifest_ddl_contract.py`): `started_at`/`finished_at`/`fetched_at` each appear in exactly one `CREATE TABLE` in their file, so `test_application_supplied_times_are_datetime` genuinely guards all three against a future flip back to `TIMESTAMP`. (Filed a separate, narrower gap in the new test's multi-table handling in this round's review — it doesn't affect this rule's columns, only `gmt_create`/`gmt_modified` in the one file with two tables.)

---

totalfrank: Confirmed and fixed in `96d255e6`. This was a real hole, and your failure scenario reproduces exactly as described.

I verified both halves against the old test before changing it:

- Reverting `ac_bot_config_manifest_apply.gmt_modified` to a bare `timestamp NOT NULL` — the round-1 bug's exact shape, one column over — **passed**.
- Flipping that table's `gmt_create` to `datetime` — **also passed**.

Both because the dict comprehension kept the last match, so `ac_bot_config_manifest_apply_lock`'s still-correct declarations masked them. The apply table is where the blocking finding lived, so the one table whose timestamp declarations most needed guarding was the one silently skipped.

`_tables()` now splits the comment-stripped body on `CREATE TABLE` boundaries and returns per-table column maps. `_declarations()` survives as a flat view for the bare-`TIMESTAMP` sweep, but with `table.column` keys so shadowing is impossible. Both injected regressions now fail — each in the assertion that should catch it, and only that one.

I also added `test_every_table_in_every_file_is_parsed`, pinning the six tables these files declare. Your finding's real lesson is that the assertions are only as complete as the parser: one that quietly returned a single table for a two-table file would make them go *quiet* rather than fail, which is how this survived a review round. That test fails loudly instead.

880 passed across `bot_config_manifest` and `task_queue`. Test-only commit; no DDL or source changed.

---

src/backend/tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py
totalfrank: Confirmed and fixed in `96d255e6`. Your failure scenario reproduces exactly as written, and I ran both variants against the old test before changing it:

- `ac_bot_config_manifest_apply.gmt_modified` → bare `timestamp NOT NULL` (the round-1 bug's shape, one column over): **passed**, as you predicted.
- That same table's `gmt_create` → `datetime`: **also passed**.

Both because the dict kept the last match, so `ac_bot_config_manifest_apply_lock`'s correct declarations masked them. The apply table is where the round-1 blocking finding lived, so the table whose timestamp declarations most needed guarding was the one silently skipped.

Took the per-table route you suggested. `_tables()` splits the comment-stripped body on `CREATE TABLE` boundaries and returns per-table column maps; `_declarations()` remains as a flat view for the bare-`TIMESTAMP` sweep but with `table.column` keys, so shadowing is impossible there too. Both injected regressions now fail — each in the assertion that should catch it, and only that one.

One addition beyond the suggestion: `test_every_table_in_every_file_is_parsed` pins the six tables these files declare. The lesson I take from this finding is that the assertions are only ever as complete as the parser feeding them, and a parser that quietly returned one table for a two-table file would make them go *quiet* rather than fail — which is exactly how this survived a review round. That failure mode now has its own test.

880 passed across `bot_config_manifest` and `task_queue`. Test-only commit; no DDL or source changed.

(My first reply to this landed on the round-2 thread by mistake — same content, wrong place.)

---

totalfrank:
## Review round 3 — 1 new finding (non-blocking), fixed in `96d255e6`

Reviewed `ddeba211`. The finding was a real hole in the test added last round, and it sat precisely where the coverage was supposed to be.

**The bug:** `_declarations()` built one dict per *file*, keyed by bare column name. `2026_08_31_bot_config_manifest_apply.sql` declares two tables — the apply record and its lock — and both carry `gmt_create`/`gmt_modified`. A dict comprehension keeps the last match, so every assertion only ever saw the *lock* table's declarations; the apply table's audit columns were checked by nothing. That is the same table the round-1 blocking finding lived in.

**Verified before fixing, both variants of the reviewer's scenario against the old test:**

| Injected regression | Old test |
| --- | --- |
| apply table's `gmt_modified` → bare `timestamp NOT NULL` (round-1 bug's shape, one column over) | **passed** |
| apply table's `gmt_create` → `datetime` | **passed** |

Both now fail, each in the assertion that should catch it and only that one.

`_tables()` splits the comment-stripped body on `CREATE TABLE` boundaries and returns per-table column maps. `_declarations()` survives for the bare-`TIMESTAMP` sweep with `table.column` keys, so shadowing is impossible there too.

**One addition beyond the suggestion.** `test_every_table_in_every_file_is_parsed` pins the six tables these files declare. The lesson of this finding is that the assertions are only as complete as the parser feeding them — a parser that quietly returned one table for a two-table file would make them go *quiet* rather than fail, which is how this survived a round. That failure mode now has its own test.

880 passed across `bot_config_manifest` and `task_queue`. Test-only commit; no DDL or source changed.

### Cumulative status

Six findings across three rounds, from two automated reviewers. **All verified against the source, all real, none disputed.** Three were defects in the DDL itself, three were gaps in the tests guarding it — including two in tests I had just written, both found by asking "what would make this test pass when it shouldn't".

Every thread is resolved. What still needs a human:

- **No approval on the current head.** `FreddieSun` approved `06df08d6`; that was dismissed by the round-1 push, and there have been three pushes since. `xianmuyq` is now a requested reviewer.
- **The DDL has never been executed against OceanBase.** No instance is reachable from this environment. Three rounds of review agreeing about `GLOBAL`, `AUTO_INCREMENT_MODE` and the implicit-default rule is not the same as observing them — every one of those claims is reasoned from the deployed `ac_bots` schema and documented engine behaviour. **Apply this to a non-production tenant before trusting it.**

---

totalfrank (commented):
Reviewed `ddeba211` → `96d255e6` (this round: `96d255e6`, test-only — reworks `_tables()`/`_declarations()` in `test_manifest_ddl_contract.py` to scope per `CREATE TABLE` block instead of per file, adds `test_every_table_in_every_file_is_parsed`).

This was the last open item, already addressed by the previous round's reply on the thread but not yet marked resolved — closing that out here. Verified it directly rather than taking the reply's word for it: ran the four contract tests against this head (all pass), then injected both regressions the finding named — `ac_bot_config_manifest.gmt_modified` reverted to bare `TIMESTAMP NOT NULL` (round-1 bug shape, one column over) and that table's `gmt_create` flipped to `datetime` — and confirmed each now fails in the correct assertion instead of being masked by the lock table's declarations. No new issues found in the rework.

- Blocking (Codex) — `fetched_at`/`started_at`/`finished_at` → `TIMESTAMP` corrupts app-supplied instants: **fixed** (`acb6c1c5`).
- Non-blocking (Codex) — `ac_task_queue` DDL contradicted README's "no DDL checked in": **fixed** (`acb6c1c5`).
- Blocking (reviewer, round 1) — `started_at` bare `TIMESTAMP NOT NULL` risks implicit `ON UPDATE CURRENT_TIMESTAMP` corruption: **fixed** (`acb6c1c5`, reverted to `DATETIME`).
- Non-blocking (reviewer, round 1) — README Provisioning section needed to describe the queue DDL as a reference, not an authority: **fixed** (`acb6c1c5`).
- Non-blocking (reviewer, round 2) — no test guarded the `DATETIME` choice on `started_at`/`finished_at`/`fetched_at`: **fixed** (`ddeba211`, adds `test_manifest_ddl_contract.py`).
- Non-blocking (reviewer, round 3) — `_declarations()` merged columns across multiple `CREATE TABLE` statements in one file, masking the apply table's own `gmt_create`/`gmt_modified` behind the lock table's: **fixed and verified** (`96d255e6`), thread resolved this round.

**Open: 0.** Nothing new found this round. PR is clean as of `96d255e6`.

---

totalfrank:
## Review round 4 — no issues flagged

Reviewed `96d255e6` (test-only rework of `_tables()`/`_declarations()`). **No new findings.** The reviewer re-ran both injected regressions itself rather than taking my reply's word for it, and confirmed each now fails in the correct assertion. Open findings: 0.

**Four rounds, six findings, none disputed:**

| Round | Finding | Fixed in |
| --- | --- | --- |
| 1 | Codex P1 — `TIMESTAMP` corrupts app-supplied instants | `acb6c1c5` |
| 1 | Codex P2 — queue DDL contradicted the ORM's legacy index | `acb6c1c5` |
| 1 | Blocking — bare `TIMESTAMP NOT NULL` implicit `ON UPDATE` on `started_at` | `acb6c1c5` |
| 1 | README "no DDL is checked in" contradiction | `acb6c1c5` |
| 2 | No test guarded the `DATETIME` choice | `ddeba211` |
| 3 | That test merged columns across two tables, masking the apply table's | `96d255e6` |

Three were defects in the DDL; three were gaps in the tests guarding it, two of those in tests written during this review. Every thread is resolved.

### One reviewer has not seen the current code

**Codex last reviewed `06df08d6`** — three commits ago. It triggers on open-for-review, draft→ready, and explicit mention, but *not* on push, so it never saw the changes made in response to its own P1 finding, nor either contract test.

That is the one gap I can still close before a human merges, and it is worth closing: Codex found the most severe defect in this PR, and the columns it flagged are exactly the ones that changed afterwards.

@codex review

### Still needs a human

- **CI is not fully green yet on `96d255e6`** — 8 checks passed, `Backend unit tests` and `Singlebox coverage` still running as of this comment. I am watching; if either fails I will fix it rather than leave it.
- **No approval on the current head.** `FreddieSun` approved `06df08d6`; the round-1 push dismissed it and three pushes have landed since. `xianmuyq` is a requested reviewer.
- **This DDL has never been executed against OceanBase.** No instance is reachable from this environment. Four rounds of reviewers agreeing about `GLOBAL`, `AUTO_INCREMENT_MODE` and the implicit-default rule is not the same as observing them — every such claim here is reasoned from the deployed `ac_bots` schema and documented engine behaviour. **Apply it to a non-production tenant before trusting it.** That is the one item no amount of further review can retire.

---

FreddieSun (approved):

PR #1938, opened 2026-09-06
refactor(backend): give the apply engine's outbound ports a home and two named implementations

totalfrank:
## Problem

Several pairs in the manifest apply engine look like duplicates — `ActivationPort` / `DirectActivationServiceProtocol`, `SkillPackageUploadPort` / `LocalSkillUploadServiceProtocol`. None of them are. But in every case the real relationship was stated nowhere in source.

`MaterialiserPorts` types four of its fields by a narrow port. What was actually bound to those fields could only be learned by walking the DI graph — field type → `_arca_ports()` → the provider → `binder.bind(...)`. `src/backend` runs no static type checker (mypy/pyright are configured for `src/gateway`, `src/baas`, `src/proxy` only), so a structurally-satisfied Protocol here is verified by nothing: not at import, not at construction, not in CI.

Four concrete symptoms:

- **Activation was asymmetric.** `RecordOnlyActivation` stated `project=False`; ARCA bound `DirectActivationService` raw and projected only because `project` *defaults* to `True`. A grep for `project=` across the whole codebase returned exactly one caller — one family stated its delivery semantics, the other left them to a parameter default.
- **Upload and resource had only one named side.** `StoreSkillPackagePort` / `StoreResourcePort` existed for teclaw; the ARCA side had no name, and the raw Service API it bound carried methods with no meaning during an apply.
- **Identity had a hand-rolled drift guard.** Its DI provider carried `if not isinstance(service, ManifestIdentityPort): raise TypeError(...)`, with a comment that "the port has no implementation relationship to the service, so nothing else would notice a renamed method until mid-apply" — a manual substitute for a declared base, and a weaker one, since `isinstance` against a `runtime_checkable` Protocol compares method names only.
- **Four constructor parameters were annotated `Any`** — `McpMaterialiser.mcp_auth_service` and all three of `SkillsMaterialiser`'s services, including the two ports this PR is about.

Naming compounded it: `StoreSkillPackagePort`, `StoreIdentityPort` and `StoreResourcePort` are concrete adapters — not Protocols, not even ABCs — whose names say "Port".

## Solution

**A `core/ports` package.** A port names a relationship between two modules and belongs to neither. Kept inside the consumer it is unreachable by the provider (`core.skill_center` must not import `core.bot_config_manifest`); kept inside the provider it drifts toward the service's full surface, which is what it exists not to be. Its README defines an outbound port against the two things it gets confused with — a Service API Protocol (provider-owned, full surface) and a repository protocol.

**Two implementations per port, named for where the write lands** — the axis every delivery family splits on: whether the write reaches the bot's device, or stays in platform state that the composed artifact delivers.

| Port | device (ARCA) | platform (teclaw) | narrows |
|---|---|---|---|
| `ActivationPort` | `DeviceActivation` | `PlatformActivation` | hides `project` |
| `SkillPackageUploadPort` | `DeviceSkillPackageUpload` | `PlatformSkillPackageUpload` | 3 → 2 methods |
| `IdentityFilePort` | `DeviceIdentity` | `PlatformIdentity` | 15 → 3 |
| `ResourceFilePort` | `DeviceResource` | `PlatformResource` | 7 → 3 |

The pairs are not the same shape underneath, and the naming is chosen for that reason. The activation two are wrappers over one `DirectActivationService` differing only in the `project` they forward. The other three share no body at all — one writes files onto a container, the other objects into the managed-files store. Same axis, different depth; the axis is what lets a reader predict which one a family gets.

The three device wrappers are new. Each wrapped service is a domain service with its own Service API and HTTP routes, so none can itself be the device implementation. They encode no parameter, but they are not pass-throughs — each hides surface a materialiser has no business reaching, which is the same narrowing `ActivationPort` performs on `project`, applied to methods.

`DeviceIdentity` and `DeviceResource` import their services under `TYPE_CHECKING` only, and that is load-bearing rather than stylistic: both service modules reach the device dispatcher graph at module scope, and importing one eagerly from a module the apply service imports walks into a circular import. Verified — importing either wrapper pulls in no device or DI module.

**Port members are `@abstractmethod`.** This is what makes a declaration load-bearing rather than decorative: with plain `...` stubs a dropped method is filled in by the inherited stub, so the name still resolves and the call silently returns `None` — the exact failure `test_protocol_base_ordering.py` was written after shipping (`CollaboratorService`, 500s instead of 404s).

The three renamed platform adapters declare their port with `_StorePort` **first**: a Protocol base ahead of it makes `__init__` resolve through `typing`'s `_no_init_or_replace_init`, and `test_protocol_base_ordering.py:_members()` filters dunders, so it would not catch that.

**`MaterialiserPorts` fields deliberately left alone.** `script_service`, `mcp_auth_service` and `capability_reader` bind the same object on both bundles and always will — `script` is unsupported on teclaw, permission checks and capability state are platform-level regardless of family — so a port there would abstract over a variation that does not exist, and all three are already Protocols. `cli_tool_service` already splits by family one level down (`CliToolDeliveryPort` with `ArcaCliToolPort` / `TeclawCliToolPort`). `package_validator` and `entry_fetcher` are not outbound boundaries at all.

**Alternatives rejected.** Deleting a pair down to one type — impossible without a lie: `PlatformActivation` would need a `project` parameter whose only correct value is `False`, and `PlatformSkillPackageUpload` has no use for `upload_local_skill_files` (it fails even a runtime name check against the Service API). Making the Service API Protocols inherit the ports — shorter, and LSP-correct for the upload pair, but a provider's Protocol inheriting a consumer-owned contract contradicts the rule `core/ports/README.md` states, and for activation it produced an asymmetric pair where one side declared the port and the other widened it.

## Validation

- Full `tests/community/core` + `tests/community/api` — **10855 passed, 56 skipped, 2 failed**. Both failures are `test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]` and reproduce identically on this branch's parent commit — pre-existing, unrelated (NFS snapshot handling in `service_bot`).
- Full `tests/community` on an earlier head: **17665 passed, 63 skipped**, same 2 pre-existing failures.
- `tests/community/architecture` + `tests/community/core/bot_config_manifest` — **1125 passed**.
- New `test_narrow_ports_are_declared.py` covers all eight implementations. Mutation-tested: removing a base class fails it, and dropping a delegated method fails the construction check — both would have passed silently before. It also pins that no port leaks its narrowed member.
- Three arch gates needed the new package registered: `BOUNDARY_SIGNIFICANT_MODULES`, `_STRUCTURAL_NON_BUSINESS` in `flow_coverage.py` (alongside `repository` and `bot_config_surface` — a package of abstract stubs has no behaviour for an e2e flow to drive), and the module README's `internal_dependencies`, which counts `TYPE_CHECKING` imports.
- flake8 clean on all changed files (`F401,F811,F821,F841,E999,E702,E703,E711,E712,E722`).

## Compatibility and risk

No behaviour change. Every write reaches the same destination with the same `project` value it had before; ARCA's `project=True` was previously implicit in the parameter default and is now stated by `DeviceActivation`. No signature, binding key, or DI wiring semantics change.

The one new failure mode is intended: the eight implementations are unconstructable if they omit a port member, surfacing at DI startup rather than at the call site. All eight implement their full surface today.

Rollback is the revert; nothing persisted or migrated.

## Related issues

Follow-ups deliberately out of scope: moving the four device adapters out of `bot_config_manifest/apply/` into a `core/adapters` sibling (raised in review, left for a separate change); the `cli_tools` ports, which already have the structure under family-named types; the same redundant quoted annotations in `entry_fetch.py`, `materialisers/identity.py`, `materialisers/skills.py` and `create_job.py`, none of which are in this diff; and the `materialisers` → `materializers` spelling rename.

https://claude.ai/code/session_01ABVRHuonyNh7SX3Hq8pa2z


totalfrank:
**Two red checks — not this PR's, and not re-running.**

Failing: **Community Boundary Check** and **Build & Test**, both jobs of workflow run [34018396789](https://github.com/inclusionAI/Avernet/actions/runs/34018396789) on head `541794d6`.

**Cause.** I opened this PR against `main` by mistake; the default branch here is `dev`, and `main` shares no merge base with this branch. That mis-targeting pulled in `.github/workflows/taskguard-ci.yml`, which is scoped to pull requests against `main` / `fudp_open` and to paths under `src/evolverun/taskguard/**`. The PR has since been retargeted to `dev`, where that workflow does not apply.

**Why it is not this diff.** Every finding is a TypeScript import under `src/evolverun/taskguard/` — `yuque-adapter`, `agentmind-adapter`, `approval-card-web-poller`, `dev-workflow-callback`, `baas-call`, in `src/index.ts`, `src/dispatch.ts` and `src/platform/default-executor.ts`. This diff is nine Python and Markdown files under `src/backend/`, and touches no `.ts` file at all. For comparison, #1935 (a normal `dev`-targeted PR) runs 10 checks and neither of these two is among them.

**No re-run.** Re-running would confirm nothing: the workflow is trigger-scoped away from `dev`-targeted PRs, so the correct outcome is that it does not run, not that it passes. The trigger config plus the #1935 comparison establishes that more directly than a repeat would. Both runs are stale against the retargeted base and will be superseded by the next push.

**Everything else.** Green so far: Engine, Sandbox-proxy, Gateway unit tests, Open Core frontend checks, Validate PR title. Backend unit tests, BCS unit tests, BaaS unit tests, BCS architecture checks, Singlebox coverage and BCS e2e are still running.

Locally the full `tests/community` suite is **17665 passed, 63 skipped, 2 failed** — the two being `test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]`, which reproduce identically on this branch's parent commit and are unrelated to this change.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/identity_files.py
```diff
@@ -0,0 +1,94 @@
...
+from agentclaw.community.core.ports.identity_file_port import IdentityFilePort
+
+if TYPE_CHECKING:  # pragma: no cover — see the module docstring
+    from agentclaw.community.core.services.identity import IdentityService
+
+
+class DeviceIdentity(IdentityFilePort):
+    """The ARCA identity road: files onto the bot's live container."""
+
+    def __init__(self, inner: "IdentityService") -> None:
```
totalfrank-personal: Do you need forward reference for IdentityService? If not, let's get rid of it.

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/identity_files.py
```diff
@@ -0,0 +1,94 @@
...
+
+from typing import TYPE_CHECKING, Any
+
+from agentclaw.community.core.ports.identity_file_port import IdentityFilePort
+
+if TYPE_CHECKING:  # pragma: no cover — see the module docstring
+    from agentclaw.community.core.services.identity import IdentityService
+
+
+class DeviceIdentity(IdentityFilePort):
```
totalfrank-personal, line 39: Does it make sense for this file to leave under apply directory? I mean, apply seems to be something scoped to the manifest apply area but DeviceIdentity seems to be a bigger thing. Does it make sense for this to live somewhere else under core?

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/identity_files.py
totalfrank: Right — the quotes are redundant. The module has `from __future__ import annotations` (line 29), so every annotation is already stored unevaluated; the string literal adds nothing. Dropped.

The `TYPE_CHECKING` import above it does have to stay, for a different reason: `core/services/identity.py` reaches the device dispatcher graph at module scope, and importing it eagerly from here — this module is imported by the apply service, which the DI graph builds — walks into a circular import. Verified after the change: importing `DeviceIdentity` still pulls in no device or DI module, and the annotation still resolves to `IdentityService`.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/identity_files.py
totalfrank: Agreed that `apply/` is the wrong scope label. `DeviceIdentity` narrows a general core service (`IdentityService`) to a general port (`core/ports/IdentityFilePort`); nothing about it is manifest-apply-specific. Same for `DeviceActivation` and `DeviceSkillPackageUpload` — and you already made the point yourself that the other pipelines calling `DirectActivationServiceProtocol` directly could be wrapped in `DeviceActivation` for symmetry. They cannot be, sitting under `bot_config_manifest/apply/`.

The mirror of the ports argument applies: an adapter belongs to neither the service it wraps nor the one consumer that happens to use it today.

There is an asymmetry worth naming before we move anything, though — the two sides of each pair are **not** equally general:

| | wraps | scope |
|---|---|---|
| `DeviceActivation`, `DeviceIdentity`, `DeviceSkillPackageUpload` | general core services | general |
| `PlatformActivation` | `DirectActivationService` | general |
| `PlatformIdentity`, `PlatformSkillPackageUpload` | `ManagedFilesStore` | genuinely `bot_config_manifest` |

So a `core/adapters/` package (sibling to `core/ports/`) takes four of the six cleanly, and the two store-backed ones stay in `managed_files/` where their store lives. Moving those two as well would make `core.adapters` import `core.bot_config_manifest`, which is the wrong direction.

That splits two pairs across packages. I think that is honest rather than a smell — it reflects that one implementation is general and the other is tied to a manifest-owned store — but it is a real cost and worth your call.

Proposal: do the move together with `resource_port` (the last field of this shape, `ResourceFileService` 7 methods → port 3), so all four pairs land in one commit instead of churning the tree twice. Happy to do it in this PR or a follow-up — say which.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/registry.py
```diff
@@ -196,12 +196,12 @@ def build_materialisers(
...
     mcp_auth_service: "MCPAuthServiceProtocol",
-    identity_service: "ManifestIdentityPort",
+    identity_service: "IdentityFilePort",
     upload_service: "SkillPackageUploadPort",
     capability_reader: "BotCapabilityStateReaderProtocol",
     package_validator: "SkillPackageValidator",
     entry_fetcher: "EntryFetcher",
-    resource_service: "ManifestResourcePort",
+    resource_service: "ResourceFilePort",
     cli_tool_service: "CliToolService",
```
totalfrank-personal: Do you really need the forward reference for them? If not, let's get rid of them.

totalfrank-personal (commented):
One single comment, please take a look.

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/registry.py
totalfrank: No — redundant, all ten. `registry.py` has `from __future__ import annotations` (line 10), so every annotation is already stored unevaluated and the string literals bought nothing over the bare names. Dropped in `b740c491`.

The `TYPE_CHECKING` block above them stays, and it is the part actually doing the work: the registry is imported by anything that just wants the ordering table, so naming these types at runtime would pull the bot-configuration graph in behind them. Verified after the change — importing `registry` pulls in no device, DI or `cli_tools.service` module, and the annotations still read back as their type names.

Scope note: the same redundancy exists elsewhere in the package — `entry_fetch.py` (11), `materialisers/identity.py` (5), `materialisers/skills.py` (6), `create_job.py` (2). All predate this PR and none are in its diff, so I left them rather than widening it. Happy to clean them up in a follow-up if you want.

---

totalfrank (commented):
Reviewed the full diff against `dev` (merge-base `ef60c3b`), commits `541794d`..`b740c491` (7 commits).

**Files covered:** all 35 changed files — the new `core/ports/` package (`activation_port.py`, `identity_file_port.py`, `resource_file_port.py`, `skill_package_upload_port.py`, `README.md`), the four new/renamed device and platform implementations (`apply/activation_delegates.py`, `apply/identity_files.py`, `apply/resource_files.py`, `apply/skill_package_upload.py`, `managed_files/ports.py`), the wiring updates (`config_manifest_apply_service.py`, `di/modules/bot_management_module.py`, `di/modules/manifest_fetch_module.py`, `apply/registry.py`, `apply/delivery.py`, `apply/materialisers/{mcp,skills}.py`), the two arch-gate registrations (`test_module_boundaries.py`, `flow_coverage.py`), the new `test_narrow_ports_are_declared.py`, and the updated test fixtures/imports across `tests/community/core/bot_config_manifest/**`.

**What I checked:**
- **Correctness** — traced every renamed/moved symbol (`ActivationPort`, `IdentityFilePort`/`ManifestIdentityPort`, `ResourceFilePort`/`ManifestResourcePort`, `SkillPackageUploadPort`, `Device*`/`Platform*`) end-to-end from the port declarations through both delivery-family wiring paths (`config_manifest_apply_service._device_ports` and `manifest_fetch_module`'s teclaw factory) to confirm the forwarding signatures, keyword names, and defaults match exactly what each device/platform implementation delegates to. No mismatches found. Grepped the whole tree for leftover references to the deleted names (`ManifestIdentityPort`, `ManifestResourcePort`, `StoreIdentityPort`, `StoreResourcePort`, `StoreSkillPackagePort`, `RecordOnlyActivation`, `apply.identity_port`/`resource_port`/`upload_port`/`activation_port`) — none remain outside historical `specs/*.md` docs, which predate and are outside this diff.
- **Contract breakage** — `DirectActivationServiceProtocol` gained `@abstractmethod` on all six members; confirmed its one implementer, `DirectActivationService`, defines all six, so this doesn't break construction. The `_StorePort`-before-`Protocol` base ordering in `managed_files/ports.py` (called out in the PR description as load-bearing for `__init__` resolution) is correct in all three renamed classes.
- **Error/failure handling, concurrency** — no changes to control flow, only type-level narrowing and delegation; nothing new here.
- **Architecture boundaries** — the new `core/ports` package is registered in `BOUNDARY_SIGNIFICANT_MODULES` and `_STRUCTURAL_NON_BUSINESS`, and ships its own `README.md` with a `Context Boundary` block and accurate `internal_dependencies` (including the `TYPE_CHECKING`-only service imports in `identity_files.py`/`resource_files.py`).
- **Test coverage** — `test_narrow_ports_are_declared.py` covers all eight implementations (base declared, members abstract, constructable, narrowing preserved); existing materialiser/delivery tests were updated in place for the renames rather than duplicated.

No findings.


---

src/backend/tests/community/architecture/test_narrow_ports_are_declared.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+@pytest.mark.parametrize(
+    ("port", "impl"), _PORT_IMPLEMENTERS, ids=lambda o: getattr(o, "__name__", str(o))
+)
+def test_implementer_is_constructable(port, impl) -> None:
+    """No leftover abstract members — the gate above is armed, not tripped."""
+    assert not impl.__abstractmethods__, (
+        f"{impl.__name__} does not implement {sorted(impl.__abstractmethods__)} "
+        f"from {port.__name__}; it cannot be instantiated."
```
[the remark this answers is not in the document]
totalfrank: Fixed in 961038d — `test_device_port_forwarding.py` drives every method of all four device delegates through a recording double and asserts arguments (including `project`) arrive unchanged. Verified this covers exactly the two failure cases named (`DeviceActivation` swapping `project`, `DeviceIdentity` dropping `stage`) plus argument-swap generally.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/README.md
```diff
@@ -604,7 +604,7 @@ provides:
   - FetchedEntry
   - EntryFetchError
   - scope_of
-  - ManifestIdentityPort
+  - IdentityFilePort
```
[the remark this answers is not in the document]
totalfrank: Fixed in 961038d — `IdentityFilePort` dropped from `bot_config_manifest`'s `provides` list; `core/ports/README.md` is now the only module claiming it.

---

src/backend/tests/community/core/bot_config_manifest/apply/test_device_port_forwarding.py
```diff
@@ -0,0 +1,210 @@
...
+    """A delegate re-exposes the port and nothing else.
+
+    The point of each wrapper is to narrow — 15 methods to 3 for identity, 7 to
+    3 for resources. A public method beyond the port's would be surface the
+    materialiser was meant not to reach.
+    """
+    extra = {
+        m for m in vars(impl)
+        if not m.startswith("_") and callable(vars(impl)[m])
+    } - set(_port_methods(port))
```
totalfrank, line 213: **Non-blocking:** `test_device_delegate_adds_no_surface_of_its_own` is vacuous for the two activation cases (`(ActivationPort, DeviceActivation)`, `(ActivationPort, PlatformActivation)`).

`DeviceActivation` and `PlatformActivation` define nothing of their own beyond `_PROJECT` — all six forwarding methods live on their shared base `_DelegatingActivation` (`apply/activation_delegates.py`). `vars(impl)` only inspects the class's own `__dict__`, so for these two `extra` is always the empty set regardless of what `_DelegatingActivation` actually exposes. The other three cases (`DeviceIdentity`, `DeviceResource`, `DeviceSkillPackageUpload`) define their methods directly and are genuinely checked.

Today there's no stray surface on `_DelegatingActivation`, so this doesn't fail — but the test silently gives up the guarantee its own docstring claims ("A public method beyond the port's would be surface the materialiser was meant not to reach") for half the pairs it parametrizes over. A future public method added to the shared base wouldn't be caught here.

Suggested fix — walk the MRO instead of the class's own dict:
```suggestion
    extra = {
        m for m in dir(impl)
        if not m.startswith("_") and callable(getattr(impl, m))
    } - set(_port_methods(port))
```

---

totalfrank (commented):
Reviewed the new commit, `961038d` (7 commits since the last round were already covered; this round is `b740c491`..`961038d`, 1 commit): adds `test_device_port_forwarding.py` and removes `IdentityFilePort` from `bot_config_manifest`'s `provides`, addressing both findings Codex posted after the last round.

**Status checklist (all rounds):**
- `identity_files.py` forward reference for `IdentityService` — fixed, resolved (round 3).
- `DeviceIdentity`/adapter location under `apply/` vs a general `core/adapters` — author's call (deferred to follow-up), resolved (round 3).
- `registry.py` redundant quoted annotations (10 params) — fixed, resolved (round 6).
- Codex P1: no behavioral conformance test for device delegate forwarding — **fixed this round**, resolved.
- Codex P1: `bot_config_manifest` still claimed `IdentityFilePort` under `provides` after the move — **fixed this round**, resolved.
- New (this round): `test_device_delegate_adds_no_surface_of_its_own` is vacuous for the two activation cases, since `vars(impl)` only sees the subclass's own `__dict__` and both delegating methods live on the shared `_DelegatingActivation` base — non-blocking, open.

**Open: 1** (non-blocking test-coverage gap, no production impact).

---

src/backend/tests/community/core/bot_config_manifest/apply/test_device_port_forwarding.py
totalfrank: Correct, and worth more than "non-blocking" — the assertion could not fail for that case at all. Confirmed before fixing: `vars(DeviceActivation)` yields **zero** public callables, because all six methods are inherited from `_DelegatingActivation` and the subclass adds only `_PROJECT`. So `extra` was `set() - port_methods` — empty by construction, whatever the shared base exposed. The other three delegates define their methods directly, which is why the suite looked sound.

Took the suggested fix (`dir()` walks the MRO) in `2eb266e6`, and mutation-tested it rather than assuming: leaking a public `leaked_helper` onto `_DelegatingActivation` now fails `[ActivationPort-DeviceActivation]`, and passed silently before.

One correction to the finding itself, for the record: `(ActivationPort, PlatformActivation)` is not among the parametrised cases — `_DEVICE_IMPLS` covers the four `Device*` delegates only, so it was one vacuous case, not two. The diagnosis and the fix are right regardless.

1151 tests pass.

---

totalfrank (commented):
Reviewed the new commit this round: `2eb266e6` (1 commit since the last summary at `961038d3`) — fixes the one open finding from the previous round by walking the MRO (`dir()`) instead of the class's own `__dict__` in `test_device_delegate_adds_no_surface_of_its_own`. Verified: the fix is exactly the suggested diff, `dir(impl)` on a class does walk its MRO, and empirically confirmed (standalone repro) that `ABCMeta`-provided attributes like `register` do not leak into `dir()` of a concrete subclass, so no false positives are introduced. No new issues found in this commit — it's test-only, minimal, and matches its own commit message's claim.

**Status checklist (all rounds):**
- `identity_files.py` forward reference for `IdentityService` — fixed, resolved (round 3).
- `DeviceIdentity`/adapter location under `apply/` vs a general `core/adapters` — author's call (deferred to follow-up), resolved (round 3).
- `registry.py` redundant quoted annotations (10 params) — fixed, resolved (round 6).
- Codex P1: no behavioral conformance test for device delegate forwarding — fixed, resolved (round 7).
- Codex P1: `bot_config_manifest` still claimed `IdentityFilePort` under `provides` after the move — fixed, resolved (round 7).
- `test_device_delegate_adds_no_surface_of_its_own` vacuous for the activation cases (only `DeviceActivation`, not `PlatformActivation` — corrected in round 7's reply) — **fixed this round**, resolved.

**Open: 0.** PR is clean as of this round (`2eb266e6`).

---

FreddieSun (approved):

PR #1870, opened 2026-09-03
feat(backend): cli_tools — platform-managed command-line tools (W9)

totalfrank:
## Problem

`cli_tools` was the last unshipped category of the bot config manifest. W1 already parsed the whole vocabulary including the mandatory `digest`, and the artifact shape landed early with W12 (#1734) — but nothing installed a tool, recorded what a bot had, or exposed it to the agent, so the capability resolver refused the category at `PUT`.

Two constraints shaped the design, both verified in the tree:

- **No file-write path can set an executable bit.** The BaaS device uploads through `POST /api/file/upload` with a `target_path` and no mode field, and `DeviceFileSystem` exposes no `chmod`.
- **A tool carries state the platform must keep** — the pinned `sha256`, which archive member was selected, the delivered `md5`, the version. Once the platform owns that record, the record has to be the only way in.

## Solution

**CLI tools are platform-managed**, and **the engine owns placement**.

```
      POST/GET/DELETE .../cli-tools          manifest apply / W13 creation
                   │                                      │
                   └──────────────┬───────────────────────┘
                                  ▼
                          CliToolService          ← the only thing that
             (fetch · verify · store · record       does the work
                    · deliver · replace)
                                  │
                ┌─────────────────┼─────────────────┐
                ▼                 ▼                 ▼
        ac_bot_cli_tool     OSS tool store    delivery, per family
          (metadata)         (the bytes)      ARCA: install by name
                                              teclaw: artifact refs
```

**One implementation, two callers.** The routes and the `cli_tools` materialiser both reach `CliToolService`; neither reimplements a step. A test installs the same wrong-architecture binary through each door and asserts they fail identically — that is what makes "one implementation" a fact rather than a claim.

**The pipeline, and the order is the design:** fetch → digest → unpack → select `subpath` → verify ELF → md5 → **store in OSS** → deliver → record. The OSS write comes before delivery because on teclaw it *is* the delivery, and on ARCA it makes a redelivery possible without re-fetching a source URL that may have rotated. The row is written last, because the table is the platform's claim that the bot has the tool and a claim made before the engine accepted it can be false. A failed delivery records nothing and discards the object it just stored, whose key is derived rather than recorded and would otherwise be collected there or never.

[...]

PR #1943, opened 2026-09-06
feat(engine): implement the ARCA CLI tool endpoints (W9 #1477)

totalfrank:
## Problem

The platform half of `cli_tools` shipped in #1870 — fetch, pin verification, archive member selection, architecture validation, the management API, the apply materialiser, and `ArcaCliToolPort`, which calls `/api/cli/install`, `/api/cli/delete`, `/api/cli/list` and `/api/cli/replace` on the bot's engine.

`src/engine` had no `/api/cli` router at all.

So every apply of a manifest declaring `cli_tools` on an ARCA bot reached the engine, hit `DeviceAdapterEndpointNotFoundError`, and failed the category — cleanly and with a correct report, but the tools never arrived. Because a category is written all-or-nothing (work-items §3.2), one declared tool was enough to fail `cli_tools` for that bot on every apply, forever. **W9 read as delivered and was inert in practice.**

teclaw needs no write endpoints — its artifact is the delivery — so ARCA was the only family blocked, and it is the family carrying today's users.

## Solution

The engine half: a `CliToolsService` protocol, one engine-agnostic filesystem implementation, the `/api/cli` router, and bindings on both community engines.

**One shared service, not per-engine adapters.** CLI placement never touches the gateway; it is local filesystem work (write, `chmod +x`, list, delete). Per-engine adapters — the shape `skills` uses, because it relays to the OpenClaw gateway — would have been copies differing by one constant.

**Three orderings carry the contract:**
- `install` writes to a temp file, chmods, then `os.replace`s. Until the rename, the target either does not exist or still holds the previous tool, so an interrupted install never leaves a half-written file reachable under the command's name.
- `replace_all` installs everything named *before* pruning anything unnamed. Pruning first would publish an intermediate state where a kept tool is briefly gone — the exact window this endpoint exists to avoid. An entry that failed, at any stage, is not pruned either, so a failed replacement degrades to *unchanged*, never *removed*.
- `list` reads the directory every call and hashes bytes on disk. A replay of the last write would be a tautology; the value is catching what a replay cannot.

**`404` stays reserved** for "this engine build has no CLI endpoints". An unknown tool answers `200 + success:false + error:"not_found"`; a 404 there would make one bad name look like a permanently broken engine. An engine without the capability refuses with `501` via the existing guard.

## Where the tools land

`LocalCliToolsService` takes a **required concrete path**, and each engine names its own literal where it binds the service. No resolver, no lookup table, no environment read inside `core/` — one line per engine, which is the only place a location can change:

| Engine | Directory |
|---|---|
| `openclaw` | `/home/admin/.openclaw/cli` |
| `claude_code` | `/home/admin/.aicoding/cli` |

Both are the deployment's own constants, confirmed with the deployment owners. `.aicoding` is worth calling out: it is neither of the two candidates the repository points at — `.claude_code` (that engine's skills layout root) or OpenClaw's tree (where `start_claude_code.sh` sets `RELAY_DEFAULT_CWD`) — so inferring it from the community image would have been wrong.

**Known trade-off, accepted deliberately.** These are shared per host, not per bot. On singlebox, where BaaS builds a workspace per bot *and* per engine, several bots would use one tool directory and a whole-set `replace` from any of them would delete the others' tools. Harmless on ARCA, where a container holds one bot. Recorded in the spec and §4 A2 with a follow-up for a per-bot location if `cli_tools` is ever run on singlebox.

## Validation

Full engine suite: **2533 passed, 5 skipped** (skips pre-existing). CI is green on the current head, including the 90% changed-line coverage gate.

Six automated review rounds plus two rounds of human review. **Every finding fixed, each with a regression test:**

*Hostile-directory hardening — the tool directory lives inside the bot's container, so its contents are untrusted input:*
- pruning fed disk-sourced names through the caller-input validator, so an agent-created file with a trailing space made every future `replace_all` for that bot fail — permanently, since that file could never be removed by that path
- reads followed symlinks, turning `download` into an arbitrary read of any file the engine account could open
- a FIFO under the tool directory made `download`'s blocking `os.open` wait forever, so enough requests would occupy the shared `to_thread` executor and stall all CLI operations
- the `.part` scratch filter hid a legally-named tool (`unpack.part`) from listing *and* pruning, stranding it on the bot after its removal from the manifest

*Integrity:*
- an entry whose base64 would not decode never reached the service, so the prune deleted the installed tool it was meant to update — "your update failed" silently became "your tool is gone"
- `delete` swallowed non-`FileNotFoundError` unlink failures and reported success, so the apply report could claim a removed tool was gone while it was still callable (AGENTS.md L266)
- unserialised writes let a replacement prune a tool a concurrent install had just reported as placed

*Design, from human review:*
- the directory was resolved through a lookup table and environment overrides; that indirection is gone, along with the environment reads in `core/` that Codex flagged (AGENTS.md L73) — the workspace is now read at the engine composition root, which L72 names as the approved place
- every `CliToolsService` member is `@abstractmethod` and implementations subclass explicitly, so a dropped method fails at construction rather than inheriting a `...` stub that returns `None`

**Open findings: 0.**

## Compatibility and risk

Purely additive on the engine side; no flag, no migration. The capability declaration is the switch — an engine that neither assigns `_cli_tools` nor declares `CLI_*` refuses with 501, and `validate_capabilities()` refuses to start an engine whose declaration and assignment disagree. Deploy order is irrelevant: the platform caller already handles the "no endpoints" case, so an unchanged engine behaves exactly as today.

`ApiResponse` gains an optional `error` field. The envelope already serialises nulls for its four other optional fields across ~150 call sites, so this adds no new kind of change for existing consumers.

Scope is `openclaw` and `claude_code` — the only `Engine` subclasses in this repository. `plugins/aicoding` and `plugins/hermes` carry ports, not engines; if the corp build ships them as engines they each need the same binding.

## Follow-ups

- End-to-end assertion on a live ARCA bot (the one acceptance criterion not verifiable here — it needs a deployed bot and belongs platform-side)
- A per-bot location if `cli_tools` is ever run on singlebox
- `aicoding` / `hermes` bindings if the corp build ships them as engines
- PATH injection, which makes the tool directory's placement load-bearing rather than a convention

## Spec

- `src/engine/specs/2026-09-06-arca-cli-tool-endpoints/spec.md`
- `src/engine/specs/2026-09-06-arca-cli-tool-endpoints/plan.md`
- `src/engine/specs/2026-09-06-arca-cli-tool-endpoints/tasks.md`

Authoritative contract: `src/backend/docs/bot-config-manifest/engine-requirements.zh-CN.md` §4 A2, updated in this PR to match what shipped (the fifth endpoint, `md5` on `list`, the 501 note, and the per-engine directories) — the reconciliation the engine handoff spec asked for.

## Related issues

Relates to #1477 (W9 `cli_tools`).

---

https://claude.ai/code/session_01XjRpKZvkzRXFC8ZJVRqJWC


src/engine/specs/2026-09-06-arca-cli-tool-endpoints/plan.md
```diff
@@ -0,0 +1,271 @@
...
+`POST` for delete, not `DELETE`: the name travels in a body, and proxies strip
+bodies from `DELETE` (`arca_port.py:_DELETE_METHOD`).
+
+## Key Files & Functions
+
+```python
+# src/engine/community/core/cli_tools/protocol.py (new)
+@runtime_checkable
+class CliToolsService(Protocol):
+    async def install(self, name: str, data: bytes) -> None: ...
```
totalfrank-personal, line 83: Make sure all methods here in the protocol is annotated with @abstractmethod and for the implementation of this protocol, it needs to explicitly extends this protocol.

src/engine/specs/2026-09-06-arca-cli-tool-endpoints/plan.md
```diff
@@ -0,0 +1,271 @@
...
+
+```diff
+# src/engine/community/engines/openclaw/engine.py:169
+     self._skills = OpenClawSkillsAdapter(self._port)
++    self._cli_tools = LocalCliToolsService(openclaw_cli_dir)
+```
+
+```python
+# src/engine/community/core/cli_tools/directories.py (new)
+def openclaw_cli_dir() -> Path:
```
totalfrank-personal, line 186: What about other engines? Do we want helper method like this for them as well? Because I believe different engines have different workspace and our design is to put the cli directory parallel to it.

totalfrank-personal (commented):
Two comments, please take a look.

src/engine/specs/2026-09-06-arca-cli-tool-endpoints/plan.md
totalfrank: Done in fe43da3b.

Every `CliToolsService` member is now `@abstractmethod`, and the plan states that implementations subclass the Protocol explicitly (`class LocalCliToolsService(CliToolsService)`) rather than relying on structural satisfaction. Added the reasoning so it survives into the implementation: this repo runs no static type checker, so a bare `...` stub is silently *inherited* in place of a method an implementation dropped — the name resolves and the call returns `None`. `@abstractmethod` turns that into a failure at construction instead.

This is the same rule the backend states for its outbound ports in `src/backend/src/agentclaw/community/core/ports/README.md`, so I cited it there.

Also carried into `tasks.md` as two separate "done when" conditions — one on Task 1 for the decorators, one on Task 2 for the explicit subclassing — so neither can be quietly skipped during implementation.

---

src/engine/specs/2026-09-06-arca-cli-tool-endpoints/plan.md
totalfrank: You're right, and I checked — the workspaces really do differ:

- OpenClaw: `$OPENCLAW_WORKSPACE_DIR`, else `~/.openclaw/workspace` (`plugin_api/workspace_root.py:21`)
- Claude Code: `<home>/.claude_code/workspace`, no env override (`plugins/claude_code/layout_pool.py:45`)

So the single `openclaw_cli_dir()` I had would have put Claude Code's tools under OpenClaw's tree. Fixed in fe43da3b by splitting the rule from the input, since "parallel to the workspace" is the part that's shared:

```python
def cli_dir_beside(workspace: Path) -> Path:
    """The rule, stated once: a bot's `cli/` is its workspace's sibling."""
    return workspace.parent / "cli"

def openclaw_cli_dir() -> Path: ...      # from workspace_root_strict()
def claude_code_cli_dir(home: Path = Path("/home/admin")) -> Path: ...
```

Each engine supplies only its own workspace; a third ARCA engine adds one more resolver here and touches nothing else. Task 5 now also asserts the two engines resolve to **different** directories, so a future engine can't silently inherit OpenClaw's.

Worth noting for scope: only `openclaw` and `claude_code` have `Engine` subclasses in this repo — `aicoding` and `hermes` carry ports but no engine — so those two are the whole surface here. If the corp build has the other two, they'll each need a resolver and a binding, and that isn't covered by this spec.

---

src/engine/src/engine/community/api/cli/router.py
```diff
@@ -0,0 +1,189 @@
...
+            data = base64.b64decode(item.content_b64, validate=True)
+        except (binascii.Error, ValueError) as error:
+            undecodable.append(
+                {"name": item.name, "success": False,
+                 "message": f"content_b64 is not valid base64: {error}"}
+            )
+            continue
+        payloads.append(CliToolPayload(name=item.name, data=data))
+
+    results = await _service().replace_all(payloads)
```
totalfrank: **Blocking:** a `replace` request with one item whose `content_b64` fails to decode silently deletes any *existing* tool that shares that item's `name`, instead of leaving it unchanged.

`item.name` for anything that fails to decode is only recorded in `undecodable` (line 120) and never appended to `payloads` (the `continue` on line 124 skips it). So it never reaches `LocalCliToolsService.replace_all`, never appears in `results`, and is therefore absent from `keep = {result.name for result in results}` in `_replace_all_sync` (`service.py:276`). The prune loop right after that deletes everything on disk not in `keep` — including a tool this request never actually touched.

Concrete repro: install `foo` successfully, then `POST /api/cli/replace` with `tools=[{"name": "foo", "content_b64": "not-base64"}]`. The response reports `foo` as a decode failure (`success: false`), which reads as "this apply couldn't update foo" — but `foo` is actually gone from the bot's filesystem. This contradicts the endpoint's documented contract (both here and in `service.py`'s `_replace_all_sync` docstring) that a failed install degrades to "unchanged", never "removed" — that guarantee only holds for names that make it as far as `_install_sync`.

Suggested fix: seed the service's `keep` set with every requested name up front, not just the ones that decoded, e.g. pass `{t.name for t in request.tools}` into `replace_all` (or a new `preserve` argument) so pruning only removes names truly absent from the request.

---

totalfrank (commented):
Reviewed commits `d4f0e5c`..`845e3cc` (the full PR: spec/plan/tasks docs, `cli_tools` core package + `LocalCliToolsService`, the `/api/cli` router, engine bindings for openclaw/claude_code, and the reconciled contract doc). Two earlier review threads (protocol conformance, per-engine directories) were already resolved by the author during the spec phase.

Checked: correctness of the filesystem service (atomic install, replace-then-prune ordering, symlink/hostile-directory hardening, concurrency locking), the HTTP wire contract (404 vs 200+`error`, capability gating, base64 handling), the engine/capability wiring (`BaseEngine`, `Capability`, `EngineManager`, both engine bindings), and test coverage of the new behavior. One blocking issue found in the `/api/cli/replace` handler's interaction with the service's prune step — see inline comment.

---

src/engine/src/engine/community/core/cli_tools/service.py
```diff
@@ -0,0 +1,368 @@
...
+        """
+        try:
+            path.unlink()
+        except FileNotFoundError:
+            return
+        except OSError as error:
+            # Replaced by a directory between the scan and here, or immutable.
+            # One stubborn entry must not abort the rest of the prune.
+            log.warning("[cli_tools] could not remove %s: %s", path, error)
+            return
```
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in 2692aac6. This was a real integrity bug and the AGENTS.md rule is exactly on point.

`delete` now treats **only** `FileNotFoundError` as success; every other `OSError` propagates and the router reports the failure. The distinction matters because "already absent" genuinely is the contract's success case (§4 A2), while "read-only filesystem" is a tool that is still callable on the bot.

Pruning is split off into `_prune_path` rather than sharing the same behaviour, because the two have opposite requirements. A prune walks names read off disk, and an earlier review round found that one unremovable entry aborting the loop left the replacement half-applied *and* permanently broken for that bot. So prune keeps its per-entry isolation — but the failures no longer vanish: they come back on a new `ReplaceOutcome.prune_failures` and the router surfaces them in the envelope's `message`.

That last part is the piece your comment's "or a per-item result" pointed at. Pruned names were never in the request, so they have no verdict slot in `results` — and the platform's parser ignores names it did not send, so putting them there would drop them silently. The envelope message is the one place the information survives, and without it a replacement reads as a clean sweep while a tool the manifest dropped is still runnable.

Two regression tests: `test_delete_propagates_a_failure_instead_of_reporting_success` and the prune assertion in `test_an_unremovable_entry_does_not_abort_the_rest_of_the_prune`.

---

src/engine/src/engine/community/api/cli/router.py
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in 2692aac6 — this was the most serious of the four, since it turns "your update failed" into "your tool is gone" while the response says only the former.

`replace_all` gains `also_keep`, and the router passes every undecodable name through it, so such an entry lands in the service's `keep` set without being installed. It now degrades to *unchanged*, exactly as an entry that reached `_install_sync` and failed there already did — the guarantee the docstring claimed but only delivered for names that got that far.

Two regressions, because the bug needs both layers to see it: `test_also_keep_spares_a_name_the_caller_could_not_prepare` at the service, and `test_an_undecodable_entry_does_not_delete_the_installed_tool` end-to-end through the router, which installs `mycli`, replaces it with an undecodable payload, and asserts the original bytes survive.

---

src/engine/src/engine/community/core/cli_tools/service.py
```diff
@@ -0,0 +1,368 @@
...
+        nothing about what that path points at. The tool directory lives
+        inside the bot's container, where the agent can create a symlink, so
+        following one here would turn the download endpoint into an arbitrary
+        read of any file the engine account can open. ``O_NOFOLLOW`` makes the
+        final component's symlink an error, and a link reads as "no such
+        tool", which is exactly what it is.
+        """
+        try:
+            fd = os.open(str(path), os.O_RDONLY | os.O_NOFOLLOW)
+        except OSError:
```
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in 2692aac6.

Reads now open `O_RDONLY | O_NOFOLLOW | O_NONBLOCK` and check `stat.S_ISREG` on the descriptor before reading; anything else — FIFO, socket, device — reads as "no such tool", which is what it is. `O_NONBLOCK` is a no-op for regular files, so the only behaviour it changes is the one that was hanging.

The threat model is right: the tool directory is inside the bot's container, so the agent can create a FIFO there, and `read_tool` backs the `CLI_DOWNLOAD` capability. Enough concurrent downloads of it would occupy the shared `to_thread` executor and stall every CLI operation for that bot.

`list` was already safe — `_names_on_disk` uses `entry.is_file(follow_symlinks=False)`, which is `S_ISREG` — but `read_tool` reaches a path directly, so it needed its own guard. `test_a_fifo_does_not_block_the_worker_thread` covers it with an `asyncio.wait_for` timeout so a regression hangs the test rather than passing quietly, plus a router-level check that the response is the ordinary `200 + not_found`.

---

src/engine/src/engine/community/api/cli/router.py
totalfrank: Confirmed and fixed in 2692aac6, along the lines you suggested — `replace_all` gains an `also_keep` argument and the router passes the undecodable names through it.

I went with *only* the undecodable names rather than `{t.name for t in request.tools}`. Seeding with every requested name would also spare names that reached `_install_sync` and failed there, which the service already handles by keeping them in `keep` via their `results` entry — so passing them again is redundant, and `also_keep` staying narrow keeps its meaning precise: "requested, but never got far enough to produce a verdict."

Your repro is now a test verbatim (`test_an_undecodable_entry_does_not_delete_the_installed_tool`): install `foo`, replace it with `not-base64`, assert the original bytes are still on disk. Plus `test_also_keep_spares_a_name_the_caller_could_not_prepare` at the service layer.

Good catch on the docstring being narrower than it read — the "degrades to unchanged, never removed" guarantee did only hold past `_install_sync`. Both the protocol and `_replace_all_sync` docstrings now say what `also_keep` is for and why the gap existed.

---

src/engine/src/engine/community/core/cli_tools/directories.py
```diff
@@ -0,0 +1,146 @@
...
+def cli_dir_for(engine: str) -> Path:
+    """Where ``engine`` keeps this bot's tools, overrides first.
+
+    Order: ``BOT_CLI_DIR_<ENGINE>``, then ``BOT_CLI_DIR``, then this engine's
+    :data:`ENGINE_CLI_DIRS` entry, then :func:`default_cli_dir`.
+    """
+    for var in (cli_dir_env_var(engine), CLI_DIR_ENV):
+        override = os.environ.get(var)
+        if override and override.strip():
+            return Path(override.strip())
```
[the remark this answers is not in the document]
totalfrank: Not changing this one, and leaving the thread open for @totalfrank to rule on rather than resolving it myself. The rule you cite is real; I think following it here would break a requirement and contradict the local convention. Reasons, so the call can be made on evidence:

**1. Late binding is the requirement, not an implementation detail.** The whole point of this module is that the directory can be retuned without a rebuild — `BOT_CLI_DIR_CLAUDE_CODE` exists precisely because Claude Code's production location is still being settled with the deployment owners. A value loaded and validated at bootstrap freezes at process start. The service holds a *callable* for the same reason: BaaS injects `OPENCLAW_WORKSPACE_DIR` at spawn, after the engine object is constructed, so a construction-time read would capture the wrong value even for the default path.

**2. The sibling this module extends already does exactly this.** `plugin_api/workspace_root.py` reads `OPENCLAW_WORKSPACE_DIR` from `os.environ` at call time, and `default_cli_dir()` is built directly on it. Routing the override through config while the base path it derives from stays an env read would put the same decision in two layers.

**3. `core/` in this engine reads env in six other places today** — `skills/skills_repo_download.py`, `cron/services/notify.py`, `bash/base.py`, `aicoding/workspace_service.py` (twice), `session_favorite/repository.py`. Whatever the rule says, the established convention in this exact layer is the opposite. Bringing one new module into compliance while its neighbours and its own dependency stay out doesn't move the boundary; it just makes this file inconsistent with the file beside it.

**4. AGENTS.md also says** "Do not add speculative abstraction or configurability" (L262) and "Do not refactor unrelated code" (L263). No config-injection path exists for this layer, so honouring L73 here means building one — and to be coherent it would have to cover all six neighbours plus `workspace_root`.

What I did do: the overrides are validated where they're read — blank or whitespace-only is ignored rather than resolving tools to the filesystem root — and the resolution order is documented at the top of the module with the tuning instructions.

If the intent is that `core/` genuinely stops reading env, I'd rather do it as one change across all seven call sites with a real config seam, and I'm happy to open that separately. Doing it for this module alone is the version I don't think is worth it.

---

src/engine/src/engine/community/core/cli_tools/service.py
```diff
@@ -0,0 +1,404 @@
...
+        """
+        path = self._path_of(name)
+        try:
+            path.unlink()
+        except FileNotFoundError:
+            log.info("[cli_tools] delete name=%s: already absent", name)
+            return
+        log.info("[cli_tools] deleted name=%s", name)
+
+    def _prune_path(self, path: Path) -> str | None:
```
totalfrank, line 212: `nit:` this commit renames `_delete_path` to `_prune_path`, but `validate_tool_name`'s docstring at `service.py:84` still says "see :meth:`LocalCliToolsService._delete_path`" — a method that no longer exists. Worth updating that cross-reference to `_prune_path` while it's fresh.

---

totalfrank (commented):
Reviewed the new commit: `2692aac6` (three defects from the previous automated review round: undecodable-entry deletion, delete-reports-success-on-failure, FIFO worker-thread hang).

Verified independently by reading the current code and running the full engine suite locally (`2545 passed, 5 deselected`, matching the commit's claim) plus the targeted `cli_tools`/`cli_router` tests (72 passed).

- ✅ Fixed — `service.py` `_delete_sync`: only `FileNotFoundError` is treated as success; other `OSError`s now propagate. Prune isolation preserved via new `_prune_path`, failures surfaced through `ReplaceOutcome.prune_failures` → router `message`. Regression: `test_delete_propagates_a_failure_instead_of_reporting_success`. *(resolving)*
- ✅ Fixed — `router.py` `/replace`: undecodable entries are now passed to `replace_all(..., also_keep=...)`, so they degrade to "unchanged" instead of being pruned. Regression: `test_an_undecodable_entry_does_not_delete_the_installed_tool` (router) + `test_also_keep_spares_a_name_the_caller_could_not_prepare` (service). *(resolving — this was filed as two separate threads, one from Codex and one from my own earlier review; both cover the same fix)*
- ✅ Fixed — `service.py` `_read_sync`: reads now open `O_NONBLOCK` and require `stat.S_ISREG`, so a FIFO can no longer park a `to_thread` worker. Regression: `test_a_fifo_does_not_block_the_worker_thread` / `test_download_of_a_fifo_does_not_hang`. *(resolving)*
- ⬜ Still open — `directories.py` `cli_dir_for()` reads raw environment state in core-domain code rather than through the configuration boundary (Codex, `AGENTS.md:L73-L74`). Not touched by this commit and no reply claims it was addressed; leaving open.
- New (this round) — `nit:` stale docstring cross-reference to the now-renamed `_delete_path` (see inline comment).

**Still open: 1** (the configuration-boundary finding on `cli_dir_for()`).

---

src/engine/src/engine/community/core/cli_tools/service.py
totalfrank: Fixed in 25f931f7 — now points at `_prune_path`.

Worth doing rather than deferring: that cross-reference is where a reader is sent to learn *why* disk-sourced names skip `validate_tool_name`, which is exactly the subtlety the earlier hostile-directory fix turned on. A dead method name there sends them looking for something that doesn't exist, at the moment they most need the explanation.

Grepped for other stragglers from the rename — `_delete_path` appears nowhere else in the tree.

---

totalfrank (commented):
Reviewed commits this round: `25f931f7` — the only commit since the last review round, and exactly the one-line docstring fix (`_delete_path` → `_prune_path`) already replied to and resolved inline. No other changes in it, and no new defects found.

- ✅ Fixed — Protocol conformance: every `CliToolsService` member `@abstractmethod`, implementations subclass explicitly. Resolved during spec phase.
- ✅ Fixed — Per-engine cli directories: `cli_dir_beside()` shares the rule, each engine supplies its own workspace. Resolved during spec phase.
- ✅ Fixed — `service.py` `_delete_sync`: only `FileNotFoundError` is treated as success; other `OSError`s now propagate. Resolved.
- ✅ Fixed — `router.py` `/replace` + `service.py` `also_keep`: an undecodable entry no longer causes the prune to delete the tool it was meant to update; degrades to unchanged. Resolved (two threads — Codex's and my own duplicate finding).
- ✅ Fixed — `service.py` `_read_bytes_nofollow`/`_read_sync`: reads now use `O_NONBLOCK` + `S_ISREG`, so a FIFO in the tool directory can no longer park a `to_thread` worker. Resolved.
- ✅ Fixed — stale `_delete_path` → `_prune_path` docstring cross-reference. Resolved in `25f931f7`.
- ⬜ Still open — `directories.py` `cli_dir_for()` reads raw environment state in core-domain code rather than through the configuration boundary (Codex, `AGENTS.md:L73-L74`). Position held once already, with reasoning (late-binding is the actual requirement here, six other `core/` call sites already read env directly, no config-injection seam exists for this layer). No commit touches it this round and no reply since — left open, awaiting @totalfrank's ruling as previously flagged.

**Still open: 1** (the configuration-boundary finding on `cli_dir_for()`, awaiting author's call).

---

src/engine/src/engine/community/engines/openclaw/engine.py
```diff
@@ -167,6 +175,11 @@ def __init__(
         self._node = OpenClawNodeAdapter(self._port)
         self._mcp = OpenClawMcpAdapter(self._port)
         self._skills = OpenClawSkillsAdapter(self._port)
+        # CLI tools need no port: placing a command is local filesystem work.
+        # Where they land is a deployment fact — override with BOT_CLI_DIR_OPENCLAW
+        # or BOT_CLI_DIR, or add an entry to ENGINE_CLI_DIRS; see
+        # core/cli_tools/directories.py.
+        self._cli_tools = LocalCliToolsService(cli_dir_resolver("openclaw"))
```
totalfrank-personal: I don't like this approach. Let's drop cli_dir_resolver and all code reachable only through it.

Then LocalCliToolsService takes a required parameter which is the path so that each engine has to specify it when it sets its _cli_tools?

src/engine/src/engine/community/core/cli_tools/directories.py
```diff
@@ -0,0 +1,146 @@
+"""Where a bot's command-line tools live — **and how to change it.**
```
totalfrank-personal: If you look at the other comment, I don't think we need this file anymore, right?

totalfrank-personal (commented):

src/engine/src/engine/community/engines/openclaw/engine.py
totalfrank: Done in f0af4db9 — exactly as you described. `LocalCliToolsService(directory: Path)`, required and concrete, and each engine names its own at the binding site:

```python
# engines/openclaw/engine.py  and  engines/claude_code/engine.py
self._cli_tools = LocalCliToolsService(workspace_root().parent / "cli")
```

`cli_dir_resolver` is gone, and with it everything only reachable through it — the whole `directories.py`, `ENGINE_CLI_DIRS`, and the `BOT_CLI_DIR` / `BOT_CLI_DIR_<ENGINE>` overrides. Net −12 tests, since the tuning-precedence suite went with them.

**And I owe you a correction.** I'd argued the callable was needed because BaaS injects `OPENCLAW_WORKSPACE_DIR` *after* the engine object is constructed. That's wrong. BaaS builds the child environment and hands it to `Popen` before the adapter process starts (`_process_manager.py:680,830` — the comment there even says the adapter reads it to decide its path root), so the variable is in `os.environ` from process start and a construction-time read sees the right value. The late binding bought nothing; I should have checked that before defending it.

One thing I did keep, because it's load-bearing rather than indirection: both engines derive from `workspace_root()` rather than a literal. BaaS builds that path per bot **and** per engine, so a constant would give every bot on a singlebox host the same tool directory — and a whole-set `replace` from either would delete the other's tools. `test_two_bots_never_share_a_tool_directory` pins it. If you'd rather see a literal there anyway, say so and I'll swap it.

---

src/engine/src/engine/community/core/cli_tools/directories.py
totalfrank: Right — deleted in f0af4db9. With the path required at the binding site there was nothing left in it that anything reached.

Also removed the docs that described the knobs it provided: §4 A2's override-precedence table is replaced by a short statement that the落点 lives at each engine's binding line, and the spec's *Resolved* entry says the same. Leaving those behind would have documented three environment variables that no longer exist.

---

src/engine/src/engine/community/core/cli_tools/directories.py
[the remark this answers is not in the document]
totalfrank: Reopening this to say you were right, and that my rebuttal was partly wrong on the facts.

f0af4db9 removes the environment reads from `core/` entirely — not by adding a config seam, but by deleting the resolver: `LocalCliToolsService` now takes a required concrete path and each engine states its own where it binds the service. The engine composition root is where the workspace gets read, which is what AGENTS.md L72 names as the approved place.

The specific claim I got wrong was point 1. I argued a construction-time read would capture the wrong value because BaaS injects `OPENCLAW_WORKSPACE_DIR` after the engine object exists. It doesn't: BaaS builds the child environment and passes it to `Popen` before the adapter process starts (`_process_manager.py:680,830`), so it is set from process start. Late binding wasn't buying anything, and I should have verified that before leaning on it.

Points 2–4 still describe the codebase accurately — `plugin_api/workspace_root.py` and six other `core/` modules do read env directly — but they were arguments for consistency with a convention, not against your rule, and they don't survive the first point falling over.

---

totalfrank (commented):
Reviewed the new commit: `f0af4db9` — the only commit since the last review round (`25f931f7`). It drops `cli_dir_resolver` and `core/cli_tools/directories.py` entirely per @totalfrank-personal's review comments, making `LocalCliToolsService` take a required concrete `Path` and binding it at each engine's composition root (`openclaw/engine.py`, `claude_code/engine.py`).

Verified independently by reading the current code (not just the replies already on the threads):
- `directories.py`, `ENGINE_CLI_DIRS`, `BOT_CLI_DIR`/`BOT_CLI_DIR_<ENGINE>`, and `cli_dir_resolver` are gone with no dangling references anywhere in the tree (grepped the full repo).
- `LocalCliToolsService.__init__` and `_dir()` are consistent with the new `Path`-not-`Callable` contract — no leftover call-site expects the old callable.
- The construction-time-read argument holds: `workspace_root()` returns a concrete `Path` (never `None`), and `_process_manager.py` sets `OPENCLAW_WORKSPACE_DIR` on the child `env` dict *before* `Popen` for both the gateway and the adapter process, so an engine reading it at `__init__` sees the right value — the commit's factual correction to its own earlier rebuttal checks out.
- Tests were updated consistently (`TestDirectories`/`TestTuning` removed with the module they tested; `test_two_bots_never_share_a_tool_directory` added at the engine level).
- CI: **Engine unit tests** is green on this head, matching the commit's claimed `2533 passed, 5 skipped`. (Backend unit tests / Singlebox coverage were still running at review time — not a blocker for this comment-only review.)

No new defects found in this commit.

- ✅ Fixed — Protocol conformance (`@abstractmethod` + explicit subclassing). Resolved, spec phase.
- ✅ Fixed — Per-engine cli directories (workspaces differ per engine). Resolved, spec phase.
- ✅ Fixed — `service.py` `_delete_sync` swallowed non-`FileNotFoundError` unlink failures. Resolved in `2692aac6`.
- ✅ Fixed — `router.py` `/replace` + `service.py` `also_keep`: undecodable entry no longer deletes the tool it was meant to update. Resolved in `2692aac6` (two threads, Codex + duplicate).
- ✅ Fixed — FIFO in the tool directory could park a `to_thread` worker forever. Resolved in `2692aac6` (`O_NONBLOCK` + `S_ISREG`).
- ✅ Fixed — `nit:` stale `_delete_path` docstring cross-reference. Resolved in `25f931f7`.
- ✅ Fixed — `directories.py` `cli_dir_for()` read raw env in core-domain code (Codex, `AGENTS.md:L73`). Held once with reasoning, then correctly conceded after @totalfrank-personal's request to remove the resolver outright; `f0af4db9` deletes the module and moves the read to each engine's composition root. Confirmed fixed by independent code read this round.
- ✅ Fixed — `directories.py` no longer needed as a file. Deleted in `f0af4db9`, confirmed.

**Still open: 0.** The PR is clean as of `f0af4db9` — no open findings, no new defects this round.

---

src/engine/src/engine/community/engines/claude_code/engine.py
```diff
@@ -137,6 +145,11 @@ def __init__(
         self._session = ClaudeCodeSessionAdapter(self._port)
         self._mcp = ClaudeCodeMcpAdapter(self._port)
         self._skills = ClaudeCodeSkillsAdapter(self._port)
+        # Beside this bot's workspace, as OpenClaw does. **This is the line to
+        # change once the deployment owners confirm where Claude Code's tools
+        # belong in production** — the community image is not what production
+        # deploys, so its layout is not evidence about production.
+        self._cli_tools = LocalCliToolsService(workspace_root().parent / "cli")
```
totalfrank-personal: Let's do not use workspace_root(). Instead, let's hard code it for now. For claude_code, it's /home/admin/.aicoding/cli

src/engine/src/engine/community/engines/openclaw/engine.py
```diff
@@ -167,6 +175,13 @@ def __init__(
         self._node = OpenClawNodeAdapter(self._port)
         self._mcp = OpenClawMcpAdapter(self._port)
         self._skills = OpenClawSkillsAdapter(self._port)
+        # CLI tools need no port: placing a command is local filesystem work,
+        # so the only per-engine fact is the directory — stated here, once.
+        # Beside this bot's workspace: BaaS injects that per bot and per engine
+        # (so two bots on one host never share a tool directory), falling back
+        # to the image's ~/.openclaw/workspace. **This is the line to change if
+        # OpenClaw's tool location moves.**
+        self._cli_tools = LocalCliToolsService(workspace_root().parent / "cli")
```
totalfrank-personal: Let's do not use workspace_root(). Instead, let's hard code it for now. For claude_code, it's /home/admin/.openclaw/cli

totalfrank-personal (commented):

src/engine/src/engine/community/engines/claude_code/engine.py
totalfrank: Done in 67a79dbb — `/home/admin/.aicoding/cli`, as a module-level literal:

```python
CLAUDE_CODE_CLI_DIR = Path("/home/admin/.aicoding/cli")
...
self._cli_tools = LocalCliToolsService(CLAUDE_CODE_CLI_DIR)
```

Worth having asked — `.aicoding` is neither candidate I'd have picked from reading the repo. The two I could see were `.claude_code` (from the skills layout root) and OpenClaw's tree (from `start_claude_code.sh` pointing `RELAY_DEFAULT_CWD` there), and both would have been wrong. The test now pins the constant and asserts it doesn't collide with OpenClaw's.

---

src/engine/src/engine/community/engines/openclaw/engine.py
totalfrank: Done in 67a79dbb — `OPENCLAW_CLI_DIR = Path("/home/admin/.openclaw/cli")`, a module-level literal, bound the same way as Claude Code's.

I read the comment's "for claude_code" as a copy-paste from the sibling thread, since this one is on `openclaw/engine.py` and the path is OpenClaw's. Say so if you meant something else.

Both are written as literals rather than derived, so the location reads as a property of the image and each engine has exactly one line to change.

**One consequence, recorded rather than argued** — you said "for now", so this is a note, not a push-back. The `workspace_root()` derivation was what kept bots apart on singlebox, where BaaS builds the workspace per bot *and* per engine (`_workspace.py:64-72` → `…/{bot_id}/{engine}/workspace`). With a shared constant, every bot on one host uses the same tool directory, so a whole-set `replace` from any of them deletes the others' tools. It's in the spec's *Resolved* entry and §4 A2 as a known trade-off, with a follow-up for a per-bot location if `cli_tools` is ever run on singlebox. Harmless on ARCA, where a container holds one bot.

---

totalfrank (commented):
Reviewed the new commit since the last round: `67a79dbb` — hard-codes each engine's CLI directory (`/home/admin/.openclaw/cli`, `/home/admin/.aicoding/cli`) in place of the `workspace_root()`-derived path, per @totalfrank-personal's two review comments on `f0af4db9`. Checked: the removed `workspace_root` imports have no other use in either `engine.py` (grepped both files), no leftover references to the deleted resolver/lookup-table anywhere in the tree, both edited files parse cleanly, and the updated `openclaw/tests/test_f2_baseline_parity.py` assertions match the new constants. **Engine unit tests** is green on this head (CI), matching the commit's claimed `2533 passed, 5 skipped`. No new defects found.

- ✅ Fixed — Protocol conformance (`@abstractmethod` + explicit subclassing). Resolved, spec phase.
- ✅ Fixed — Per-engine cli directories (workspaces differ per engine). Resolved, spec phase.
- ✅ Fixed — `service.py` `_delete_sync` swallowed non-`FileNotFoundError` unlink failures. Resolved in `2692aac6`.
- ✅ Fixed — `router.py` `/replace` + `service.py` `also_keep`: undecodable entry no longer deletes the tool it was meant to update. Resolved in `2692aac6`.
- ✅ Fixed — FIFO in the tool directory could park a `to_thread` worker forever. Resolved in `2692aac6`.
- ✅ Fixed — `nit:` stale `_delete_path` docstring cross-reference. Resolved in `25f931f7`.
- ✅ Fixed — `directories.py` `cli_dir_for()` read raw env in core-domain code. Held once, then conceded and resolved in `f0af4db9` (resolver deleted, path moved to the composition root).
- ✅ Addressed (human review, `f0af4db9`) — drop `cli_dir_resolver`, require a concrete `Path` at each engine's binding site. Done.
- ✅ Addressed (human review, `67a79dbb`) — hard-code `claude_code`'s CLI dir to `/home/admin/.aicoding/cli`. Done, this round.
- ✅ Addressed (human review, `67a79dbb`) — hard-code `openclaw`'s CLI dir to `/home/admin/.openclaw/cli`. Done, this round.

**Still open: 0.** The PR is clean as of `67a79dbb` — no open findings, no new defects this round.

---

FreddieSun (approved):
lgtm

PR #1949, opened 2026-09-07
fix(engine): normalize early-final followup session keys

evilcoda:
## Problem
Early-final followup runs are matched using the original session key verbatim. OpenClaw may add the default agent:main: prefix and lowercase the key, causing the real followup response to be dropped and the held empty final to be returned instead.

## Solution
Normalize case and whitespace and remove only the default-agent prefix for the early-final followup comparison. Keep exact equality after normalization; other agents, sessions and users remain isolated. Leave normal runId matching, announce handling, other session fallback branches and BaaS unchanged.

## Validation
- Rebased onto latest origin/dev (36b898b54).
- OpenClaw client tests: 57 passed.
- Changed executable production-line coverage: 100% (1/1); the multiline comparison is recorded as one executable line by coverage.py.
- Parameterized coverage for equal keys, prefix variants, case/whitespace normalization, different agents/sessions/users, substring collisions and invalid event keys.
- Pre-push Python SAST gate passed.
- Full singlebox/E2E not run locally; remote CI pending.

## Compatibility and risk
No HTTP schema, configuration or database changes. The behavior change is limited to linking a followup run while an early empty final is pending.

### DIFF
--- a/src/engine/src/engine/community/openclaw/client/gateway_client.py
+++ b/src/engine/src/engine/community/openclaw/client/gateway_client.py
@@ -717,7 +717,11 @@ class OpenClawGatewayClient:
                     return announce_matches_parent_session(payload)
                 if pending_early_final is not None and payload.get("state") not in ("final", "error", "aborted"):
                     payload_session_key = payload.get("sessionKey")
-                    return payload_session_key == session_key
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #1951, opened 2026-09-07
feat(bot-public): resolve agent-scope callback visibility from reques…

rudyLi:
## Problem

  The `POST /openapi/v1/collaboration/bots/{bot_uuid}/public` publish endpoint opens an approval ticket and persists the requested `visibility` on the block, to be
  The `POST /openapi/v1/collaboration/bots/{bot_uuid}/public` publish endpoint opens an approval ticket and persists the requested `visibility` on the block, to be
  applied later by the AGREE callback. For the `agent` scope, that callback did not honor the caller's requested visibility — it derived the target BCS `visibility`
  from the bot's top-level `friend_check_in_strategy` field (`"public"` when the strategy was `"OPEN"`, else `"protected"`). As a result an agent-scope publish could
  not land the explicit visibility the user requested, and a publish with no explicit `visibility` was auto-promoted to `public` whenever the BCS strategy flag
  happened to be `OPEN`, regardless of intent.

  ## Solution

  In `BotPublicService._apply_callback_decision`, the AGREE branch now writes the BCS visibility field from the user's requested visibility persisted on the approval
  block (`block["visibility"]`, defaulting to `"protected"`) — for both `user` and `agent` scopes. The agent-specific `friend_check_in_strategy` lookup is removed;
  the two scopes collapse to one line: `body[field] = block.get("visibility") or "protected"`. The scope→BCS-field mapping (`user` → `user_visibility`, `agent` →
  `visibility`, via `_BCS_VISIBILITY_FIELD_BY_SCOPE`) is unchanged, and the friend_ext sub-block status write plus the AGREE `view_friend_deps` lift to
  `view_scope_*_friend_deps` are untouched.

  Rejected: keeping a per-scope branch. After this change agent and user have identical visibility-write behavior, so a single line is the correct shape; a branch
  would only restate the same expression.

  ## Validation

  - `ruff check` on the two changed files → All checks passed.
  - `pytest tests/community/core/bot_public/test_bot_public_service.py` → 140 passed.
  - `pytest tests/community/adapters/http/openapi_v1/test_collaboration_bots_endpoints.py` → 3 passed.
  - Replaced the two `friend_check_in_strategy`-driven agent callback tests with three block-driven ones
  (`test_callback_agree_agent_writes_block_visibility_to_visibility`, `test_callback_agree_agent_uses_protected_from_block`,
  `test_callback_agree_agent_defaults_protected_without_block_visibility`). Fixtures deliberately keep `friend_check_in_strategy: "OPEN"` to prove it no longer drives
  the outcome.
  - Not run: the full backend suite and the pre-push gate (the latter is push-time only). `tests/community/endpoints/test_openapi_public_bcs.py` collects 0 under
  standard pytest — it uses `public_bcs_*` function names that don't match the `test_*` filter; pre-existing and unaffected by this change.

  ## Compatibility and risk

  Behavior change for the agent-scope publish callback. Previously, an agent-scope AGREE with `friend_check_in_strategy=OPEN` and no explicit request `visibility`
  wrote BCS `visibility=public`; now it writes `protected` (or the explicit request value). This is the intended effect of resolving visibility from the user's
  request rather than the strategy flag. No API path, request schema, response (`BcsPublishResult`), DB, or config change. Rollback: revert the single method edit.

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_public/services/bot_public_service.py
+++ b/src/backend/src/agentclaw/community/core/bot_public/services/bot_public_service.py
@@ -948,11 +948,9 @@ class BotPublicService(BotPublicServiceProtocol):
           (public_user_approval/public_agent_approval) 的 status 写成 last_operate
           (AGREE/DISAGREE/CANCEL)。
         - AGREE 时同时翻 BCS 可见性字段 (_BCS_VISIBILITY_FIELD_BY_SCOPE):
-            - public_scope=user → user_visibility = block.visibility (block 存的请求值,
-              缺省 protected);
-            - public_scope=agent → visibility = "public" if
-              BCS friend_check_in_strategy=="OPEN" else "protected" (BCS top-level
-              friend_check_in_strategy; agent 要考虑它, user 不考虑)。
+            - user → user_visibility, agent → visibility, 都取 block.visibility
+              (审批时存的请求值, 缺省 protected) —— 一律以请求值为准, 不再同步查询
+              BCS top-level friend_check_in_strategy 来推导 agent 的目标可见性。
         所有变更合一 PATCH (friend_ext + 可见性字段) 回。BCS 跳过 (非 prod/pre 或
         凭据空) 时静默返回。
         """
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM.

PR #1953, opened 2026-09-07
fix(bcs): resolve historical conversation IDs from session metadata

RaymondHX:
## Problem
After `/new` rolls a DingTalk conversation over to a new BCS session, `bcs-cli channel conversation-id --session <old-session>` returns no results even though the old session retains its conversation ID in metadata.

## Solution
When no stored conversation mappings exist for the requested session, resolve the historical route from `meta.channel` and filter by its source channel. Existing mappings retain precedence. Historical lookup works without the original binding and does not recreate mappings or change outbound routing. Use the fallible `SessionRepoPort::try_get` path to propagate database query and row decoding errors instead of reporting empty results. Preserve the legacy `get` interface for existing callers. Document the additive Service API behavior; HTTP authorization and response shape remain unchanged.

## Validation
- `cargo test --manifest-path src/bcs/Cargo.toml -p bcs-channel -p bcs-session-store --quiet`: 139 passed, including rollover, metadata fallback, query and decoding failures, and memory/SQL repository contracts.
- `cargo test -p bcs-session --quiet`: 27 passed.
- `cargo test -p bcs-http --test session_files_member_gate_contract channel_conversation_lookup --quiet`: 4 passed.
- All listed tests passed on the follow-up error-propagation fix.
- Full workspace and Singlebox E2E suites were not run locally. The previous PR head passed BCS unit and E2E CI; Singlebox Coverage failed. CI must validate the updated head.
- Commit and push hooks skipped with `--no-verify` per the repository publishing workflow; tests ran explicitly.

## Compatibility and risk
No CLI argument, response schema, database schema, or configuration changes. Requires deploying the BCS service update. Historical results use session update time as `last_active_at`; incomplete metadata yields an empty result. Reverting this change restores mapping-only lookup.

### DIFF
--- a/src/bcs/crates/service-api/bcs-service-api/src/application/channel.rs
+++ b/src/bcs/crates/service-api/bcs-service-api/src/application/channel.rs
@@ -176,6 +176,10 @@ pub trait ChannelService: Send + Sync {
         channel_type: Option<ChannelType>,
     ) -> Result<Vec<ChannelBinding>, ChannelUseCaseError>;
     /// Query the channel conversation mappings associated with a BCS session.
+    /// When no stored mappings remain, return the historical `meta.channel` route,
+    /// filtering by its `source` instead of requiring the binding to still exist.
+    /// Missing source/binding/conversation IDs yield no result. Historical routes
+    /// use the session update time as `last_active_at` and never restore live mappings.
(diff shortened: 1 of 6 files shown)


questchaos (approved):
LGTM

PR #1954, opened 2026-09-07
chore(bcs): stop tracking bcs-config-prod.toml

zh-bupt:
## Problem

`src/bcs/configs/bcs-config-prod.toml` is deployment-local configuration but is tracked in the repository. It was added to `.gitignore`, however gitignore rules do not take effect for files already tracked, so local prod config changes keep showing up in git status and risk being committed.

## Solution

- Add `src/bcs/configs/bcs-config-prod.toml` to `.gitignore` (alongside the already-ignored `bcs-config.toml`).
- Remove the file from the git index with `git rm --cached`; the local working-copy file is preserved.

## Validation

- `git ls-files src/bcs/configs/bcs-config-prod.toml` returns nothing after the change; local file still present on disk.
- No code paths read the tracked file at build time from git; deployment provides its own config.

Fresh clones will no longer contain `bcs-config-prod.toml`; deployments must supply it (same treatment as the already-ignored `bcs-config.toml`). Rollback: revert the commit.


questchaos (approved):
LGTM

carolynli (approved):
LGTM.

PR #1956, opened 2026-09-07
fix(bcs): write MySQL-compatible gmt_modified timestamps in group store

zh-bupt:
## Problem

On MySQL deployments, every group mutation transaction (e.g. `remove_group_participant` → `GroupServiceImpl::remove_participant`) fails with:

```
ERROR 22007 (1292): Incorrect datetime value: '2026-09-04T13:24:20.462Z' for column 'gmt_modified'
```

`bcs-group-store` formats the `gmt_modified` bind parameter as an RFC3339 string (`T` separator + `Z` suffix). SQLite stores the column as TEXT so it never fails locally, but the MySQL `bcs_groups.gmt_modified` column is `timestamp`, which rejects this format.

## Solution

`db_timestamp_from_millis` now emits `YYYY-MM-DD HH:MM:SS.mmm` (UTC) instead of RFC3339:

- MySQL `timestamp`/`datetime` columns parse this format directly.
- SQLite read paths (`strftime('%s', gmt_modified)` via `DbSqlFlavor::unix_ts`) parse both the new format and existing RFC3339 rows, so no migration is needed for stored data.
- Matches the existing precedent in `bcs-event-store/src/timestamp.rs`.

This single function feeds every `gmt_modified` binding in the group-store mutation transaction (patch/status/participant add-remove/mode/routing-policy/service-spec paths), so all group updates on MySQL are fixed at once.

## Validation

- New unit test `test_db_timestamp_from_millis_is_mysql_compatible` locks the format (no `T`, no `Z`).
- `cargo test --package bcs-group-store`: 87 tests pass.
- `cargo test` for bcs-group, bcs-session, bcs-app-group, bcs-app-session: all pass.
- `cargo check --workspace --all-targets`: clean.
- Not run: live MySQL instance verification (no MySQL environment locally); the format is the one MySQL documents for datetime literals.

Write-format-only change; read paths on both backends accept old and new values. Rollback: revert the commit (MySQL writes would break again).

### DIFF
--- a/src/bcs/crates/services/bcs-group-store/src/lib.rs
+++ b/src/bcs/crates/services/bcs-group-store/src/lib.rs
@@ -19,7 +19,7 @@ use bcs_event_store::{
     EventAppendTransactionPlan, GroupDeletionEventTransactionPlan,
     GroupProvisioningEventTransactionPlan,
 };
-use chrono::{SecondsFormat, TimeZone, Utc};
+use chrono::{TimeZone, Utc};
 use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::RwLock;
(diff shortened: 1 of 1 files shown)


RaymondHX (approved):
LGTM

PR #1959, opened 2026-09-07
fix(agent): ensure taskguard extension directory is owned by admin

fudp:
## Problem
TaskGuard workflow pack synchronization fails at runtime because /opt/openclawExt/taskguard/packs/ is owned by root, while the OpenClaw gateway runs as the admin user (uid 10001). The plugin cannot write downloaded packs into the directory.

## Solution
Add chown -R admin:admin /opt/openclawExt/taskguard in the runtime stage of docker/agent/avernet.dockerfile, right after the admin user is created. This gives the admin user write access to the packs directory and any future runtime-writable subdirectories.

## Validation
- Local docker build reaches the new RUN layer without error.
- After container start, ls -ld /opt/openclawExt/taskguard shows admin:admin ownership.
- Pack synchronization no longer fails with EACCES.

## Related issues
N/A


fudp:
Closing this PR as requested.

PR #1961, opened 2026-09-07
fix(claude-code): repair local skill file IO and storage addresses

FreddieSun:
## Problem

Community Claude Code Local Skill uploads call Node relay file RPCs that do not exist. Its Backend also derives private uploads from the shared OpenClaw repository root, so a file-only repair can store a package outside the existing Claude Skill layout.

## Solution

Implement the Claude Code file port with local filesystem I/O in the shared Engine/runtime container. Preserve binary content, recursive directory listings, filesystem errors and symlink-safe cleanup. Use the existing Engine layout to permit source files and active Skill links, while keeping SDK configuration outside that active-root permission.

Resolve Legacy Local Skill roots through a narrow `LocalSkillStorageResolver` Plugin API consumed by the existing package factory. The default provider preserves configured addresses; the community provider selects the existing Claude Code layout. Public Skill business logic and `WorkspacePathFactory` gain no deployment-profile flag. Package creation, reopening and request-scoped services use the same root; existing locator/device adapters and symlink mapping remain in place. Pool resolution retains precedence.

Historical locators are not rewritten. Incompatible locators fail the existing package-root checks before replacement writes. This change does not migrate old packages or change their persisted identity.

## Validation

- Standards and Spec reviews passed after fixing their findings: real DI consumer coverage and active-link cleanup permissions.
- Targeted Backend factory/upload/contract/architecture tests and Engine file/adapter tests passed. New Plugin API type checks passed; factory-wide Pyright reports only the same two errors reproduced at the base revision.
- Real Backend package factory and storage through Engine HTTP: nested/binary create, same-name replacement, metadata-failure rollback, mapping publish/verify and active-link unlink preserving source content passed. Repository/device transport boundaries use test doubles; active root is pre-created, so this is not fresh-container initialization or API-level activation proof.
- Enterprise Claude Code file regression: 9 tests passed. OCB composition and its pinned public gitlink were inspected; enterprise image/deployment compatibility remains unverified.
- Final full Backend/Engine gates and GitHub CI are running for the narrowed change; results will be updated before handoff.
- Cross-service lifecycle regression is manually runnable with `bash scripts/ci/claude_code_file_lifecycle.sh`. GitHub workflows are unchanged.

## Compatibility and risk

Requires Engine and Claude runtime to share the filesystem. Chat relay behavior, shared Skill repository synchronization, other engines' configured roots and existing Pool state are preserved. Upload remains inactive until explicitly activated.

No Bot creation metadata, ACK arguments, Dockerfile/startup changes, cwd changes, general workspace/identity/config address redesign or new symlink protocol are included. No merge or cloud deployment is performed.

## Spec

[Claude Code Local Skill file I/O and storage addresses](docs/superpowers/specs/2026-09-07-claude-code-local-files-spec.md)

## Related issues

Alternative to the unimplemented relay backup proposal #1955; that issue remains open as a backup.


src/engine/src/engine/community/plugins/claude_code/_file.py
```diff
@@ -1,56 +1,123 @@
...
+            if not root.exists():
+                raise FileNotFoundError("Directory does not exist")
+            if not root.is_dir():
+                raise NotADirectoryError("Path is not a directory")
+            excluded = exclude_dirs or set()
+            entries: list[dict] = []
+
+            def visit(directory: Path) -> None:
+                for child in sorted(directory.iterdir()):
+                    checked = self._file_path(str(child))
```
totalfrank: **Bug:** `visit()` in `file_list_dir` calls `self._file_path(str(child))` unguarded for every directory entry. `_file_path` calls `target.resolve()` and raises `PermissionError` if the resolved (symlink-followed) path isn't under one of `self._file_roots`. A dangling symlink whose target also doesn't resolve under a root, or a symlink an agent created via its own bash tool that points outside the configured roots, will make this raise and propagate all the way out of `list_files()`/`file_list_dir()` uncaught — failing the *entire* listing for that directory, not just skipping/flagging the one bad entry.

**Fails when:** A workspace directory (which the Claude Code agent can write to via arbitrary bash, e.g. `ln -s /etc some_dir/link`) contains one symlink pointing outside the configured engine roots (or a dangling symlink whose target simply doesn't exist). Listing that directory — even non-recursively — then throws `PermissionError`/`FileNotFoundError` instead of returning the other legitimate entries. This contradicts the port's own doc contract ("Recursive listings never follow directory symlinks" implies such entries are still listed, just not recursed into) and isn't covered by the existing symlink tests, which only exercise write/remove escape, not listing.

**Fix:** Guard the per-child check and treat an out-of-root/unresolvable entry as non-fatal (e.g. catch `OSError`/`PermissionError` from `self._file_path(str(child))` and either skip the entry or report it without following/statting it), rather than letting one bad symlink abort the whole directory listing:
```python
def visit(directory: Path) -> None:
    for child in sorted(directory.iterdir()):
        try:
            checked = self._file_path(str(child))
            is_dir = checked.is_dir()
            size = 0 if is_dir else checked.stat().st_size
        except OSError:
            is_dir = False
            size = 0
        ...
```


---

totalfrank (changes requested):
Reviewed `0e4c7a27a3d` (34 files, local Claude Code file I/O + workspace startup resolution). One bug found: `file_list_dir` in `_file.py` can crash an entire directory listing on a single symlink entry that resolves outside the configured engine roots (or is dangling).

---

totalfrank (dismissed):
LGTM — reviewed `5b0abe9` (27 files: local filesystem file I/O port for Claude Code, `LocalSkillStorageResolver` plugin API, DI wiring, and tests). No new bugs found in this round.

Note: my earlier finding on the unguarded `self._file_path(str(child))` call in `_file.py`'s `file_list_dir`/`visit()` (symlink escaping configured roots aborts the whole listing) is still unaddressed in this push — not re-raising here since it's already tracked on the open thread.

---

src/engine/src/engine/community/plugins/claude_code/_file.py
totalfrank: Fixed in `6887299`: `file_list_dir`'s `visit()` now lstats each child first, and only treats an out-of-root or dangling symlink as non-fatal (falls back to link-only metadata) instead of letting the exception abort the whole listing; non-symlink errors still propagate as before. New test `test_listing_keeps_dangling_and_outside_links_without_following_targets` covers both the dangling and outside-root cases for recursive and non-recursive listing. Resolving.

---

totalfrank (approved):
LGTM — reviewed `6887299` (3 files changed since last round: `_file.py`, `plugin_api/claude_code/file.py`, and their tests). This push fixes the previously-reported bug: `file_list_dir` no longer aborts the whole directory listing when one child is a dangling or out-of-root symlink — it now lstats first and falls back to link-only metadata for that entry, re-raising for any other error. New test covers both dangling and outside-root symlinks for recursive and non-recursive listing, plus that read/remove semantics on those entries are unchanged. No new bugs found.

---

PR #1962, opened 2026-09-07
fix(backend): retry transient BaaS skill sync

FreddieSun:
## Problem
BaaS restart publish can arrive before the Claude Code engine accepts requests. The resulting 502 prevents the Skill symlink clean/bind operation from converging, leaving the restart marker for a later publish.

## Solution
Retry only BaaS Skill symlink requests that fail with 502, 503, 504, or a request-level connection failure. Use bounded 1s, 2s, and 4s delays after the first attempt. Other failures remain immediate failures; exhausted retries preserve the existing marker behavior.

## Validation
- `uv run pytest tests/community/core/devices/test_baas_device_sync.py tests/community/core/bot_management/services/test_restart_authorization_refresh.py -q` (44 passed)
- `uv run pytest -q` (17,744 passed, 60 skipped)
- `uv run ruff check src/agentclaw/community/core/devices/services/baas_device_sync.py tests/community/core/devices/test_baas_device_sync.py`
- Standards/Spec two-axis review: no remaining findings.

## Compatibility and risk
No public API, schema, configuration, or non-BaaS provider behavior changes. At most, BaaS Skill symlink delivery waits 7 seconds before the existing failure and restart-marker fallback applies.

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_device_sync.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_device_sync.py
@@ -7,6 +7,7 @@ behavior without making the service profile-aware.
 """
 from __future__ import annotations
 
+import time
 from typing import Any, Optional
 
 from agentclaw.community.core.devices.services.baas_invoke_transport import BaasTransport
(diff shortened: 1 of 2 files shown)


totalfrank (approved):
LGTM — reviewed `e40ae85` (2 files, retry-with-backoff for BaaS Skill symlink requests plus tests). Traced the `_post_symlink_request` loop against the `(1, 2, 4, None)` delay/attempt schedule: non-retryable status codes and connection errors raise immediately, retryable ones sleep and retry, and the final attempt (`delay_seconds is None`) always re-raises rather than looping past the bound — matches the new tests' call counts and sleep sequences. The `symlinks_count == 0` vs. non-empty branches still compute the same `path`/`body` as before the refactor, and the outer exception handling in `sync_symlinks` is unchanged. No bugs found in the new code.

---

PR #1964, opened 2026-09-07
fix(backend): project default MCPs in SkillSet listing

FreddieSun:
## Problem

`GET /openapi/v1/bots/{bot_id}/skill-sets/{set_id}/mcps` returned only persisted `ac_skill_set_mcp` rows. A System Default SkillSet therefore appeared empty when its MCPs came from the Engine/Template `_defaults` policy. The existing Default projection also failed to hide an excluded MCP when that MCP had a persisted Default membership row.

## Solution

- Route Default Set reads through the established compatibility projection: persisted Default membership plus Engine/Template defaults, minus the addressed Bot's exclusions.
- Keep ordinary Set listing on the canonical membership repository.
- Keep MCP permission inspection and permission application membership-scoped so this read fix cannot apply for every platform default MCP.
- Share the same projection helper with `/skill-sets/resources` and document the Default Set exception in the Phase 2 contract.

## Validation

- `169 passed`: SkillSet management, legacy Default projection, MCP defaults, OpenAPI SkillSet adapter, and architecture tests.
- `3 passed`: live endpoint-framework cases for platform defaults, platform-default exclusions, and persisted-default exclusions.
- `ruff check`: passed for every changed Python file.
- changed-file Python SAST: passed.
- Standards/Spec review: no remaining findings.
- Full Community suite reached `17751 passed, 60 skipped`; its single failure was the 1000-line architecture cap after the first implementation. The helper was moved to the existing compatibility seam, and the failed architecture gate plus all affected suites were rerun successfully.

## Compatibility and risk

No database or response-schema change. Ordinary SkillSets and MCP permission operations retain their previous behavior. The pinned Gateway OpenAPI artifact is unchanged because the wire shape is unchanged; regenerating it currently includes unrelated repository-wide artifact drift. OCB `dev` was checked at gitlink `8aaec59e`; no Corp override of this path exists, so OCB only needs the normal Avernet gitlink update after merge.

### DIFF
--- a/src/backend/specs/2026-08-20-skill-capability-upgrade/spec.md
+++ b/src/backend/specs/2026-08-20-skill-capability-upgrade/spec.md
@@ -503,7 +503,7 @@ Canonical 前缀：`/openapi/v1/bots/{bot_id}/skill-sets`。
 | POST | `/skill-sets/{set_id}/activate` | 原子激活全部 Skill/MCP，一次 reconcile |
 | POST | `/skill-sets/{set_id}/deactivate` | 原子停用全部 Skill/MCP，一次 reconcile |
 | GET | `/skill-sets/resources` | 所有 Set 的 MCP/Default CLI 聚合 |
-| GET | `/skill-sets/{set_id}/mcps` | MCP Membership |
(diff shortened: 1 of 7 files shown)


totalfrank (approved):
LGTM.

PR #1830, opened 2026-09-02
chore(skill-center): instrument MCP projection timing

FreddieSun:
## Problem

Desktop Bot MCP add operations synchronously rebuild effective MCP state and project it to BaaS/Passport. Existing logs expose only broad projection duration, so the dominant Effective MCP computation cannot be attributed safely.

## Solution

- Add duration and cardinality logs for Default-policy resolution, exclusions, active Skills, installed MCPs, and non-default MCP enrichment.
- Add Projector sub-stage timings for identity, effective MCP collection, Passport CLI reads, Installation MCP reads, and final capability resolution.
- Preserve every query, write, ordering, and runtime call; this PR is observability-only.

## Validation

- uv run ruff check src/agentclaw/community/core/skill_center/services/skill_set_service.py src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py tests/community/core/skill_center/test_bot_runtime_projector_timing.py tests/community/core/skill_center/services/test_skill_set_service.py
- uv run pytest -q tests/community/core/skill_center/test_bot_runtime_projector_timing.py tests/community/core/skill_center/services/test_skill_set_service.py (32 passed)
- uv run pytest -q tests/community/architecture/test_module_boundaries.py (3 passed)

## Compatibility and risk

No API, DDL, cache, or runtime-delivery behavior changes. Added INFO logs only.

PR #1965, opened 2026-09-07
chore(skill-center): add actionable MCP projection timing logs

FreddieSun:
## Problem

Desktop Bot MCP mutations synchronously rebuild effective MCP state and project it to device and Passport consumers. Existing logs expose only broad projection duration, so slow Default policy, Installation, identity, Passport, or final resolution stages cannot be attributed. The original delivery PR #1830 was closed because its old required-check state no longer converged.

## Solution

- Add duration, outcome, and stage-local cardinality logs for Default capability lookup, active Sets, exclusions, active Skills, installed MCPs, and non-default enrichment.
- Add Projector timings for identity resolution, effective MCP collection, Passport CLI reads, Installation MCP reads, and final capability resolution.
- Record the exact failing substage before re-raising the original exception.
- Preserve query order, writes, Runtime calls, response contracts, and projection behavior.

## Validation

- `uv run ruff check` on the four changed files: passed.
- Focused Skill Center tests: 37 passed.
- Architecture module-boundary tests: 3 passed.
- Full Avernet Backend suite: 17,749 passed, 60 skipped.
- OCB `origin/dev@642358c13e` with `ocb-public` temporarily pointed to this commit: Corp compose equivalence, MCP adapter, DI binding parity, and Service API conformance: 61 passed.
- Standards/Spec dual review completed. The initial error-stage and cumulative-cardinality findings were fixed; both axes passed on re-review.

## Compatibility and risk

No API, DDL, cache, ordering, or Runtime-delivery behavior changes. The change adds INFO logs only. OCB validation proves the current Corp composition and contract tests accept the updated community implementation; it is not deployment or live-environment proof.

## Related issues

Supersedes closed PR #1830.

### DIFF
--- a/src/backend/src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py
+++ b/src/backend/src/agentclaw/community/core/skill_center/services/bot_runtime_projector.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import time
 from collections.abc import Mapping, Sequence
-import time
 
 from injector import inject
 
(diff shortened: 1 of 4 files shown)


totalfrank (approved):
LGTM.

PR #1969, opened 2026-09-07
feat(workflow): Add cross-instance workflow run queries and retry options support

wienyang:
Description
Enable bots to query workflow runs and execution logs across instances through shared internal APIs. Queries enforce bot and owner scope and support pagination, with optional node and log-level filters.
Move the retry request repository and routes into the shared workflow package. Preserve retry execution options in the mailbox and prevent older consumers from claiming requests whose options they cannot process.
Includes a database migration adding flow_retry_requests.options_json and an index on flow_runs(origin_bot_id, id).
Validation: all 161 workflow and shared-package tests passed.


3h-william (approved):
ok

PR #1976, opened 2026-09-07
fix(engine): activate Claude Code skills through local symlinks

FreddieSun:
## Problem

Community Claude Code can upload a local Skill successfully while leaving it undiscoverable: the bulk Skill adapter drops mappings/cleanup options, calls unsupported Relay RPCs, and turns the returned error into an empty successful result. The mapping API also raises FileNotFoundError when `.claude/skills` has not been created yet.

## Solution

- Carry complete bulk requests through the private Claude Code port and reconcile symlinks in the Engine filesystem, within configured roots. Initialize directories on activation; report actual created/updated/kept/removed results. Validate sources and conflicts before mutation, preserve ordinary files, and retain the previous link if replacement fails.
- Treat a missing active directory as an empty read-only inventory; initialize it during valid STRICT/BEST_EFFORT mapping publication.
- Replace fake Relay-success tests with HTTP/filesystem regression tests and stateful local-plugin conformance tests.

This change is limited to Engine activation. It does not change Backend policy, cwd, Bot creation, Docker/startup, or GitHub workflows. The separate Backend empty-set cleanup that hardcodes the OpenClaw directory remains outside this repair.

## Validation

- Engine full suite: 2,565 passed, 5 intentionally deselected by the existing community CI script.
- Final activation/conformance regression suite: 41 passed, including additional boundary cases added after the full run began.
- Combined coverage gate: 93.42% overall, 99.37% changed lines (473/476).
- OCB Corp Claude Code Skill/Pool tests: 10 passed against OCB origin/dev `642358c13`, using the changed public helpers.
- Standards and Spec reviews: findings fixed; both re-reviews passed.
- Review follow-ups: preserve selected/unselected package-internal symlinks; reject target and cleanup symlink ancestors before resolution (including across requests).
- GitHub CI: pending on latest commit.

## Compatibility and risk

Public HTTP request/response DTOs are unchanged. The community-only ClaudeCodeSkillsPort bulk signatures now require the request payload; its production/local implementations and adapter callers are updated together. OCB assembles its own ClaudeCodeSkillsService and does not consume this private port. Shared mapping helpers are covered by regression and Corp tests. OCB inspected gitlink: `8aaec59ee33d93d54c2b08e379fa9ae18e4d696d`; no gitlink bump or deployment is included. No architecture waiver is needed.

## Spec

`docs/superpowers/specs/2026-09-07-claude-code-skill-activation-spec.md`

### DIFF
--- a/src/engine/src/engine/community/plugins/claude_code/symlinks.py
+++ b/src/engine/src/engine/community/plugins/claude_code/symlinks.py
@@ -73,7 +73,9 @@ class LocalSkillSymlinks:
         return result
 
     def clean(self, params: dict) -> dict:
-        directories = [self._path(raw) for raw in params.get("directories", [])]
+        directories = [self._path(raw, link=True) for raw in params.get("directories", [])]
+        if any(directory.is_symlink() for directory in directories):
+            raise ValueError("Skill cleanup directory must not be a symlink")
         removed: list[str] = []
         scanned = 0
         for directory in directories:
--- a/src/engine/src/engine/community/plugins/claude_code/symlinks.py
+++ b/src/engine/src/engine/community/plugins/claude_code/symlinks.py
@@ -14,6 +14,8 @@ class LocalSkillSymlinks:
         path = Path(raw)
         if not raw.strip() or not path.is_absolute() or ".." in path.parts:
             raise ValueError("Skill path must be absolute without parent traversal")
+        if link and any(parent.is_symlink() for parent in path.parents):
+            raise ValueError("Skill target parent must not traverse a symlink")
         resolved = path.parent.resolve() / path.name if link else path.resolve()
         if not any(resolved.is_relative_to(root) for root in self.roots):
             raise ValueError("Skill path is outside the configured engine roots")
--- a/src/engine/src/engine/community/plugins/claude_code/symlinks.py
+++ b/src/engine/src/engine/community/plugins/claude_code/symlinks.py
@@ -1,4 +1,5 @@
 """Local filesystem implementation of the existing bulk Skill symlink contract."""
+from collections.abc import Iterator
 from pathlib import Path
 from typing import Any
 from uuid import uuid4
@@ -99,8 +100,18 @@ class LocalSkillSymlinks:
             for item in params.get("symlinks", [])
         ]
         desired = {self._path(item["target"], link=True) for item in mappings}
-        stale = [entry for entry in sorted(base.rglob("*"))
-                 if entry.is_symlink() and entry not in desired]
+        def active_links(directory: Path) -> Iterator[Path]:
+            # A real Skill directory owns its contents, including symlinks.
+            # Never traverse packages, even when absent from this request.
+            if (directory / "SKILL.md").is_file():
+                return
+            for entry in sorted(directory.iterdir()):
+                if entry.is_symlink():
+                    yield entry
+                elif entry.is_dir():
+                    yield from active_links(entry)
+
+        stale = [entry for entry in active_links(base) if entry not in desired] if base.exists() else []
         result = self.sync({"symlinks": mappings, "clean_target_dir": False})
         for entry in stale:
             entry.unlink()
--- a/src/engine/src/engine/community/plugins/claude_code/symlinks.py
+++ b/src/engine/src/engine/community/plugins/claude_code/symlinks.py
@@ -41,6 +41,8 @@ class LocalSkillSymlinks:
             if target.exists() and not target.is_symlink():
                 raise RuntimeError(f"Skill target is occupied: {target}")
             desired[target] = source
+        if any(other in target.parents for target in desired for other in desired):
+            raise ValueError("Skill targets cannot contain another requested target")
         result: dict[str, Any] = {"total": len(desired), "created": [], "updated": [], "kept": [], "removed": []}
         for target, source in desired.items():
             target.parent.mkdir(parents=True, exist_ok=True)
--- a/src/engine/src/engine/community/plugins/claude_code/symlinks.py
+++ b/src/engine/src/engine/community/plugins/claude_code/symlinks.py
@@ -1,0 +1,105 @@
+"""Local filesystem implementation of the existing bulk Skill symlink contract."""
+from pathlib import Path
+from typing import Any
+from uuid import uuid4
+
+
+class LocalSkillSymlinks:
+    def __init__(self, roots: tuple[Path, ...], base_dir: Path) -> None:
+        self.roots = roots
+        self.base_dir = base_dir
+
+    def _path(self, raw: str, *, link: bool = False) -> Path:
+        path = Path(raw)
+        if not raw.strip() or not path.is_absolute() or ".." in path.parts:
+            raise ValueError("Skill path must be absolute without parent traversal")
+        resolved = path.parent.resolve() / path.name if link else path.resolve()
+        if not any(resolved.is_relative_to(root) for root in self.roots):
+            raise ValueError("Skill path is outside the configured engine roots")
+        return resolved
+
+    def sync(self, params: dict) -> dict:
+        desired: dict[Path, Path] = {}
+        # Validate the whole request before changing any existing links.
+        for item in params.get("symlinks", []):
+            source = self._path(item["source"])
+            target = self._path(item["target"], link=True)
+            if target in self.roots or target == source:
+                raise ValueError("Skill target cannot replace a root or its source")
+            if target in desired:
+                raise ValueError(f"Duplicate Skill target: {target}")
+            if not source.is_dir() or not (source / "SKILL.md").is_file():
+                raise RuntimeError(f"Skill source is missing or unreadable: {source}")
+            try:
+                with (source / "SKILL.md").open("rb") as skill_file:
+                    skill_file.read(1)
+            except OSError as error:
+                raise RuntimeError(f"Skill source is unreadable: {source}") from error
+            for parent in target.parents:
+                if parent.exists() and not parent.is_dir():
+                    raise RuntimeError(f"Skill target parent is occupied: {parent}")
+            if target.exists() and not target.is_symlink():
+                raise RuntimeError(f"Skill target is occupied: {target}")
+            desired[target] = source
+        result: dict[str, Any] = {"total": len(desired), "created": [], "updated": [], "kept": [], "removed": []}
+        for target, source in desired.items():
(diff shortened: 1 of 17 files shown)


src/engine/src/engine/community/plugins/claude_code/symlinks.py
```diff
@@ -0,0 +1,111 @@
...
+
+        def absolute(raw: str) -> str:
+            path = Path(raw)
+            if not raw.strip() or path.is_absolute() or ".." in path.parts or path == Path("."):
+                raise ValueError("Skill path must be a non-empty relative path without traversal")
+            return str(base / path)
+
+        mappings = [
+            {"source": absolute(item["source"]), "target": absolute(item["target"])}
+            for item in params.get("symlinks", [])
```
totalfrank, line 104: **Bug:** `sync_relative`'s stale-detection scans the *entire* `base` tree recursively (`base.rglob("*")`) for symlinks not in `desired`, instead of scoping to the parent directories of the currently-desired targets (the way `sync()`'s own `clean_target_dir` cleanup does at line ~66-70, and the way the in-memory fake `LocalClaudeCodePluginImpl.skills_sync_symlinks` does by only ever touching its own `_skill_links` bookkeeping dict). Because this method forces `source` paths to also resolve under `base` (via `absolute()` above), any skill's source content that itself contains a symlink — anywhere under `base`, at any depth — gets treated as "stale" and deleted on the very next `/api/skills/symlink` call, even though it isn't a target this reconciliation manages at all.

**Fails when:** two skills are pooled under `base` (e.g. `base/sources/skill-a/`, `base/sources/skill-b/`), and `skill-a`'s own content includes a symlink (e.g. a shared-resource link inside the skill directory). A reconciliation call like `{"symlinks": [{"source": "sources/skill-a", "target": "active/skill-a"}]}` builds `desired = {base/active/skill-a}`. `base.rglob("*")` also walks into `base/sources/skill-a/`, finds the internal symlink there, sees it isn't in `desired`, and deletes it as "stale" — silently destroying part of the skill's own source content that the request never asked to touch.

**Fix:** scope `stale` to the parent directories of `desired` (mirroring `sync()`'s `clean_target_dir` loop), or track/only ever remove previously-managed target links (as the in-memory fake plugin does), rather than recursively scanning every symlink under `base`.

---

totalfrank (changes requested):
Reviewed `1437fe7` (17 files: new `symlinks.py` local FS reconciliation, adapter/plugin wiring, activation layout fixes, tests). One bug found: `LocalSkillSymlinks.sync_relative`'s stale-symlink cleanup scans the whole `base` tree recursively instead of scoping to the desired targets' parent directories, which can delete symlinks that are part of a skill's own source content rather than managed target links.

---

src/engine/src/engine/community/plugins/claude_code/symlinks.py
```diff
@@ -0,0 +1,111 @@
...
+class LocalSkillSymlinks:
+    def __init__(self, roots: tuple[Path, ...], base_dir: Path) -> None:
+        self.roots = roots
+        self.base_dir = base_dir
+
+    def _path(self, raw: str, *, link: bool = False) -> Path:
+        path = Path(raw)
+        if not raw.strip() or not path.is_absolute() or ".." in path.parts:
+            raise ValueError("Skill path must be absolute without parent traversal")
+        resolved = path.parent.resolve() / path.name if link else path.resolve()
```
totalfrank, line 19: **Bug:** A `target` whose parent path already contains a *pre-existing* skill symlink (created by an earlier, separate `sync`/`sync_relative` call) is silently resolved through that symlink and accepted, so the new link ends up written inside the resolved symlink's destination — another skill's own source/content directory — instead of under the intended active-link tree.

**Fails when:** Call 1: `sync({"symlinks":[{"source": "<pool>/skills/a", "target": "<active_root>/a"}]})` creates `<active_root>/a -> <pool>/skills/a`. Call 2 (separate request): `sync({"symlinks":[{"source": "<pool>/skills/b", "target": "<active_root>/a/nested"}]})`.

In `_path(..., link=True)` (line 16), `path.parent.resolve()` follows the existing `a` symlink, so `target` resolves to `<pool>/skills/a/nested` — a real path under `<pool>/skills/a`, not under `active_root`. Because `self.roots` also includes `layout.pool_root.parent.parent` (see `load_claude_code_file_roots` in `config.py`), that resolved path still passes the "inside configured roots" check on line 17-18. The parent-occupancy loop at lines 42-44 only rejects a parent that `exists()` and `not is_dir()`; `<pool>/skills/a` *is* a real directory (after the resolve already walked through the symlink), so it passes. `target.exists()` for `.../a/nested` is False, so the "occupied" check on line 45-46 is skipped too. Note this check runs on the *already-resolved* `target`, so it can no longer see that the path originally passed through a symlink — checking `parent.is_symlink()` there would not catch this case.

The mutation loop then does `target.parent.mkdir(parents=True, exist_ok=True)` (no-op) and `target.symlink_to(source, ...)`, physically creating `nested` **inside `<pool>/skills/a`** — skill `a`'s own source directory — polluting/corrupting it with an entry unrelated to that skill. This is reachable directly through the public `/api/skills/symlink/bindpath` endpoint via two ordinary sequential requests, no extra privilege beyond the existing capability guard.

**Fix:** The check needs to happen before symlinks are resolved away. In `_path`'s `link=True` branch, walk `path.parent`'s components against the raw (unresolved) path and reject if any intermediate segment is itself a symlink (a valid target parent chain should be made only of plain directories, never a pass-through skill link), rather than calling `path.parent.resolve()` directly and only re-checking directory-ness of the *already-resolved* result afterward.

---

totalfrank (changes requested):
Reviewed `6827613` (17 files, the new `LocalSkillSymlinks` local-filesystem symlink reconciliation and its callers). One bug found: a nested target under an existing skill symlink resolves through it and writes into another skill's source directory instead of the active-link tree.

---

src/engine/src/engine/community/plugins/claude_code/symlinks.py
FreddieSun: Fixed in 4a9564ade. The regression reproduces the reported deletion before the fix.

Relative reconciliation now stops at real directories containing SKILL.md and never follows symlink directories. This preserves internal links in both selected and unselected Skill packages, while still reconciling stale active links outside packages. Restricting cleanup only to the current targets' parents would leave stale links behind on empty-set or cross-directory reconciliation, so those existing semantics are retained.

The new HTTP regression checks a selected package, an unselected package, removal of an old active link, and a subsequent empty-set cleanup; all package-internal links remain readable. The activation/conformance suite passes (36 tests), and both incremental reviews passed. Full Engine/CI checks are running for this commit.

totalfrank (changes requested):
Reviewed `4a9564ad` (this push touches only `symlinks.py`'s `sync_relative` stale-detection, plus its regression test and the spec doc).

- The stale-detection fix is correct: `active_links` now stops at any directory containing `SKILL.md` and never recurses through a symlink, so package-internal symlinks (e.g. `sources/<skill>/assets/shared`) are preserved on both a selective sync and a subsequent empty-set cleanup. The new test `test_relative_reconcile_preserves_all_skill_package_symlinks` exercises exactly this and would fail on the prior `base.rglob("*")` implementation. No new bugs in this push.
- Previously reported issue in `_path` (`link=True` branch resolving through an existing symlink in an intermediate path segment, letting a second `sync` request write into another skill's source directory — https://github.com/inclusionAI/Avernet/pull/1976#discussion_r3948460100) is untouched by this commit and still reachable: `load_claude_code_file_roots()` in `config.py` still includes `pool_root.parent.parent` among the roots, so the resolved-through-symlink path still passes the roots check. Keeping this open pending a fix.

---

src/engine/src/engine/community/plugins/claude_code/symlinks.py
FreddieSun: Fixed in a2f93e65c. Target ancestor symlinks are now rejected on the original path before resolve(), including links created by earlier requests. The final target itself may still be a symlink and is safely kept/replaced. The in-memory implementation enforces the same boundary.

Added HTTP regressions for two sequential bindpath requests and two sequential relative-sync requests; both reproduced source-directory writes before the fix and now return 400 with the original link and source untouched. Added matching local-plugin conformance coverage. The focused suite passes (39 tests); incremental Standards/Spec re-reviews passed. Latest-head CI is running.

Related follow-up in c4e80de6f: cleanup now also rejects symlink directories/ancestors rather than following an active link into package content. HTTP and local regressions reproduce and prevent that deletion. The final focused suite passes 41 tests; changed-line coverage is 99.37%. Both incremental reviews passed.

totalfrank (approved):
LGTM — reviewed `c4e80de` (17 files, symlink activation reconciliation + adapter/port plumbing).

Both previously-reported bugs are verified fixed on this head:
- `sync_relative`'s stale-link scan now stops recursion at any real directory containing `SKILL.md` (`active_links()` in `symlinks.py`), so it no longer deletes symlinks that live inside a skill's own source content.
- `_path(..., link=True)` now rejects any symlink in `target.parent`'s *raw* (unresolved) ancestor chain before calling `resolve()`, so a target can no longer be silently written inside another skill's directory through a pre-existing active link from an earlier request.

No new bugs found in the rest of this round's diff (adapter/port param plumbing, `config.py`'s new `load_claude_code_skills_root`, the `layout_activation.py` directory-initialization additions, and the local in-memory plugin double).

---

totalfrank (approved):
LGTM — reviewed `a2f93e6` (17 files, symlink-based Skill activation for Claude Code).

Both bugs flagged in the previous round are fixed in this push and verified:
- `sync_relative`'s stale-detection no longer recurses into real Skill package directories (stops at any dir containing `SKILL.md`), so a package's own internal symlinks survive reconciliation of unrelated targets. Covered by `test_relative_reconcile_preserves_all_skill_package_symlinks`.
- `_path(..., link=True)` now rejects any symlink among the *raw* target's parent components before resolving, closing the path where a second, separate `sync`/`sync_relative` call could write through an already-activated skill symlink into another skill's source directory. Covered by `test_separate_request_cannot_write_through_existing_skill_link` (both absolute and relative variants).

I ran the full Engine suite locally against this commit (`uv run pytest src`): 2568 passed, 5 skipped, no failures — consistent with the PR's own validation numbers. No new bugs found in the changed code.

Out of scope (FYI): none.

---

PR #1977, opened 2026-09-07
fix(bot-config-manifest): repair mis-provisioned ac_bot_config_manifest_apply

totalfrank:
## Problem

`GET /openapi/v1/bots/{bot_id}/config-manifest/last-apply` fails with:

```
mysql.connector.errors.ProgrammingError: 1054 (42S22): Unknown column 'ac_bot_config_manifest_apply.apply_id' in 'field list'
```

`SHOW CREATE TABLE ac_bot_config_manifest_apply` in production returns the apply **lock**'s body (`holder_user_id`, `lock_token`, `UNIQUE KEY uk_manifest_apply_lock`) and no `apply_id`, `status`, `report`, `started_at` columns at all. The ORM and the repo DDL agree with each other; the database was provisioned with the lock DDL under the record table's name.

The root cause surfaced when re-provisioning: the DDL review standard (研发规范) that gates OceanBase provisioning refuses the canonical record DDL on three findings:

- `trigger` is a SQL keyword and cannot be a column name.
- `started_at` is DATETIME, which the standard forbids (with BIT, FLOAT, DOUBLE, ENUM, SET).
- `finished_at`, same.

So the record table could never have been created as declared, while the lock table (which has none of those) went through.

## Solution

Two parts, one PR:

**1. Make the record DDL pass the standard**, with the ORM moving in lockstep:

- Column renamed `trigger` → `apply_trigger`. The ORM maps it via `Column("apply_trigger", ...)` so the Python attribute, the pydantic record field and the API field all stay `trigger`. No caller changes.
- `started_at` / `finished_at` → `TIMESTAMP` with explicit `DEFAULT CURRENT_TIMESTAMP` / `NULL DEFAULT NULL`. The application fills both from `datetime.now()` and reads them back through the same session, so the TIMESTAMP round trip is the identity; the explicit defaults are what stop the server attaching an implicit `ON UPDATE CURRENT_TIMESTAMP`. The ORM keeps `DateTime`, as it already does for `gmt_*` over TIMESTAMP.
- Canonical DDL (`2026_08_31_bot_config_manifest_apply.sql`) and its header prose updated to say why.
- `test_manifest_ddl_contract.py`: the apply file leaves `_APPLICATION_SUPPLIED`; a new `test_apply_tables_pass_the_ddl_review_standard` holds both rules for this table. Parser now accepts `CREATE TABLE IF NOT EXISTS`.

**2. Repair migration** `2026_09_07_repair_misprovisioned_bot_config_manifest_apply.sql` for any environment carrying the mis-shaped table: detects it via `information_schema` (record table present, no `apply_id`), renames it into place as the lock when the lock is missing or aside to a quarantine name when not, then `CREATE TABLE IF NOT EXISTS` with the corrected DDL. Idempotent; nothing dropped.

Rejected alternative: keeping DATETIME and arguing with the standard. `skill_center/sql/2026_09_03_align_space_skill_timestamps_with_gmt.sql` already made the same DATETIME→TIMESTAMP move for the same reason.

Out of scope, flagged: `ac_manifest_content.fetched_at` is also DATETIME and will hit the same rule when that table is provisioned. Its value is UTC-normalised by the application, so it needs its own reasoning; left for a separate change.

## Validation

- `test_manifest_ddl_contract.py`: 5 passed locally (stdlib-only, run with `--noconftest`). The three original failures reproduce without the parser change.
- `ruff check` clean on the touched Python files.
- Column-by-column comparison of the production `SHOW CREATE TABLE` output against both bodies in the canonical DDL: it matches the lock table exactly.
- **Not run:** the migration and the corrected DDL against OceanBase. No MySQL/OceanBase in the sandbox. Production is being re-provisioned by hand from the corrected CREATE TABLE statements; the migration is for other environments.

## Compatibility and risk

Schema and ORM change together. The renamed column is only ever addressed through the ORM, so deploying the app without the new DDL (or vice versa) fails loudly on the first apply, not silently. No production rows exist in the record table (it never existed correctly), so there is nothing to backfill. Rollback: revert the PR and recreate the table from the previous DDL, which the standard will refuse.

Note: `REL20260908` rejects direct pushes (repository rule "changes must be made through a pull request"), hence this PR from a working branch into that release branch.

https://claude.ai/code/session_01SLdSEfH7LzYhxp3TrKj6qN

### DIFF
--- a/src/backend/tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py
+++ b/src/backend/tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py
@@ -65,7 +65,7 @@ def _tables(path: Path) -> dict[str, dict[str, str]]:
     )
     starts = [
         (match.group(1), match.start())
-        for match in re.finditer(r"CREATE TABLE\s+`(\w+)`", body)
+        for match in re.finditer(r"CREATE TABLE\s+(?:IF NOT EXISTS\s+)?`(\w+)`", body)
     ]
     assert starts, f"{path.name}: no CREATE TABLE found"
 
--- a/src/backend/tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py
+++ b/src/backend/tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py
@@ -30,12 +30,23 @@ _SQL_DIR = (
 
 #: Columns filled by the application, which must stay DATETIME. TIMESTAMP reads
 #: the bound naive value as session-local and converts it, so an instant the
-#: application already normalised to UTC (``fetched_at``) or took from
-#: ``datetime.now()`` (the apply times) is stored shifted by the session offset.
+#: application already normalised to UTC (``fetched_at``) is stored shifted by
+#: the session offset.
+#:
+#: The apply table's started_at/finished_at used to be listed here and are
+#: not any more: the DDL review standard refused them as DATETIME (see
+#: ``_DDL_STANDARD_FORBIDDEN_TYPES``), and their round trip through one
+#: session is the identity, so TIMESTAMP loses nothing the application reads.
 _APPLICATION_SUPPLIED = {
-    "2026_08_31_bot_config_manifest_apply.sql": ("started_at", "finished_at"),
     "2026_08_31_manifest_content.sql": ("fetched_at",),
 }
+
+#: The DDL review standard (研发规范) that gates provisioning on OceanBase.
+#: ac_bot_config_manifest_apply was refused on exactly these two rules -- a
+#: column named `trigger`, and DATETIME started_at/finished_at -- which is why
+#: the record table never existed in production while its lock table did.
+_DDL_STANDARD_FORBIDDEN_TYPES = {"bit", "float", "double", "datetime", "enum", "set"}
+_DDL_STANDARD_FILES = ("2026_08_31_bot_config_manifest_apply.sql",)
 
 #: Filled by the database itself, so TIMESTAMP's conversion is a no-op round
 #: trip and matches ac_bots. Every table here has both.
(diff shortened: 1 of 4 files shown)


src/backend/tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py
```diff
@@ -30,13 +30,24 @@
...
-    "2026_08_31_bot_config_manifest_apply.sql": ("started_at", "finished_at"),
     "2026_08_31_manifest_content.sql": ("fetched_at",),
 }
 
+#: The DDL review standard (研发规范) that gates provisioning on OceanBase.
+#: ac_bot_config_manifest_apply was refused on exactly these two rules -- a
+#: column named `trigger`, and DATETIME started_at/finished_at -- which is why
+#: the record table never existed in production while its lock table did.
+_DDL_STANDARD_FORBIDDEN_TYPES = {"bit", "float", "double", "datetime", "enum", "set"}
+_DDL_STANDARD_FILES = ("2026_08_31_bot_config_manifest_apply.sql",)
```
totalfrank: Non-blocking: `_DDL_STANDARD_FILES` only lists `2026_08_31_bot_config_manifest_apply.sql`, so `test_apply_tables_pass_the_ddl_review_standard` never runs against the new `2026_09_07_repair_misprovisioned_bot_config_manifest_apply.sql`, even though that file embeds its own full, independently-maintained copy of the same `CREATE TABLE` body (verified identical to the canonical file today). If a future edit changes one file's column types or reintroduces a keyword column name without updating the other, this test — whose whole purpose is to catch exactly that kind of drift ("what make a future 'tidy this up for consistency' edit fail loudly instead of silently corrupting an audit column") — will pass while the repair file has silently drifted out of DDL-standard compliance.

Suggested fix:
```suggestion
_DDL_STANDARD_FILES = (
    "2026_08_31_bot_config_manifest_apply.sql",
    "2026_09_07_repair_misprovisioned_bot_config_manifest_apply.sql",
)
```
(`test_apply_tables_pass_the_ddl_review_standard` indexes into `tables["ac_bot_config_manifest_apply"]` unconditionally, which both files declare, so no other change is needed.)

---

totalfrank (commented):
Reviewed commits d934d47, 5c87ec4, cf6c921 (full PR, first review pass) across `apply_models.py`, both `.sql` files, and `test_manifest_ddl_contract.py`.

Checked: correctness of the column rename (`trigger` attribute → `apply_trigger` physical column, verified no other caller references the old physical name), the DATETIME→TIMESTAMP change against the existing implicit-ON-UPDATE guard test, the repair migration's detect/rename/recreate logic (idempotency across all three starting states: correctly-provisioned, mis-shaped-with-no-lock, mis-shaped-with-lock), and DDL contract test coverage. Ran the (stdlib-only) DDL contract tests locally against the PR head — all 5 pass.

One non-blocking test-coverage gap, left as an inline comment.

---

totalfrank:
**Review round 1 — 1 finding, fixed.** Reviewed `cf6c921`; fix pushed as `ab58af3`.

- **Fixed:** `_DDL_STANDARD_FILES` now also lists `2026_09_07_repair_misprovisioned_bot_config_manifest_apply.sql`, so the repair migration's embedded `CREATE TABLE` body is held to the same keyword-name and forbidden-type rules as the canonical DDL. Verified the test still passes against both files (5 passed locally, stdlib-only run). Thread resolved.

Nothing disputed. CI was green on `cf6c921`; round 2 is open on `ab58af3`.

---

FreddieSun (approved):

totalfrank:
@claude review

Round 2 on `ab58af3`: the only change since your round-1 pass on `cf6c921` is the one-line fix you suggested, adding the repair migration to `_DDL_STANDARD_FILES` in `test_manifest_ddl_contract.py`. CI is green on this head.

---

PR #1980, opened 2026-09-07
feat(backend): Unify the return values of the connection interface

liaoxianhao:
…ssage interface to expose the metadata field


totalfrank (approved):
LGTM.

PR #1989, opened 2026-09-07
feat(aicoding/cron): engine_properties bag + reuse_session for agentTurn cron

panqiwp:
## Summary
Add an `engine_properties` bag to the cron task API so aicoding's agentTurn cron can reuse an existing conversation or start a fresh one each run (`reuse_session`, bool, default `true`).

## Changes
- `EngineProperties` schema with `reuse_session` (`extra="allow"` for future engine switches).
- `engine_properties` (opaque dict) added to create/update task requests and internal job models.
- Router passes it through on create/update and carries the existing value when other payload fields are updated.

## Scope
- Only aicoding consumes it (flatten/renest lives in the parent repo, PR'd separately).
- openclaw / claude_code / teclaw ignore it — no behavior change.

## Tests
Router passthrough + carry-over, models/schemas, and cross-engine adapter regression all pass.

### DIFF
--- a/src/engine/src/engine/community/api/cron/router.py
+++ b/src/engine/src/engine/community/api/cron/router.py
@@ -186,7 +186,13 @@ async def create_task(request: CreateTaskRequest):
             bot_id=request.bot_id,
             session_target="isolated",
             enabled=request.enabled,
-            notify=notify
+            notify=notify,
+            # 引擎特有参数配置
+            engine_properties=(
+                request.engine_properties.model_dump(exclude_none=True)
+                if request.engine_properties is not None
+                else None
+            ),
         )
 
         job = await cron_api.add_job(create_request)
@@ -234,6 +240,13 @@ async def update_task(task_id: str, request: UpdateTaskRequest):
                     update_data["payload"]["model"] = request.model
                 if request.runtime is not None:
                     update_data["payload"]["runtime"] = request.runtime
+                # engine_properties 不在 payload 内（由各引擎按需展平），而 relay 的
+                # cron.update 对 payload 是整体替换，因此这里须把现存 engine_properties
+                # 原样带回，避免改 message/timeout/model/runtime 时丢掉 reuse_session 等
+                # 引擎特有参数。仅当本次未显式提供 engine_properties 时才回带。对 openclaw /
+                # claude_code 等不使用 engine_properties 的引擎是 no-op（其值为 None）。
+                if request.engine_properties is None and existing_job.engine_properties:
+                    update_data["engine_properties"] = existing_job.engine_properties
             else:
                 raise HTTPException(status_code=404, detail="Task not found")
 
@@ -245,6 +258,9 @@ async def update_task(task_id: str, request: UpdateTaskRequest):
             if request.notify.user_ids is not None:
                 notify_patch.user_ids = request.notify.user_ids
             update_data["notify"] = notify_patch
+
+        if request.engine_properties is not None:
+            update_data["engine_properties"] = request.engine_properties.model_dump(exclude_none=True)
 
         if not update_data:
             raise HTTPException(status_code=400, detail="No fields to update")
--- a/src/engine/src/engine/community/api/cron/schemas.py
+++ b/src/engine/src/engine/community/api/cron/schemas.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 from typing import Optional
 
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, Field
 
 
 class NotifyRequest(BaseModel):
@@ -19,14 +19,18 @@ class NotifyRequest(BaseModel):
 
 
 class EngineProperties(BaseModel):
-    """引擎专属属性 bag。
+    """引擎特有参数配置。
 
-    各引擎只读取自己关心的字段；未声明的字段经 ``extra="allow"`` 原样保留，
-    这样 aicoding 后续新增引擎级开关时无需再改本层 schema。当前唯一使用者是
-    aicoding 引擎的 agentTurn 任务：``reuse_session`` 控制「是否复用已有会话」
-    （默认 True 复用已有会话；False=每次触发新开一个会话）。其它引擎忽略本对象。
+    每个引擎级开关都必须在此声明为显式具名字段（非开放 bag），确保契约有版本、
+    有校验、随变更一并评审。当前唯一字段：
+
+    - ``reuse_session``（aicoding agentTurn）是否复用已有会话：默认 True 复用；
+      False=每次触发新开一个会话。
+
+    消费方：corp aicoding 引擎（``corp/engines/aicoding``，仅在内部全量 checkout，
+    不在 GitHub 社区版内）。社区版引擎（openclaw / claude_code）不读取本对象——
+    这是显式 no-op-by-design，而非静默丢弃（详见 ``plugin_api/cron/README.md``）。
     """
-    model_config = ConfigDict(extra="allow")
     reuse_session: Optional[bool] = Field(
         default=None,
         description="（aicoding agentTurn）是否复用已有会话：默认 True 复用；False=每次触发新开会话",
--- a/src/engine/src/engine/community/api/cron/schemas.py
+++ b/src/engine/src/engine/community/api/cron/schemas.py
@@ -9,13 +9,28 @@ from __future__ import annotations
 
 from typing import Optional
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field
 
 
 class NotifyRequest(BaseModel):
     """通知配置请求"""
     enabled: bool = Field(default=False, description="是否启用通知")
     user_ids: Optional[list[str]] = Field(default=None, description="通知目标用户 ID 列表")
+
+
+class EngineProperties(BaseModel):
+    """引擎专属属性 bag。
+
+    各引擎只读取自己关心的字段；未声明的字段经 ``extra="allow"`` 原样保留，
+    这样 aicoding 后续新增引擎级开关时无需再改本层 schema。当前唯一使用者是
+    aicoding 引擎的 agentTurn 任务：``reuse_session`` 控制「是否复用已有会话」
+    （默认 True 复用已有会话；False=每次触发新开一个会话）。其它引擎忽略本对象。
+    """
+    model_config = ConfigDict(extra="allow")
+    reuse_session: Optional[bool] = Field(
+        default=None,
+        description="（aicoding agentTurn）是否复用已有会话：默认 True 复用；False=每次触发新开会话",
+    )
 
 
 class CreateTaskRequest(BaseModel):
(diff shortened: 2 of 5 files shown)


totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0
- Target branch: `dev` (integration branch — not a `REL*` release branch, so the release-branch gate does not apply here).
- Classification: feature (new `engine_properties` bag + `reuse_session` on the cron Plugin API / HTTP contract).
- Gate 0: **pass** (`dev` is the correct default target for new work). Full architecture review below still applies.

## Findings

### 1. `arch.rules.md` Rule 16 ("Changes Propagate") — Invariant — contract shipped with no implementation in this change, silently dead on the one in-repo consumer

- `src/engine/src/engine/community/plugin_api/cron/models.py:29,68,87` adds `engine_properties: Optional[dict[str, Any]]` to `CronJob`, `CreateJobRequest`, and `UpdateJobRequest` — a shared Plugin API contract (`CronService` Protocol, `src/engine/src/engine/community/core/cron/protocol.py`) implemented in-repo by `ClaudeCodeCronAdapter` (`src/engine/src/engine/community/core/adapters/claude_code/cron.py`), which is what the `"aicoding"` engine alias resolves to (`core/engine/naming.py:47`, `manager.py:169-180`).
- That adapter's wire builders never read the new field:
  - `_build_add_params()` (`core/adapters/claude_code/cron.py:203-213`) forwards only `name/schedule/payload/sessionTarget/enabled/delivery` to the relay — `request.engine_properties` is dropped.
  - `_build_update_patch()` (`core/adapters/claude_code/cron.py:215-243`) builds the same reduced key set — `request.engine_properties` is dropped.
  - `_job_from_dict()` (`core/adapters/claude_code/cron.py:92-124`) never sets `engine_properties` when parsing a relay job back — so `CronJob.engine_properties` is always `None` coming out of this repo's only implementation.
- That last point makes the router's own carry-over logic dead in production: `router.py:248-249` (`if request.engine_properties is None and existing_job.engine_properties: update_data["engine_properties"] = existing_job.engine_properties`) can only fire if `existing_job.engine_properties` is truthy, which `_job_from_dict` never produces. The new tests (`test_unrelated_update_preserves_existing_engine_properties`, `test_cron_router.py:344-351`) only pass because they mock `get_job` to return a job with `engine_properties` set directly — bypassing the real (missing) relay round-trip entirely, so they give false confidence about behavior that doesn't exist end to end (Rule 15).
- The PR body confirms this is intentional and explicit: *"aicoding 适配器的展平/归位逻辑（`reuse_session` ⇄ relay `reuseSession`）在父仓 teamclaw/ocb 的 `corp/engines/aicoding/cron.py`，另单独提 PR。"* — the actual consuming implementation lives in a different, closed/downstream repository, in an unlinked, not-yet-merged follow-up PR. Rule 16 requires a structural change to "identify affected consumers, affected implementations... compatibility status" as part of *this* change, and "merge is blocked when required impact analysis is absent." A prose note deferring the only implementation to an out-of-tree PR is not propagation — it means merging this to `dev` publishes a public HTTP/Plugin API field (`POST/PUT /api/cron ... engine_properties.reuse_session`) that any community-edition (`claude_code`/`aicoding`) user can set, get a `200`, and have it silently discarded forever, since this repository's own adapter has no path to ever pick it up without a further code change here.
- Required fix: either (a) land the `ClaudeCodeCronAdapter` wiring (`_build_add_params`/`_build_update_patch`/`_job_from_dict` reading/writing `reuseSession` on the relay wire format) in this same PR, with a conformance-level test that exercises the adapter (not just the router with a mocked `CronService`), or (b) if the real engine implementation is genuinely out-of-tree for this repository, keep the contract change out of `dev` until it lands together with a working in-repo (or explicitly documented no-op-by-design) implementation, per Rule 25's requirement that a declared contract have conformance tests all relevant implementations pass.

### 2. `arch.rules.md` Rule 1 ("The API Specification Is the Single Authority") — Invariant — open-ended `extra="allow"` bag on a public contract

- `src/engine/src/engine/community/api/cron/schemas.py:29` (`model_config = ConfigDict(extra="allow")` on `EngineProperties`) is documented at line 24 as deliberately allowing undeclared fields through so that "aicoding 后续新增引擎级开关时无需再改本层 schema" (future engine-level toggles need no further schema change). This institutionalizes exactly what Rule 1 forbids: "implementations must not introduce undocumented semantics that other components depend on," and the required check that "contract changes update specs/docs in the same change." An `extra="allow"` public request field is a standing invitation for engine-specific keys to flow through the HTTP boundary and the Plugin API `dict[str, Any]` bag with no schema, no validation, and no doc update ever required — each new key becomes a de facto, unversioned wire contract negotiated only in adapter code and PR prose (the `reuseSession` RFC-226 reference in the PR body being the only "spec" for the one field that exists today).
- Required fix: drop `extra="allow"`; require each engine-specific property to be declared as an explicit, named, optional field on `EngineProperties` (as `reuse_session` already is) so every addition goes through schema/doc review, matching how `payload`'s per-`kind` shape is otherwise handled elsewhere in this router.

### 3. `arch.rules.md` Rule 22 ("Context Boundaries Are Explicit") — Policy — Plugin API doc not updated

- `src/engine/src/engine/community/plugin_api/cron/README.md` documents `CronJob`/`CreateJobRequest`'s fields and "Change impact" (it was updated for the earlier `owner_id`/`bot_id` addition) but is untouched by this diff even though `engine_properties` is added to all three DTOs in the same file it describes. Minor on its own, but it compounds finding #1: there is now no in-repo document that says what `engine_properties` is for, who's expected to populate it, or that today no in-repo adapter does.

## Out-of-scope notes
- The `reuse_session` semantics themselves (default-True, per-agentTurn session reuse toggle) are a product/business decision and not reviewed here.
- `router.py`'s "kind defaults to agentTurn" and payload-shape handling are pre-existing and unchanged by this diff.

---

totalfrank (changes requested):
## Architecture Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0 — Target branch

Target: `dev` (not a `REL*` release branch) → Gate 0 does not apply; branch policy is satisfied trivially. Classification: feature (`feat(aicoding/cron)`) landing on the integration branch, which is the normal route.

### Scope note

The diff is deliberately narrow: it threads an opaque `engine_properties` bag from the HTTP schema (`api/cron/schemas.py`) through the Plugin API DTOs (`plugin_api/cron/models.py`), with no new `if engine_type == ...` branching anywhere and no engine-named modules under `core/`. I verified against `core/adapters/openclaw/cron.py` and `core/adapters/claude_code/cron.py` that neither adapter reads or forwards `engine_properties` (their `_build_add_params`/`_build_update_patch`/`_job_from_dict` whitelist fields explicitly), so the "zero behavior change for other engines" claim in the PR description holds. That part of the change is structurally sound.

Two findings below block approval.

---

### Findings

**1. `src/engine/src/engine/community/api/cron/router.py:229-250` — aicoding-relay-specific merge policy embedded in the shared, engine-agnostic HTTP adapter**

```python
if request.command is not None or request.timeout_secs is not None or request.model is not None or request.runtime is not None:
    existing_job = await cron_api.get_job(task_id)
    if existing_job:
        update_data["payload"] = existing_job.payload.copy()
        ...
        # engine_properties 不在 payload 内（由各引擎按需展平），而 relay 的
        # cron.update 对 payload 是整体替换，因此这里须把现存 engine_properties
        # 原样带回，避免改 message/timeout/model/runtime 时丢掉 reuse_session 等
        # 引擎特有参数。
        if request.engine_properties is None and existing_job.engine_properties:
            update_data["engine_properties"] = existing_job.engine_properties
```

`api/cron/router.py` is the delivery adapter shared by every engine (openclaw, claude_code, aicoding). This block exists solely to work around one specific fact about the **aicoding relay's** `cron.update` wire behavior ("整体替换" — whole-record replace), as stated in the PR description and the comment itself. That is domain/transport policy belonging to one engine, hard-coded into the shared adapter — a violation of **arch.rules.md Rule 7 (Invariant)**: "delivery adapters... must not own domain policy," and the Structural Review Red Flag "an adapter contains domain policy." Contrast with how `core/adapters/openclaw/cron.py::_build_update_patch` correctly keeps its own patch-assembly quirks local to the openclaw adapter rather than leaking them into the shared router.

This also isn't just a placement issue — it's an *incomplete* implementation of the very safety guarantee it's trying to provide: the carry-over only fires inside the `command`/`timeout_secs`/`model`/`runtime` branch. A request that updates only `schedule`, only `notify`, or only `name`/`enabled` (without also touching `engine_properties`) never reaches this code path, so on the assumption stated in the comment (relay replaces the whole record), those updates would silently drop `reuse_session` anyway. An inline, reverse-engineered assumption about an external, undocumented wire contract — applied inconsistently — is exactly the "hidden wire protocol" pattern the constitution's Rule 1 requires to be an explicit, documented, conformance-tested contract instead.

**Required fix direction:** Move the "preserve existing `engine_properties` across a partial update" responsibility into the aicoding engine's own `CronService` implementation (the adapter that actually knows the relay's replace semantics), the same way `core/adapters/openclaw/cron.py` and `core/adapters/claude_code/cron.py` each own their engine-specific patch assembly. The shared router should stay a thin, uniform pass-through of whatever `engine_properties` the caller supplied (or omitted), full stop.

**2. `src/engine/src/engine/community/plugin_api/cron/models.py` — Plugin API contract change with no propagation/doc update**

`CronJob`, `CreateJobRequest`, and `UpdateJobRequest` (lines 15-34, 58-71, 80-90) each gain a new `engine_properties: Optional[dict[str, Any]]` field. This is a Plugin API contract change (per `arch.rules.md` Rule 3, this module is explicitly the Plugin API boundary between `core/cron` and the notification plugins). Its own `plugin_api/cron/README.md` states the required practice in black and white:

> "`CronJob` / `CreateJobRequest` now also carry optional `owner_id` and `bot_id` fields so adapters can round-trip the creator/bot context..." / "Changing a field ripples to `core/cron` producers and every notification impl that renders a run record." / "The `owner_id` / `bot_id` fields are additive and optional (`None` by default), so existing callers remain compatible..."

That is the exact precedent for this exact kind of change, and the PR does not touch the README at all — the "provides," "Change impact," and "Compatibility" sections say nothing about `engine_properties`. This violates **Rule 1 (Invariant)** — "contract changes update specs/docs in the same change" — and **Rule 16 (Invariant)** — "PRs touching... Plugin APIs... must include propagation analysis: affected consumers, affected implementations, affected deployments/configuration, compatibility status... merge is blocked when required impact analysis is absent."

**Required fix direction:** Update `plugin_api/cron/README.md` to document `engine_properties` the same way `owner_id`/`bot_id` are documented (what it provides, that it's additive/optional/`None`-default, and which consumers — the notification impls, `core/cron` producers — are unaffected because they don't read it yet).

---

### Out-of-scope notes

- The `EngineProperties` Pydantic model (`api/cron/schemas.py`) uses `extra="allow"` specifically so aicoding can add engine-level switches later without a schema change here. That's a pragmatic, bounded choice consistent with this file's existing `payload`/`state` opaque-dict conventions — not flagged, but worth knowing it further widens the untyped surface at the one place (`plugin_api`) that Finding 2 says needs a documented contract.
- `api/cron/router.py` living under `api/` while being a FastAPI HTTP router (rather than under `adapters/http/`, per the constitution's recommended layout in Rule 8) is a pre-existing, repo-wide naming convention (every sibling under `api/` — `bash`, `bot`, `session`, etc. — is the same shape), not something this PR introduces or entrenches further; not gated here.
- The actual aicoding `reuse_session` ⇄ relay `reuseSession` interpretation logic lives entirely outside this repository (per the PR description, in `teamclaw/ocb`'s `corp/engines/aicoding/cron.py`), so its correctness and conformance testing cannot be verified from this diff — flagging only that this repo's contract (Finding 2) should still be self-describing regardless of where the consumer lives.

https://claude.ai/code/session_015Mp8NvKcbvYitESo7LyDHS

---

totalfrank (approved):
LGTM.

PR #1994, opened 2026-09-07
fix(backend): retry transient MCP device delivery

FreddieSun:
## Problem

MCP device delivery can race BaaS/Desktop device startup. A Bot may resolve successfully and then return `503 NO_ACTIVE_DEVICES` when the MCP create request is sent. The existing retry loop made only three total attempts over roughly three seconds, so short readiness gaps caused the user configuration flow to fail and roll back.

The same helpers are used for allow-list declaration, create, update, and delete delivery, but their retry handling was duplicated and retried permanent errors inconsistently.

## Solution

- Centralize MCP device mutation retry handling behind one transport-neutral helper.
- Use an initial attempt plus three bounded retries at 1, 2, and 4 seconds.
- Retry request failures and server-side `500/502/503/504` responses, including explicit `NO_ACTIVE_DEVICES`.
- Fail permanent 4xx responses immediately.
- Preserve the existing create-conflict fallback to update and the existing configuration rollback semantics after retry exhaustion.
- Keep the existing `DeviceActivatedEvent` full MCP projection as the recovery path for already committed configuration; no lifecycle or public contract changes are introduced.

## Validation

- Targeted Device/MCP regression: 198 passed.
- Relevant architecture gates: 11 passed.
- Backend full suite: 17,804 passed, 60 skipped.
- Ruff on changed files: passed.
- Targeted Backend SAST on changed files: passed.
- `git diff --check`: passed.

## Compatibility and risk

No Service API, Plugin API, HTTP contract, database schema, or configuration changes. The existing rule remains unchanged: when every applicable runtime delivery fails, the user MCP configuration write is rolled back. Transient failures may now keep the synchronous request open for up to seven seconds before that rollback. OCB's ARCA and BaaS adapters consume the unchanged DeviceSync interface and require only a later ocb-public gitlink update after merge.

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/mcp_device_payload.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/mcp_device_payload.py
@@ -12,11 +12,6 @@ from agentclaw.community.log import get_logger
 
 logger = get_logger()
 
-# 重试配置
-MAX_RETRY_ATTEMPTS = 3  # 最大重试次数
-RETRY_BACKOFF_BASE = 1  # 退避基数（1s, 2s, 4s）
-
-
 class DeviceMCPConfig:
     """MCP Server 配置（符合远端设备接口）"""
 
--- a/src/backend/src/agentclaw/community/core/devices/services/mcp_device_transport.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/mcp_device_transport.py
@@ -6,6 +6,7 @@ share payload, retry, and result behavior.
 """
 from __future__ import annotations
 
+from collections.abc import Callable
 import time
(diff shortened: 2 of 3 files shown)


totalfrank (approved):
LGTM — reviewed `9f0d855` (3 files: the shared `_request_with_transient_retry` helper, its call sites in `_create_mcp`/`_update_mcp`/`_delete_mcp`/`filter_servers`, and the expanded test suite). Traced the retry/attempt counting, the `stop_on_already_exists` short-circuit (including the 500-with-"already exists"-body case), and the permanent-4xx-fails-immediately path against both the new and pre-existing tests — behavior matches the PR description and no regressions found in the new code.

---

PR #1996, opened 2026-09-07
fix(backend): make installation read gate DB-configurable

FreddieSun:
## Problem
The completed-backfill migration gate was static YAML, so an emergency rollback required a Backend release.

## Solution
Read `skill_installation/default_sync_only` from the environment-scoped `ac_common_config` record on every Effective Read. Missing, disabled, malformed, or unreadable values retain full installation repair.

## Validation
- `pytest tests/community/di/modules/test_installation_read_config.py tests/community/core/skill_center/test_bot_capability_state_reader.py tests/community/architecture/test_architecture_compliance.py -q` (29 passed)
- `ruff check` on changed Python modules
- `git diff --check`

## Compatibility and risk
The default is unchanged: full repair. The new gate becomes active only through an enabled boolean DB record for the deployment environment. Default/exclusion synchronization, new-Bot initialization, and operational backfill remain complete.


totalfrank (approved):
LGTM — reviewed `9ba0828` (5 files: installation_read_config.py, installation_read_config_module.py, docs, and tests). The new `InstallationReadConfig.default_sync_only` property correctly fails safe to `False` on a missing/disabled record, a read exception, or a non-bool value, and `get_current_env()` already normalizes `pre`/`prepub` and `prod`/`gray` before the DB lookup so the env-scoped query is correct. `api.common_config_service.CommonConfigServiceProtocol` is a re-export of the same class used in `installation_read_config.py`, so the two import paths resolve to one type for DI purposes — no circular import between `common_config` and `skill_center`. No bugs found in the new code.

---

PR #2005, opened 2026-09-08
fix(backend): align ac_bot_startup_script DDL with the deployed table

totalfrank:
## Problem

`POST /openapi/v1/bots/with-manifest` reported `APPLY_FAILED` with the `script`
category aborted:

```
write failed: (mysql.connector.errors.DatabaseError) 1366 (HY000):
Incorrect integer value for column 'script' at row 1
```

`SHOW CREATE TABLE ac_bot_startup_script` shows the deployed table differs from
`sql/2026_08_10_bot_startup_script.sql` in four places. All four are
transcription errors made while applying that file by hand, not decisions:

| | file declared | deployed | effect |
| --- | --- | --- | --- |
| `script` | `mediumtext` | `mediumint(9)` | **every script write rejected** — the error above |
| `uk_tenant_script_key` | `(avernet_tenant, script_key)` | `(avernet_tenant, env, entity_id, bot_id)` | **no index on `script_key`** — every read is a full scan |
| `script_key` | `char(64)` | `varchar(256)` | none |
| `gmt_create` / `gmt_modified` | `datetime` | `timestamp` | none today |

The second one matters more than it looks. The repository filters `get`,
`upsert` and `delete` on `script_key` **alone** — that is the stated reason the
surrogate exists (`_script_key`'s docstring: *"filtering on the surrogate is
what keeps a lookup on the one index the table has"*). The deployed key carries
this key's name with the pre-surrogate design's columns, so it enforces the
right constraint while leaving the column every query filters on unindexed, on a
table `_build_create_bot_payload` reads on every payload it composes.

The root cause of the hand-transcription is the one #1977 found on the apply
table: the DDL review standard (研发规范) that gates OceanBase provisioning
refuses `DATETIME`. This file as written could not have provisioned the table it
describes.

## Solution

**One principle throughout: this file must read back identically to `SHOW CREATE
TABLE`.** An index or a column type that exists in the database and not in this
file is invisible drift, and invisible drift is exactly how the
`mediumint`/`mediumtext` error survived. So the file is updated to describe the
table that actually exists after repair, rather than an idealised one.

**`gmt_create` / `gmt_modified` → `timestamp`.** Required by the DDL review
standard, and independently the correct half of the rule this module already
tests: a database-filled column round-trips through TIMESTAMP's session-offset
conversion unchanged, while an application-bound one would be stored shifted.
Both are database-filled — the ORM maps them to `func.now()`, which renders
server-side as `now()`, and the upsert re-stamps `gmt_modified` the same way —
so nothing the application reads is lost. Matches both sibling tables in
`bot_config_manifest/sql/`.

**`script_key` → `varchar(256)`.** The value is always 64 lowercase hex
characters, so the column is wider than anything that can be written to it. The
width is mirrored rather than tightened because this file's job is to say what
the table *is*. CHAR's blank-padding was not relied on anywhere.

**Both unique keys are now declared.** The original key holds the name the
surrogate wanted while spanning the pre-surrogate columns, and dropping an index
is not available in the deployed environments — so the surrogate was added
alongside it as `uk_tenant_script_key_v2`, and the file now carries both:

```sql
UNIQUE KEY `uk_tenant_script_key`    (`avernet_tenant`, `env`, `entity_id`, `bot_id`),
UNIQUE KEY `uk_tenant_script_key_v2` (`avernet_tenant`, `script_key`)
```

They are equivalent — `script_key` is injective over the three columns the older
key spans, so a row accepted by one is accepted by the other, and nothing
depends on which raises the `IntegrityError` the upsert retries on. Only the
surrogate is indexed the way the repository reads, which is why it had to be
added at all.

⚠️ **Declaring the older key has a cost, recorded in the file:** its key is 5456
utf8mb4 bytes (64+20+1024+256 chars), past InnoDB's 3072-byte cap. OceanBase
accepts it; stock MySQL/InnoDB would refuse this `CREATE TABLE` outright. That
makes this file OceanBase-targeted, which it already was in practice — and that
cap is the same constraint that made the surrogate necessary in the first place.
A fresh InnoDB environment must omit the older key and loses nothing by doing so.

**The ORM keeps declaring only the surrogate** — it is the constraint the code
relies on, the one that fits the cap, and `create_all` has no older key to
preserve.

**The ORM model keeps `DateTime`**, as `apply_models.py` does for the same
reason. It cannot disagree with the deployed column because it never describes
it: the OceanBase schema is operator-provisioned (`create_schema: false`), so
the `.sql` file creates these and the ORM emits no DDL there. For reads and
writes the declaration is inert either way — SQLAlchemy's `TIMESTAMP` is a
subclass of `DateTime`, both bind and return `datetime`.

Where it is **not** inert: a community deployment on MySQL with `create_schema:
true` bootstraps through `create_all(mysql=True)`, which emits `DATETIME` from
this declaration. That gap is repo-wide rather than this table's — every model
here is declared this way — and switching this one column would only half-close
it, since `default=`/`onupdate=` are client-side constructs: the bootstrap emits
neither `DEFAULT CURRENT_TIMESTAMP` nor `ON UPDATE CURRENT_TIMESTAMP` for any
table here whatever the type says. Closing it properly is a repo-wide follow-up.

**`test_manifest_ddl_contract.py` now covers `bot_startup_script/sql/`.** The
rule it states is about a column-type choice under OceanBase, not about one
module, and this table — the manifest's storage for its `script` category — sat
outside its directory and so outside its coverage. That is exactly where the
drift turned up, found by reading a `SHOW CREATE TABLE` rather than by anything
in CI.

## Validation

- `tests/community/core/bot_config_manifest/`, `tests/community/repository/bot/`,
  `tests/community/core/bot_startup_script/`, `tests/community/plugins/local/` —
  1305 passed
- Confirmed the extended contract test actually catches this drift: reverting
  `gmt_create` to `datetime` fails with
  `ac_bot_startup_script.gmt_create is declared datetime, not timestamp`
- Confirmed the DDL parser still reads the table correctly with two `UNIQUE KEY`
  lines — 11 columns parsed, no key line mis-read as a column
- Verified the `DateTime`/`TIMESTAMP` claim by compiling both against the MySQL
  and SQLite dialects rather than asserting it, and verified that widening
  `script_key` does **not** trip `prepare_for_mysql`'s index-prefixing
  (320 of 768 budget chars, `_prefix_lengths` returns `None`)
- Not run: anything requiring OceanBase/MySQL. The suite is SQLite, which has no
  TIMESTAMP/DATETIME distinction and does not parse the OceanBase modifiers —
  which is why these rules are asserted against the DDL as text.

## Compatibility and risk

No application behaviour changes: the column types are declarations, and the
constraint names are not referenced anywhere but the model and the DDL.

The two breaking repairs are applied directly to the affected environments:

```sql
ALTER TABLE `ac_bot_startup_script`
  MODIFY COLUMN `script` mediumtext NOT NULL COMMENT '脚本正文（清空即删行）';

ALTER TABLE `ac_bot_startup_script`
  ADD UNIQUE KEY `uk_tenant_script_key_v2` (`avernet_tenant`, `script_key`) GLOBAL;
```

Neither drops anything. `MODIFY COLUMN` is re-runnable; `ADD UNIQUE KEY` is not
(it fails with "Duplicate key name" once the key exists), which is why these are
applied rather than shipped as a re-runnable script — matching them to #1977's
idempotent repair would need an `information_schema` guard, since there is no
`ADD KEY IF NOT EXISTS`.

Worth checking whether prod carries the same drift: if `pre` was transcribed by
hand, prod likely was too, and the `mediumint` typo would take out startup
scripts there identically.

## Related issues

Same class of finding as #1977, and the same root cause (the DDL review standard
refusing the canonical file, so the table was created by hand).

https://claude.ai/code/session_01C5EuqWQKRuw3yL7Rhz4cmd

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_startup_script/repository/models.py
+++ b/src/backend/src/agentclaw/community/core/bot_startup_script/repository/models.py
@@ -97,10 +97,23 @@ class BotStartupScriptModel(Base):
     #: Written by the repository, never by a caller. Hex sha256 is 64 ASCII
     #: characters, and the tenant is carried alongside rather than hashed in, so
     #: the isolation boundary stays visible in the key itself.
+    #:
+    #: Declared 256 to match the deployed column, which is ``varchar(256)`` --
+    #: wider than the 64 characters that can ever be written here. The width is
+    #: mirrored rather than tightened because this declaration's job is to say
+    #: what the table *is*; the value's real width is stated above, and the
+    #: repository is what holds it to 64 by only ever writing a hex digest.
     script_key = Column(
-        String(64), nullable=False, comment="唯一键代理：sha256(env|entity_id|bot_id)"
+        String(256), nullable=False, comment="唯一键代理：sha256(env|entity_id|bot_id)"
     )
 
+    # ``DateTime`` here, ``timestamp`` in the DDL, and the two are not in
+    # conflict: these types only shape the local-mode ``create_all`` schema,
+    # and SQLite has no TIMESTAMP/DATETIME distinction to shape. The column
+    # type that matters is the one in the .sql file, which states the rule --
+    # database-filled times are TIMESTAMP. Every sibling model in this repo is
+    # declared the same way; switching this one to ``TIMESTAMP`` would make it
+    # the outlier without changing a single byte of behaviour.
     gmt_create = Column(
         DateTime, default=func.now(), nullable=False, comment="创建时间"
     )
@@ -112,11 +125,19 @@ class BotStartupScriptModel(Base):
         comment="修改时间",
     )
 
+    # ``_v2`` because the plain name is occupied in the deployed environments
+    # by a key on (avernet_tenant, env, entity_id, bot_id) -- this key's name
+    # carrying the pre-surrogate design's columns -- which cannot be dropped
+    # there, so the surrogate key is added alongside it. The suffix is not a
+    # second constraint: the two enforce the same logical uniqueness, because
+    # script_key is injective over the three columns it hashes. Declared here
+    # under the name the deployed key actually has, so create_all, the .sql
+    # file and every environment agree on one name.
     __table_args__ = (
         UniqueConstraint(
             "avernet_tenant",
             "script_key",
-            name="uk_tenant_script_key",
+            name="uk_tenant_script_key_v2",
         ),
     )
 
--- a/src/backend/src/agentclaw/community/core/bot_startup_script/repository/models.py
+++ b/src/backend/src/agentclaw/community/core/bot_startup_script/repository/models.py
@@ -107,13 +107,25 @@ class BotStartupScriptModel(Base):
         String(256), nullable=False, comment="唯一键代理：sha256(env|entity_id|bot_id)"
     )
 
-    # ``DateTime`` here, ``timestamp`` in the DDL, and the two are not in
-    # conflict: these types only shape the local-mode ``create_all`` schema,
-    # and SQLite has no TIMESTAMP/DATETIME distinction to shape. The column
-    # type that matters is the one in the .sql file, which states the rule --
-    # database-filled times are TIMESTAMP. Every sibling model in this repo is
-    # declared the same way; switching this one to ``TIMESTAMP`` would make it
-    # the outlier without changing a single byte of behaviour.
+    # ``DateTime`` here, ``timestamp`` in the DDL. Deliberate, and the same
+    # split apply_models.py states for the same reason.
+    #
+    # It cannot disagree with the deployed column, because it never describes
+    # it: the OceanBase schema is operator-provisioned (``create_schema:
+    # false``), so the .sql file creates these and the ORM emits no DDL there.
+    # For reads and writes the declaration makes no difference either --
+    # SQLAlchemy's ``TIMESTAMP`` is a subclass of ``DateTime``, both bind and
+    # return ``datetime``, and the driver hands back a ``datetime`` for a
+    # TIMESTAMP column whichever is declared.
+    #
+    # Where it is NOT a no-op, and this is the part worth knowing: a community
+    # deployment on MySQL with ``create_schema: true`` bootstraps through
+    # ``create_all(mysql=True)``, which emits DATETIME from this declaration.
+    # That gap is repo-wide rather than this table's -- every model here is
+    # declared this way -- and switching this one column would only half-close
+    # it, since ``default=``/``onupdate=`` are client-side constructs: the
+    # bootstrap emits neither DEFAULT CURRENT_TIMESTAMP nor ON UPDATE
+    # CURRENT_TIMESTAMP for any table here, whatever the type says.
     gmt_create = Column(
         DateTime, default=func.now(), nullable=False, comment="创建时间"
     )
--- a/src/backend/src/agentclaw/community/core/bot_startup_script/sql/2026_08_10_bot_startup_script.sql
+++ b/src/backend/src/agentclaw/community/core/bot_startup_script/sql/2026_08_10_bot_startup_script.sql
@@ -6,7 +6,10 @@
 -- "no script" are the same state.
 --
 -- Key is (avernet_tenant, script_key), where script_key is a sha256 of the
--- logical key (env, entity_id, bot_id). The tenant is in the key because
+-- logical key (env, entity_id, bot_id). The table also carries an older unique
+-- key over those three columns directly -- both are declared below, both are
+-- really in the deployed table, and they are equivalent; only the surrogate is
+-- indexed the way the repository reads. The tenant is in the key because
 -- ac_bots is itself tenant-scoped, so a bot_id is unique only within a tenant —
 -- legacy "default" bots carry documented residual cross-tenant collision on
 -- that identifier. Without the tenant here, two such bots would share one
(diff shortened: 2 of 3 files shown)


totalfrank (commented):
Reviewed commits `3cc0b9d`..`3190db0` (full PR diff, first review round).

Files covered:
- `src/backend/src/agentclaw/community/core/bot_startup_script/repository/models.py`
- `src/backend/src/agentclaw/community/core/bot_startup_script/sql/2026_08_10_bot_startup_script.sql`
- `src/backend/tests/community/core/bot_config_manifest/test_manifest_ddl_contract.py`

Checked: correctness of the type/width changes against the ORM↔DDL split the contract test itself states, contract breakage (searched the repo for other referrers of the renamed `uk_tenant_script_key` constraint and of `script_key`/`char(64)` — none outside the two files this diff already updates together), test coverage (traced `_SQL_DIRS`/`_tables`/`_declarations` to confirm the new `bot_startup_script/sql` directory is actually parsed and asserted, not just added to a list), and architecture boundaries (no transport/framework leakage, no new env access). Concurrency, security/secrets, and Python-optional-typing rules don't apply — this diff touches only DDL comments, one column width/type pair, a constraint name, and test scoping.

No blocking or non-blocking code issues found.

One process note, not a code defect and not anchorable to a line: the PR body says the two breaking repairs (`script` → `mediumtext`, the missing index) are applied directly to environments and are "not in this diff." The PR's own cited precedent for this exact failure class, #1977, shipped its environment repair as a committed, idempotent SQL script (`2026_09_07_repair_misprovisioned_bot_config_manifest_apply.sql`) that CI's DDL contract test can see. Doing the same here would leave an auditable, idempotent trail for the `mediumint`→`mediumtext` and index repairs instead of an unrecorded manual step — worth considering, not required for this PR.

---

totalfrank:
**Review round 1 — no issues flagged.** Reviewed `3190db0` (the DDL alignment plus the comment correction that follows it). No blocking or non-blocking code issues found. CI green: 10/10 checks on this head.

On the process note — shipping the environment repair as a committed idempotent script, the way #1977 did:

The point is well taken, and that file existed in an earlier draft of this branch (`2026_09_08_startup_script_surrogate_index.sql`, covering both repairs with the detect/idempotency reasoning in its header). It was dropped at the author's direction, after the `mediumtext` repair was already being applied by hand and only the index statement was still outstanding. So this is a decision that was made deliberately rather than an oversight — but it was made before the auditability argument was on the table, so I've put it back to the author.

If it goes back in, it should cover both repairs rather than just the index:

```sql
ALTER TABLE `ac_bot_startup_script`
  MODIFY COLUMN `script` mediumtext NOT NULL COMMENT '脚本正文（清空即删行）';

ALTER TABLE `ac_bot_startup_script`
  ADD UNIQUE KEY `uk_tenant_script_key_v2` (`avernet_tenant`, `script_key`) GLOBAL;
```

One caveat worth recording either way, since it is a difference from #1977's script: this pair is **not** fully re-runnable. `MODIFY COLUMN` is idempotent, but `ADD UNIQUE KEY` fails with "Duplicate key name" once the key exists. #1977's repair could be written re-runnably because it detects its starting state through `information_schema` and uses `CREATE TABLE IF NOT EXISTS`; there is no `ADD KEY IF NOT EXISTS` to match that, so an equivalent here would need the same `information_schema` guard to earn the same property. Worth doing properly if we commit it, rather than committing a script that looks idempotent and is not.


---

totalfrank (commented):
Reviewed the new commit since the last round: `3190db0e`..`dd3122d6` ("fix(backend): declare both unique keys in the startup-script DDL").

**Part A — open findings from round 1:** none. Round 1 posted zero inline threads; its only content was a body-level process note (committing the environment DDL repair as an idempotent migration, following the #1977 precedent) explicitly framed as "worth considering, not required for this PR." Nothing to verify or resolve there — it's advisory and still stands as the author's call.

**Part B — review of the new commit:** this commit only touches `sql/2026_08_10_bot_startup_script.sql`, re-adding the original `uk_tenant_script_key (avernet_tenant, env, entity_id, bot_id)` key alongside the surrogate `uk_tenant_script_key_v2`, with extensive comments.

Checked:
- The stated key-width math (`64+20+1024+256 chars × 4 = 5456 bytes`) against the actual column declarations — correct.
- Whether this `.sql` file is ever executed by application code (it isn't — `CommunityDatabase.bootstrap()` calls `core/schema.py`'s `create_all()` against the ORM `Base.metadata`, never this file; these `sql/` files are read only as text by the DDL contract tests). So the fact that this file's `CREATE TABLE` would fail on stock InnoDB (the reintroduced key is past InnoDB's 3072-byte cap) — which the commit's own comment already discloses at length — doesn't affect any live code path, and matches the PR's stated goal of mirroring the actually-deployed OceanBase table exactly.
- Whether the repository's `IntegrityError` retry logic in `startup_script.py` depends on which constraint name raises — it doesn't (bare `except IntegrityError`), so the redundant second key is functionally inert as claimed.
- Test coverage: `test_manifest_ddl_contract.py` wasn't touched by this commit and none of its assertions are affected by the new key line (the parser only matches lines starting with a backtick'd column name, not `UNIQUE KEY` lines).

No new defects found.

**Status:**
- Round 1 findings: 0 filed, 1 non-anchored advisory note (not required, no action needed) — not open.
- Round 2 (`dd3122d6`): reviewed, no new findings.
- Still open: 0.

The PR is clean as of this round.

---

totalfrank:
**Review round 2 — no issues flagged. Loop complete.** Reviewed `dd3122d` (declaring both unique keys). No new defects; 0 findings open across both rounds. CI green: 10/10 checks on this head, including the two long ones (Backend unit tests, Singlebox coverage).

**Across the loop:**

| Round | Head | Outcome |
| --- | --- | --- |
| 1 | `3190db0` | No code issues. One advisory process note (commit the environment repair as an idempotent script, per #1977). |
| 2 | `dd3122d` | No new defects. Round-1 note addressed — see below. Nothing open. |

**What changed in response to round 1.** The advisory note was taken, but not in the shape it proposed. Rather than adding a second SQL file for a table that already has one, the canonical `2026_08_10_bot_startup_script.sql` now declares *both* unique keys — the original `uk_tenant_script_key (avernet_tenant, env, entity_id, bot_id)` alongside the surrogate `uk_tenant_script_key_v2 (avernet_tenant, script_key)`. That achieves what the note was after: the repair's end state is recorded in version control where CI's DDL contract test can see it, and the file goes back to satisfying its own contract of reading back identically to `SHOW CREATE TABLE`. The repair statements themselves remain in the PR body, along with the reason they are applied rather than shipped as a script — `ADD UNIQUE KEY` has no `IF NOT EXISTS`, so it cannot be made re-runnable without an `information_schema` guard.

**Three things round 2 verified that had been assertions until then**, which is the part worth keeping:

- The 5456-byte key-width math, against the actual column declarations
- That this `.sql` file is never executed by application code — `CommunityDatabase.bootstrap()` runs `create_all()` against the ORM `Base.metadata`, never these files — so the InnoDB incompatibility the reintroduced key creates touches no live code path
- That the repository's retry uses a bare `except IntegrityError` and so does not depend on *which* constraint raises, which is the load-bearing assumption behind "the second key is inert"

**Still open for a human, not blocking:** whether a fresh InnoDB environment provisioning from this file is a case anyone needs — it would have to omit the older key, losing nothing. And separately, the repo-wide `create_all(mysql=True)` gap described in the PR body, which is not this table's to fix.

Awaiting review from @FreddieSun and @xianmuyq.


---

FreddieSun (approved):
LGTM

PR #2011, opened 2026-09-08
fix(bcsfuse): include success in worker config responses

wienyang:
## Problem
Worker config responses omit `success`, but frontend-nextgen requires it to be a boolean. This causes valid responses with `fusion_enable: true` to fail validation, hiding eligible Bots from fusion mode.

## Solution
Add `success: true` to the shared response model for GET and PUT worker config endpoints. Preserve existing fields and error behavior, and update the API documentation.

No frontend or gateway changes are required.

## Validation
- Added regression coverage for both enabled and disabled fusion configurations.
- All 34 contract model tests passed.
- `git diff --check` and the lint-only pre-push gate passed.
- Live deployment and browser verification have not been performed.

### DIFF
--- a/src/bcsfuse/docs/api/api-reference.md
+++ b/src/bcsfuse/docs/api/api-reference.md
@@ -1252,13 +1252,22 @@ curl http://localhost:8765/v1/workers/my-bot-001/profiles/quality \
 
 ```json
 {
+  "success": true,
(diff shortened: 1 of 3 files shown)


wienyang:
@codex review

archer910517-ctrl (approved):
LGTM

PR #2012, opened 2026-09-08
fix(clawweb): install dynamic OCB workspace locally

3h-william:
## Summary

Fix ClawWeb local startup when using a dynamically linked composition workspace.

## Problem

The startup script used `npm ci` for both the Avernet workspace and the linked composition workspace. The latter does not maintain a committed lockfile because its workspace dependencies are resolved dynamically, so a stale or missing lockfile could block startup.

## Changes

- Keep `npm ci` for the Avernet workspace with its committed lockfile.
- Use `npm install` for the dynamically linked composition workspace.
- Preserve the existing package-based composition and startup behavior.

## Validation

Verified local startup with Node.js 22:

- Package boundary validation passed.
- All workspace packages built successfully.
- Database connection succeeded.
- ClawWeb started successfully on port `5173`.

## Scope

This change only affects the local startup script. It does not modify business logic, CI behavior, or deployment behavior.


archer910517-ctrl (approved):
LGTM

PR #2018, opened 2026-09-08
fix(bcs): allow session self-leave without manage rights

zh-bupt:
## Problem

`DELETE` session participant (V1 `remove_session_participant`) required session-manage authorization for every removal. A plain Human participant — not a group manager and not the session creator — deleting its own actor (a voluntary leave) was rejected with `Principal may not manage this Session`. The sibling group endpoint already supports self-service leave (Design §8.7), so the two surfaces were inconsistent.

## Solution

- Add `principal_can_act_as` to the V1 session facade, mirroring the same-named `bcs-app-group` helper: true when the target actor is the caller's own Human actor, or a Bot the caller owns via `created_by` / a creator relation edge.
- `delete_participant` now branches on `is_self`: a self-leave only loads the session (no manage check); removing any other actor still requires `load_session_for_manage`.
- The driver / ManagerWorker-manager pin is untouched: the legacy removal path still rejects removing those structural roles, so self-leave cannot drop them either.
- The group facade (`remove_group_participant`) was checked and already implements this — no change needed there.

## Validation

- New tests (bcs-app-session): plain participant self-leave succeeds and emits BotLeft; owner withdraws its own Bot without manage rights; non-manager removing another actor is still forbidden; driver owner's self-leave is still rejected (invalid_request).
- Updated `owned_bot_does_not_grant_session_participant_removal_permission` → `owned_bot_removal_is_owner_self_leave_not_management`: the owner-removal leg now asserts success per the new contract; also re-seeded the session via `session_repo.create` because the `create_session` helper ignores the roster argument (it inherits the group roster, which lacked the bot under test — the old assertion passed vacuously).
- `cargo test --package bcs-app-session --package bcs-app-group`: all pass (75 + 85 + sub-suites).
- `cargo check --workspace --all-targets`: clean.

### DIFF
--- a/src/bcs/crates/application/v1/bcs-app-session/src/lib.rs
+++ b/src/bcs/crates/application/v1/bcs-app-session/src/lib.rs
@@ -164,6 +164,26 @@ impl SessionServiceImpl {
         Ok(false)
     }
 
+    /// Whether the Principal acts as `actor_id` itself: the same Actor, or
+    /// a Human acting for a Bot it owns / has a creator relation edge to.
+    /// Mirrors the sibling `bcs-app-group` helper of the same name.
+    async fn principal_can_act_as(
+        &self,
+        principal: &Principal,
+        actor_id: &str,
+    ) -> Result<bool, ApplicationError> {
+        if principal.actor_id() == actor_id {
+            return Ok(true);
+        }
+        match principal {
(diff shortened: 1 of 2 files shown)


questchaos (approved):
LGTM

PR #2019, opened 2026-09-08
feat(bot-config-manifest): source protocols, resources over git, oss_aksk

totalfrank:
Follow-up to W6 (`resources`) and W7 (named + git sources). Specs live in `src/backend/specs/2026-09-08-manifest-source-protocols/`.

## What this is

A manifest declares **where content comes from**. That "where" used to be spelled by which key a source happened to carry — `git:` or `url:` — so the protocol was a side effect of a shape rather than a declared fact. Nothing could enumerate it, the capabilities endpoint could not publish it, and each category re-derived it its own way. Every defect below starts there.

Three changes:

1. **A source declares a `protocol`.** Exactly two: `git` and `oss`.
2. **`resources` gains git and named sources.** It was the one fetching category locked to inline sources.
3. **The (category × protocol) matrix is one table in code**, read by the `PUT` validator *and* published by the capabilities endpoint, with a test that drives a real request per cell.

The third is what keeps the other two honest.

## The matrix as shipped

| category | `content` | `oss` | `git` |
| --- | --- | --- | --- |
| `identity` | ✅ | ✅ | ✅ |
| `resources` | ✅ | ✅ | ✅ **(new)** |
| `skills` | ❌ not a package | ✅ *(digest required)* | ✅ |
| `cli_tools` | ❌ not an executable | ✅ *(digest required)* | ✅ **(digest not required)** |
| `mcp` | — registry reference, no source axis | | |
| `engine_config` | — no materialiser, whole category refused | | |

Reaching a source **by name** (`from:`) is not a third column: a named source declares a protocol and that protocol's cell governs.

## Defects closed

| | |
| --- | --- |
| **D1** | `resources` could not use a named source at all — `from:` was refused even for a URL source, so the declare-once-reference-many mechanism excluded workspace files |
| **D2** | the published example in `examples.zh-CN.md` returned 422. A test now reads that document out of the file and PUTs it |
| **D3** | an entry `subpath` on a git source passed `PUT` and failed at apply, so one git source served exactly one entry |
| **D4** | no way to reach a private object store: `CredentialType.OSS_AKSK` was refused at write and the row held one secret |
| **D5** | the `cli_tools` digest rule read the source *form*, so `from:` a git source classified as `named` and was charged the object-store pin |

Deferred, unchanged: `categories[].aborted` means "abandoned at resolve/plan" while §B.2.5 defines it as "an entry failed". Real, separately tracked.

## Found while building

**`cli_tools` was the one construct that passed `PUT` and failed at apply.** Its materialiser read `decl.source_url`, which for a `from:` entry holds the source *name*, and put that on the wire as a URL. Opening the matrix cell without fixing it would have shipped "accepted" and "appliable" as different sets. Two consequences followed: its resolve-time digest belt had to key on the protocol too (a belt refusing what the surface accepts is the same defect wearing the other hat), and an unpinned git-sourced declaration now has **no convergence key** — `("", subpath)` would have made every git tool at one path compare equal, so a moved ref would have survived as `unchanged`.

## Found in review, and fixed here

Three from an automated review and two from a code review, all verified against the code before fixing. **Every guard added for these was checked against the *unfixed* version first** — each one fails there and passes here, so none is a test that would have let its bug through.

| | |
| --- | --- |
| **P1** | **Git-sourced tools collided on one object key.** The store embeds a digest fingerprint in the key so a new version never overwrites the object a surviving row points at — and a git declaration carries no digest, which fingerprints to the constant `"0"`. A rejected delivery would roll the row back to already-overwritten bytes and publish a binary the engine refused. The acquired bytes are hashed when the declaration pins nothing. |
| **P1** | **`subpath` was inert on an oss file entry** — the exact failure mode this change set exists to remove. Refused at `PUT` (`subpath_without_archive`) rather than applied, because applying it would give `subpath` two meanings on one protocol. The rule: on `oss`, `subpath` selects a member of an archive, so it requires one. Also moves a `cli_tools` refusal that previously fired mid-apply. |
| **P2** | **`keep_last` could not restore a git-delivered tree.** It is the *default* policy and promises the bot keeps running what it has, but a transient git outage failed the whole resources category under a message about an `unpack` a git source may not carry. The canonical tree form is now self-describing (`acm-tree-v1`) and reversible. |
| — | **A doomed oss directory entry paid for its fetch.** Routing resources through `fetch_declared` had moved the `unpack` check after the fetch, so an entry guaranteed to be rejected spent up to 200 MiB against the apply's byte budget and lock TTL. A new shared `declared_protocol()` answers the protocol from the declaration with no network. |
| — | Added the missing regression test for the inline `source: {protocol: git, …}` road. |

## In-place spelling change

`schema_version` stays `1`. The feature is pre-release, so there is no installed base a compatibility road would protect; a source using `git:`/`url:` as the protocol key is refused at `PUT` with a message naming the replacement. In-repo fixtures and doc snippets are updated in the same change.

## Two places the plan bent

**`subpath` on the `oss` road does not compose into the URL.** The plan and `manifest-schema.zh-CN.md` §2.3 both said an oss `url` is a prefix an entry's `subpath` appends to. Nothing ever implemented that, and shipped skills/`cli_tools` behaviour is the opposite — `subpath` selects *inside the fetched archive*. Two meanings for one word split by category is what this change exists to remove, so the rule shipped is the single one: **`subpath` selects within what the source delivered.** Git delivers a tree, so source's and entry's compose (D3). `oss` delivers one object, so its `url` addresses that object and two objects are two sources. The docs now say that, and say plainly that declare-once-reference-many is git's property.

**No `fetch/oss_source.py`.** AK/SK signing produces *headers*, and `headers_for(url)` is already the seam the guarded fetcher calls. Signing lives in `credentials/signing.py` (botocore's `SigV4Auth`, not hand-rolled); the one guarded road keeps its ceilings, timeout budget, redirect policy and per-hop re-authorization. A parallel transport would have been a second copy of every limit.

Full list in `tasks.md` under "Implementation notes — where the plan bent".

## Contract changes to existing tests

Each is a test that asserted a defect:

- `test_named_and_git_sources_are_refused_for_resources_entries` → replaced by its inverse (D1)
- `test_entry_level_subpath_on_a_git_source_is_refused` → replaced by composition tests (D3)
- `test_reserved_types_are_refused_at_write[oss_aksk]` → `oss_aksk` is implemented; `basic` still holds the case
- the two `mode` cases moved from a url source to a git source — `mode` rules on ref movement and an object store has no ref
- `cli_tools` fixtures writing `from: <a URL>` corrected to `source:` — they only worked because a source name was being fetched as a URL

## Security

- secrets never enter the manifest, `script`, or a URL; userinfo is refused at `PUT` and at fetch
- `access_key_id` reads back (an identifier rotation cannot be operated blind); its secret half has **no representation** in any response, log or apply report
- `allowed_prefixes` matches on whole path segments — `…/team/content` does not authorize `…/team/content-secret` — and bounds a signing credential exactly as it bounds a header one
- mismatched credential shapes are refused, not dropped: silence would leave a caller believing they configured signing

## Verification

- `tests/community/core/bot_config_manifest`: **956 green** (865 before this change)
- `tests/community/endpoints`: green, including the per-cell matrix conformance suite
- full `tests/community` run: green apart from two `service_bot` skill-artifact cases that fail identically on `dev` and on this branch's merge base — outside CI's backend job scope, not this PR's
- the published example document is validated on every run, read from the file rather than copied

https://claude.ai/code/session_019dzBBLChdsLffbjx71e2zM


src/backend/specs/2026-09-08-manifest-source-protocols/spec.md
```diff
@@ -0,0 +1,237 @@
...
+An entry names a source and, where the protocol addresses a tree, the part it
+wants:
+
+```yaml
+manifest:
+  resources:
+    - path: data/faq.csv
+      from: content
+      subpath: kb/faq.csv
+    - path: data/kb/
```
totalfrank-personal: This means the entire directory and its sub directories right?

src/backend/specs/2026-09-08-manifest-source-protocols/spec.md
```diff
@@ -0,0 +1,237 @@
...
+    url: https://artifacts.example-corp.com/tools/
+    auth: oss-artifacts
+```
+
+An entry names a source and, where the protocol addresses a tree, the part it
+wants:
+
+```yaml
+manifest:
+  resources:
```
totalfrank-personal, line 145: You might also want to give an example of oss protocol here as well.

src/backend/specs/2026-09-08-manifest-source-protocols/spec.md
```diff
@@ -0,0 +1,237 @@
...
+
+```yaml
+schema_version: 2
+
+sources:
+  content:
+    protocol: git
+    url: https://code.example-corp.com/team/content.git
+    ref: v1.2.0
+    auth: corp-git-content
```
totalfrank-personal, line 125: I did not see you mention the either SHA here or the place where this source is referenced? We need that right?

src/backend/specs/2026-09-08-manifest-source-protocols/spec.md
```diff
@@ -0,0 +1,237 @@
...
+  either on a git source is **refused at `PUT`** rather than silently ignored —
+  a field that appears to configure something and does nothing is the failure
+  mode D3 and Appendix C's `cli_tools` trap both have.
+
+## Behaviour changes
+
+### Sources declare a protocol (`schema_version: 2`)
+
+```yaml
+schema_version: 2
```
totalfrank-personal: Version could be 1, we are testing it ourselves, not released yet.

totalfrank-personal (commented):
A few comments, please take a look.

src/backend/specs/2026-09-08-manifest-source-protocols/spec.md
totalfrank: Agreed — done in 190082c9. `schema_version` stays `1` and the whole v2/compat design is withdrawn.

That collapses a good chunk of the work: plan §2 loses the dual-spelling `parse_source(raw, *, schema_version)` and the v1→v2 mapping table, and group B loses the equivalence test. The `protocol:` spelling now replaces `git:`/`url:` in place, and a source using the old form is refused at `PUT` with a message naming the replacement (acceptance criterion 7). B4 became "update every in-repo fixture and doc snippet using the old spelling", plus a test asserting `SUPPORTED_SCHEMA_VERSIONS` is untouched so a later drive-by bump gets caught.

Recorded under "Decisions taken" as withdrawn-on-review rather than deleted, so the reasoning is still visible.

---

src/backend/specs/2026-09-08-manifest-source-protocols/spec.md
totalfrank: Right, that was missing. Added after the source example in 190082c9:

`ref` takes **a tag, a branch, or a full commit SHA** — writing a SHA is how you pin absolutely. Whichever you write, the platform resolves it to one commit per `(url, ref)` per apply and records it in the apply report's `sources[]` row as `resolved_sha`, which is the answer to "which version is actually running". The example now carries a `# tag, branch, or a full commit SHA` comment on the `ref` line.

The related half — what happens when a tag *moves* between applies — is `mode`, not a digest: `strict` refuses a ref resolving to a different commit than the last apply saw, `non_strict` (the default) allows it and reports the move. That's also why git needs no `digest` while `oss` does: the commit SHA is already a recorded content identity, and a `sha256` over git bytes has no stable meaning across the checkout/archive/canonical-zip roads.

---

src/backend/specs/2026-09-08-manifest-source-protocols/spec.md
totalfrank: Added in 190082c9 — the entry example now has three entries covering both protocols and both shapes:

```yaml
manifest:
  resources:
    - path: data/faq.csv       # one file, from git
      from: content
      subpath: kb/faq.csv
    - path: data/kb/           # a whole tree, from git — no packaging
      from: content
      subpath: kb/
    - path: data/pricing.csv   # one object, from oss
      from: artifacts
      subpath: reference/pricing.csv
```

`artifacts` is the `protocol: oss` source already declared above it, so the example exercises both without a second sources block. Also queued as E3: the same `oss` entry goes into `examples.zh-CN.md` when the docs are corrected.

---

src/backend/specs/2026-09-08-manifest-source-protocols/spec.md
totalfrank: Yes — recursive, every file at every depth. It was implied rather than stated, which is exactly the kind of thing that ends up as a surprise; now spelled out in 190082c9:

> A `path` ending in `/` is a **directory entry**: the source subtree is delivered under it **recursively — every file at every depth**, and the declared area is replaced wholesale, so a file that disappears upstream disappears from the workspace on the next apply.

The wholesale-replacement half matters as much as the recursion: the declared `path` is an *area* the manifest owns, so this is a tree replacement, not a merge. Acceptance criterion 2 and task D4 both say it now, and D4's test asserts nested subdirectories land and that a file removed upstream disappears on the next apply.

Worth knowing this is the shape only git gives you cheaply. On `oss` a directory still arrives as an archive (`unpack: zip|tar.gz`), because one HTTPS GET fetches one object — which is why the examples doc calls the git form 免打包.

---

src/backend/specs/2026-09-08-manifest-source-protocols/plan.md
```diff
@@ -0,0 +1,201 @@
...
+
+`apply/entry_fetch.py` — `fetch_declared()` currently refuses an entry
+`subpath` on the git road. It instead **composes**: `source.subpath` joined
+with `entry.subpath`, normalised and re-checked by the same
+`relative_path_refusal` predicate the schema uses (no second, weaker rule), and
+handed to `GitEntrySource.subpath`. One checkout per `(url, ref)` per apply is
+already cached in the source session, so N entries off one source stay one
+fetch and one `resolved_sha`.
+
+## ③ The OSS road
```
totalfrank-personal, line 99: Can we also have some code snippets here to show how we are going to register different credentials here(e.g. for OSS, for git).

Have some example code here would be more intuitive.

totalfrank-personal (commented):
One comment, please take a look.

src/backend/specs/2026-09-08-manifest-source-protocols/plan.md
totalfrank: Good call — §③ described the model change and never showed a single call. Added in the latest push: worked examples for both, since they're the same endpoint one field apart.

A git host token, presented in a header while cloning:

```http
PUT /openapi/v1/bots/source-credentials/corp-git-content
```
```json
{
  "type": "header",
  "header_name": "PRIVATE-TOKEN",
  "secret": "<token>",
  "allowed_prefixes": ["https://code.example-corp.com/team/content"]
}
```

An object-store key pair — `access_key_id` is the new field, `secret` carries the secret key:

```json
{
  "type": "oss_aksk",
  "access_key_id": "LTAI5t...",
  "secret": "<secret key>",
  "allowed_prefixes": ["https://artifacts.example-corp.com/tools/"]
}
```

Plus the redacted read-back shape (the id comes back, the secret has no representation at all), and how each is referenced from a source by **name** — `auth: corp-git-content` / `auth: oss-artifacts` — so the manifest and the apply report never carry a value.

Three things the examples make visible that the prose had buried: this endpoint group is **tenant-level and takes no `user_id`**, unlike every other route in the feature; `allowed_prefixes` matches on **whole path segments**, which is what stops `…/team/content` authorizing `…/team/content-secret`; and the service **refuses** mismatched shapes rather than ignoring the extra field. C4 grew to cover one endpoint case per type plus those three refusals.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/resources.py
```diff
@@ -162,53 +163,71 @@ async def resolve(
...
-                source_url = entry.get("source")
-                if not isinstance(source_url, str) or not source_url:
-                    failures.append(
-                        ResolveFailure(
-                            path, "a directory entry must declare 'source'"
-                        )
-                    )
-                    continue
                 try:
                     fetched = await self._fetch_entry(
```
totalfrank, line 184: **Non-blocking:** For the OSS (non-git) directory road, this diff now fetches the entry (`_fetch_entry` → `fetch_declared` → a real guarded network fetch) *before* validating `unpack`/`strip_components` (moved to lines ~189-208). Pre-diff, those two fields — plus `source` — were validated before any network call.

Trigger: an apply-time document that predates or skipped the PUT validator (the module's own docstring calls out "W8 lifecycle points" hand-built applies) declares `{"path": "wrap/", "source": "<oss url>"}` with no `unpack` or an invalid `strip_components`. `resolve()` now spends a full fetch (bandwidth, apply-fetch budget, and the apply-lock TTL this module's docstring already worries about for large archives) on an entry that is guaranteed to fail once `unpack_kind`/`strip` are checked afterward.

Suggested fix: validate `unpack_kind` and `strip_components` before calling `_fetch_entry` on the non-git branch, deferring the fetch until those checks pass — keep the git branch's flow unchanged since it must fetch first to know it's git.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/cli_tools.py
```diff
@@ -264,4 +292,29 @@ async def write(
         return tuple(results)
 
 
+def _is_git_sourced(ctx: ApplyContext, entry: dict[str, Any]) -> bool:
```
totalfrank: nit: `_is_git_sourced`'s inline-object branch (an entry with `"source": {"protocol": "git", ...}` written directly, not via `from:`) isn't exercised by any test in `test_cli_tools_materialiser.py` — only the named (`from:`) road is covered by `test_a_tool_from_a_named_git_source_applies_without_a_digest`. The logic itself looks correct (`raw = entry.get("source")` falls through to the same `parse_source` check), just worth a matching regression test for the inline form.

---

totalfrank (commented):
Reviewed commits eb9ee30..69a34af (full PR range, merge-base 589b88d1..69a34af5) — 44 files, focused on the implementation and test files under `bot_config_manifest/` (schema/sources.py, support_matrix.py, entry_fetch.py, the resources/cli_tools materialisers, capabilities.py, credentials/{models,service,signing}.py, the SQL migration, and the openapi/source_credentials adapter) plus their new/changed tests.

Checked: correctness of the support-matrix/parser logic, capabilities-vs-PUT-validator contract consistency, DB write-failure handling in the credential service, SigV4 signing for secret leakage (never returned/logged), `allowed_prefixes` segment-boundary matching, architecture isolation of the new botocore usage, `T | None` widening justification, and test coverage of the new source-name/URL-confusion fix (D5) and the resources-over-git fix (D1).

Two findings posted inline: one non-blocking behavioral regression (OSS directory entries in `resources.py` now fetch before validating `unpack`/`strip_components`, reversing the pre-diff order) and one nit (missing test for the inline-object git-source road in `cli_tools.py`, though the logic itself is correct). Everything else — the D1/D5 fixes, the support-matrix single-source-of-truth design, secret handling, and doc/code/test alignment — held up under review.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/cli_tools.py
```diff
@@ -153,12 +153,18 @@ async def resolve(
             seen.add(name)
 
             decl = CliToolDecl.from_entry(entry)
-            if not decl.digest:
+            if not decl.digest and not _is_git_sourced(ctx, entry):
```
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in f6c1db82. `fingerprint("")` is `"".rsplit(":",1)[-1][:16] or "0"` → the constant `"0"`, so every revision of one git-sourced tool wrote to `…/mycli.0` — against the store's own docstring, which says the key embeds the fingerprint "so this never overwrites the object a *different* version's row still points at".

`_record_one` now derives the digest the store and the row are keyed by:

```python
content_digest = decl.digest or ("sha256:" + hashlib.sha256(data).hexdigest())
```

It only ever fills in the missing case — a declared digest has already been proved by the fetch pipeline, so hashing there would be re-proving it.

Two guards, both verified to fail on the unfixed version: two different git revisions of one tool write two distinct keys and both objects survive (so a rollback has real bytes to point back at), and the persisted row carries a full `sha256:` + 64 hex.

The convergence rule is unaffected — the *declaration* still has no digest at plan time, so a git-sourced tool still never plans `unchanged` and re-acquires every apply, which is what stops a moved ref surviving.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/resources.py
```diff
@@ -256,66 +271,105 @@ async def resolve(
...
-                    ctx, source_url, entry, path, _FETCH_CATEGORY_FILE
+                    ctx, entry, path, _FETCH_CATEGORY_FILE
                 )
+                if isinstance(fetched, GitEntrySource):
+                    data = await asyncio.to_thread(
+                        self._git_file, ctx, fetched, path
+                    )
+                    note = fetched.moved_note()
+                else:
+                    data, note = fetched.content, fetched.fallback_reason
```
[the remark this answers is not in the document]
totalfrank: Real, and it was the exact failure mode this change set exists to remove — a field that looks like configuration and governs nothing. Rejected at validation (your second option) in f6c1db82, because applying the selector would have given `subpath` two meanings on one protocol.

The rule, stated once: **on `oss`, `subpath` selects a member of an archive, so it requires one.**

| category | `oss` + `subpath` | why |
| --- | --- | --- |
| `skills` | ✅ | a skill *is* a package; this road always unpacks |
| `cli_tools` | needs `unpack` | without it the object *is* the command |
| `resources` (directory) | ✅ | already required to declare `unpack` |
| `resources` (file) | ❌ `subpath_without_archive` | no archive — your case |
| `identity` | ❌ same | an identity file is one text body, never an archive |

`git` is unaffected: it delivers a tree, so `subpath` always has something to select within, in every category and with no `unpack` anywhere.

The `cli_tools` row is worth noting separately — the acquisition already refused a subpath with no `unpack`, but only once the apply was running. That was the surface accepting what apply rejects, so the check moved to `PUT`.

Both docs updated (`manifest-schema.zh-CN.md` §2.2, `user-manual.zh-CN.md` §5.3), and eight cases pin it — the three refusals, the three shapes that legitimately unpack, and `git` across all four categories.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/resources.py
```diff
@@ -162,53 +163,71 @@ async def resolve(
...
+                    try:
+                        members = await asyncio.to_thread(
+                            self._git_members, ctx, fetched, path
+                        )
+                    except EntryFetchError as exc:
+                        failures.append(ResolveFailure(path, exc.reason))
+                        continue
+                    note = fetched.moved_note()
+                else:
+                    unpack_kind = entry.get("unpack")
```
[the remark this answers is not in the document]
totalfrank: Both halves confirmed, and fixed in f6c1db82 by making the stored form reversible — the first of your two suggestions, since it is the one that keeps the promise `keep_last` makes.

This mattered more than P2: `keep_last` is the **default**, and it promises the bot keeps running what it has. A transient git outage instead failed the whole resources category, under a message telling the caller to declare an `unpack` a git source may not even carry.

`_canonical_tree_bytes` now emits a magic header and `_decode_tree_bytes` reads it back:

```python
_TREE_MAGIC = b"acm-tree-v1\n"
```

Self-describing rather than sniffed from the receipt URL: the fallback arrives as plain bytes with no idea what shape they are, and "guess from the URL" is not identification. The version digit lets the framing change later without a stored copy being decoded under the wrong rules. `_decode_tree_bytes` returns `None` — an answer, not a fault — for anything that is not one, including a truncated one, so a bad receipt never half-delivers a tree.

The directory branch checks for a stored tree before the archive road, so the members come straight back with the fallback reason on each intent (§9.6).

Four guards: the codec round-trips including a payload containing the framing characters and an empty file (why it is length-prefixed, not delimiter-joined); a zip and a truncated tree both decode to `None`; the keep_last path delivers the members with the reason attached; and a real archive still takes the unpack road unchanged.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/resources.py
totalfrank: Right, and I'd taken the ordering as forced when it wasn't. Fixed in 267eb9b6.

The reason the fetch moved first was that the protocol was only known once the result came back — a `GitEntrySource` or a `FetchedEntry`. But the protocol is a property of the *declaration*, so it can be answered with no network at all:

```python
def declared_protocol(ctx, entry) -> SourceKind | None
```

Read through the same parser the `PUT` validator and `fetch_declared` already use, so it cannot become a fourth derivation of "which protocol is this". `None` means "cannot say from the declaration alone" (a malformed or undeclared source), and the caller falls through to the fetch, which raises the real error with the real message.

The archive check is now one pure function (`_archive_refusal`) asked twice: before the fetch on the `oss` road, and after it for the `None` case. Two calls of one function, never two rules — the shape the codebase already uses for `relative_path_refusal`.

Your trigger is pinned directly: an apply-time `{"path": "wrap/", "source": "<oss url>"}` with no `unpack` now fails with `stub.calls == []` — no fetch paid for. Same for a negative `strip_components`. A third case holds the git road, which must still fetch first and declares no `unpack` at all.

`cli_tools` now reads the same helper instead of the private `_is_git_sourced` copy it had.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/cli_tools.py
totalfrank: Fair catch, and a pointed gap given what this PR is about: `from:` a git source and an inline `source: {protocol: git, …}` being *one source declared two ways* is the whole of D5, so proving it on the named road alone left the claim half-tested.

Added in 267eb9b6 — `test_an_inline_git_source_is_exempt_from_the_digest_belt_too` asserts the inline form resolves with no `digest`, applies, and reaches the transport as the repository URL on the git road.

The function it covered has since been replaced: `_is_git_sourced` is gone and `cli_tools` calls the shared `declared_protocol` (see the thread on resources.py:167), so both materialisers now answer the protocol the same way. The new test covers the inline branch there.

---

totalfrank (commented):
Reviewed commit f6c1db8 (the only new commit since the last review, on top of 69a34af5) — `fix(bot-config-manifest): three defects from review`, touching `resources.py`, `cli_tools/service.py`, `schema/entries.py`, docs, and their tests.

**Status checklist (all rounds):**

1. Directory-entry recursion semantics (spec.md) — fixed, resolved
2. OSS protocol example missing (spec.md) — fixed, resolved
3. `ref`/`resolved_sha` undocumented (spec.md) — fixed, resolved
4. `schema_version` v2/compat withdrawn (spec.md) — fixed, resolved
5. Credential registration examples missing (plan.md) — fixed, resolved
6. [Codex P1] git tool storage key collides on empty digest — addressed in f6c1db8: acquired bytes now hashed (`content_digest = decl.digest or sha256(data)`); verified against code and the new regression tests (`test_two_revisions_of_a_git_tool_do_not_share_one_object_key`, `test_a_git_tool_records_a_real_content_digest`), not mine to resolve
7. [Codex P1] `subpath` silently ignored on OSS file entries — addressed in f6c1db8: `check_source_subpath` now refuses it at `PUT` for resources/identity/cli_tools; verified against code and tests, not mine to resolve
8. [Codex P2] `keep_last` couldn't restore a git-delivered tree — addressed in f6c1db8: canonical tree bytes are now self-describing (`_TREE_MAGIC`) and reversible (`_decode_tree_bytes`); verified round-trip and fallback tests, not mine to resolve
9. **Still open** — `resources.py` OSS directory road fetches before validating `unpack`/`strip_components` (non-blocking, reordering regression vs. pre-diff). Code at those lines is unchanged in f6c1db8 and the commit message makes no claim about it, so left open per instructions rather than re-filed.
10. **Still open** — `cli_tools.py` inline-object git-source road (`"source": {"protocol": "git", ...}`) has no regression test (nit). Not touched by f6c1db8, no claim of a fix, left open.

**New commit review:** read `resources.py`, `cli_tools/service.py`, `cli_tools/models.py`, `cli_tools/declarations.py`, and `schema/entries.py` in full; ran the full `tests/community/core/bot_config_manifest` suite locally against this commit (952 passed). The `check_source_subpath` gate correctly narrows only the `(oss, no-archive)` cell in every category (skills/git always short-circuit before it), the tree-canonicalisation round-trip is a correct inverse of its encoder including the empty-tree case, and `content_digest` does not disturb `CliToolDecl.convergence_key` (still keyed on the *declared* digest, so git-sourced tools keep re-acquiring every apply as before). No new correctness, security, or architecture-boundary issues found.

Count still open: **2** (both mine, non-blocking/nit, pre-existing from the prior round).

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/cli_tools.py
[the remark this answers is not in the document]
totalfrank: Verified fixed — `content_digest` now hashes the acquired bytes when the declaration has none (`service.py:209-211`), so two revisions of a git-sourced tool get distinct store keys. `test_two_revisions_of_a_git_tool_do_not_share_one_object_key` and `test_a_git_tool_records_a_real_content_digest` both pass. Resolving.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/resources.py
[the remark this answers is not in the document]
totalfrank: Verified fixed — `check_source_subpath` (`schema/entries.py:520-553`) refuses `subpath` on a non-archive `oss` entry at `PUT`, wired into resources, identity, and cli_tools. The parametrized suite in `test_manifest_schema.py` covers all three refusal cases plus the accept cases (skills, unpack-declared cli_tools/resources, all four categories over git) — all pass. Resolving.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/resources.py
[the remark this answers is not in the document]
totalfrank: Verified fixed — the canonical tree form is now self-describing (`_TREE_MAGIC`) and reversible via `_decode_tree_bytes`, and the resolve loop recognizes a decoded stored tree before falling into the oss-archive branch. `test_keep_last_delivers_a_stored_git_tree_instead_of_failing`, the round-trip test, and the truncated/non-tree-decodes-to-None test all pass. Resolving.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/resources.py
totalfrank: Verified fixed — `declared_protocol` answers oss-vs-git from the declaration with no network call, and the archive-field checks now run before `_fetch_entry` on the oss road. `test_an_oss_directory_entry_is_refused_before_the_network_is_touched` and `test_an_invalid_strip_components_is_also_refused_before_the_fetch` both assert `stub.calls == []`; the git road is confirmed unaffected by `test_a_git_directory_entry_still_fetches_without_declaring_unpack`. Resolving.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/cli_tools.py
totalfrank: Addressed — `test_an_inline_git_source_is_exempt_from_the_digest_belt_too` now covers the inline `source: {protocol: git, ...}` road alongside the already-covered named (`from:`) road, and it passes. Resolving.

---

totalfrank (commented):
Commits reviewed this round: `f6c1db82` (fix: three defects from review), `267eb9b6` (perf: refuse a doomed oss directory entry before fetching) — the delta since my previous review at `69a34af5`.

Spec/plan doc review (totalfrank-personal, human):
- ✅ directory-entry recursion semantics — resolved (spec clarified in `190082c9`)
- ✅ missing oss example in spec entry list — resolved (added in `190082c9`)
- ✅ missing `ref`/resolved-SHA explanation — resolved (added in `190082c9`)
- ✅ `schema_version` should stay 1 — resolved (v2/compat design withdrawn in `190082c9`)
- ✅ missing credential registration code examples — resolved (added in `43b52044`)

Codex automated review (at `69a34af5`):
- ✅ **P1** git tool storage-key collision (`cli_tools.py`, empty digest → constant fingerprint) — **fixed** in `f6c1db82`, verified: `content_digest` now hashes acquired bytes; tests pass
- ✅ **P1** `subpath` silently ignored on an oss file entry (`resources.py`) — **fixed** in `f6c1db82`, verified: refused at `PUT` via `subpath_without_archive`; 4-category test matrix passes
- ✅ **P2** `keep_last` couldn't restore a git-delivered tree (`resources.py`) — **fixed** in `f6c1db82`, verified: canonical tree form now self-describing/reversible; tests pass

My own review (at `69a34af5`):
- ✅ **Non-blocking** oss directory fetch happens before `unpack`/`strip_components` validation — **fixed** in `267eb9b6`, verified: `declared_protocol` gates the fetch; `stub.calls == []` asserted for the doomed-entry case
- ✅ **nit** missing test for inline `source: {protocol: git, ...}` road in `cli_tools.py` — **addressed** in `267eb9b6`, verified: `test_an_inline_git_source_is_exempt_from_the_digest_belt_too` added and passing

**New commits (`f6c1db82`, `267eb9b6`) — fresh review:** no new defects found. Ran the full `tests/community/core/bot_config_manifest` suite (956 passed) plus the two materialiser test files directly. Walked the new `declared_protocol` helper (shared by both materialisers, reads through the same parser as `PUT` and `fetch_declared` — no fourth derivation), the `_archive_refusal`/`_TREE_MAGIC`/`_decode_tree_bytes` additions in `resources.py`, and the digest derivation in `cli_tools/service.py`. Coverage is thorough (round-trip, truncation, both accept/refuse branches, git-road non-regression).

**Open findings: 0.** PR is clean as of this round (`267eb9b6`).

---

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/_support.py
```diff
@@ -52,8 +59,15 @@ class Context:
...
-    #: against. Populated before any entry is walked.
+    #: against. Populated before any entry is walked. Every declared name is
+    #: here, including one whose declaration failed to parse — so a ``from``
+    #: naming it is not *also* reported as undeclared, which would be one
+    #: mistake answered as two.
     source_names: set[str] = field(default_factory=set)
+    #: The declarations that parsed, by name. An entry's ``from`` reads its
+    #: protocol from here, which is what makes a named source resolve to a
+    #: support-matrix cell rather than needing a column of its own.
+    sources: dict[str, "SourceDecl"] = field(default_factory=dict)
```
totalfrank-personal: Do you need this forward reference here? If not, let's get rid of it.

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/sources.py
```diff
@@ -0,0 +1,291 @@
...
+
+    protocol: SourceKind
+    url: str
+    #: git only — a tag, a branch, or a full commit SHA. ``None`` means the
+    #: repository's default head.
+    ref: str | None = None
+    #: git only — the part of the repository's tree this declaration
+    #: addresses. Composes with an entry's own ``subpath`` (source's first,
+    #: then the entry's), which is what lets one source serve many entries.
+    subpath: str | None = None
```
totalfrank-personal, line 108: Can you give a concrete example here for the git path so that it's more intuitive on how we interpreter these? Right now it's a little bit confusing.

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/sources.py
```diff
@@ -0,0 +1,291 @@
...
+    needs to know where in a document it was called from.
+    """
+
+    suffix: str
+    code: str
+    message: str
+
+
+def parse_source(
+    raw: Any,
```
totalfrank-personal: If this is always a Mapping, then let's not use Any.

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/signing.py
```diff
@@ -0,0 +1,89 @@
...
+    SigV4Auth(
+        Credentials(access_key_id, secret_access_key),
+        _SERVICE,
+        region or DEFAULT_REGION,
+    ).add_auth(request)
+    signed = dict(request.headers)
+    return {name: signed[name] for name in _SIGNED_HEADERS if name in signed}
+
+
+__all__ = ["DEFAULT_REGION", "sign_headers"]
```
totalfrank-personal, line 93: What is this for and what is it doing? I did not get it.

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_fetch.py
```diff
@@ -493,12 +507,12 @@ def fetch_declared(
 
         keep_last = entry.get("on_fetch_failure", "keep_last") == "keep_last"
         name: Optional[str] = None
-        decl: Optional["Mapping[str, Any]"] = None
+        raw: Optional["Mapping[str, Any]"] = None
```
totalfrank-personal: If we don't need forward reference, let's get rid of it.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/signing.py
totalfrank: Good question to have asked — it was pointing at two real bugs, not just unclear code.

**The signer was the wrong one.** S3 requires `x-amz-content-sha256` on every SigV4 request, and only `S3SigV4Auth` emits it; the generic `SigV4Auth` I used does not. Against real S3 the fetch would have failed as an opaque 403.

```
SigV4Auth   -> ['Authorization', 'X-Amz-Date']
S3SigV4Auth -> ['Authorization', 'X-Amz-Content-SHA256', 'X-Amz-Date']
```

**And the line you asked about would have dropped it anyway.** It was an allowlist over the signer's output:

```python
_SIGNED_HEADERS = ("Authorization", "X-Amz-Date", "X-Amz-Security-Token")
return {name: signed[name] for name in _SIGNED_HEADERS if name in signed}
```

My reasoning was "don't present a header we didn't intend to" — which is the wrong caution here. A signature covers a *specific set* of headers, so presenting a subset is not a weaker request, it is an invalid one, failing the same opaque way with nothing pointing at the filter that caused it. The signer is the authority on what its own signature needs, so the allowlist is gone and its output is returned whole.

Fixed in 743aa9f2. Worth noting the existing test asserted the exact set `{Authorization, X-Amz-Date}` and so **pinned the bug** — it now asserts what must travel and what must not (the secret key never does), which is its actual subject, and a new case pins the three headers S3 requires.

If line 89 was the `__all__` rather than the return: that is just the module's public surface, matching `entry_fetch.py`, `support_matrix.py` and the other modules in this package. Nothing load-bearing.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/_support.py
totalfrank: No — and the comment I wrote justifying it was simply wrong. It claimed `sources` imports this module, so a runtime import would close a cycle. It does not: `sources.py`'s only in-package import is `support_matrix`, so nothing here was ever circular.

Direct import in 743aa9f2, and the `TYPE_CHECKING` block and the quotes on `dict[str, SourceDecl]` go with it. Suite green, which is the proof there was no cycle to break.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_fetch.py
totalfrank: Right — redundant twice over, in fact. `Mapping` is imported at runtime at the top of the module, *and* the file has `from __future__ import annotations`, so every annotation is already a string. The quotes bought nothing.

Removed in 743aa9f2, along with the other quoted `Mapping` annotations I'd added in the same file.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/sources.py
totalfrank: It is deliberately *not* always a `Mapping` — that is the parameter's whole job. `raw` is whatever `yaml.safe_load` produced at that position in a caller-authored document: a string, a list, `None`, anything. The first thing the function does is narrow it:

```python
if not isinstance(raw, Mapping):
    add("", "invalid_source", "a source must be a mapping")
    return None, tuple(violations)
```

But your instinct is right about `Any` — it was the wrong way to say "anything". `Any` *silences* the checker; `object` says the same thing and makes the narrowing mandatory, which is exactly the contract this function has with its callers. Changed to `object` in 743aa9f2, with the reasoning in the docstring so the next reader does not have to re-ask.

`_parse_protocol`'s `add` callback had the same smell — it was `Any` and is now `Callable[[str, str, str], None]`.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/schema/sources.py
totalfrank: Agreed — the prose described a rule without ever showing it. Worked example added to the field in 743aa9f2:

```yaml
sources:
  content:
    protocol: git
    url: https://code.example.com/team/content.git
    ref: v1.2.0
    subpath: kb          # ← this field
manifest:
  resources:
    - path: data/faq.csv
      from: content
      subpath: faq.csv   # ← composes to "kb/faq.csv"
    - path: data/prices/
      from: content
      subpath: prices    # ← composes to "kb/prices", a whole subtree
```

Plus the sentence that makes it click: **both entries read one checkout of one commit.** The composition changes which path is read *out of* the tree, never which tree is fetched — so `(url, ref)` is still cloned once per apply and the report still carries one `sources[]` row with one `resolved_sha`, however many entries name the source.

That is the whole point of the composition (D3): before it, `subpath` beside a git source was refused, so one source served exactly one entry and a second file from the same repository needed a duplicate source block carrying the same `url`, `ref` and `auth` — the drift named sources exist to remove.

---

totalfrank (commented):
Commit reviewed this round: `743aa9f2` (`fix(bot-config-manifest): sign with S3SigV4Auth, drop the header allowlist`) — the delta since my previous full-round review at `267eb9b6`.

**Status checklist (all rounds, all 15 threads resolved):**

1. Directory-entry recursion semantics (spec.md) — fixed, resolved
2. OSS protocol example missing (spec.md) — fixed, resolved
3. `ref`/`resolved_sha` undocumented (spec.md) — fixed, resolved
4. `schema_version` v2/compat withdrawn (spec.md) — fixed, resolved
5. Credential registration examples missing (plan.md) — fixed, resolved
6. [Codex P1] git tool storage key collides on empty digest — fixed in `f6c1db8`, resolved
7. [Codex P1] `subpath` silently ignored on OSS file entries — fixed in `f6c1db8`, resolved
8. [Codex P2] `keep_last` couldn't restore a git-delivered tree — fixed in `f6c1db8`, resolved
9. OSS directory road fetched before validating `unpack`/`strip_components` — fixed in `267eb9b6`, resolved
10. Missing test for inline `source: {protocol: git, ...}` road in `cli_tools.py` — fixed in `267eb9b6`, resolved
11. `_support.py` unnecessary `TYPE_CHECKING` forward reference for `SourceDecl` — fixed in `743aa9f2` (direct import), resolved
12. `sources.py` — no worked example for the git `subpath` composition rule — fixed in `743aa9f2` (docstring example added), resolved
13. `sources.py` `parse_source(raw: Any)` should narrow away from `Any` — fixed in `743aa9f2` (`object` + mandatory `isinstance` narrowing; `_parse_protocol`'s callback typed too), resolved
14. `entry_fetch.py` redundant quoted `Mapping` annotations — fixed in `743aa9f2`, resolved
15. `credentials/signing.py` — unclear signer code, turned out to hide two real bugs: wrong signer (`SigV4Auth` instead of `S3SigV4Auth`, missing `x-amz-content-sha256`) and a header allowlist that would have dropped a required header even with the right signer — both fixed in `743aa9f2`, resolved

**New commit (`743aa9f2`) — fresh review:** read the full diff (`entry_fetch.py`, `credentials/signing.py`, `schema/_support.py`, `schema/sources.py`, `test_service.py`). Verified two things independently rather than taking the commit message at its word:

- The `_support.py` → `sources.py` import is not actually circular: traced the package's internal import graph (`sources` → `support_matrix` → `capabilities`; nothing in that chain imports `_support`) and confirmed by importing `schema._support` directly in a worktree at this commit — it succeeds, so the claim in the commit message holds.
- `S3SigV4Auth.add_auth` only ever injects `Authorization`, `X-Amz-Date`, `X-Amz-Content-SHA256` (and `X-Amz-Security-Token` when a token is present, which this credential type never carries) into `request.headers` — no `Host` or other header sneaks in via `dict(request.headers)`, so dropping the allowlist doesn't leak anything unintended. Matches the new `test_the_signature_carries_every_header_s3_requires` assertion.

No new correctness, security, or architecture-boundary issues found. The `_SIGNED_HEADERS` removal and blank-line-only diff in `entry_fetch.py`/`sources.py` are pure typing/doc cleanup with no behavior change.

**Open findings: 0.** PR is clean as of this round (`743aa9f2`). CI: all completed checks green; `Backend unit tests` and `Singlebox coverage` were still in progress at review time.

---

FreddieSun (approved):

PR #2022, opened 2026-09-08
fix(bcsfuse): use the shared profile store for fusion

wienyang:
## Problem
Profile fusion queried a separate local SQLite store even in MySQL runtime mode. As a result, it reported “profile not found” for profiles that existed and were active in MySQL, returning no recommendation.

## Solution
Resolve the profile content store from the application registry so fusion uses the same provider as profile CRUD and activation.

Fail explicitly when an application context exists but the provider is missing. Preserve standalone SQLite behavior when no application context is present.

## Validation
- All 43 related contract tests passed.
- Regression tests cover shared-provider reads, stale local cache, missing provider, and standalone behavior.
- `git diff --check` and the lint-only pre-push gate passed.
- The reporter deployed the fix and confirmed fusion works.

## Compatibility and risk
No database migration, configuration, frontend, or gateway changes are required. Redeploy BCSFuse to apply the fix.

### DIFF
--- a/src/bcsfuse/README.md
+++ b/src/bcsfuse/README.md
@@ -178,6 +178,10 @@ export LLM_REASONING_MODEL="claude-3-5-sonnet-20241022"
 
 BCSFuse requires MySQL 8.0+ for runtime mode. The schema is auto-created on startup.
 
+Profile fusion reads the `worker_profile_content_store` registered in the application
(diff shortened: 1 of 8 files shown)


archer910517-ctrl (approved):
LGTM

PR #2024, opened 2026-09-08
fix(repair): run target commands as admin

XIEAN99:
## Summary
- port the approved ARCA direct-access Repair change from REL20260908 to dev
- run every target Bot diagnostic read and approved container command as admin
- fail closed when the transport cannot safely switch to admin
- document the effective admin identity in the deep-diagnostics permission copy

## Verification
- Repair service tests: 307 passed
- full ClawWeb workspace type checks passed
- public credential boundary check passed
- no credentials or AK/SK added


3h-william (approved):
ok

PR #2039, opened 2026-09-08
fix(bcs): refresh undercover game panel artifacts

ziminz19:
## Problem
The undercover game panel was still reading session messages from the legacy endpoint and could silently omit node output failures. The published panel artifact also needed a version bump.

## Solution
- Refresh the panel from completed node artifacts instead of session messages.
- Surface node-output and non-JSON response failures to the panel.
- Update the package metadata and lockfile to version 1.4.1.
- Refresh the panel contract and integration fixtures.

## Validation
- `npm run typecheck`
- `npm run test:umd`
- `npm run test:focused`
- `npm run test:integration`
- OCB pre-push gate (lint-only mode)

## Compatibility and risk
The panel now expects the configured API base URL and reverse proxy to return JSON for game endpoints. Existing successful API responses remain compatible.


questchaos (dismissed):
LGTM

questchaos (approved):
LGTM

PR #2040, opened 2026-09-08
fix(workflow): keep run history independent of metric window

wienyang:
## Problem
Run history inherited the default 7-day metric window, hiding older runs despite a nonzero lifetime count.

## Solution
Query paginated run history without date bounds. Keep the 7/30-day window for metrics only, without resetting the history page.

## Validation
ClawWeb workspace build, workflow TypeScript check, two focused regression tests, and git diff --check passed.


3h-william (approved):
ok

ZhaosenLin (approved):
lgtm

PR #2048, opened 2026-09-08
fix(backend): tolerate concurrent BaaS resource uploads

FreddieSun:
## Problem
Concurrent overwrite uploads to the same BaaS Bot path can interleave as follows: both reads observe 404, one request creates the file, and the other request then receives HTTP 400 from `list(file_path)`. `BaasDeviceFileSystem.exists()` propagated that expected race as a public upload 500, which caused the upstream orchestration to enter its cleanup flow.

## Solution
Classify the Engine not-directory verdict at the BaaS transport boundary. When `exists()` encounters that verdict, it re-reads the exact path once. It returns `True` only when the concurrent writer is verified; otherwise it re-raises the original error.

The existing upload behavior remains last-writer-wins. `list_dir()` behavior, 404 absence handling, and 401/5xx propagation are unchanged.

## Validation
- `ruff check` on the three changed files: passed
- 150 focused Backend unit, adapter, contract, and architecture tests: passed
- Standards review: passed, no remaining findings
- Spec review: passed, no remaining findings
- Full Backend suite: 18,008 passed, 60 skipped; one unrelated task-discovery test crossed local midnight (`2026-09-08` to `2026-09-09`) and failed its date assertion, then passed immediately in a fresh process
- GitHub CI: in progress

## Compatibility and risk
No HTTP API, schema, database, or public protocol contract changes. The behavior change is limited to the verified read-404/list-400/read-success race. A genuine 400 is still re-raised unchanged. This does not add distributed locking or change the separately documented database-record concurrency boundary.

OCB enterprise code has no override of this filesystem implementation. Enterprise rollout still requires a separate OCB `ocb-public` gitlink bump and deployment after this PR is merged.

### DIFF
--- a/src/backend/src/agentclaw/community/core/devices/services/baas_device_filesystem.py
+++ b/src/backend/src/agentclaw/community/core/devices/services/baas_device_filesystem.py
@@ -18,7 +18,10 @@ import httpx
 
 from agentclaw.community.log import get_logger
 from agentclaw.community.core.devices.services.device_filesystem import DeviceFileSystem
-from agentclaw.community.core.devices.services.baas_invoke_transport import BaasTransport
+from agentclaw.community.core.devices.services.baas_invoke_transport import (
+    BaasTransport,
+    is_not_directory_error,
+)
 
 logger = get_logger()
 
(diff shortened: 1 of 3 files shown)


totalfrank (approved):
LGTM — reviewed `5bf26c9` (3 files: `exists()`'s single re-read on a 400 "not a directory" verdict, the new `is_not_directory_error` classifier, and the two added tests covering the recovered-race and still-genuinely-bad-400 paths). Traced the full `exists()`/`read_file()`/`list_dir()` control flow against this diff, including how the 400-vs-404 branches interact with the pre-existing `read_file`-first probe: the re-read only returns `True` when content is actually found, and a genuine 400 with no recovered file still propagates unchanged. No bugs found in the new code.

---

PR #2068, opened 2026-09-09
fix(workflow): reserve release numbers and resume interrupted publication

wienyang:
## Problem
Deploying a saved workflow can reuse an occupied deployment number (for example #9). The legacy API retries by changing only the version, and Git may already contain the tag when history insertion fails.

## Solution
Reserve a deployment number and version transactionally before Git publication, keyed by workflow and saved commit. Keep the history row pending and unavailable to run/rollback until the client confirms the exact tag and Pack are ready. Repeated reservations and completions are idempotent; finishing an older release cannot displace a newer active release. Legacy duplicate inserts return 409 without silently rewriting versions. Includes the unmerged version-metadata and snapshot-identity correction from #2054, with signed INTEGER bounds validation.

## Validation
- Avernet workspace build and check passed under Node 22.23.2.
- 28 focused workflow tests passed, including real HTTP/SQLite reservation, concurrency, retry, pending visibility, mismatch rejection and legacy collision tests.
- Actual ClawMind API client + Avernet Express routes + SQLite integration passed.
- Matching ClawMind changes passed 80 focused tests and build.
- git diff --check and repository pre-push gate passed.

## Compatibility and risk
Apply migration 120 (workflow_release_reservations) and deploy the server before updating ClawMind. Managed databases that prohibit application DDL need the SQL documented in the module contract. Existing history and tags are preserved; no production data was changed. Git publication remains outside the DB transaction and the authenticated client verifies it before confirmation. MySQL/ZDAS transaction behavior has not been exercised against a live production database.

## Spec
src/evolverun/clawweb/public/modules/workflow/docs/internal-version-contract.md

## Related issues
Includes the changes from #2054; this PR can deliver the complete server update.


3h-william (approved):
ok

PR #2069, opened 2026-09-09
fix(resources): show complete workspace tree

evilcoda:
## Problem

The resource browser hides dotfiles, system directories, and root identity files, and replaces nested local-skill paths with a synthetic `skills-local` root node. The current requirement is for both legacy and OpenAPI resource pages to show the workspace exactly as stored.

## Solution

- Preserve every directory entry returned by the device filesystem, including dotfiles and system paths.
- Remove the synthetic `skills-local` root projection so real `skills/skills-local` and `skills-pool/skills-local` hierarchies are visible.
- Keep path traversal checks, read-only write/delete policy, and directory-download filtering unchanged.
- Update shared-service, legacy-route, and OpenAPI-route tests.

## Validation

- `uv run ruff check` on all changed Python files: passed.
- Resource-related backend suites: 324 passed.
- Pre-push plaintext-secret scan and Python SAST: passed.

## Compatibility and risk

Both `/api/resources/files` and `/openapi/v1/bots/{bot_id}/resources` intentionally adopt the same unfiltered listing behavior. Response schemas and file operation permissions are unchanged.

### DIFF
--- a/src/backend/src/agentclaw/community/core/services/resource_file_service.py
+++ b/src/backend/src/agentclaw/community/core/services/resource_file_service.py
@@ -12,8 +12,8 @@ What the service owns (moved out of ``adapters/http/resources/file_router.py``):
 - resolve the device context (draft via ``resolve_for_bot``; publish-stage via
   ``resolve_for_binding`` — added in a follow-up task) and dispatch by the
   ``workspace`` namespace;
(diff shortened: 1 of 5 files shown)


totalfrank (approved):
LGTM.

PR #2071, opened 2026-09-09
fix(backend): prevent deletion of referenced Skill assets

### LINKED ISSUE #2061: fix(skill-center): prevent deletion of referenced Skill assets
## Problem Statement

Skill Installation is the source of truth for a Bot's Effective Capabilities, but several existing Skill hard-delete paths can destroy an `ac_skill` asset without first proving that no Installation, SkillSet Membership, Draft, Publication, Version, or Service Artifact lineage still references it. The result is a dangling Installation: the control plane says a Skill is effective while the asset can no longer be resolved or delivered to Runtime.

The defect is not limited to one HTTP endpoint. Generic deletion, Bot-local deletion, name-based cascade deletion, Git synchronization cleanup, and Bot lifecycle cleanup each reach a hard-delete primitive with incomplete blocker handling. A service-only precheck is insufficient because another request can add an Installation after the check and before deletion.

Pre-release auditing found eight dangling Skill Installations across three live OpenClaw Bots. All eight Skill assets are absent, the corresponding active SkillSet Memberships remain, and the current containers have no dangling symlink for those missing assets. The repair must stop future corruption and separately reconcile the historical database state without treating implicit deletion as Deactivation.

## Solution

Adopt a fail-closed Asset Deletion rule: a Skill asset may be irreversibly deleted only when it has no Blocking References. Asset Deletion must never implicitly deactivate a Bot or remove a saved Membership.

Every Skill hard-delete primitive and caller will enforce the same invariant. The delete transaction locks the Skill row, rechecks all database-backed blockers, and either deletes an unreferenced asset or raises the domain conflict `SKILL_ASSET_IN_USE`. Installation writers use the same Skill-first lock order, closing the check-then-delete race.

Center Skill Offline retains its existing impact preview, command-time recheck, and locked guard. Git synchronization that discovers a missing source while references remain preserves the Skill asset and Desired State, reports `SOURCE_MISSING_IN_USE`, and lets Runtime express the unavailable source through its existing pending state. No new persisted source-health state is introduced.

Historical dangling data is repaired through an explicit, audited operation after the prevention fix is deployed: export the exact affected rows, remove dangling Installations and Memberships that point to nonexistent Skills, and verify the affected live Bots through Effective Read and Runtime Projection. Because current container inspection found no dangling runtime entries for these eight assets, the correction does not delete container files.

## User Stories

1. As a Bot owner, I want deleting a Skill to be rejected while my Bot still uses it, so that an asset operation cannot silently remove an Effective Capability.
2. As a Bot owner, I want a SkillSet Membership to remain intact unless I explicitly remove it, so that deleting an asset cannot silently rewrite my saved capability organization.
3. As a Bot owner, I want a clear conflict response when a Skill is still referenced, so that I know which relationship must be removed first.
4. As a Bot owner, I want deletion conflicts to expose safe blocker counts, so that I can understand the scope without seeing another user's Bot identifiers.
5. As a Local Skill owner, I want to deactivate an active Local Skill before deleting it, so that Installation and Runtime converge before its content is destroyed.
6. As a Local Skill owner, I want inactive SkillSet Memberships to block deletion too, so that saved but currently disabled configuration is not silently discarded.
7. As a user of a shared Repo Skill, I want a temporary Git source disappearance not to automatically deactivate my Bot, so that a repository mistake does not cause an irreversible capability change.
8. As a Git synchronization operator, I want a referenced missing source reported as `SOURCE_MISSING_IN_USE`, so that the sync run distinguishes a blocked removal from a successful deletion.
9. As a Git synchronization operator, I want blocked removals excluded from the deleted count, so that operational metrics reflect what actually happened.
10. As a Runtime operator, I want a referenced Skill with unavailable source content to remain `PENDING`, so that the system can recover when the source returns.
11. As a Space Skill owner, I want the existing Offline impact preview to continue showing blockers, so that the product workflow does not change.
12. As a Space Skill owner, I want the Offline command to recheck blockers while committing, so that a stale preview cannot permit an unsafe lifecycle change.
13. As a Service Bot owner, I want published Artifact lineage to block destructive Skill removal, so that restart, scale, and rollback remain reproducible.
14. As a platform operator, I want all hard-delete primitives to enforce the same invariant, so that an internal caller cannot bypass an HTTP-layer check.
15. As a platform operator, I want the delete and activation paths to use one lock order, so that concurrent activation cannot create a dangling Installation.
16. As a platform operator, I want Bot deletion to explicitly remove that Bot's Installations before deleting Bot-owned assets, so that lifecycle cleanup does not weaken ordinary Asset Deletion.
17. As a platform operator, I want failed upload compensation to keep deleting newly created unreferenced assets, so that the new guard does not leak failed uploads.
18. As a platform operator, I want historical dangling Installations and Memberships exported before correction, so that the repair is auditable and reversible from evidence.
19. As a platform operator, I want the correction to avoid container writes when no dangling runtime entry exists, so that unrelated working Skills are not disturbed.
20. As a maintainer, I want Asset Deletion, Deactivation, Skill Offline, Membership, and Installation to remain distinct domain concepts, so that future changes do not recreate the same semantic shortcut.

## Implementation Decisions

- `Installation` remains the fact that a capability is currently effective for one Bot; absence is inactive. Runtime filesystem state is not Installation.
- `Membership` is a saved SkillSet reference and is a Blocking Reference even when the SkillSet is inactive.
- `Asset Deletion` is irreversible destruction of an unreferenced Skill. It does not remove Installation or Membership as a convenience.
- `Deactivation` explicitly removes an Effective Capability and performs Runtime Projection while retaining the Skill asset.
- A Blocking Reference includes Installation, SkillSet Membership, Default Membership, Draft, Publication Attempt, published Version, and Service Artifact lineage.
- All generic, Bot-local, name-based cascade, and Bot-owned hard-delete primitives must fail closed when Blocking References exist. Callers cannot rely solely on a transport-layer precheck.
- The delete transaction locks the exact Skill before checking database-backed blockers. Skill activation and Membership addition continue locking the same Skill before creating references. The canonical order is Skill, then reference rows, then delete or insert.
- A blocked public deletion returns HTTP 409 with domain code `SKILL_ASSET_IN_USE`, a user-readable message, and blocker counts grouped by type. Cross-user Bot, owner, SkillSet, or Artifact identifiers are not exposed. A caller may receive identifiers only for resources it is already authorized to view.
- A missing Skill and an in-use Skill remain distinct outcomes. Repository contracts raise the domain conflict for the latter instead of collapsing it into `False`.
- Center Skill Offline is not reimplemented. It continues to use the existing impact query for product preview and the existing command-time plus locked rechecks for authority.
- When Git synchronization finds that a Repo Skill source disappeared, it may hard-delete the Skill only if the common Asset Deletion guard finds no blockers. Otherwise it preserves the asset and Desired State, increments failure or blocked accounting, returns `SOURCE_MISSING_IN_USE`, and emits an actionable operator log.
- Missing Git source health is not persisted in a new database column. Runtime continues to use its existing `MANAGED_SOURCE_MISSING` and `PENDING` result until the source is restored or references are explicitly removed.
- Bot deletion is a separate lifecycle workflow. It explicitly deletes that Bot's Skill and MCP Installations before deleting Bot-owned Skill assets. It does not bypass or weaken the common Asset Deletion rule.
- Upload compensation remains allowed to delete the newly created Skill because no reference should exist. If a reference unexpectedly exists, compensation fails visibly rather than deleting a referenced asset.
- The deprecated legacy SC batch-delete surface returns HTTP 410. Coordinating an external destructive SC call with local references cannot be made atomic without a durable reservation, so the unsafe endpoint is retired rather than guarded by a check-then-delete window.
- Legacy BFF Local deletion delegates to the recoverable LocalSkillDeleteService; Bot-not-ready now fails closed instead of using the old metadata-only fallback.
- No schema migration is required for the prevention fix.
- Historical correction is not an application startup migration and is not an automatic full-table scrub. It uses a frozen, reviewed manifest and records the result per row.
- The known pre-release manifest contains Installation IDs `1221`, `1222`, `1223`, `1455`, `2598`, `2599`, `2600`, and `2601`, spanning Skill IDs `1104067`, `1105159`, `1106303`, `1106304`, `1106607`, `1106608`, `1106684`, and `1108186`.
- Historical correction removes every current Membership that points to those nonexistent Skill IDs, including the additional live Membership for owner `271376` on Bot `default`. Deprecated backup tables are not an authority and are not used to reconstruct Skill assets.
- Container inspection is repeated immediately before correction. If a corresponding runtime entry unexpectedly appears, the run stops for that Bot instead of deleting or modifying the entry implicitly.
- The Avernet fix targets `dev`, and requires a separate OCB gitlink bump for enterprise deployment. OCB Corp currently has no independent Skill hard-delete implementation, but its integration and tests remain part of delivery validation.

## Testing Decisions

- Prefer the highest behavioral seam: public Local/Generic Skill deletion and Git synchronization responses. Tests assert externally visible status, domain code, counts, and retained state rather than private helper calls.
- Public deletion of a Skill with an Installation returns 409 `SKILL_ASSET_IN_USE`; the Skill, Installation, Membership, and runtime intent remain unchanged.
- Public deletion of a Skill referenced only by an inactive SkillSet also returns 409 and leaves the Membership unchanged.
- Public deletion of an unreferenced Skill succeeds and removes only that asset's permitted records.
- Response tests prove that cross-user blocker identifiers are absent while authorized self-owned identifiers may be present.
- Git synchronization with a missing unreferenced source deletes the asset and counts it as deleted.
- Git synchronization with a missing referenced source preserves the Skill and all references, reports `SOURCE_MISSING_IN_USE`, and does not count it as deleted.
- Runtime projection tests preserve the existing `MANAGED_SOURCE_MISSING` / `PENDING` semantics; this change must not introduce another runtime state.
- Repository transaction tests use real overlapping transactions where supported: deletion locks the Skill, concurrent activation or Membership addition waits, and the loser observes either a deleted Skill or a blocker without producing a dangling row.
- Repository tests cover all hard-delete primitives, not only the Git Sync caller.
- Bot-local delete tests cover Direct Installation, active Membership, inactive Membership, Default Membership, and the no-reference success path.
- Bot cleanup tests prove Installations are explicitly removed before Bot-owned assets and that cleanup remains idempotent.
- Upload compensation tests prove an unreferenced partially created Skill is removed and that unexpected blockers are surfaced.
- Center Offline impact and command tests remain green, including the locked blocker recheck, proving no parallel deletion policy was introduced.
- Tenant and environment isolation tests prove blockers are evaluated only in the current tenant and environment while never allowing a same-ID row in another scope to mask or authorize deletion.
- Architecture and ownership tests ensure Installation SQL remains owned by the canonical capability desired-state module and that transport adapters contain no deletion policy.
- OCB conformance and DI tests verify the enterprise composition continues to use the updated Avernet contracts without a Corp bypass.
- Pre-release correction validation records before/after counts, verifies the eight dangling Installations and all invalid Memberships are absent, verifies no other Installation changed, and checks the four affected live Bots through Effective Read and Runtime Projection.
- Production correction is preceded by a fresh read-only audit; pre-release row IDs or counts are never assumed to apply to production.

## Out of Scope

- Building a new cross-Bot asynchronous Retirement workflow.
- Automatically deactivating every Bot when a shared Repo source disappears.
- Restoring deleted Skill assets from deprecated backup tables or incomplete historical metadata.
- Adding a persisted `source_status` field or a new Runtime state.
- Changing Center Skill Offline, publication, version, lease, or Artifact contracts beyond retaining their existing blocker semantics.
- Cleaning Installations belonging only to soft-deleted Bots.
- Cleaning unrelated stale Default SkillSet Memberships.
- Automatically running historical correction in every environment.
- Modifying current valid container symlinks or Skill content during the known pre-release correction.

## Further Notes

- The current production table has no foreign key from Installation to Skill, so database constraints neither reject Skill deletion nor cascade Installation cleanup. Application invariants are therefore mandatory.
- The known eight Installation rows were created between 2026-08-25 and 2026-08-28, before the current cutover observation window. They demonstrate historical corruption, while the present hard-delete primitives show the defect can still recur.
- Current pre-release inspection found all three affected Bots active and all active-root symlink targets readable; `find -xtype l` returned no dangling symlink. The missing assets are already absent from runtime projection because they cannot be resolved.
- Some invalid Skill IDs remain referenced by active SkillSets beyond the three Bots carrying dangling Installations. The correction manifest must therefore be generated from nonexistent Skill identity, not only from the eight Installation rows.
- Deployment, historical data mutation, and production correction are separate authorized stages. A merged or green PR is not evidence that prevention is deployed or that data has been repaired.

FreddieSun:
## Problem

Skill hard-delete primitives could remove `ac_skill` while Bot Installations or other durable references still existed. This created dangling `ac_bot_skill_installation` rows that could no longer resolve to a runtime Skill asset. Several legacy deletion paths also silently removed SkillSet membership or performed external/device work before the database deletion invariant was rechecked.

## Solution

- Add one fail-closed deletion invariant at `SkillRepository`: lock the exact Skill row, count durable blockers, and reject deletion with `SKILL_ASSET_IN_USE` when Installation, SkillSet membership, legacy grants, Space binding/grants, Draft facts, Versions, or Publication Attempts still reference the asset.
- Reuse the same guard from generic, Bot-local, Git-sync, legacy cascade, and Bot cleanup paths.
- Make Bot cleanup explicitly purge that Bot's Skill/MCP Installations before deleting owner-scoped SkillSets and Local Skills; stop cleanup if the purge fails.
- Route BFF Local deletion through the canonical Local delete service so storage quarantine and DB deletion compensate safely. Unknown historical locators fail closed before device I/O.
- Preserve source-missing Repo assets when they remain referenced and return structured Git-sync blocker details.
- Permanently retire the unsafe legacy batch-delete surfaces with HTTP 410.
- Keep Space Skill identity when any Publication Attempt exists, including failed attempts.
- Map the conflict to an explicit OpenAPI/BFF 409 response and document the invariant and ownership boundary.

## Validation

- Focused post-rebase suite: `202 passed, 27 skipped`.
- Broader affected suites before final rebase: `1754 passed, 27 skipped`.
- Backend full suite reached completion once: `18167 passed, 60 skipped, 2 failed`; both failures were caused by this change (safe offline-Local delete contract and module-size gate), were fixed, and their exact tests now pass.
- A second full local run was stopped at `5423 passed, 15 skipped` at the user's request to create the PR and use GitHub CI as the final full gate.
- Changed Python Ruff: passed.
- `git diff --check`: passed.
- Standards/Spec dual review: no remaining P0-P2 findings.

Real overlapping `SELECT ... FOR UPDATE` behavior cannot be proven by the local SQLite suite; row-lock ordering is covered structurally and should be exercised by the relational integration/CI environment.

## Compatibility and risk

- No schema migration and no historical data mutation.
- Unreferenced canonical assets remain deletable.
- Referenced assets now return a deterministic conflict instead of being deleted.
- A Local Skill whose storage cannot be safely located is retained rather than metadata-only deleted.
- Legacy batch-delete endpoints now return 410 and require callers to migrate to canonical per-asset lifecycle APIs.
- OCB contains no independent Corp hard-delete implementation for this seam; enterprise deployment still requires the normal OCB gitlink bump after this PR is merged.

## Spec

- `docs/adr/0004-block-withdrawal-when-bots-reference-skill.md`
- `src/backend/src/agentclaw/community/core/skill_center/CONTEXT.md`

## Related issues

Closes #2061

### DIFF
--- a/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
+++ b/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
@@ -2404,11 +2404,56 @@ async def delete_skill(
         ("git://", "center://")
     )
     if is_governed_content_source:
-        service = skill_service_factory.create()
+        persisted_owner_id = str(skill.get("user_id") or "")
+        persisted_bot_id = str(skill.get("bolt_id") or "")
+        collaborator_authorized = bool(
+            ctx.metadata.get("skill_delete_collaborator_authorized")
+        )
+        if persisted_owner_id:
+            if not persisted_bot_id:
+                raise HTTPException(
+                    status_code=409,
+                    detail="Skill is missing its Bot identity",
+                )
+            if bot_id and bot_id != persisted_bot_id:
+                raise HTTPException(
+                    status_code=409,
+                    detail={
+                        "error_code": "SKILL_BOT_CONTEXT_MISMATCH",
+                        "message": "删除技能必须使用 Skill 持久化归属的 Bot",
+                        "bot_id": persisted_bot_id,
+                    },
+                )
+            scoped_bot = bot_repo.get_by_id_and_owner(
+                persisted_bot_id, persisted_owner_id
+            )
+            if not scoped_bot:
+                raise HTTPException(status_code=404, detail="Bot not found")
+            active_engine = str(
+                scoped_bot.get("active_engine") or DEFAULT_ENGINE_TYPE
+            )
+            if engine_type and engine_type != active_engine:
+                raise HTTPException(
+                    status_code=409,
+                    detail={
+                        "error_code": "SKILL_ENGINE_CONTEXT_MISMATCH",
+                        "message": "删除技能必须使用 Bot 当前的生效引擎",
+                        "active_engine": active_engine,
+                    },
+                )
+            service = skill_service_factory.create(
+                bot_id=persisted_bot_id,
+                bot_owner_id=persisted_owner_id,
+                engine_type=active_engine,
+            )
+        else:
+            service = skill_service_factory.create()
         try:
             success = await service.delete_skill(
                 skill_id,
                 user_id=current_user_id,
+                authorized_bot_owner_id=persisted_owner_id or None,
+                collaborator_authorization_verified=collaborator_authorized,
             )
             if not success:
                 raise HTTPException(status_code=404, detail="Skill not found")
--- a/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
+++ b/src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
@@ -94,20 +94,29 @@ from agentclaw.community.api.bot_service import BotServiceProtocol
 from agentclaw.community.api.skill_set_management_service import (
     SkillSetManagementServiceProtocol,
 )
+from agentclaw.community.api.local_skill_delete_service import (
+    LocalSkillDeleteServiceProtocol,
+)
 from agentclaw.community.core.repository.protocols.bot import BotRepository
 from agentclaw.community.core.bot_management.services.engine_resolver import (
     resolve_engine_for_bot,
     resolve_runtime_engine_for_bot,
 )
 from agentclaw.community.core.skill_center.errors import (
+    LocalSkillActiveError,
+    LocalSkillEditBusyError,
+    LocalSkillEditLockUnavailableError,
+    LocalSkillEditPausedError,
+    LocalSkillLayoutRollbackError,
     LocalSkillNotFoundError,
+    LocalSkillNotReadyError,
     LocalSkillRuntimeSyncError,
-    SkillDeleteConsistencyError,
+    LocalSkillStorageError,
+    SkillAssetInUseError,
     SkillReferencedBySkillSetError,
     SkillRuntimeNameConflictError,
     SkillSetControlPlaneConflictError,
 )
-from agentclaw.community.core.bot_management.errors import BotLookupAmbiguousError
 from agentclaw.community.core.skills_pool.edit_guard import (
     SkillsPoolEditGuard,
     SkillsPoolEditLockUnavailableError,
@@ -2360,12 +2369,12 @@ async def delete_skill(
     ctx: RequestContext = Depends(get_request_context),
     skill_repo: SkillRepository = Injected(SkillRepository),
     bot_repo: BotRepository = Injected(BotRepository),
-    path_factory: WorkspacePathFactory = Injected(WorkspacePathFactory),
     skill_service_factory: SkillServiceFactoryProtocol = Injected(
         SkillServiceFactoryProtocol
     ),
-    resolver: DeviceContextResolver = Injected(DeviceContextResolver),
-    edit_guard: SkillsPoolEditGuard = Injected(SkillsPoolEditGuard),
+    local_delete: LocalSkillDeleteServiceProtocol = Injected(
+        LocalSkillDeleteServiceProtocol
+    ),
 ) -> MessageResponse:
     """Delete a skill.
 
@@ -2416,6 +2425,15 @@ async def delete_skill(
                     "skill_set_ids": e.skill_set_ids,
                 },
             ) from e
+        except SkillAssetInUseError as e:
+            raise HTTPException(
+                status_code=409,
+                detail={
+                    "error_code": "SKILL_ASSET_IN_USE",
+                    "message": "请先解除该 Skill 的生效或能力集引用，再删除 Skill",
+                    "blockers": e.blocker_counts,
+                },
+            ) from e
         except ValueError as e:
             error_msg = str(e)
             if "无权删除" in error_msg or "Permission denied" in error_msg:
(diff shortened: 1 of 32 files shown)


src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
```diff
@@ -2435,124 +2453,69 @@ async def delete_skill(
...
+                status_code=502, detail="Skill storage operation failed"
+            ) from e
 
+    raise HTTPException(
+        status_code=409,
+        detail={
+            "error_code": "SKILL_SOURCE_IDENTITY_INVALID",
+            "message": "历史 Skill 缺少可安全删除的内容定位，请联系管理员处理",
+        },
+    )
```
totalfrank, line 2563: **Bug:** `DELETE /api/skills/{skill_id}` now only handles `git_path` starting with `local://` (via `local_delete.delete_local_skill`) or the shared-market case (`user_id` unset + `git://`/`center://`). Every other reachable state — a Skill the caller owns whose `git_path` is `git://...` or `center://...` — falls through to this final branch and always returns 409 `SKILL_SOURCE_IDENTITY_INVALID`, regardless of whether it has any blocking references.

**Fails when:** a user publishes a Space Skill (`skill_publish_service.py` sets `git_path = f"center://{skill_uuid}"` on the existing owned row while leaving `user_id` set once `local_status == "PUBLISHED"`, see lines ~168-171) and later tries to delete it via this endpoint with zero references (no Installation/Membership/Draft/etc.). Before this PR, the route built a full `SkillService` via `skill_service_factory.create(active_dir=..., repo_dir=..., local_dir=..., ...)` and called the generic `service.delete_skill(...)`, which handles any `git_path` scheme (see `SkillService.delete_skill` in `skill_service.py`, still exercised directly by `test_skill_service_delete.py::test_delete_preflights_the_canonical_asset_guard` with `git_path="center://uuid"` + `user_id="u-1"`). After this PR, that generic call path is gone from the HTTP route, so the same request now always 409s with "请联系管理员处理" instead of deleting the unreferenced skill.

The same gap applies to any user-owned `git://...` skill (see the `git-default` / `user_id="owner"` fixture in `test_skills_endpoints.py`).

**Fix:** either route non-`local://` owned Skills back through the generic `SkillService.delete_skill(...)` path (it already calls `require_unreferenced_for_delete` and now fails closed on `SkillAssetInUseError`), or, if this restriction is intentional, that needs to be a deliberate product decision rather than a byproduct of the routing rewrite — worth confirming with the author.

---

totalfrank (changes requested):
Reviewed `184ace0` (32 files). One finding: the rewritten `DELETE /api/skills/{skill_id}` route in `skill_center/skills.py` drops the generic (non-`local://`, owned) delete path, so owned `git://`/`center://` Skills (e.g. a published Space Skill) can no longer be deleted via this endpoint even with zero blocking references.

---

src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
FreddieSun: Good catch — the compatibility finding is valid. Legacy BFF can persist owner-owned `git://` Skills, and the legacy publish flow converts an owner-owned row to `center://` without clearing `user_id`; both were incorrectly falling through to `SKILL_SOURCE_IDENTITY_INVALID` after the route rewrite.

Fixed in `064ebb024` with the narrower safe boundary:

- all recognized `git://` / `center://` assets, shared or owner-owned, enter the guarded generic delete path;
- authorization and `require_unreferenced_for_delete` remain mandatory, with the repository repeating the blocker check under the Skill row lock;
- Repo/Center sources never resolve DeviceFS or delete an active/local entry merely because `user_id` is populated;
- `local://` remains on `LocalSkillDeleteService`, and empty/unknown locators still fail closed with 409.

Added route and service regression coverage for owner-owned Git and Center rows and verified no DeviceFS calls. Focused validation: 174 passed, 17 skipped; the narrower route/service set is 59 passed.

src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
```diff
@@ -2391,10 +2400,10 @@ async def delete_skill(
     if not skill:
         raise HTTPException(status_code=404, detail="Skill not found")
     persisted_git_path = str(skill.get("git_path") or "")
-    is_shared_source = not skill.get("user_id") and persisted_git_path.startswith(
+    is_governed_content_source = persisted_git_path.startswith(
         ("git://", "center://")
     )
-    if is_shared_source:
+    if is_governed_content_source:
```
totalfrank, line 2406: **Bug:** The `064ebb024` fix for the previous finding widened `is_governed_content_source` to also cover *owned* `git://`/`center://` Skills, but it still calls `service.delete_skill(skill_id, user_id=current_user_id)` (a few lines below) without `authorized_bot_owner_id`/`collaborator_authorization_verified`. This drops the collaborator-authorization path that `SkillService._can_delete_skill` requires for a non-owner to delete an owned Skill.

**Fails when:** an ADMIN-role Bot collaborator (not the Skill owner) deletes an owned, unreferenced `center://` Skill (e.g. a published Space Skill on a Bot they collaborate on — the route's own docstring says this should work: "协作者（admin 角色）在持锁时可删除"). The route-level `_FailClosedSkillMutationPermissionInterceptor` authorizes them and sets `ctx.metadata["skill_delete_collaborator_authorized"] = True`, but this branch never reads that flag or resolves `bot.owner_id` to pass through. Inside `_can_delete_skill` (skill_service.py:2099), the caller is not `skill_admin()`, not `skill.user_id`, and the third (collaborator) branch requires `authorized_bot_owner_id` to be truthy — which is `None` here — so it returns `False`, and the route raises `ValueError` → HTTP 403 "无权删除此技能：您不是该技能的创建者，且没有管理员权限", even though the interceptor already authorized the request.

Contrast with the sibling `local://` branch further down, which re-derives authorization independently inside `LocalSkillDeleteService._authorize` (calling `self._collaborators.check_collaborator_permission(...)` when `user_id != owner_id`) — collaborators still work there. Only this governed-content branch lost collaborator support in the rewrite.

The new regression test `test_owner_can_delete_unreferenced_governed_content_without_device_io` only exercises the *owner* deleting their own Skill, so it doesn't catch this.

**Fix:** resolve the Bot (as the pre-PR code did) and pass `authorized_bot_owner_id=str(bot.get("owner_id") or "")` and `collaborator_authorization_verified=bool(ctx.metadata.get("skill_delete_collaborator_authorized"))` into `service.delete_skill(...)` for this branch too, or otherwise re-derive collaborator authorization the same way the `local://` branch does.

---

totalfrank (changes requested):
Reviewed `064ebb024` (32 files, diffed against merge-base `227613a5`). One finding: the fix for the prior review's `SKILL_SOURCE_IDENTITY_INVALID` gap widened the governed-content delete branch to owned `git://`/`center://` Skills but dropped the collaborator-authorization plumbing (`authorized_bot_owner_id`/`collaborator_authorization_verified`), so an ADMIN-role Bot collaborator can no longer delete an owned, unreferenced governed Skill — a capability the route's own docstring and the sibling `local://` path still support.

---

src/backend/src/agentclaw/community/adapters/http/skill_center/skills.py
FreddieSun: Agreed — this was a real Phase 1 compatibility regression. The governed-content branch restored owner deletion but failed to forward the trusted ADMIN collaborator authorization already established by the fail-closed interceptor.

Fixed in `a7bb6eb81`:

- owner-backed `git://` / `center://` rows are resolved with the exact persisted `(bot_id, owner_id)` via `get_by_id_and_owner` (never global `get_unique_by_id`);
- the SkillService is bound to that persisted owner and active engine;
- `authorized_bot_owner_id` and the trusted `skill_delete_collaborator_authorized` flag are forwarded to `_can_delete_skill` for its existing four-way consistency check;
- caller-supplied `bot_id` / `engine_type` mismatches still fail closed;
- shared ownerless Skills remain restricted to global Skill admins, and Phase 2 Space assets remain blocked by Binding/Grant/Attempt/Version references.

Added coverage for ADMIN collaborator success, missing trusted flag returning 403, owner success, Phase governed sources no DeviceFS, and formal Space blockers remaining 409. Focused route/service/OpenAPI tests: 37 passed; broader delete/repository suite before push: 174 passed, 17 skipped. Ruff and diff checks pass.

totalfrank (approved):
LGTM.

PR #2080, opened 2026-09-09
fix(bot-config-manifest): let the deployment declare an internal object store

totalfrank:
## Problem

`SourceCredentialService` refuses an `oss_aksk` endpoint whose host resolves to a private address. That guard is deliberate — without it `https://169.254.169.254/` is a storable endpoint and the platform connects to it on the next apply.

An **internal** object store endpoint resolves exactly that way, and the service already carried the escape hatch for it: `endpoint_allow_hosts`, whose own comment says it is *"the deployment's transport allowlist — the same value the guarded fetcher takes, so a legitimately internal object store is declared in one place rather than two that can disagree."*

Nothing ever passed through it. The composition root constructed the service with `repository`, `vault` and `fail_closed` only, so the parameter took its `()` default in every profile, and the one place a deployment declares such a host — `user_config.bot_config_manifest.fetch_transport_allowlist` — governed the fetch road alone.

The consequence is not a degraded path but a closed one: a legitimately internal object-store endpoint was **unregistrable by any configuration**, because no yaml key reached the check that refused it. Found while registering a credential against an internal OSS endpoint during end-to-end testing of the manifest feature.

## Solution

Two commits.

**`376a155` — the wiring.** Inject `BotConfigManifestConfig` into the credential-service provider and pass `fetch_transport_allowlist` through as `endpoint_allow_hosts`. Both modules already live in the same injector and the config is already a singleton provider, so this is the value that existed reaching the consumer that documented needing it — no new config key, no second list to keep in sync with the fetcher's.

**`216ed22` — the provider moves to `ManifestFetchModule`.** The R9 file-size guard failed the first commit: 17 added lines took `bot_management_module.py` from 999 to 1015, past the 1000-line cap. With one line of headroom, no amount of trimming the addition would have fit, which left three options — an allowlist entry, splitting a 999-line DI module, or moving the provider to where it belongs.

The third is the one that explains the bug rather than working around it. `SourceCredentialService` lives in `core/bot_config_manifest/credentials/` and now reads `BotConfigManifestConfig`; both belong to `ManifestFetchModule`, whose docstring already says the manifest config cluster lives there because `config_module` sits at the cap too. The provider was in the bot-management module because it landed there with W4, and that distance is *why* the allowlist went unwired for a wave — the config was one module away, so passing it was never the obvious thing to do.

The provider moves whole with the three imports only it used; the repository binding stays with its repository siblings. `bot_management_module.py` lands at 968 and `manifest_fetch_module.py` at 682, both under the cap with no allowlist entry — that list is meant to shrink, not to grow for a five-line wiring fix.

## Validation

Local, before each push:

- `tests/community/core/bot_config_manifest/` + source-credential endpoint + `tests/community/architecture/` — 1318 passed
- `tests/community/di/` + the file-size guard — 348 passed
- `ruff check` clean on every touched file

The regression test was mutation-checked in **both** homes: removing the `endpoint_allow_hosts=` argument fails it on `assert () == ('store.corp.internal', 'objects.corp.internal')`, and it passes with the argument restored.

It asserts the value that **arrives** at the resolved service rather than that a keyword appears in the provider's source. That matters twice over: the source-text form would not have caught the original defect (an omitted argument with a default type-checks, resolves, and passes every unit test of the service — including the existing one that proves the allowlist works when it *is* passed), and it makes the test indifferent to which module binds the provider, which is what let it survive the move unchanged.

## Compatibility and risk

No schema, contract or migration impact. Behaviour is unchanged for every deployment that declares no `fetch_transport_allowlist` (the default stays `()`), and for those that do, the credential surface now honours the same declaration the fetch road already honoured. The provider move changes no binding key — both modules register together unconditionally in `di/container.py` — so resolution is identical. Rollback is reverting the two commits; nothing persists differently.

---
_Generated by [Claude Code](https://claude.ai/code/session_019dzBBLChdsLffbjx71e2zM)_

### DIFF
--- a/src/backend/src/agentclaw/community/di/modules/bot_management_module.py
+++ b/src/backend/src/agentclaw/community/di/modules/bot_management_module.py
@@ -240,13 +240,6 @@ from agentclaw.community.core.repository.implementations.bot.manifest_content im
 from agentclaw.community.core.repository.implementations.bot.source_credential import (
     SourceCredentialRepository,
 )
-from agentclaw.community.api.source_credential_service import (
-    SourceCredentialServiceProtocol,
-)
-from agentclaw.community.core.bot_management.token_vault import TokenVault
-from agentclaw.community.core.bot_config_manifest.credentials.service import (
-    SourceCredentialService,
-)
 
 from agentclaw.community.core.repository.implementations.bot.config_manifest_apply import (
     BotConfigManifestApplyLockRepository,
@@ -973,43 +966,3 @@ class BotManagementModule(Module):
         self, svc: DataInitService
     ) -> DataInitServiceProtocol:
         return svc
-
-    @singleton
-    @provider
-    @inject
-    def _source_credential_service_protocol(
-        self,
-        repository: SourceCredentialRepositoryProtocol,
-        vault: TokenVault,
-        manifest_config: cfg.BotConfigManifestConfig,
-    ) -> SourceCredentialServiceProtocol:
-        """W3 (#1471): the profile decides the fail-closed posture.
-
-        Production columns (corp, community) refuse credential writes when
-        the vault has no master key — TokenVault's plaintext passthrough
-        is right for local, catastrophic for tenant tokens at rest. The
-        local/test columns keep the permissive default; corp_test runs the
-        Mist-backed vault anyway and benefits from the same guard.
-
-        The transport allowlist is the SAME value ``GuardedFetcher`` takes,
-        from the same ``user_config.bot_config_manifest`` block. It has to
-        reach here too: the endpoint guard refuses a host that resolves to a
-        private address, which is precisely what an internal object store
-        endpoint does — and without this the deployment has no way to say so.
-        The escape hatch existed on the service and nothing passed through it,
-        so a legitimately internal endpoint was unregistrable by any
-        configuration, and the one place the deployment declares such a host
-        governed the fetch road only.
-        """
-        from agentclaw.community.di.profile import DeployProfile
-
-        fail_closed = DeployProfile.detect() in (
-            DeployProfile.CORP,
-            DeployProfile.COMMUNITY,
-        )
-        return SourceCredentialService(
-            repository,
-            vault,
-            fail_closed=fail_closed,
-            endpoint_allow_hosts=manifest_config.fetch_transport_allowlist,
-        )
--- a/src/backend/src/agentclaw/community/di/modules/bot_management_module.py
+++ b/src/backend/src/agentclaw/community/di/modules/bot_management_module.py
@@ -981,6 +981,7 @@ class BotManagementModule(Module):
         self,
         repository: SourceCredentialRepositoryProtocol,
         vault: TokenVault,
+        manifest_config: cfg.BotConfigManifestConfig,
     ) -> SourceCredentialServiceProtocol:
         """W3 (#1471): the profile decides the fail-closed posture.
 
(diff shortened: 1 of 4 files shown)


totalfrank (commented):
Reviewed commit `376a155` (the only commit on this PR).

Diff is clean — no findings.

**Files covered:**
- `src/backend/src/agentclaw/community/di/modules/bot_management_module.py` — the `_source_credential_service_protocol` provider gains `manifest_config: cfg.BotConfigManifestConfig` and threads `endpoint_allow_hosts=manifest_config.fetch_transport_allowlist` into `SourceCredentialService`.
- `src/backend/tests/community/di/test_bot_management_module_wiring.py` — new regression test asserting the value that arrives at the resolved service.

**What I checked:**
- **Correctness**: verified `BotConfigManifestConfig.fetch_transport_allowlist` is `tuple[str, ...]`, matching `SourceCredentialService.__init__`'s `endpoint_allow_hosts` parameter type exactly — no coercion needed.
- **Contract/wiring**: confirmed `manifest_config` is resolvable in this provider — the `@inject` decorator on the provider method means the injector supplies it; `BotConfigManifestConfig` is bound as a `@singleton @provider` in `ManifestFetchModule`, and both `BotManagementModule` and `ManifestFetchModule` are registered unconditionally together in `di/container.py`, so this doesn't introduce an unresolvable binding in any profile. `manifest_config` is already used as a sibling provider's parameter in the same file (line 747), so the pattern is established.
- **Test coverage**: the new test resolves the real `SourceCredentialServiceProtocol` through `test_injector` after rebinding `BotConfigManifestConfig`, and asserts against the constructed service's `_endpoint_allow_hosts` rather than the provider's source text — this would have caught the original defect (an omitted keyword argument that type-checks and passes every other unit test).
- **Blast radius**: only the DI composition root constructs `SourceCredentialService` outside of tests; the constructor parameter has a default, so no other call site needed updating.
- **Architecture boundaries**: no transport/framework leakage into core, no new environment access, no schema/contract change — this only completes existing config plumbing between two modules that already share an injector.

Nothing outside this diff was in scope, and I didn't find anything in scope to flag.

---

xianmuyq (dismissed):

FreddieSun (approved):
LGTM

totalfrank (commented):
Reviewed the new commit `216ed22` (moves the provider added in `376a155` from `bot_management_module.py` to `manifest_fetch_module.py`, to stay under the R9 1000-line file cap).

**What changed:** `_source_credential_service_protocol` is removed whole from `bot_management_module.py` (47 lines) and re-added as `source_credential_service` in `manifest_fetch_module.py` (48 lines), with its regression test moved from `test_bot_management_module_wiring.py` to a new `test_manifest_fetch_module_wiring.py`. The `endpoint_allow_hosts=manifest_config.fetch_transport_allowlist` wiring from the previous commit is preserved intact through the move.

**What I checked:**
- `SourceCredentialServiceProtocol` (used as the new provider's return type) is not among the imports added by this commit, but it was already imported in `manifest_fetch_module.py` at head (pre-existing, consumed by `manifest_entry_fetcher`) — no `NameError` risk.
- `SourceCredentialRepositoryProtocol`'s binding (`to=SourceCredentialRepository`) stays untouched in `bot_management_module.py`, so the moved provider can still resolve it across modules — consistent with `BotManagementModule` and `ManifestFetchModule` being registered together in `di/container.py`.
- No dangling references to the removed `_source_credential_service_protocol` name anywhere else in the tree.
- `test_bot_management_module_wiring.py` is restored to byte-identical with the base branch (the test added in `376a155` was fully removed, not left stale or duplicated).
- Line counts match the commit message's claim: `bot_management_module.py` 968 lines, `manifest_fetch_module.py` 682 lines, both under the 1000-line cap.
- The moved test still asserts against the resolved service's `_endpoint_allow_hosts`, not source text, so it still would have caught the original defect.

Nothing to flag — this is a mechanical, correct move.

## Status

**Commits reviewed this round:** `216ed22` (new since last review, which covered `376a155`)

**Findings checklist:**
- Round 1 (`376a155`): no findings — clean.
- Round 2 (`216ed22`): no findings — clean.

**Open findings: 0**

PR is clean as of `216ed22`.

---

PR #2087, opened 2026-09-09
fix(frontend): allow users to change their message view and inherit g…

quechao76:
## Problem

Users could not change their own message view in Group or Session management unless they were viewing as the group owner. Creating a session with the main button also forced `full`, overriding the user's group-level setting.

## Solution

- Allow users to switch their own Human member view between `full` and `participant` in both management panels.
- Preserve existing group-owner controls.
- Make the main New Session button inherit the user's group-level view.
- Keep explicit view selection available in the dropdown.
- Apply inheritance to both chat and service-invocation sessions.

## Validation

- Passed 10 regression tests covering self-service controls, owner controls, default inheritance, and explicit overrides.
- Passed ESLint for changed files.
- Passed `git diff --check`.
- Re-ran all 10 regression tests after rebase.
- Browser save and creation flows were not manually verified.

## Compatibility and risk

No API or database schema changes. Existing sessions retain their current view settings. New sessions use the existing backend inheritance behavior when no explicit view is supplied.

### DIFF
--- a/src/frontend/src/pages/GroupChat/components/CreateSessionControl.tsx
+++ b/src/frontend/src/pages/GroupChat/components/CreateSessionControl.tsx
@@ -4,8 +4,7 @@ import {
   PopoverContent,
   PopoverTrigger,
 } from '@/components/ui/popover';
-import { cn } from '@/utils/utils';
-import { Check, ChevronDown, Plus } from 'lucide-react';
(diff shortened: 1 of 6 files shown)


ClaaaaaaaaaireR (approved):
LGTM

PR #2089, opened 2026-09-09
refactor(backend): make the manifest oss road a core module over oss2

totalfrank:
## Problem

The manifest `oss` transport was the only one of the three fetch roads that went through `plugin_api/`: `ObjectStoreClientFactory` (a `Plugin` Protocol), an S3/boto3 community implementation, an in-memory local implementation, and a DI module to select between them. `bot_config_manifest/fetch/` already holds `GuardedFetcher` (HTTPS) and `GitCheckout` (git) as ordinary core classes. The plugin seam existed so a deploy profile could select a different implementation, and after this change there is only one.

Live testing also proved the S3 implementation unusable against the deployment that matters: boto3 signs AWS SigV4, the Aliyun internal endpoint serves the native `OSS4-HMAC-SHA256` protocol, and every read answered HTTP 400 `InvalidArgument`, "Invalid signing region in Authorization header". `oss2` is public (Apache-2.0, PyPI), so nothing forces the native implementation out of this repo.

Two report defects and one latent credential bug rode along, each of which cost a real debugging cycle:

1. The refused-request `detail` carried no error code, so `InvalidArgument`, the one word that identified the signing bug, never reached the apply report.
2. A local SDK error (`oss2.exceptions.ClientError`: missing region, empty AK/SK, malformed endpoint, empty bucket) was reported as "the object store refused the request", sending an operator to the bucket policy for a problem in their own credential row.
3. `REQUIRED_FIELDS_BY_TYPE[OSS_AKSK]` did not include `region`, but `oss2.AuthV4` raises `ClientError: The region should not be None in signature version 4.` on every read (verified against oss2 2.19.1). A credential without a region was accepted at `PUT` and failed every apply, violating this module's one rule: the surface never accepts something it cannot apply.

## Solution

**One core module, one implementation.** `core/bot_config_manifest/fetch/object_store.py` now carries `ObjectStoreTarget`, `ObjectFetchStatus`, `ObjectFetchResult` and `REFUSAL_STATUSES` (docstrings preserved: the refusal-vs-failure ruling, the redacting `__repr__`, why this is not `ObjectStoragePlugin`) plus `AliyunObjectStore`, an `oss2.AuthV4` client with 256 KiB streaming reads and a `_classify` that maps oss2 types, statuses and codes to `ObjectFetchStatus`. The module docstring that argued S3 signing was the road is rewritten to say why the native protocol is the only one.

**The seam is a plain one.** `AliyunObjectStore.get(target, key, *, byte_limit)` allocates one `oss2.Bucket` per call (one bucket, one endpoint, one credential, so a single apply can read two buckets under two credentials), and the class takes `_open` as its only constructor seam, the way `GuardedFetcher` takes a `Resolver`. `EntryFetcher` and `ManifestFetchModule` take the concrete class, exactly as they take `GuardedFetcher`. The in-memory factory is now `FakeObjectStore` in `tests/community/core/bot_config_manifest/apply/_fakes.py`, a test double rather than a shipped implementation; the 49 call sites across `apply/`, `managed_files/` and `creation/` are updated.

**Every consumer migrated.** `EntryFetcher.acquire_object`, `ManifestFetchModule` (a new `manifest_object_store` provider next to the `GuardedFetcher` one), `SourceCredentialBinding.object_store_target`, the `TestingSkillCenterModule` override, the endpoint test's seeded world and the 49 unit-test rigs all read through the new class. A repo-wide search finds no remaining reference to the plugin outside the historical spec directory.

**Deleted:** `plugin_api/object_store_client.py`, `plugins/community/object_store_client.py`, `plugins/local/object_store_client.py`, `di/modules/infrastructure/community/object_store_client.py`, its entries in `di/profile_modules.py`, the `TestingSkillCenterModule` binding, and the Rule 25 conformance suite `tests/community/contracts/test_object_store_client.py` (its rules are re-pinned against the real client in `tests/community/core/bot_config_manifest/fetch/test_object_store.py`, and the consumer-through-the-injector assertion moved to `tests/community/di/test_manifest_fetch_module_wiring.py`).

**Ported against the corp `AliyunOSSObjectStoreClient`.** The corp class was supplied after the first push and the client was reconciled with it: the same `_NOT_FOUND_CODES` / `_DENIED_CODES` / `_INVALID_REQUEST_CODES` / `_RETRYABLE_CODES` vocabularies and the same order of rulings (not found, denied, retryable, refused, else unavailable), so throttling (`QpsLimitExceeded`, 429, 408) is `UNAVAILABLE` and maskable, `PermanentRedirect` (a 301) is a refusal by code, `RequestTimeTooSkewed` and 401 are `DENIED`; the same `object 'key' in bucket 'name': …` report line; the same `(RequestError, RequestException, OSError)` handling at open and mid-stream; and `_close_quietly`, so a socket that fails to close cannot raise out of a read that already has its verdict. The corp-only `@plugin_impl(mode=Mode.PROD)` decorator and the factory wrapper are dropped, the per-target reader becoming a per-call `oss2.Bucket` inside `get`. What differs from corp is exactly the two fixes below and `region=target.region or None`, so an empty region trips the SDK's own precondition.

**The two `detail` fixes.** Every store-side sentence now appends the store's error code when one was sent (`the object was not found (NoSuchKey)`, `the credential was denied (SignatureDoesNotMatch)`, `the object store refused the request (InvalidArgument)`). The code is a closed vocabulary carrying nothing sensitive; the SDK message is still never copied. `oss2.exceptions.ClientError` gets its own sentence, "the request could not be built from the credential; check its endpoint, region and key pair", keeps `NOT_FOUND` (a refusal `keep_last` must not mask), and is ruled on first, ahead of the 4xx-range branch, because the SDK reports it with a negative status.

**The latent bug.** `region` joins `REQUIRED_FIELDS_BY_TYPE[OSS_AKSK]`, the API field description no longer claims "Omit to take that client's own default", and `ObjectStoreTarget.region` becomes a required `str`. A row written before this rule is refused at binding time, before any client is built, with a `CredentialError` that names the credential to rotate (the client also maps an empty region to the SDK's own precondition, as defence in depth).

**Constraints that survive:** the cap is enforced while streaming (`TOO_LARGE` after no more than `byte_limit` plus one chunk, never buffer-then-measure, the stream closed on every exit); `detail` and logs carry bucket, key, status and error code only; the refusal/failure split; and no store-side outcome raises out of the read (a `ClientError` at bucket construction is inside the same `try`).

**Architecture guards touched, stated plainly for review:**

- `tests/community/architecture/test_no_data_infra_vendor_in_core.py` forbids the bare tokens `Aliyun` and `oss2` anywhere under `core/`. This change names both in `core/bot_config_manifest/fetch/object_store.py` and the class name in `core/bot_config_manifest/apply/entry_fetch.py`, so both files are added to that guard's per-file allowlist with a justification: the `oss` road is a tenant-bucket transport, not a B3 data-plane concern, and with one implementation there is no neutral name to hide the vendor behind. If reviewers would rather keep core vendor-free by name, the alternative is a neutral class name plus a lazy import; the guard's substring match on `oss2` in the import line would still need the allowlist entry.
- `tests/community/architecture/test_community_only_boot.py` listed `oss2` among corp-only packages. It is now a declared community dependency, so it leaves that list; the community-only boot proof still blocks the remaining corp packages and still builds the community injector.

**Dependencies.** `pyproject.toml` adds `oss2>=2.18`; `boto3` stays (it backs `CommunityS3ObjectStorage`, the platform's own bucket). `uv.lock` gains `oss2`, `crcmod`, `aliyun-python-sdk-core` and `aliyun-python-sdk-kms`. The configured index (`mirrors.aliyun.com`) is not reachable from this environment, so the four entries were resolved against pypi.org and inserted with the registry and package URLs rewritten to the mirror's identical paths; hashes are index-independent. `uv export --frozen` parses the result, and the Backend unit tests job (which installs with `uv sync --frozen`) passed on it. A maintainer with mirror access can confirm with `uv lock --check`.

## Validation

Run in a venv installed from the lock's pinned versions plus oss2 2.19.1, from `src/backend`:

- `pytest tests/community/core/bot_config_manifest/fetch/test_object_store.py`: 44 passed. Pins every synthesised oss2 outcome (`NoSuchKey`, `NoSuchBucket`, `AccessDenied`, `SignatureDoesNotMatch`, bare 403 `InvalidAccessKeyId`, `RequestTimeTooSkewed`, bare 401, typed and bare 400 `InvalidArgument`, 301 `PermanentRedirect`, `ClientError`, `RequestError`, 429 `QpsLimitExceeded`, 408 `RequestTimeout`, 5xx, an unparseable response, and a raw `requests`/`OSError` at open) to its status and its exact `detail` sentence including both fixes; the streaming cap (a stream that serves 512 bytes per read, 100 KiB body, 4 KiB cap: at most cap plus one chunk pulled, content `None`, stream closed); mid-read failure; a `close()` that raises; the real opener with an empty region and with a malformed endpoint (no network, the SDK's own precondition); report safety (endpoint, secret and key id never in `detail`).
- `pytest tests/community/core/bot_config_manifest tests/community/di tests/community/architecture tests/community/contracts tests/community/endpoints/test_openapi_config_manifest_apply.py tests/community/endpoints/test_openapi_source_credentials.py` on the reconciled head: 2012 passed. (On the first push the same run showed 3 failures in `test_community_only_boot` and two `test_task3_review_regressions` cold-import cases, which spawn a fresh interpreter; they failed identically on unmodified `dev` here because the project was not importable outside pytest's `pythonpath`. With the package made importable to subprocesses they pass, including the community-only boot proof with `oss2` removed from its corp-only list.)
- The review fix (binding-time refusal of a regionless row): `pytest tests/community/core/bot_config_manifest/credentials tests/community/core/bot_config_manifest/fetch tests/community/core/bot_config_manifest/apply/test_source_fetchers.py tests/community/core/bot_config_manifest/apply/test_entry_fetch.py tests/community/endpoints/test_openapi_source_credentials.py tests/community/endpoints/test_openapi_config_manifest_apply.py`: 293 passed.
- Full `pytest tests/community -n 4` on the first push: 18159 passed, 63 skipped, 3 failed. The 3 are not this change's: two `test_bot_build_service_skill_artifact` cases fail with `rsync: not found` (the binary is absent in this environment) and fail identically on unmodified `dev`; `test_openapi_bot_owner_name::test_manifest_completion_accepts_pre_upgrade_job_without_nickname` failed only under the parallel run and passes with its file run alone (18 passed).
- `ruff check` on every touched file: clean (the two pre-existing E402 findings in `source_credentials/schemas.py` are untouched).
- CI: all checks green on the first two pushes, including Backend unit tests and Singlebox coverage (one Backend job hung on a dead runner and was re-run once).

Not run: `uv lock --check` (needs the Aliyun mirror); any read against a real OSS endpoint (no credentials or network here).

## Compatibility and risk

- **The community distribution loses its S3/MinIO/R2 object-store client.** Any deployment reading a manifest `oss` source from an S3-compatible store stops working: the native client answers such an endpoint with a refusal, never silently. This is a deliberate scope reduction, one implementation being sufficient for now, not an oversight.
- **Previously valid credential writes now fail.** `PUT /source-credentials/{name}` with `type: oss_aksk` and no `region` is refused with 422. Every such credential failed every apply before this change, so this is the correct direction. Existing rows stored without a region are not migrated: they keep failing at read time, now with a per-entry error that names the credential to rotate rather than a bucket to inspect.
- `ObjectStoreTarget.region` is a required `str`. A row stored before the rule (NULL region) is refused by `SourceCredentialBinding.object_store_target` with a `CredentialError` naming the credential to rotate, before any store client is built; the entry fails, the apply does not.
- `ObjectFetchResult.detail` sentences follow the corp report line (`object 'key' in bucket 'name': …`) with a trailing ` (<Code>)` where the store sent a code; the previous community sentences are gone with the S3 client.
- Rollback is a revert; no schema or config migration is involved.

## Spec

`specs/2026-09-08-manifest-oss-client-and-delivery/` (the plugin-seam design this change supersedes; left as the historical record).

https://claude.ai/code/session_012sHakPvRTa7e4WqX62Q78C


totalfrank (commented):
Reviewed commits `5f682b1` and `d63ceb3` (full PR diff against merge-base `a153449`) — no findings.

What I checked:
- **Correctness**: `AliyunObjectStore._classify`'s ruling order (client error → not-found → denied → retryable → invalid-request → catch-all-unavailable) against the real `oss2` exception hierarchy — confirmed `ClientError` and `RequestError` both subclass `OssError` and are ruled correctly ahead of the generic handler; verified `oss2.Bucket`/`AuthV4` accept the `region` kwarg and that an empty region trips the SDK's own precondition inside the same `try`, so it's reported as a credential-row refusal rather than propagating unhandled.
- **Streaming cap**: the chunked read/abandon logic in `get()` — cap enforced before buffering, boundary-inclusive, socket closed on every exit path (including a `close()` that itself raises).
- **Contract/API surface**: `REQUIRED_FIELDS_BY_TYPE[OSS_AKSK]` now includes `region` and is enforced in `_check_mechanism_fields`/`put` (422 on missing region), matching the new endpoint test; `ObjectStoreTarget.region` is required and `SourceCredentialBinding.object_store_target` passes `row.region or ""` for legacy NULL rows.
- **DI wiring**: `ManifestFetchModule.manifest_object_store` is bound in the always-present `container.py` (not gated by `profile_modules.py`), so removing `CommunityObjectStoreClientModule` from the community profile doesn't leave `AliyunObjectStore` unbound.
- **Cleanup completeness**: grepped for `ObjectStoreClientFactory`/`object_store_client`/`InMemoryObjectStoreClientFactory` across `src` and `tests` — no stale references after the deletions.
- **Architecture guards**: the `test_no_data_infra_vendor_in_core.py` allowlist additions and the `test_community_only_boot.py` corp-package-list removal match the two files/name that actually need them.
- **Test coverage**: `test_object_store.py`'s SDK-outcome matrix, the streaming/cap/close-failure cases, and `FakeObjectStore`'s credential-mismatch-as-DENIED behavior all look correctly targeted at the classification rules the module depends on.

Not independently re-run: the test suite itself (relying on the PR's reported pytest results) and `uv lock --check` (no Aliyun mirror access here, as the PR description notes).


---

src/backend/src/agentclaw/community/core/bot_config_manifest/fetch/object_store.py
```diff
@@ -0,0 +1,534 @@
...
+"""
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import StrEnum
+from typing import Any, Callable, Optional
+
+import oss2
+import oss2.exceptions as oss_exc
+from requests import RequestException
```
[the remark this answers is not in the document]
totalfrank: Not changing this: moving the client out of `plugin_api/` and into `core/bot_config_manifest/fetch/` is the deliberate scope of the PR, not a side effect, and the reasoning is in the description under "Solution" and "Architecture guards touched".

The specific points:

- **Precedent in this package.** `fetch/guarded_fetcher.py` drives `httpx` and pins connections itself; `fetch/git_source.py` shells out to `git`. The object store is the third transport of the same kind, and it was the only one behind a plugin seam. The AGENTS.md line about transport-agnostic core is what keeps FastAPI and routers out of core; the fetch package has always owned its wire clients.
- **The plugin layer's own admission rule.** `plugins/README.md`: a component belongs there "when different runtime profiles need different implementations". After this change there is one implementation and no profile selects anything, so a Protocol plus a local/prod pair would exist only to satisfy Rule 20's pairing check, which is the situation that README says not to create.
- **The contract did not go away, it lost its plugin wrapper.** `ObjectStoreTarget`, `ObjectFetchStatus`, `ObjectFetchResult` and `REFUSAL_STATUSES` are unchanged, and `AliyunObjectStore` keeps a narrow injectable seam (`_open`) exactly as `GuardedFetcher` takes a `Resolver`. Tests substitute the store on the injector the same way the endpoint test in this PR does; the in-memory implementation moved to the test tree because it is a test double, not something to ship.
- **The vendor-term guard.** `test_no_data_infra_vendor_in_core.py` is the B3 ratchet for the platform's own data plane (ZDAS / ZCache / the platform bucket). This client reads a tenant's bucket under the tenant's credential over a protocol only that vendor's SDK speaks, so its per-file allowlist entry names the reason and is called out in the PR description for the human reviewers to weigh. If they would rather record it as a formal waiver, that is a one-line follow-up; it does not change where the code lives.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service.py
```diff
@@ -420,7 +422,11 @@ def object_store_target(self, bucket: str) -> "ObjectStoreTarget":
             secret_access_key=self._service._vault.decrypt_or_passthrough(
                 row.secret_ciphertext
             ),
-            region=row.region,
+            # Required at PUT since signature version 4 landed; a row written
+            # before that rule carries NULL, and the empty string is how the
+            # store client learns to blame the credential rather than the
+            # bucket.
+            region=row.region or "",
```
totalfrank-personal: If region is always required, should we fail early here if it's empty rather than this?

totalfrank-personal (commented):
Comments posted, please take a look

src/backend/src/agentclaw/community/core/bot_config_manifest/credentials/service.py
totalfrank: Agreed, done in f1804437. `object_store_target` now raises `CredentialError("credential '…' has no region; an object store source needs one, so rotate the credential with its region")` before any client is built, which the oss road already turns into a per-entry `EntryFetchError` the same way it handles a wrong-type credential. Only a row written before `region` became required can reach that branch, so `ObjectStoreTarget.region` is now always the stored value rather than `row.region or ""`. Covered by `test_a_row_stored_before_region_was_required_is_refused_at_binding`, which writes the legacy row straight to the repository since the service no longer accepts one.

---

totalfrank (commented):
Reviewed the new commit `f180443` (against merge base `d63ceb3`, where my last full review landed) — no findings.

**Part A — open findings:**
- Codex P1 ("keep the OSS transport out of core") — already resolved; the thread shows this was disputed with the PR's scope rationale and the thread is marked resolved. No new information in this commit changes that.
- @totalfrank-personal's review question ("should we fail early here if it's empty rather than this?" on `credentials/service.py:429`) — **fixed** in `f180443`: `object_store_target` now raises `CredentialError` when `row.region` is falsy, before `ObjectStoreTarget` is constructed, and `region=row.region or ""` became `region=row.region` since the guard makes it always truthy past that point. Verified the new `CredentialError` is caught the same way the existing wrong-credential-type check is (`source_fetchers.py:179` wraps the `object_store_target(...)` call in `except CredentialError as exc: raise EntryFetchError(str(exc)) from exc`), so this becomes a normal per-entry failure naming the credential, exactly as claimed. Covered by the new `test_a_row_stored_before_region_was_required_is_refused_at_binding`. Thread is already replied-to and resolved — nothing further needed.

**Part B — new commit review (`f180443`):**
- The guard is placed before any client/target construction and checks falsiness (`not row.region`), which correctly covers both a legacy `NULL` and a legacy empty-string region.
- No stale reads: confirmed no other call site still relies on the old `row.region or ""` fallback for building a client target (the one other `region=row.region` in `_masked()` is just the read-back record, unrelated to signing).
- New test writes the legacy row straight through the repository (bypassing the service, which already refuses a regionless row at `PUT`), correctly targeting the one path — an old row — that can still reach this code.

## Status
- Commits reviewed this round: `f1804437`
- Findings checklist:
  1. Codex P1 (core boundary) — author's call / conceded by prior reply, resolved
  2. Region-empty-string question (`credentials/service.py:429`) — fixed in `f1804437`, resolved
- Open findings: **0**

PR is clean as of this round.

---

totalfrank:
**CI status note: the "Backend unit tests" job on `f1804437` is hung, and the hang is not this PR's.**

- Job [102694289580](https://github.com/inclusionAI/Avernet/actions/runs/34420385427/job/102694289580) started at 00:13Z; its streamed log stops at 00:20:52Z with all four xdist workers going silent in the same second at 67% progress, on tests unrelated to this diff. No test failed. The same thing happened on the previous head (`d63ceb3f`): the job froze at 67% at 16:53Z, and after I cancelled and re-ran it once, it passed in 11 minutes.
- The same stall occurs on unrelated branches: run [6121](https://github.com/inclusionAI/Avernet/actions/runs/34370231302) on `codex/default-mcp-membership-guard` had its Backend unit tests job hang for 74 minutes (15:26Z to 16:40Z, cancelled) and pass on attempt 2 in 10 minutes; runs 6114 and 6111 on other branches also needed a second attempt. So this is a runner-side stall in the Unit Tests workflow, not a failure in code this PR touches.
- The full backend suite passed on this PR's first push, on the re-run of its second, and locally (`pytest tests/community -n 4`: 18159 passed, with the three environment-only failures noted in the description). The changes since then are the `oss2` client reconciliation and the binding-time `region` guard, both covered by suites that pass locally.

I have used the one re-run I allow myself on this PR, so I am not cancelling and re-queuing this job again; it will either complete or hit the workflow's default timeout. A maintainer can re-run the failed job from the Actions page if they would rather not wait, and I will keep checking the PR in the meantime. No fix for the stall exists in this branch's scope; if the root cause is found in the workflow itself, I will port it here.

---

FreddieSun (approved):
LGTM

PR #2116, opened 2026-09-10
fix(clawevolve): use origin bot owner for automatic analysis

wienyang:
## Problem
Failed-run automatic analysis prioritizes `flow_runs.user_id` when selecting the identity used for Bot permissions and runtime lookup. Gateway-triggered runs can store `gateway-client` there, even though `origin_bot_id` contains the actual Bot and owner. This makes candidate selection fail before an analysis task is created.

## Solution
- Prefer the owner from a valid, non-empty `botId:ownerId` origin identity.
- Fall back to the existing run user ID when the origin owner is missing or malformed.
- Preserve the original run record, Bot-level authorization, and manual-analysis behavior.
- Add regression coverage for service callers, differing caller/owner identities, missing or malformed origin identities, non-failed runs, and the workflow switch.

## Validation
- Confirmed the gateway-caller regression fails before the fix.
- Focused Vitest suite: 15 tests passed.
- `npm run check --workspace @avernet/clawevolve`: passed.
- `npm run build --workspace @avernet/clawevolve`: passed (bundle-size warnings remain).
- `git diff --check` and the pre-push gate: passed.
- No production tasks were retriggered; this change has not been deployed.

## Compatibility and risk
Only the automatic-analysis owner-selection path changes. Bot permissions and runtime validation still apply. No database migration, public API change, or ClawMind/OCB source change is required. Existing failed runs are not backfilled automatically.

### DIFF
--- a/src/evolverun/clawweb/public/modules/clawevolve/server/services/evolve/__tests__/auto-analysis-bot-owner.test.ts
+++ b/src/evolverun/clawweb/public/modules/clawevolve/server/services/evolve/__tests__/auto-analysis-bot-owner.test.ts
@@ -1,0 +1,65 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createFailedRunAutoAnalysisObserver } from "../run-analysis-starter.js";
+
+afterEach(() => vi.restoreAllMocks());
+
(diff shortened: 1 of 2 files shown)


fudp (approved):
LGTM

PR #2117, opened 2026-09-10
fix(backend): scope restart template persistence to aicoding strategy

maicaisongxiaocong:
## Problem

The existing AICoding restart hook only persisted strictly higher-version snapshots and required an existing template row. Consequently:

- a published policy change at the same template version could not be persisted;
- an unversioned legacy Coding Bot, including one with no template row, could not save the frontend-composed migration snapshot through restart.

This is a deliberate change to accepted inputs on an existing pipeline, not a dormant/isolated addition. The `fix` title does not remove the need for release-policy review.

## Solution

Keep backend runtime changes exclusively in `AicodingProvisioningStrategy.apply_restart_extra_configs`:

- consume nonempty dictionary snapshots only for `aicoding` / `claude_code`;
- leave publication/version selection and composition to the paired frontend;
- accept equal/lower/unversioned snapshots and reuse the existing `create_or_update_template` API;
- preserve existing token encryption and authorization-resync markers;
- report an empty save result as a strategy error, without changing the caller's existing error handling.

`BotService`, `TemplateService`, the generic provisioning interface and HTTP router remain byte-identical to the original branch baseline. No changes to lifecycle ordering, restart locks, shared persistence, or database schema.

This revision adds documentation only: a revision-2 snapshot contract and an explicit supersession notice on the historical authorization proposal. It does not add backend version validation or alter runtime logic.

## Validation

- 212 relevant backend tests passed across strategy persistence, the unchanged template service/encryption, authorization refresh, restart locking and BaaS/env handling.
- Real `TemplateService` + in-memory repository coverage: both supported engines and historical Coding types; missing row / missing ext / null ext / empty ext; subsequent same-version writes; encryption and ciphertext preservation; non-target no-op; invalid input and save failures.
- Verified the shared runtime files above are unchanged from baseline. `git diff --check` passes.
- These are not browser-to-device end-to-end tests and do not prove arbitrary API clients always send complete snapshots. No production Bot was restarted or modified for verification.

## Compatibility and risk

**Request contract:** `extra_configs.template_config` is a complete merged snapshot, not a PATCH. `confirmed_template_update` is a sibling in `extra_configs`; the entire envelope is not persisted. Existing internal `_aicoding_restart` state may be stored with the snapshot for asynchronous resync.

**Frontend obligations:** choose visible published templates; normal full upgrades still require formal publication plus a higher version; initial migration handles unversioned legacy Coding Bots; independent `envs.DOMAIN_POLICY` sync accepts visible whitelist/formal publication when the key exists, including empty values. Even field-only updates must send the full merged snapshot. The paired frontend is delivered separately from this public backend PR.

**Affected callers:** calls without a snapshot keep the previous no-write behavior. Existing complete higher-version requests still save. Direct API clients lose the prior server-side monotonic-version rejection guarantee and must not send partial snapshots. Non-target engines do not consume this configuration.

**Failure policy:** the strategy propagates save errors, but the existing `BotService.restart_bot` hook catches them, logs a warning with Bot ID/exception/stack, and continues with persisted configuration. Therefore a restart may succeed without applying the requested update. This PR does not claim caller-visible persistence errors or fail-closed restarts. Authorization resync remains asynchronous and non-atomic with the save.

**Concurrency:** delayed full snapshots may overwrite newer stored configuration, including version, business settings and domain policy. There is no CAS/config revision or atomic upsert guarantee. The existing hook still runs before the restart lock. Avoid concurrent submissions; a server-side non-regression guarantee requires a separate concurrency design, not just client selection or documentation.

**Rollout/rollback:** verify the paired frontend before rollout. Reverting code does not restore previously overwritten snapshots or remove migrated rows; restoring data requires a verified per-Bot snapshot, not blanket row deletion.

**Release gate remains open:** the target remains `REL20260910`; no merged `dev` counterpart is referenced yet. This update does not claim a Gate 0 waiver or resolution. Maintainers must confirm the required release path, including dev-first review and a referenced backport if policy requires it. The target branch has not been silently changed.

## Spec

- Current contract: `src/backend/specs/2026-09-10-aicoding-restart-template-snapshot/spec.md` (revision 2).
- Historical proposal now explicitly superseded for restart snapshot semantics: `src/backend/specs/2026-08-20-template-update-consented-authorization/spec.md`.


totalfrank (changes requested):

## 2. Gate 0

- **Target branch:** `REL20260910` — matches `^REL\d{8}` → release branch, subject to the release-branch policy.
- **Classification:** Feature work that extends an existing, live pipeline (mislabeled by title as `fix`).
- **Result: FAIL.**

Reasoning:

- `apply_restart_extra_configs` is not new/dormant code. It is invoked unconditionally on every bot restart from production code: `bot_service.restart_bot` → `strategy.apply_restart_extra_configs(ctx, extra_configs, template_service=self._template_service)` (`src/backend/src/agentclaw/community/core/bot_management/services/bot_service.py:4605`). This fails Isolation-Audit point 1 outright — it is reachable through an existing pipeline, not net-new/unreachable code — so the path (b) exception does not even apply.
- The diff does not change a defect in a minimal, referenced way. It removes a pre-existing safety invariant (`incoming_version <= stored_version` reject) and switches persistence from update-only (`update_template`) to create-or-update (`create_or_update_template`), changing what happens for bots that previously had no template row. That is new behavior on an existing code path, not a bug fix restoring a known-good prior state.
- The PR body is the unfilled template (no `Problem`/`Solution`/`Validation` content), and there is no `dev` cherry-pick reference. `git log --all` shows this commit (`f0bc6b6f`) exists nowhere except this PR — there is no `dev` counterpart. Per the constitution's Gate 0 rule: "a release-branch PR with no `dev` counterpart is a finding on its own — the change would be lost at the next release cut."
- The branch name (`feat/restart-legacy-template-migration`) is honest about what the diff is; the PR title's `fix(...)` framing does not change that classification.

**Required action:** re-target this PR to `dev`. If any portion is later found to be a genuinely isolated, dormant addition, it can be re-submitted against a REL branch through the Isolation Audit — but no such portion exists in the current diff.

## 3. Findings

**Finding 1 — Gate 0: pipeline-touching change targets a release branch**
`src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:768-812` (`apply_restart_extra_configs`)
Violated rule: Gate 0 / release-branch policy (not waivable).
Why it breaks the process: this rewrites the persistence policy of a restart flow that runs on every bot restart in production on `REL20260910`, with no `dev` review trail and no defect citation. If merged as-is, the release branch and `dev` permanently diverge in behavior for this path, and the change is lost at the next release cut.
Required fix: retarget to `dev`; open a REL-branch PR only if/when the change is cut down to a minimal, referenced fix, or resubmit as a genuinely isolated feature per the Isolation Audit (not applicable to this diff as written).

**Finding 2 — Undocumented relaxation of a cross-component contract (Rule 1 / Rule 16)**
`src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:790-793`:
```python
# The frontend owns publication/version selection and full-snapshot
# composition. Same-version and unversioned snapshots are valid (e.g.
# published env policy refresh or a legacy bot's first migration).
incoming_version = self._template_version_id(candidate)
```
The prior code enforced `incoming_version` monotonicity as a backend-side safety check against a stale/out-of-order frontend write clobbering a newer stored snapshot. This diff deletes that check entirely and replaces it with an assumption — "the frontend owns publication/version selection" — asserted only in a code comment. That assumption is a contract between the frontend and this backend method (what "full-snapshot composition" means, when a stale write is safe to accept), but it is not captured as a documented, versioned contract anywhere (no spec/schema update, no propagation analysis of the `restart_bot` caller or other `template_config` consumers, no conformance test asserting the frontend actually always sends a full, authoritative snapshot). Per Rule 1, "implementations must not introduce undocumented semantics that other components depend on," and per Rule 16, contract-affecting changes must declare propagation scope.
Required fix: either keep a backend-side invariant check (defense in depth) or, if the frontend is now the sole authority, document that contract explicitly (spec/API doc) and add a test that pins the frontend's obligation, rather than relying on an inline comment.

## 4. Out-of-scope notes

- The new `raise RuntimeError("Failed to persist coding restart template snapshot")` is caught and merely logged by the existing broad `except Exception` in `bot_service.restart_bot` ("Optional engine extensions must never block the existing restart path"), so the stated intent of surfacing persistence failures as errors is not actually observable by callers of `restart_bot`. This is a business-logic/error-handling concern at the call site, not gated here.
- Version-guard removal semantics (whether accepting equal/lower/unversioned snapshots is product-correct) are business logic and not reviewed here beyond the contract-documentation gap in Finding 2.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

This is an architecture-gate review only (target-branch policy + layering). Feature/business-logic correctness of the restart-template flow is out of scope and not assessed here.

## Gate 0 — target branch

- **Target branch**: `REL20260910` (matches `^REL\d{8}`) — a release branch.
- **PR classification**: feature/behavior-change work extending an **existing** pipeline, mislabeled as `fix`. **FAIL.**

Reasoning, by the diff and not the title:

- `apply_restart_extra_configs` in `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py` is already live for every `aicoding`/`claude_code` bot restart (gated only by `TEMPLATE_CONFIG_CONSUMING_ENGINES`, unchanged by this diff). It is not a dormant/net-new/unreachable entry point, so it can never qualify for the isolated-new-feature exception (path (b)) — it's squarely an existing pipeline.
- The diff removes the version-monotonicity safety check entirely (`stored_version`/`incoming_version` comparison and the `incoming_version < 0` / `<= stored_version` early-returns are deleted), and switches the persistence call from `update_template` (update-only) to `create_or_update_template` (create-or-update). That is new accepted behavior on a code path already shipping in production on this release branch — same/lower-version and previously-nonexistent-record snapshots are now silently persisted where they were previously rejected/no-op. This is new functionality ("legacy template migration", per the branch name), not a narrow correction of an existing defect.
- I checked whether this corresponds to an already-reviewed `dev` commit (required even for a legitimate backport): it does not. `git show origin/dev:.../aicoding/strategy.py` still contains the old `stored_version`/`incoming_version` gating and the `update_template` call verbatim — identical to `REL20260910` pre-PR. `git log --oneline origin/dev -- .../aicoding/strategy.py` and a repo-wide search for this PR's commit find no `dev` counterpart. If this lands only on `REL20260910`, the behavior change is lost at the next release cut cut from `dev` (regression by omission), and diverges the release branch from `dev` in a way that will conflict on the next merge-forward.

**Required fix**: retarget this PR to `dev`. If a release-branch fix is genuinely needed ahead of that, it must be a minimal, referenced cherry-pick of a change already reviewed and merged on `dev` — not this PR as-is.

Since this doesn't qualify for path (b), the Isolation Audit does not apply (there is no isolated feature to audit — the change modifies a shared, already-wired method in place).

## Findings

1. **`src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:772-810` — Gate 0 violation.** Feature/behavior change to an existing, already-wired restart-persistence pipeline (removal of version-gating, `update_template` → `create_or_update_template`) targets a release branch with no corresponding `dev` commit. Violates the branch-policy gate (Part 0 of review charter, enforced independently of the constitution's rule set). **Fix**: retarget to `dev`; land on a `REL*` branch only as a referenced cherry-pick of a `dev`-merged commit.

2. **Same hunk — arch.rules.md Rule 16 ("Changes Propagate"), Invariant.** This method is the shared restart-persistence contract point for `TEMPLATE_CONFIG_CONSUMING_ENGINES` (`aicoding`, `claude_code`). Changing its accepted-input semantics (dropping the anti-regression version check; allowing first-time creation via `create_or_update_template`) is a behavior/contract change that "could cause a correctly written consumer... to fail" (Rule 17's compatibility test) — e.g., a caller relying on the old contract to reject stale/duplicate restart payloads. The PR description is the unfilled template (`Problem`/`Solution`/`Validation` sections contain only their placeholder HTML comments) — no propagation analysis, no statement of affected consumers/compatibility status, no migration/rollback note. **Fix**: fill in the required PR sections with actual propagation analysis (why every existing caller of this contract still behaves safely under "always persist, no version check") before this can be evaluated for merge, on `dev`.

No violations found in categories 1–11 of the engine-variability catalog: the change stays inside the aicoding engine's own strategy module, doesn't add engine-string branching, capability sets, engine-named `core/` paths, filesystem literals, hidden wire protocols, env/mode detection, hardcoded operational policy, transport leakage, or misplaced Plugin-API usage. `TEMPLATE_CONFIG_CONSUMING_ENGINES` and the `template_service: Any` typing are pre-existing and untouched by this diff.

## Out-of-scope notes

- Whether dropping the version-monotonicity check is *correct* product behavior for legacy-template migration is a business-logic question I'm not gating on — only that it's a behavior change to shipping code that needs `dev`-first review and propagation analysis.
- The new `raise RuntimeError(...)` on empty `create_or_update_template()` result changes failure semantics of the restart hook (silent no-op → hard failure); worth the feature reviewer's attention, not an architecture concern by itself.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0

- **Target branch**: `REL20260910` (release branch, matches `^REL\d{8}`).
- **Classification**: Feature/behavior change to an existing, already-wired pipeline — **wrong target**. Not a minimal existing-behavior fix, and not an isolated/dormant feature.
- **Result**: FAIL.

Reasoning:

- `apply_restart_extra_configs` is not new code; it is the live persistence step of the confirmed-restart pipeline, called from `bot_service.py:4605` on every bot restart. This diff changes what that call does, not a narrow defect in it.
- The change is not additive/backward-compatible on an existing path — it **removes** an existing safety invariant (the `incoming_version > stored_version` monotonic gate) and **replaces** `template_service.update_template` (update-only) with `template_service.create_or_update_template` (upsert), so restart can now create template records where none existed and can persist same-or-older-version and unversioned snapshots it previously rejected. That is new capability ("legacy bot's first migration", per the branch name and new docstring), not a bug fix to existing behavior.
- **No `dev` counterpart exists.** I diffed `origin/dev`'s `strategy.py` against this PR: `dev` still has the old version-gated logic and still calls `update_template`. There is no merged/open `dev` PR referenced anywhere in this PR's body (the body is the unfilled template — see Finding 3). Per the branch policy, a release-branch change with no `dev` counterpart is a regression-by-omission risk on its own: it will be silently dropped at the next release cut off `dev`.
- This isn't eligible for the path (b) isolation exception either — it directly mutates behavior of code already reachable from production restart traffic, so "deleting the new files would leave every existing path byte-identical" fails immediately.

**Required action**: re-target this PR to `dev`. If a genuinely minimal, already-`dev`-reviewed fix exists for whatever defect motivated this, that specific fix may be cherry-picked to `REL20260910` separately, referencing the `dev` commit/PR.

## Findings

**1. `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:786-799` — contradicts the documented Service contract (arch.rules.md Rule 1 — Invariant, "The API Specification Is the Single Authority")**

```python
candidate = extra_configs.get("template_config")
if not isinstance(candidate, dict) or not candidate:
    return
...
saved = template_service.create_or_update_template(
    bot_id=ctx.bot_id,
    template_config=persisted_config,
    ...
)
```

`src/backend/specs/2026-08-20-template-update-consented-authorization/spec.md:36-38` is the authoritative spec for this exact code path and explicitly states the contract this function must implement:

> `apply_restart_extra_configs`（`strategy.py:424`）仅当 `extra_configs.get("template_config")` 存在且 `incoming_version > stored_version` 才 `update_template`。无传入则什么都不做。

The diff deletes the `incoming_version > stored_version` comparison entirely (unconditional persist for any dict), and swaps the documented `update_template` call for `create_or_update_template` (adds create/upsert semantics the spec explicitly rules out — spec.md:39-41 states there is **no** code path anywhere in the repo that fetches-and-writes-back a "latest version" config outside that gate). None of this is reflected in `spec.md` — the file is untouched by this PR. Per Rule 1, "implementations must not introduce undocumented semantics that other components depend on," and "contract changes update specs/docs in the same change." This is an unwaived Invariant violation on its own, independent of Gate 0.

**Required fix**: either keep the documented monotonic-version gate and the `update_template` (update-only) call, or — if the new upsert/no-gate behavior is actually wanted — update `spec.md` in the same change with the new contract, run it through the propagation analysis in Finding 2 below, and route it to `dev` per Gate 0.

**2. PR-wide — missing propagation analysis for a contract change (arch.rules.md Rule 16 — Invariant, "Changes Propagate")**

The PR body is the unmodified template: `Problem`, `Solution`, `Validation`, `Compatibility and risk`, and `Spec` are all still the blank HTML-comment placeholders, not filled in. This diff changes the persisted semantics of the `extra_configs.template_config` / `confirmed_template_update` restart contract — consumed by the frontend restart flow and by neighboring backend logic that reads the same persisted state (`_clear_restart_resync_marker`, `refresh_restart_authorization`, both in the same file, both operating on the record this method now creates rather than assumes-exists). Rule 16 requires the PR to identify affected consumers, affected implementations, compatibility status, and a migration plan for exactly this kind of change; none of that is present. "Merge is blocked when required impact analysis is absent" per the constitution's required checks.

**3. `src/backend/src/agentclaw/community/core/bot_management/engines/aicoding/strategy.py:797-798` — new failure-mode surfaced on a path with no documented failure contract**

```python
if not saved:
    raise RuntimeError("Failed to persist coding restart template snapshot")
```

This adds an exception path that did not exist before, on a Service-API-adjacent method (`apply_restart_extra_configs`) whose caller (`bot_service.py`, restart flow) wraps calls broadly in `except Exception`. Whether that's the intended failure handling for this hook is not specified anywhere — the spec's stated behavior for the "no-op" branches is "无传入则什么都不做" (does nothing), and this introduces an alternate raise-on-falsy-upsert branch with no documented semantics. This should be specified alongside Finding 1/2's contract update rather than left to caller-side incidental exception handling.

## Out-of-scope notes

- Whether removing the version monotonicity check is the *product-level* right call for legacy-bot migration is a business-logic question I am not gating on — only its placement (undocumented contract change, wrong branch) is in scope here.
- Token-encryption reuse via `create_or_update_template` and the new test coverage for insert-vs-update paths look reasonable in isolation; not evaluated further given the Gate 0/Rule 1 blockers above.

---

totalfrank (approved):

PR #2120, opened 2026-09-10
feat(workflow): filter run history by status and keyword

wienyang:
## Problem
The workflow overview run list has no status or keyword controls, making it difficult to locate failed runs or records from a particular Bot across a large history.

## Solution
- Add a compact status selector, keyword search (button or Enter), and reset control to the run list.
- Search run ID, initiator (`triggered_by`), and origin Bot ID server-side across all historical records.
- Combine keyword and status filters before counting and pagination; reset to the first page when filters change.
- Support both `cancelled` and `canceled` for the cancellation filter.
- Preserve the independent 7/30-day metric queries and show a distinct no-match empty state.
- Bind SQL parameters and escape LIKE wildcard characters for literal substring matching.

## Validation
- Confirmed new UI and repository regression tests failed before implementation.
- Focused workflow tests: 11 passed (overview behavior and SQLite repository filtering/pagination).
- Shared package build: passed.
- Workflow type check and build: passed.
- `git diff --check` and pre-push gate: passed.
- Live browser verification and production deployment were not performed.

## Compatibility and risk
The optional `query` parameter extends the existing run-list API; existing callers remain unchanged. No database migration or independent time filter is added. Substring search can increase query cost on large histories. This PR is independent of the automatic-analysis Bot-owner fix in #2116.

### DIFF
--- a/src/evolverun/clawweb/public/modules/workflow/docs/run-list-contract.md
+++ b/src/evolverun/clawweb/public/modules/workflow/docs/run-list-contract.md
@@ -1,0 +1,77 @@
+# Run list HTTP contract
+
+Owner: `@avernet/workflow`. Route: `GET /runs` relative to the host API mount
+(normally `GET /api/runs`). This is a read-only Service API, not an engine
+Plugin API. The optional `query` parameter is an additive, backward-compatible
+extension; the response envelope and existing filters are unchanged.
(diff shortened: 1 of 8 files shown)


fudp (approved):
LGTM

PR #2135, opened 2026-09-11
feat(clawinsight): add admin controls and enable improvement status advancement

ZhaosenLin:
## Problem

The Insight Center improvement lifecycle could not be driven from either side.

**Owner side.** `markHandled` only wrote the verification handoff for governance sources, so a manually
created (`USER_SELECTED`) improvement kept `verificationStatus=NOT_STARTED` with no `handledAt` and could
never reach the Agent verification queue. The `ACTIVE` + manual action-type branch of the detail drawer also
offered only 进入 Bot 修复 / 驳回, so an Owner who had already fixed the problem had no way to declare it.

**Admin side.** The admin surface shipped the whole contract — `req.isAdmin`/`isLogAdmin`/`isBenchAdmin`/
`isClawEvolveAdmin`/`isSuperAdmin` on the request type, every consumer (`requireAdmin`, workflow access,
bench, evolve, repair), the management web views, the `clawweb_admin_users` table and the `auth.admin*`
config keys — but nothing ever set those flags, so every `/api/insight/v1/admin/**` call answered `401` and
the management perspective was unreachable in an Avernet-only deployment. The internal identity providers and
the roster live in the closed composition (OCB: `@ocb/clawweb-archive`, `@ocb/access-admin`,
`internal/configs`), so the open version needs its own binding.

**Around it.** The admin state machine offered targets the server rejects (`待修复` from a non-archived item
always returned `409`, `已完成` was missing required fields, `已归档` went through the owner-scoped
endpoint). The app stylesheet declared only the workflow module as a Tailwind source while the
clawinsight/clawevolve sources resolve through `node_modules`, which automatic content detection skips — so
the Insight Center rendered without its cards, field grids and badges.

## Solution

**Owner side**
- Write the `用户已处理` governance event and the `PENDING` verification block for every source type, matching
  `listVerificationCandidates`, which selects by `handledAt`.
- Offer `我已自行修复，开始验收` on the `ACTIVE` + manual branch, mirroring the auto-repair branch and the
  existing `IN_PROGRESS` action.

**Admin side**
- Port the roster and identity chain into the open packages: `AdminUserRepository` (dynamic roster plus
  idempotent YAML seeding into `clawweb_admin_users`) and `adminAuthMiddleware` (identity candidates from
  `X-User-Id`, the `staff_id` cookie and the `IAM_TOKEN` `sno`/`sub` claims; cumulative role flags; the
  roster first with the configured lists as fallback; a loopback request carrying the dev user stays an
  admin so a local run needs no configuration, while a request without an identity is still rejected).
- Add `GET /api/auth/me` so the web client can learn the resolved identity and its roles.
- List only the state-machine transitions the server accepts for the current status (`ACTIVE → 系统验收`,
  `ARCHIVED → 待修复`) and hide the panel when there is none; span it across the action row.
- Deliberately not ported: the internal identity providers (Antbuservice, Asfagent, TERN), the hardcoded
  super-admin staff id and the `/api/admin-users` CRUD router. `isSuperAdmin` now follows the roster, and
  deployments designate admins through `auth.admins` / the roster.

**Earlier work in this branch**
- One-time admin execution / auto-approval endpoints for improvement items, with the repair runtime and admin
  review queue wired to them.
- Admin state-machine controls in the improvement detail drawer, repair-record links routed to the repair
  detail view, and the admin resolved API typing fixed.

**Styling**
- Declare the clawinsight and clawevolve module source directories for Tailwind (69 kB → 103 kB of previously
  missing utilities).

## Validation

- `@avernet/clawinsight`: 632 tests pass (new cases cover the Owner manual mark-handled action, the admin
  state-machine advance, the status-specific option list and hiding the panel when no transition applies).
- `@avernet/clawweb`: 15 tests pass (10 new: cumulative role flags, roster-over-config precedence, fallback
  when the roster read fails, `IAM_TOKEN` `sno`/`sub` matching, the loopback dev admin, dev mode, cookie
  identity precedence, the `prod` rejection).
- `@avernet/clawweb-shared`: 1 test passes. All three workspaces build clean.
- Manual end-to-end against a locally started public runner (`profile=public`, SQLite) in a real browser with
  no stubs: the Owner flow advanced an `ACTIVE` item to 自动验收中; the admin flow advanced another one
  through 管理视角 → 系统验收 and wrote `[管理员已处理] … [自动验证] 状态：PENDING`; an archived item offers
  only 待修复. The local items were restored to their pre-check state afterwards.
- Pre-push gate on the pushed range: plaintext secret scan passed, lint-only mode. The full coverage and
  Singlebox gates were not run locally.

## Compatibility and risk

- Additive: no contract, schema or migration change. `clawweb_admin_users` already existed; the middleware,
  repository and route are new files, and `resolveAdminConfig()` now delegates to the new
  `resolveAdminConfigFrom()` so there is a single normalization path.
- Access control: the public runner binds to `127.0.0.1` and roles come from an explicitly configured roster.
  A loopback request carrying the dev user is an admin by design, matching the internal middleware rule;
  a deployment that exposes the runner must list its admins in `auth.admins`. No credential or internal staff
  identifier is introduced.
- Known gaps left open deliberately: the `已完成` and `已归档` state-machine targets still need a server-side
  decision (required verification fields and an admin-scoped status endpoint respectively), and
  `/api/admin-users` remains unported.
- Rollback: revert the branch. The seeding is idempotent and leaves runtime grants untouched, so no data
  cleanup is required.


3h-william (dismissed):
ok

3h-william (approved):
ok

PR #2140, opened 2026-09-11
docs(apply): document domain-model shapes with per-road examples

totalfrank:
## Problem

The comments across `bot_config_manifest/apply/` were long on rationale (why a decision was made, which review round asked for it, which wave introduced it) and short on shape. A reader opening a file cold could not answer "what does this variable actually contain?" without tracing three files.

Two concrete gaps that motivated this:

- `EntryFetcher.fetch(source_url=...)` said nothing about what a `source_url` looks like. It is always a plain `https://...` URL (the legacy bare-string `source:` spelling), never `oss://` and never `git+...` — while `FetchedEntry.source_url` and every receipt `source_url` can be any of three shapes depending on the road. None of that was stated where the field is declared.
- `SourceSession._checkouts: dict[tuple[str, str], GitCheckout]` did not say what the tuple is. It is `(substituted repository url, ref)`, deliberately excluding `subpath` so two entries reading different paths out of one commit share one checkout.

## Solution

Comments and docstrings only, across all 26 Python files in `apply/` and `apply/materialisers/`. Zero behavioural change.

Every domain model (dataclass, Protocol, enum, type alias, and any `dict`/`tuple` field whose element type is not self-explanatory) now carries, in this order:

1. a one-sentence purpose;
2. field-by-field **shape** — the concrete form of the value, not just the Python type;
3. a concrete example instance in a `::` literal block, showing **both the git and object-store roads** where a model differs between them, and each meaningfully different state (`FetchedEntry` gets fresh-fetch, pinned store-hit, and `keep_last` fallback);
4. one line each for where it is created and where it is consumed, as `module.function` references.

Highlights:

- **`outcomes.py`** — trigger tables for the five `EntryOutcome` values and the four `ApplyStatus` values; the per-category `identity` key table; the `aborted`/`partially_written` truth table; a full `ApplyReport` and its `as_payload` wire shape.
- **`entry_fetch.py`** — the three `source_url` shapes spelled out with which method sees which, and the four entry points (`fetch` / `fetch_declared` / `acquire_object` / `file_bytes`) described by caller, URL shape, receipt and budget behaviour.
- **`entry_delivery.py`** — a method-by-road answer table for `EntryDelivery`, and the `TREE_MAGIC` wire format shown byte-for-byte for a two-file tree.
- **`registry.py`** — the `resolve` → `plan` → `write` contract worked end to end for `mcp`, including what an unpermitted entry does to the whole category.
- **`materialisers/*.py`** — a YAML entry example per category in both `from:` and inline `source:` forms, what `identity` is, and what `Intent.value` carries; summary table in the package `__init__`.

Existing rationale was kept where a maintainer needs it and trimmed where it was review-round, defect-number or wave-number history.

Example values were verified against the code and the tests rather than written from reading. Where I could not confirm a value I derived it by execution — the canonical tree framing, `compose_key`/`compose_subpath` input→output tables, `declared_protocol`'s answers, and `git_receipt_url`'s trailing-colon form were each produced by running the function. Three initial guesses were corrected this way (see below).

## Validation

- `uv run --no-sync python -m pytest -q -o addopts="" -W ignore tests/community/core/bot_config_manifest/` — **1059 passed**, same as the pre-change baseline.
- `uv run --no-sync ruff check src/agentclaw/community/core/bot_config_manifest/apply/` — clean.
- **Proved the diff is docstring-only mechanically** rather than by reading it: for each of the 26 changed files, parsed the base and working versions, stripped every module/class/function docstring, and compared `ast.dump` output. All 26 identical. Since comments never enter an AST, nothing executable changed.
- No file outside `bot_config_manifest/apply/` is touched. `uv.lock` and `pyproject.toml` are unmodified (`git diff --name-only` against the base returns nothing for either).
- Every line I added wraps within the surrounding style; the long lines that remain in these files were all present before this change.

Environment note: the default package index in `pyproject.toml` is an aliyun mirror unreachable from this sandbox, and `--index-url` does not override a named `default = true` index. `uv lock --default-index https://pypi.org/simple` (without `--upgrade`, so the pinned versions are preserved) followed by `uv sync` works; `uv.lock` was restored from git afterwards and verified clean before each commit.

## Compatibility and risk

None. No code, signatures, names, or ordering changed.

One behavioural constraint worth flagging for future doc edits: `test_cli_tools_materialiser.py::test_no_materialiser_names_an_engine` greps the **raw module source**, so an engine name in a `cli_tools.py` docstring fails it — even though `test_materialisers_name_no_engine.py` is AST-based and explicitly permits engine names in docstrings ("Docstrings and comments may *mention* an engine to explain a port's contract"). The two tests disagree. I worked within the stricter one and noted the constraint in the docstring; reconciling them is out of scope here.

## Observations not fixed

Noticed while reading; deliberately left alone per the comments-only scope.

1. **`FETCH_ENTRY_LIMITS` lookups are inconsistent between the two roads.** `source_fetchers.GitSourceFetcher.fetch` uses a total lookup, `FETCH_ENTRY_LIMITS[request.category]`, with a comment explaining that a missing member should raise rather than be papered over. `entry_fetch.acquire_object` does the opposite three lines of code away: `FETCH_ENTRY_LIMITS.get(category, FETCH_ENTRY_LIMITS["resources_file"])` — the silent narrowing-to-the-file-cap fallback that `DeclaredFetch.category`'s own comment says typing the field was meant to eliminate. Not currently reachable (the front door coerces to `FetchCategory` before any fetcher runs, and `StrEnum` members hash as their string value so the enum keys the string-keyed dict correctly), but the two sites state opposite policies.

2. **`FetchCategory.RESOURCES_UNPACKED` is never used as a fetch category.** Its docstring calls the enum "the closed vocabulary a fetch's category lives in", but this member is only ever read as a size constant — `FETCH_ENTRY_LIMITS["resources_unpacked"]` for `unpacked_size_limit` in `fetch/unpack.py` and `materialisers/skills.py`. No fetch is ever charged or capped under it. It is a limit wearing a category's clothes.

3. **`source_fetchers.substitute` is misnamed for what it does.** Its parameter is `source_url` and its docstring says "in a source URL", but `ObjectStoreFetcher.fetch` calls it on the bucket name and on the composed object key, neither of which is a URL. The behaviour is right; the name and signature describe a narrower job than the function has. I documented the wider actual usage rather than renaming.

4. **`source_fetchers`' module docstring referenced `:data:`FETCHERS`**, a name that does not exist — the table is `FETCHER_TYPES`. Corrected, since it is docstring text.

5. **`SourceResolution`'s docstring claimed it is "Empty in v1's URL wave — sources are inline URLs only".** That is stale: `SourceSession.adopt` fills it on every git road. Rewritten to describe what it actually carries (git sources only; the object-store road resolves no ref and contributes no row).

https://claude.ai/code/session_015J5SVbGkKQSqGPwuJWhCfY

---
_Generated by [Claude Code](https://claude.ai/code/session_015J5SVbGkKQSqGPwuJWhCfY)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_delivery.py
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_delivery.py
@@ -278,28 +278,35 @@ class EntryDelivery(Protocol):
     read) and :class:`GitDelivery` (a proven checkout). What each member
     answers, per road:
 
-    ==================  =====================  ==========================  ==========================
-    Method              Called by              ``BlobDelivery`` answers    ``GitDelivery`` answers
-    ==================  =====================  ==========================  ==========================
-    ``is_tree()``       ``skills``             ``False``                   ``True``
-    ``members()``       ``resources``          the archive unpacked, or    every file under the
-                                               a stored tree decoded       composed ``subpath``
-    ``single()``        ``resources``,         the whole blob              the one file ``subpath``
-                        ``identity``,                                      names
-                        ``cli_tools``
-    ``note()``          all four               the ``keep_last``           the moved-ref note
-                                               fallback reason
-    ``digest()``        ``cli_tools``          the content digest          ``None`` (git has no
-                                                                           declarable digest)
-    ``receipt_url()``   all four               ``https://…`` or            ``git+…@<sha>:<subpath>``
-                                               ``oss://…``
-    ``auth()``          all four               ``None`` (the fetch         the credential name
-                                               already filed it)
-    ``source_url()``    ``skills``             the fetched URL             the repository URL
-    ``content_type()``  ``skills``             what the source said        ``None``
-    ``from_store()``    ``skills``             True on a store hit         always ``False``
-    ``needs_receipt()`` all four               ``False``                   ``True``
-    ==================  =====================  ==========================  ==========================
+    ===================  ======================  =========================
+    Method               ``BlobDelivery``        ``GitDelivery``
+    ===================  ======================  =========================
+    ``is_tree()``        ``False``               ``True``
+    ``members()``        the archive unpacked,   every file under the
+                         or a stored tree        composed ``subpath``
+                         decoded
+    ``single()``         the whole blob          the one file ``subpath``
+                                                 names
+    ``note()``           the ``keep_last``       the moved-ref note
+                         fallback reason
+    ``digest()``         the content digest      ``None``; git declares no
+                                                 digest
+    ``receipt_url()``    ``https://…`` or        ``git+…@<sha>:<subpath>``
+                         ``oss://…``
+    ``auth()``           ``None``; the fetch     the credential name
+                         already filed it
+    ``source_url()``     the fetched URL         the repository URL
+    ``content_type()``   what the source said    ``None``
+    ``from_store()``     ``True`` on a store     always ``False``
+                         hit
+    ``needs_receipt()``  ``False``               ``True``
+    ===================  ======================  =========================
+
+    Which category asks which: ``skills`` alone calls ``is_tree``,
+    ``source_url``, ``content_type`` and ``from_store``; ``resources`` calls
+    ``members``; ``resources``, ``identity`` and ``cli_tools`` call
+    ``single``; ``cli_tools`` alone calls ``digest``; and all four call
+    ``note``, ``receipt_url``, ``auth`` and ``needs_receipt``.
 
     Created by: ``apply/source_fetchers`` (both fetchers) and
     ``apply/entry_fetch.fetch_declared`` on the legacy inline-URL road.
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_delivery.py
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_delivery.py
@@ -84,12 +84,72 @@ class EntryFetchError(Exception):
 
 @dataclass(frozen=True)
 class FetchedEntry:
-    """One entry's bytes, their content address, and where they came from."""
+    """One entry's bytes, their content address, and where they came from.
 
+    Three states, and a caller tells them apart by ``from_store`` and
+    ``fallback_reason``.
+
+    A **fresh fetch** — the wire moved and the bytes were filed with the
+    store::
+
+        FetchedEntry(
+            content=b"PK\x03\x04...",
+            digest="sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b...",
+            from_store=False,
+            content_type="application/zip",
+            fallback_reason=None,
+            source_url="https://example.com/tools/qc-v2.zip",
+        )
+
+    A **pinned store hit** — the entry declared a ``digest``, the platform
+    already held those exact bytes, and no network was touched. Silent by
+    design: this is the legitimate fast path, so there is no note::
+
+        FetchedEntry(
+            content=b"PK\x03\x04...",
+            digest="sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b...",
+            from_store=True,
+            content_type="application/zip",
+            fallback_reason=None,
+            source_url="oss://team-artifacts/tools/qc/v2.tgz",
+        )
+
+    A **keep_last fallback** — the source was fetched, it failed, and the
+    stored copy stood in. ``from_store`` is true *and* a reason is set, which
+    is the only combination that means "fallback"::
+
+        FetchedEntry(
+            content=b"acm-tree-v1\n...",
+            digest="sha256:2c26b46b68ffc68ff99b453c1d30413413422d70...",
+            from_store=True,
+            content_type=None,
+            fallback_reason=(
+                "delivered from the platform's stored copy (keep_last): "
+                "the git fetch failed"
+            ),
+            source_url=(
+                "git+https://code.example.com/team/content.git"
+                "@4f2a9c1b8e7d6a5c4b3a2918f7e6d5c4b3a29187:kb"
+            ),
+        )
+
+    Created by: ``apply/entry_fetch.EntryFetcher`` — ``fetch``,
+    ``acquire_object`` and ``_git_keep_last``.
+    Consumed by: ``apply/entry_delivery.BlobDelivery``, which is the only thing
+    that wraps one.
+    """
+
+    #: The entry's bytes, whole and in memory. For a stored git tree these are
+    #: the canonical framing :func:`canonical_tree_bytes` produced, not a file.
     content: bytes
+    #: ``"sha256:"`` followed by 64 lowercase hex characters, over ``content``.
     digest: str
     #: True when the platform's own copy answered — no network was touched.
+    #: True for both a pinned hit and a ``keep_last`` fallback;
+    #: ``fallback_reason`` is what separates them.
     from_store: bool
+    #: What the source said the bytes are, e.g. ``"application/zip"``, or
+    #: ``None`` when it said nothing. The object-store road never sets one.
     content_type: Optional[str] = None
     #: Set only when the platform's copy answered as a ``keep_last``
     #: FALLBACK — the source was fetched and failed, the stored bytes stood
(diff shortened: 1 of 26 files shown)


FreddieSun (approved):
LGTM

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/triggers.py
```diff
@@ -1,20 +1,39 @@
...
+========================  =================================================
+``explicit``              ``POST …/config-manifest/apply``, through
+                          ``adapters/http/…/bots/config_manifest_apply``.
+                          Also ``start_apply``'s default, so any caller
+                          naming no trigger records this one.
+``put``                   ``PUT …/config-manifest`` on an existing bot,
+                          through
+                          ``adapters/http/…/bots/config_manifest_support``.
+``create:pre_container``  The creation job's first phase, through
+                          ``creation.ManifestCreationSteps`` and
```
totalfrank, line 17: Non-blocking: this cross-reference names a class that doesn't exist. There is no `ManifestCreationSteps` anywhere in the codebase (verified by repo-wide search). The actual creation-job class in `creation.py` is `BotCreationManifestSeam` (implementing `ManifestCreationSeam` from `bot_management/manifest_seam.py`). Since this whole docstring is new, a reader following this reference will hit a dead end.

```suggestion
``create:pre_container``  The creation job's first phase, through
                          ``creation.BotCreationManifestSeam`` and
                          ``create_job``. Applies ``PRE_CONTAINER`` only.
```

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/outcomes.py
```diff
@@ -188,29 +392,64 @@ def as_dict(self) -> dict[str, Any]:
...
+                ),
+            ),
+            notes=(),
+        )
 
-    Apply's only output. It is not a projection of some richer internal state —
-    there is no richer state, because §2.7 makes the per-entry records the whole
-    of what apply knows.
+    Created by: ``apply/orchestrator`` at the end of an apply, and by
+    ``apply/apply_task`` for the background path.
```
totalfrank, line 427: Non-blocking: `apply/apply_task.py` never constructs an `ApplyReport` — it only forwards the task payload to `BotConfigManifestApplyService.run_apply_task`. The actual background-path creator is `services/config_manifest_apply_service.py` (via `apply/carry_forward.py`). As written, a reader chasing this reference into `apply_task.py` won't find an `ApplyReport` construction there.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/outcomes.py
```diff
@@ -272,9 +548,14 @@ def entry_identity(entry: Any, index: int) -> str:
 def derive_status(categories: tuple[CategoryResult, ...]) -> ApplyStatus:
     """The summary, computed once every decision has already been made.
 
+    Takes every :class:`CategoryResult` the apply produced and answers one of
+    the three terminal :class:`ApplyStatus` values (never ``RUNNING``).
+
+    Called by: ``apply/orchestrator`` and ``apply/apply_task``, once, at the end.
```
totalfrank, line 554: Non-blocking: same issue as the `ApplyReport` "Created by" note above — `apply/apply_task.py` has no import of `outcomes` and never calls `derive_status`. That call happens in `services/config_manifest_apply_service.py` / `apply/carry_forward.py` for the background path, not in `apply_task.py`.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_delivery.py
```diff
@@ -162,11 +274,49 @@ def moved_note(self) -> Optional[str]:
...
+    ``from_store()``     ``True`` on a store     always ``False``
+                         hit
+    ``needs_receipt()``  ``False``               ``True``
+    ===================  ======================  =========================
+
+    Which category asks which: ``skills`` alone calls ``is_tree``,
+    ``source_url``, ``content_type`` and ``from_store``; ``resources`` calls
+    ``members``; ``resources``, ``identity`` and ``cli_tools`` call
+    ``single``; ``cli_tools`` alone calls ``digest``; and all four call
+    ``note``, ``receipt_url``, ``auth`` and ``needs_receipt``.
```
totalfrank, line 309: Non-blocking: this "which category asks which" attribution has three mismatches against the actual materialiser code:

- `materialisers/skills.py` also calls `single()` (line ~526) and `members()` (line ~564), but the text excludes `skills` from both.
- `cli_tools/service.py` also calls `is_tree()` (line ~689), but the text says `skills` is the sole caller.

Since this table is meant to let a reader trust "member X has exactly one caller family" (as the paragraph right below it claims), the wrong attributions undercut that guarantee.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/registry.py
```diff
@@ -102,27 +195,55 @@ def ok(self) -> bool:
...
+    call: a preview that cannot write is one that is *missing the call*, rather
+    than one that is disciplined about not making it.
     """
 
+    #: One per resolved intent, in declaration order.
     entries: tuple[PlannedEntry, ...] = ()
     #: Identities present in the area and no longer declared — what overwriting
-    #: removes. Reported separately from entry outcomes because a removal has no
-    #: declared entry to attach to.
+    #: removes, e.g. ``("old",)``. Sorted by the materialisers that produce
```
totalfrank, line 244: Non-blocking: "Sorted by the materialisers that produce them" doesn't hold for all producers. `materialisers/resources.py`'s `plan()` builds `removals` as `tuple(intent.identity for intent in intents if intent.value is _DECLARED_TREE)`, preserving declaration order — never `sorted()`. `mcp.py`, `identity.py`, `skills.py`, and `cli_tools.py` do sort theirs, but `resources` is the exception this blanket claim misses.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/registry.py
```diff
@@ -204,26 +349,38 @@ def build_materialisers(
...
-    put it. ``engine_config`` arrives when X2/T3 lets it back in; until then a
-    document declaring it takes the orchestrator's no-materialiser path: an
-    expected state, not a gap.
+    both construct services at import time and pull the bot-configuration graph
+    into anything that merely wants the ordering table.
+
+    The map is keyed by each materialiser's own ``construct`` rather than by a
+    name written here, so a materialiser cannot be registered under the wrong
+    key. The fetch-side dependencies (``package_validator``,
+    ``entry_fetcher``) exist because ``skills`` and ``identity`` materialise
```
totalfrank, line 378: Non-blocking: this explains `entry_fetcher` as existing because `skills` and `identity` materialise fetched bytes, but `resources` also consumes it and also materialises fetched bytes — this same function constructs `ResourcesMaterialiser(resource_service, entry_fetcher)` a few lines below (line 415), and `materialisers/resources.py` fetches named (`from:`) sources through the same fetch funnel. The docstring should name all three categories, not two.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/materialisers/cli_tools.py
```diff
@@ -104,7 +160,23 @@ def context_for(ctx: ApplyContext) -> CliToolContext:
...
+    ``FAILED`` row out of ``write`` rather than aborting the category::
+
+        resolve -> ResolveResult(intents=(Intent(
+                       identity="mycli",
+                       value=CliToolDecl(name="mycli", ...)),))
+        plan    -> CategoryPlan(
+                       entries=(PlannedEntry(<that intent>, "created"),),
+                       removals=("retired-tool",))
+        write   -> (EntryResult(ManifestCategory.CLI_TOOLS, "mycli",
+                                EntryOutcome.CREATED),)
```
totalfrank, line 178: Non-blocking: `write()` can never emit `EntryOutcome.CREATED`. It maps outcomes through `_OUTCOMES` (defined a few dozen lines above in this same diff), which sends `CliToolStatus.INSTALLED` to `EntryOutcome.UPDATED` — and that module comment explicitly says "`INSTALLED` maps to `UPDATED`, not `CREATED`". This worked example directly contradicts the module's own adjacent comment. Should be `EntryOutcome.UPDATED`.

```suggestion
        write   -> (EntryResult(ManifestCategory.CLI_TOOLS, "mycli",
                                EntryOutcome.UPDATED),)
```

---

totalfrank (commented):
Reviewed commits 5118291, 0aa6e45, and f25ba12 (docs-only, comments/docstrings across 26 files in `bot_config_manifest/apply/`). Since the entire diff is documentation, I focused review on whether the new prose/examples are factually accurate against the real code rather than style.

Six confirmed inaccuracies found, all non-blocking (nothing runtime-visible changed, but each would mislead a future reader):

- `triggers.py`: cross-reference to a class (`ManifestCreationSteps`) that doesn't exist.
- `outcomes.py` (x2): `ApplyReport`/`derive_status` "created by / called by" notes wrongly attribute the background path to `apply/apply_task.py`, which never touches either.
- `entry_delivery.py`: the per-category "which category calls which method" table omits `skills` from `single()`/`members()` callers and `cli_tools` from `is_tree()` callers.
- `registry.py` (x2): the `CategoryPlan.removals` "sorted by the materialisers" claim doesn't hold for `resources` (preserves declaration order); `build_materialisers`'s rationale for `entry_fetcher` omits `resources`, which also consumes it.
- `materialisers/cli_tools.py`: a worked example shows `write()` producing `EntryOutcome.CREATED`, which the module's own `_OUTCOMES` map (and its adjacent comment) says can never happen — `INSTALLED` always maps to `UPDATED`.

Checked and confirmed accurate: the `cli_tools.py` engine-naming constraint against both `test_cli_tools_materialiser.py::test_no_materialiser_names_an_engine` and `test_materialisers_name_no_engine.py`; identity/`Intent.value` derivation per category; the registry `resolve→plan→write` worked example including the unpermitted-entry branch; budget/TTL figures; the tree wire-format byte layout; and the large majority of the cross-reference and worked-example claims across all 26 files.

---

PR #2144, opened 2026-09-11
feat(dormant): publish personal bot lifecycle OpenAPI

FreddieSun:
## Summary

- publish owner-scoped personal Bot recycle, activate, and status endpoints under `/openapi/v1`
- share typed dormant lifecycle services with the existing page and internal operations APIs while preserving their response contracts
- exclude Teclaw from dormant governance, add OpenAPI lifecycle audit records, and publish the generated Gateway schema

## API

- `POST /openapi/v1/bots/{bot_id}/recycle`
- `POST /openapi/v1/bots/{bot_id}/activate`
- `GET /openapi/v1/bots/{bot_id}/status`

All routes use the existing OpenAPI principal/admission stack, require OWNER access to the addressed personal Bot, and accept optional `owner_id` for exact default-Bot addressing.

## Validation

- Backend focused suite: 505 passed
- Gateway schema/served OpenAPI suites: 26 passed
- Backend full CI gate: 18,492 passed, 60 skipped; total line coverage 89.00%; changed-line coverage 96.48% (192/199)
- Singlebox dormant lifecycle acceptance: 2 passed
- Standards/Spec dual-axis review completed and high-priority findings fixed
- `git diff --check` passed; generated schema slices match the live OpenAPI dump

PR #2145, opened 2026-09-11
feat(cli-tools): accept the binary as a multipart upload and drop the HTTPS-GET fetch road

totalfrank:
## Problem

`POST /openapi/v1/bots/{bot_id}/cli-tools` took `source` (an HTTPS URL) plus
`auth`, and the platform fetched it. That shape predates declared sources: the
manifest side was upgraded to `protocol: git|oss` objects and bare URLs are
refused at `PUT`, but this API was not. It is also the odd one out — every
other category's management API takes the content itself (skills: a zip;
resources: file bytes; identity: a text body; mcp: a config body).

That call was also the last caller of the HTTPS-GET road in `apply/`. #2144
("drop the legacy bare-string source road") landed on `dev` while this branch
was being written and removed everything on that road *except* the road itself,
naming this endpoint as the reason `EntryFetcher.fetch` had to stay. With the
endpoint converted, it does not.

## Solution

**The API uploads; the manifest declares.** After this change there are exactly
two ways bytes reach `CliToolService`, and they converge on one pipeline the
moment the bytes are in hand:

1. **Manifest apply** — the `cli_tools` materialiser resolves the entry through
   `EntryFetcher.fetch_declared` (git or oss), like every other fetching
   category.
2. **Upload API** — the route receives the bytes and hands them over.

`EntryFetcher.fetch(...)` and `EntryFetcher._fetch(...)` are deleted, and with
them the last reason this module holds a `GuardedFetcher` at all — the
constructor drops it.

### API contract (breaking)

Before — `application/json`:

```json
{"name": "mycli", "source": "https://cdn.example.com/mycli.tar.gz",
 "digest": "sha256:…", "unpack": "tar.gz", "subpath": "…", "version": "1.4.2",
 "auth": "cdn-prod"}
```

After — `multipart/form-data`: a `file` part (the binary, or a `zip` /
`tar.gz`), plus form fields `name` (required), `digest` (required), `unpack`,
`subpath`, `version`.

- **Removed: `source` and `auth`.** There is nothing to fetch and no credential
  to fetch under.
- `digest` stays **mandatory** and is verified against the uploaded bytes
  before anything else — on this road it is the client's statement of what it
  meant to send, so a truncated or swapped upload is a `422` / `422110` rather
  than an installed executable nobody named.
- The response (`CliTool`) and the `409110` / `409111` / `422110` / `404`
  semantics are unchanged.
- **New: `413` / `413110`** when the upload exceeds `FETCH_ENTRY_LIMITS["cli_tools"]`
  (200 MiB — the same cap a manifest-declared fetch gets). Enforced while the
  body is read: `Content-Length` refuses up front, and the stream is bounded
  through the Space Skill upload's `_bounded_request_stream`, so an over-cap
  body is never buffered whole.

**This is breaking for API clients and for the web UI**, which must switch to
multipart. Note: nothing under `src/frontend` or `src/frontend-nextgen`
references `cli-tools` today, so there is no frontend call site to update in
this repo — the follow-up is for whatever client is built against the old
shape.

### Uploaded bytes file no W11 content-store receipt

A receipt is the platform's record of **a fetch** — what was acquired, from
which address, under which credential, for which apply. An upload was not
fetched: there is no source address to file it under and no credential to
attribute it to, and inventing an `upload://` identity would put a shape into
the lineage that no `keep_last` or audit query can act on. The install is
already recorded twice — the `ac_bot_cli_tool` row and the platform's own OSS
copy of the bytes — which is what every later read (delivery, redelivery,
removal, drift) actually uses. The git road's `file_bytes` receipt is untouched.

### `ac_bot_cli_tool.source`

The column is NOT NULL and is write-only (nothing in the codebase reads it
back). It is no longer populated from a URL the service might fetch: it now
carries `CliToolDecl.declared_address` — the `from` name, the git URL, or
`bucket/key` for a manifest entry, and the empty string for an upload, which
declared no source at all. The column itself is left alone.

## Validation

```
cd src/backend
  tests/community/endpoints/test_openapi_cli_tools.py \
  tests/community/adapters/http/openapi_v1/
```

**2709 passed, 2 skipped** on the merged head. Also ran
`tests/community/architecture/`, `tests/community/core/config_compose/test_cli_tools_refs.py`
and `tests/community/core/resources/test_cli_tools_absent_from_listings.py` —
green, since the removed constructor argument and the shared apply fakes reach
beyond the three paths above.

`uv.lock` is not in the diff.

### Merged with `dev` (#2144 overlap)

#2144 landed mid-flight and does part of what this branch's second commit did.
The merge resolves **toward `dev`** wherever it had already converted a suite,
because that is the version everything else on `dev` is written against: its
`apply/_fakes.py` object-road rig (`FakeObjectCredentials`, `seeded_object_store`),
its conversions of the skills / identity / resources / apply-engine /
skill-port / first-artifact suites, its `declares_source` naming, and its
refusal message for a plain-string `source` (which names both `protocol: git`
and `protocol: oss`) all stand. This branch's own delta is re-applied on top:
every rig drops the `GuardedFetcher` argument.

Two of `dev`'s cases and two of this branch's tested the same two things; the
pair from `dev` is kept and this branch's duplicates dropped. `dev`'s
`test_a_placeholder_in_a_declared_address_is_substituted` is **deleted**: it
asserts `intent.value.source_url`, a field this branch removes, and the
materialiser it covers no longer substitutes anything — the replacement is
`test_the_entry_reaches_the_fetch_funnel_unrewritten`.

### Tests converted or deleted

`tests/community/endpoints/test_openapi_cli_tools.py`
- install cases → multipart (`form_data` + `files`); the `EntryFetcher.fetch`
  stub is gone with the road, so the install cases now send a real ELF and the
  platform really verifies it.
- **added**: archive upload with `subpath`; digest mismatch → `422110`;
  `unpack` without `subpath` → `422110`; over-cap upload → `413110`.

`tests/community/adapters/http/openapi_v1/test_cli_tools_upload_limits.py` (new)
- the half a request/response assertion cannot see: an endless, length-less
  body is cut off mid-stream and the endpoint never runs. The cap is
  monkeypatched down — moving 200 MiB to prove the platform will not move
  200 MiB is the one shape this test must not take — and a second case pins
  `_UPLOAD_LIMIT` against `FETCH_ENTRY_LIMITS`.

`tests/community/core/bot_config_manifest/cli_tools/test_service.py`
- `_decl()` now builds from a manifest entry (`protocol: oss`), so the pipeline
  cases stay on the manifest road they always exercised.
- the three insert-only cases (the API's 409) → `install_upload`, which is the
  road that contract belongs to.
- `test_install_enforces_the_declared_sha256` rewritten: the belt hashes the
  bytes now rather than trusting what the delivery said about itself, so the
  case delivers *different bytes* instead of a lying digest.
- `test_the_fetch_goes_through_the_entry_fetcher_under_this_category` →
  `…_through_the_declared_source_door`, asserting the resolved declaration
  rather than a URL.
- **added**: upload installs without acquiring anything; an upload that does not
  match its digest; an uploaded archive unpacks identically; an uploaded binary
  for another architecture; upload defaults to insert-only; a decl with no
  entry is refused on the manifest road.

`tests/community/core/bot_config_manifest/cli_tools/_fakes.py`
- `FakeEntryFetcher.fetch` deleted with the method it doubled, so the double
  cannot answer a call the real class no longer has; its `fetch_declared`
  refusal keeps `dev`'s wording.

`tests/community/core/bot_config_manifest/apply/test_cli_tools_materialiser.py`
- `_entry()` → an inline `protocol: oss` declaration.
- `test_a_placeholder_in_a_declared_address_is_substituted` →
  `test_the_entry_reaches_the_fetch_funnel_unrewritten` (above).
- `test_the_api_and_apply_refuse_the_same_hostile_declaration` → the API arm is
  an upload.
- **added**: a bare-URL source fails the entry rather than being fetched.

`tests/community/core/bot_config_manifest/apply/test_entry_fetch.py`
- Every `EntryFetcher.fetch` case was triaged. **Ported to `acquire_object`**
  (the policy is the pipeline's, not the transport's, and the object road is
  where it lives now): pinned store-hit, mismatched receipt re-reads, unpinned
  re-reads, pinned hit survives a dead source, keep_last with no receipt,
  keep_last never disagrees with a pin, a read failure without keep_last, the
  actor as modifier, the missing blob heals, the corrupt blob fails loudly,
  both store-fault translations, keep_last naming both halves, and both budget
  cases. **Deleted** (they pinned `GuardedFetcher`'s own seams, which this
  module no longer has): placeholder substitution before the transport, the
  credential binding as injector/policy, the credential value never leaking
  through an error, per-hop prefix-authorization refusal, a refused transport.
  **Deleted as duplicates** of oss cases that already existed: unpinned
  keep_last falls back to the last digest, a refusal never triggers keep_last.
  `test_the_funnel_requires_a_category_by_keyword` now checks the two entry
  points that remain, and `dev`'s plain-URL refusal case asserts the stronger
  property that there is no URL transport left to have been asked.

Every rig that built `EntryFetcher(fetcher=GuardedFetcher(...))` dropped the
argument: the two apply-service suites, the apply task, the three creation
suites, the two managed-files suites and `test_source_fetchers.py`.

## Compatibility and risk

- Breaking for callers of `POST …/cli-tools` (above). No stored data changes;
  no migration.
- `schema/entries.py`'s refusal of a string `source` is untouched: it is now the
  only place the old spelling is recognised, and its message is the one users see.
- The apply-side consequence of a stored bare-string `source` landed with #2144,
  not here; this branch only removes the transport that road used.

## Out of scope (found while working)

- `fetch/guarded_fetcher.py` now has no consumer in this package. Left in place
  per the task's non-goals, along with its DI provider in
  `manifest_fetch_module.py`; removing both is a clean follow-up.
- One `source_url` spelling survives in `core/bot_config_manifest/cli_tools/`:
  `_acquire` passes `file_bytes(..., source_url=delivery.receipt_url())`. That is
  the fetcher's own keyword for a **receipt identity** (`git+…`), not a declared
  address — the declaration carries no URL anywhere. Renaming that parameter would
  touch the `resources` and `skills` materialisers too, so it is left alone.
- `CliToolDecl` still carries `auth` and `keep_last`, read from the entry by
  `from_entry`. Nothing reads them any more (`fetch_declared` reads both off the
  entry itself). They are declared metadata rather than dead plumbing, so they
  stay; dropping them is a separate tidy.
- `specs/2026-09-03-manifest-cli-tools/plan.md` still shows the old
  `body: CliToolInstall` signature. Historical spec, left as written.

---

https://claude.ai/code/session_01RjsJ9Sy55CzqW4WApFHSk1

---
_Generated by [Claude Code](https://claude.ai/code/session_01RjsJ9Sy55CzqW4WApFHSk1)_


totalfrank (commented):
Reviewed commits `920cc4c`, `512dba3`, `aeddbc5`, `cfbc11e` (head) against merge-base `1e10700`. No findings — the diff is clean.

What I checked:
- **Correctness**: the multipart upload path in `cli_tools.py` (`CliToolUploadRoute`, header/stream size bounding, `_too_large` translation), the unified `_shape`/`_install_bytes` pipeline in `service.py`, and the deleted `EntryFetcher.fetch`/`_fetch` HTTPS-GET road in `entry_fetch.py`.
- **Contract breakage**: the breaking API change (JSON→multipart, dropped `source`/`auth`, new `413`/`413110`) is documented in the PR body, `contracts.py`/`responses.py` error tables, and both `docs/bot-config-manifest/*.zh-CN.md` files, all consistently.
- **Error handling**: digest mismatches, oversized uploads, and store/engine failures all return typed outcomes rather than swallowing; verified `acquire_object`/legacy pin-check still independently verifies the digest on a fresh object-store read, so the belt-and-braces recompute in `_shape` is redundant but not a regression.
- **Concurrency/resource safety**: the bounded multipart stream reads via `_bounded_request_stream` (reused from `spaces/multipart_limits.py`) correctly refuses before buffering an over-cap body; `install_upload`'s `expect_absent=True` insert-then-UNIQUE-constraint race handling is unchanged from the existing pattern.
- **Architecture boundaries / DI**: `manifest_fetch_module.py` correctly drops the `GuardedFetcher` singleton now that no consumer remains in this package.
- **Test coverage**: new/converted tests cover the upload success, archive+subpath, digest mismatch (422), unpack-without-subpath (422), and over-cap (413) cases, plus a dedicated stream-cutoff test (`test_cli_tools_upload_limits.py`) for the part a response-only assertion can't see.

Python type contracts and security/hygiene had nothing new to flag in this diff.

---

totalfrank (commented):
Reviewed the two commits pushed since the last round: `509899c5` (merge from `dev`) and `5679220b` (head).

**New commits, checked against the PR's actual scope** (merge-base `5d1bfc40`..`5679220b`, which still matches the PR metadata: 35 files, +1449/-1182):
- `509899c5` merges in `dev`'s unrelated "dormant bot lifecycle" work (`bot_dormant/`, `bcs` crates, etc.). None of it touches this PR's files (`cli_tools`, `entry_fetch`, `manifest_fetch_module`, or their tests) — confirmed via `git diff` scoped to those paths. Out of scope for this review.
- `5679220b` removes the resources-apply endpoint test's dead `EntryFetcher.fetch` stub override, which would otherwise bind over a method the class no longer has (deleted in `512dba3`). Checked the surrounding fixture: `archive` is still used to seed the object store, the updated comment accurately describes the now-real oss road, and no other reference to the removed `fetch`/`_fetch`/`FetchedEntry` imports remains in the function. Clean, matches its commit message.
- Verified no other caller in the PR-scoped diff still references the deleted `EntryFetcher.fetch`/`_fetch` (the one remaining `fetcher.fetch(...)` in `entry_fetch.py:396` is the per-protocol `git`/`oss` fetcher from `source_fetchers.py`, a different object — legitimate).

## Status

- Round 1 (`cfbc11e`, previous review): **no findings**.
- Round 2 (this round, `509899c5`..`5679220b`): **no findings**. No open threads existed to verify.

**Open findings: 0.** PR is clean as of `5679220b`.

---

FreddieSun (approved):
LGTM

PR #2152, opened 2026-09-11
feat(bcs): add undercover tie-break PK

ziminz19:
## Problem

Undercover voting could not resolve ties through a dedicated PK phase, leaving tied players without a defined follow-up flow and making the panel state inconsistent with the referee state machine.

## Solution

- Add a documented tie-break PK phase for tied highest vote counts.
- Track PK attempts and history separately from regular rounds, including stale-submission protection.
- Enforce PK candidates and disallow abstentions in PK ballots.
- Update referee prompts, public result artifacts, panel contracts, and UI rendering for PK state.
- Add focused regression coverage for PK lifecycle, validation, retries, multi-way ties, final-round behavior, and human-player elimination.

## Validation

- `python3 -m unittest test_tie_break_pk.py` — 10 tests passed.
- `npm run test:focused` in `src/bcs/assets/panel` — passed.

## Compatibility and risk

The change is scoped to the undercover game profile and BCS panel assets. Existing regular voting remains supported; PK metadata is additive to the panel contracts and public artifacts.


questchaos (approved):
LGTM

vzvince (approved):

PR #2153, opened 2026-09-11
refactor(apply): give each fetch road one home and rename the dispatcher

totalfrank:
## Problem

`apply/entry_fetch.py` held three methods on `EntryFetcher` that did not belong
to a dispatcher:

- `_git_keep_last` — one caller, `source_fetchers.GitSourceFetcher.fetch`.
- `acquire_object` — one caller, `source_fetchers.ObjectStoreFetcher.fetch`.
- `file_bytes` — the git road's receipt, filed *on behalf of* the materialisers,
  always behind an `if delivery.needs_receipt():` the caller had to remember.

Both fetchers were constructed with their owner and reached back through it
(`self._owner._credentials`, `self._owner._objects`, `self._owner._content`),
which is why `source_fetchers.py` carried a `TYPE_CHECKING` import of the module
that imports it and why private attributes were read across module boundaries.
And `fetch_declared` and `declared_protocol` each spelled out the same "find the
raw declaration for this entry" lookup — two places for one rule to be true in,
where the second exists precisely to predict what the first will do.

## Solution

Each road gets one home, a delivery files its own receipt, and the class that is
left is a dispatcher named for what it does.

### Where each method lives

| Before | After |
| --- | --- |
| `EntryFetcher._git_keep_last` | `GitSourceFetcher._keep_last` (`apply/source_fetchers.py`) |
| `EntryFetcher.acquire_object` | `ObjectStoreFetcher._acquire` (`apply/source_fetchers.py`) |
| `EntryFetcher.file_bytes` | `GitDelivery.file` (writes) / `BlobDelivery.file` (no-op) (`apply/entry_delivery.py`) |
| `EntryFetcher.fetch_declared` | `DeclaredSourceResolver.resolve` (`apply/source_resolver.py`) |
| `FetchContext`, `scope_of` | `apply/fetch_context.py` (new) |
| the declaration lookup, inline in two functions | `source_resolver._raw_declaration`, read by both |

`ObjectStoreFetcher(content, credentials, objects)` and
`GitSourceFetcher(content, credentials)` now take their collaborators directly,
bound once by `build_fetchers(content, credentials, objects)`. Nothing reaches
through an owner, and `source_fetchers.py` has no `TYPE_CHECKING` block left.

`fetch_context.py` exists because the fetchers, the dispatcher **and** the
deliveries all read `FetchContext`/`scope_of`: with `scope_of` now called at
runtime from `source_fetchers.py` and `entry_delivery.py`, leaving it on the
dispatcher would close a real import cycle rather than a typing-only one.

### One behaviour-preserving fix, called out

`acquire_object` read its per-entry cap as

```python
limit = FETCH_ENTRY_LIMITS.get(category, FETCH_ENTRY_LIMITS["resources_file"])
```

a `.get` whose default would have silently narrowed an unrecognised category to
the 100 MiB file cap. `category` is the `FetchCategory` the front door already
coerced, and `fetch/limits.py` binds that enum to `FETCH_ENTRY_LIMITS` with its
own module-level assertion — so the lookup is now the total
`FETCH_ENTRY_LIMITS[category]`, and the parameter is annotated `FetchCategory`.
That is exactly how the git road's `file_limit=` line already read the table;
the two are now consistent. No reachable call changes value: every caller comes
through the front door's coercion.

### `receipt_url` / `auth` callers kept: none

Grepped `apply/`, `cli_tools/` and `services/`: the only callers of
`EntryDelivery.receipt_url()` and `EntryDelivery.auth()` were the five
`file_bytes` sites. All three of `needs_receipt()`, `receipt_url()` and `auth()`
are therefore gone from the protocol and from both implementations, and nothing
was kept. `GitEntrySource.receipt_url()` — the dataclass helper, not the
protocol member — stays: `GitDelivery.file` is now its one caller.

### The three renames, for grepping

`EntryFetcher` → `DeclaredSourceResolver`; `fetch_declared` → `resolve`; `apply/entry_fetch.py` → `apply/source_resolver.py`.

`EntryFetchError`, `FetchContext`, `scope_of` and `declared_protocol` keep their
names.

## Validation

```
cd src/backend
  tests/community/endpoints/test_openapi_cli_tools.py
```

**1063 passed** (1061 on the base commit; net +2 from the new
`BlobDelivery.file` test and the object-road "files once" test, against the
`file_bytes`/`needs_receipt` assertions that were converted rather than
duplicated). `ruff check` is clean on every file this PR touches.

Post-change greps, all empty as intended:

- `grep -rn "EntryFetcher\b\|fetch_declared\|entry_fetch\b" src tests --include=*.py`
- `grep -n "_owner\|TYPE_CHECKING" .../apply/source_fetchers.py`
- `grep -n "def acquire_object\|def _git_keep_last\|def file_bytes" .../apply/source_resolver.py`
- `grep -rn "needs_receipt\|receipt_url()\|\.auth()" .../apply/materialisers/ .../cli_tools/`

`git status --short` lists no `uv.lock`.

### Zero behaviour change

The same bytes reach the same store under the same receipt identity, credential
name, category, apply id and entry identity on both roads; the budget is charged
at the same points and for the same amounts; every error message is byte-for-byte
what it was, including the asymmetry `_raw_declaration` preserves (a `from`
naming a source that does not parse still gets the parser's reason, not "not
declared under 'sources'"). The one changed expression is the
`FETCH_ENTRY_LIMITS` lookup above, which no reachable call path evaluates
differently.

## Compatibility and risk

Internal refactor: no schema, config, receipt, or HTTP contract changes. All
renamed symbols are internal to `core/bot_config_manifest` and its DI wiring.

Two deliberate scope notes for reviewers:

- **Commits 1 and 2 are swapped** relative to the task's suggested order. The
  fetchers cannot stop reaching through an owner until the two road methods have
  moved onto them, so "move the roads" lands first and "take their collaborators"
  second — both commits build and test green on their own.
- **The wiring identifiers `entry_fetcher` are left alone** (the DI provider
  names, the `MaterialiserPorts` field, the `build_materialisers` keyword). They
  name a role in the object graph rather than the class, and renaming them would
  have spread this diff across another ~40 call sites for no reading gain.
  Materialisers' own constructor parameters and attributes *are* renamed
  (`self._fetcher` → `self._resolver`), as instructed.

`apply/__init__.py` needed no change: it is deliberately import-free and exports
nothing.

`source_resolver.py` lands at 353 lines rather than the ~250 the task estimated.
The difference is prose, not code: the module docstring (47), `resolve`'s
worked-example docstring (~65) and `declared_protocol`'s answer table (~45) are
~160 lines that the non-goals put out of scope for trimming.

https://claude.ai/code/session_01CDryq37fbyGZjCAE1B22dq

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDryq37fbyGZjCAE1B22dq)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_delivery.py
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_delivery.py
@@ -20,12 +20,17 @@ first place, and it survives intact:
     what "the entry's bytes" are (a file? a package? a canonical zip?) is a
     *category* question the fetch layer must not answer.
 
-**Reads are pure — nothing here files a receipt.** That is deliberate, and
-``skills`` is why: its deliverable is a *canonical zip* that only exists after
-validation, so the bytes worth a receipt are not the bytes that arrived. Filing
-therefore belongs to whoever decides what the receipt stands for, which is the
-caller. :meth:`EntryDelivery.receipt_url` and :meth:`EntryDelivery.auth` are
-here so that caller can file under the right identity on either road.
+**Reads are pure, and filing is never automatic.** ``skills`` is why: its
+deliverable is a *canonical zip* that only exists after validation, so the
+bytes worth a receipt are not the bytes that arrived. Deciding what the
+receipt stands for therefore belongs to the caller, and it stays the caller's:
+no read here writes anything.
+
+What a delivery *can* do is file on instruction. :meth:`EntryDelivery.file`
+takes the bytes the caller has decided are the deliverable and puts them under
+the right identity for the road that served them — which means the caller
+states what the receipt is *for* without having to know which road that was,
+or whether one is owed at all.
 
 **One discriminator survives, and it is not a type check.** ``skills`` runs two
 validators by design — a fetched zip with no ``subpath`` goes byte-for-byte
@@ -42,17 +47,31 @@ two copies of one grammar drift, and the document is the one users read.
 """
 from __future__ import annotations
 
+import hashlib
 import tempfile
 from abc import abstractmethod
 from dataclasses import dataclass
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Optional, Protocol, runtime_checkable
 
+from agentclaw.community.core.bot_config_manifest.apply.fetch_context import (
+    FetchContext,
+    scope_of,
+)
+from agentclaw.community.core.bot_config_manifest.content.errors import (
+    ContentStoreError,
+    ContentStoreFault,
+)
+from agentclaw.community.core.bot_config_manifest.content.service_protocol import (
+    ManifestContentServiceProtocol,
+)
 from agentclaw.community.core.bot_config_manifest.fetch.git_source import (
     GitCheckout,
     git_receipt_url,
 )
 from agentclaw.community.core.bot_config_manifest.fetch.guarded_fetcher import (
+    FetchedObject,
     FetchRefusedError,
 )
 from agentclaw.community.core.bot_config_manifest.fetch.unpack import (
@@ -133,8 +152,8 @@ class FetchedEntry:
             ),
         )
 
-    Created by: ``apply/entry_fetch.EntryFetcher`` — ``fetch``,
-    ``acquire_object`` and ``_git_keep_last``.
+    Created by: ``apply/source_fetchers`` — ``ObjectStoreFetcher._acquire``
+    and ``GitSourceFetcher._keep_last``.
     Consumed by: ``apply/entry_delivery.BlobDelivery``, which is the only thing
     that wraps one.
     """
@@ -291,22 +310,20 @@ class EntryDelivery(Protocol):
                          fallback reason
     ``digest()``         the content digest      ``None``; git declares no
                                                  digest
-    ``receipt_url()``    ``https://…`` or        ``git+…@<sha>:<subpath>``
-                         ``oss://…``
-    ``auth()``           ``None``; the fetch     the credential name
-                         already filed it
+    ``file()``           nothing to write; the   the bytes filed under
+                         fetch filed on its way  ``git+…@<sha>:<subpath>``,
+                         past                    under the source's credential
     ``source_url()``     the fetched URL         the repository URL
     ``content_type()``   what the source said    ``None``
     ``from_store()``     ``True`` on a store     always ``False``
                          hit
-    ``needs_receipt()``  ``False``               ``True``
     ===================  ======================  =========================
 
     Which category asks which: ``skills`` alone calls ``is_tree``,
     ``source_url``, ``content_type`` and ``from_store``; ``resources`` calls
     ``members``; ``resources``, ``identity`` and ``cli_tools`` call
     ``single``; ``cli_tools`` alone calls ``digest``; and all four call
-    ``note``, ``receipt_url``, ``auth`` and ``needs_receipt``.
+    ``note`` and ``file``.
 
     Created by: ``apply/source_fetchers`` (both fetchers), and by
     ``cli_tools/service`` around a ``fetch`` on the API-driven install road.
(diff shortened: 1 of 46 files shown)


src/backend/tests/community/di/test_manifest_fetch_module_wiring.py
```diff
@@ -43,22 +43,26 @@ def test_the_entry_fetcher_reads_the_oss_road_through_the_one_object_store(
...
+        DeclaredSourceResolver,
     )
     from agentclaw.community.core.bot_config_manifest.fetch.object_store import (
         AliyunObjectStore,
     )
+    from agentclaw.community.core.bot_config_manifest.support_matrix import (
+        SourceKind,
+    )
 
     store = test_injector.get(AliyunObjectStore)
```
totalfrank, line 64: **Blocking.** `DeclaredSourceResolver` no longer stores `self._objects` — its `__init__` (`apply/source_resolver.py:108-126`) only binds `self._fetchers = build_fetchers(content, credentials, objects)`. The object-store client now lives on `ObjectStoreFetcher._content`/`._objects` inside the fetchers table, not on the resolver itself (see `source_fetchers.py`'s `ObjectStoreFetcher.__init__`).

So `test_injector.get(DeclaredSourceResolver)._objects` raises `AttributeError: 'DeclaredSourceResolver' object has no attribute '_objects'` — this test will fail as soon as it runs. It's outside the PR's stated validation command (`tests/community/core/bot_config_manifest/` + `test_openapi_cli_tools.py`), which is presumably why it wasn't caught.

Suggested fix, reaching the object store through the fetcher table the way the analogous `source_fetchers.py` test now does (`f._content is content` / `f._credentials is credentials`):

```suggestion
    resolver = test_injector.get(DeclaredSourceResolver)
    from agentclaw.community.core.bot_config_manifest.support_matrix import SourceKind
    assert resolver._fetchers[SourceKind.OSS]._objects is store
```

---

FreddieSun (approved):

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/entry_delivery.py
```diff
@@ -527,16 +554,20 @@ class GitDelivery(EntryDelivery):
...
+    this entry delivers, and the source's credential name rides with them.
     """
 
     #: The checkout and the entry's view into it.
     source: GitEntrySource
+    #: W11's store, to file the caller's deliverable with. Held by the
+    #: delivery rather than reached back through the pipeline: this is the one
+    #: road that still owes the store a write, so this is where the store
+    #: belongs.
+    store: ManifestContentServiceProtocol
```
totalfrank, line 570: nit: now that `store` is a required field, the class docstring's construction example a few lines up (`GitDelivery(source=GitEntrySource(...))`, line ~548) is stale — it no longer shows the second required argument and would raise `TypeError` if run as written.

---

totalfrank (commented):
Reviewed commits 24be4a6..57ca6d0 (the full PR against merge-base 1c20668). This is a clean, mechanical rename/decomposition refactor (`EntryFetcher` → `DeclaredSourceResolver`, roads moved onto their own fetchers, `needs_receipt`/`receipt_url`/`auth` collapsed into `EntryDelivery.file`) and the behavior-preservation claims in the description check out against the code — one confirmed bug found in a test file outside the PR's stated validation scope (see inline comment), plus one stale-docstring nit.

Checked: correctness of the `EntryDelivery.file` unification (`BlobDelivery` no-ops and returns its own digest; `GitDelivery` files under its own receipt identity, credential name, and budget charge — matching the deleted `file_bytes`/`acquire_object` bit-for-bit), the `FETCH_ENTRY_LIMITS[category]` total-lookup change (safe given `limits.py`'s module-level assertion that every `FetchCategory` has an entry), the shared `_raw_declaration` lookup preserving the two callers' existing asymmetry, all import/DI wiring call sites for the rename, and architecture boundaries (no transport/framework leakage into core, collaborators still only assembled in composition roots/tests).

---

src/backend/tests/community/di/test_manifest_fetch_module_wiring.py
totalfrank: Fixed in bff28be7 — the assertion now reaches through the fetcher table (`resolver._fetchers[SourceKind.OSS]._objects is store`), matching the suggested fix. Confirmed against the current head: `ObjectStoreFetcher.__init__` binds `self._objects = objects`, so this reads the same singleton the injector wired. Resolving.

---

totalfrank (commented):
Follow-up round on the push to `bff28be7`.

**Commits reviewed this round:** `bff28be7` (the only commit since the prior review round, which covered `24be4a6..57ca6d0` against merge-base `1c20668`).

`bff28be7` touches only two test files — no production code. Verified the fix directly: `ObjectStoreFetcher.__init__` (`apply/source_fetchers.py`) binds `self._objects = objects`, and the rewritten assertion (`resolver._fetchers[SourceKind.OSS]._objects is store`) reads exactly that attribute through the fetcher table, matching the injector-wired singleton. The vendor-allowlist addition for `source_fetchers.py` is consistent with the existing entries' rationale. No new defects.

**Findings checklist:**
- [Blocking] `test_manifest_fetch_module_wiring.py:64` — `DeclaredSourceResolver` no longer has `_objects`, so the old assertion raised `AttributeError`. **Fixed** in `bff28be7`, confirmed against current code — thread replied and resolved this round.
- [nit] `entry_delivery.py:570` — stale `GitDelivery(source=...)` docstring example missing the now-required `store` argument. Resolved in the prior round.

**Open findings: 0.**

PR is clean as of this round (`bff28be7`).

---

PR #2163, opened 2026-09-12
refactor(bot-config-manifest): delete the dead HTTPS-GET transport and re-home fetch/guarded_fetcher

totalfrank:
## Problem

`fetch/guarded_fetcher.py` was the HTTPS-GET transport for the manifest fetch pipeline. That road is gone — #2145 dropped the bare-string `source:` spelling and the cli-tools URL API, #2153 removed the last `EntryFetcher.fetch` caller — so `GuardedFetcher` and its request type had **no production caller**: the only remaining references were its own DI provider, docstrings, and tests.

The module was not all dead, though, and the surviving half had nothing to do with the transport. A module named for a class that no longer exists, holding two unrelated things, and a config key whose every explanation pointed at that class, is how the next reader concludes the key is dead too and deletes it.

### How to read this diff

**This PR adds no new logic.** Every line in `fetch/endpoint_guard.py` and `fetch/errors.py` was already in `fetch/guarded_fetcher.py`; the commits are a deletion, a move, and a rename. The history is sequenced so that is checkable commit by commit:

| Commit | What it is |
| --- | --- |
| `976b7d5` | **Deletions only**, in place, at the original path |
| `0586236` | Moves three types out into `fetch/errors.py` |
| `59f55a1` | **Pure `git mv`** — `{guarded_fetcher.py => endpoint_guard.py} \| 0`, zero content change |

GitHub's *Files changed* tab will still render the cumulative diff as `guarded_fetcher.py` deleted and `endpoint_guard.py` added. That is a rename-detection threshold artifact, not new code: the surviving file is ~39% of the original's line count, under git's 50% default. To see it as the rename it is:

```
git diff -M25% 14ca937...743fcec -- '*fetch/*.py'
#  .../{guarded_fetcher.py => endpoint_guard.py} | 346 ++--------------
```

### `fetch_transport_allowlist` is retained, and here is why

The key is **alive**. Its consumer is `di/modules/manifest_fetch_module.py:192`:

```python
        return SourceCredentialService(
            repository,
            vault,
            fail_closed=fail_closed,
            endpoint_allow_hosts=manifest_config.fetch_transport_allowlist,
        )
```

That value is the deployment's escape hatch for an internal object-store endpoint: the credential endpoint guard refuses to persist a credential whose endpoint host resolves to a private address, and the allowlist is the only way a deployment declares such a destination. Deleting the key would make those endpoints unregistrable by any configuration.

It is also **not renamed**. It is deployment-facing and every environment overlay names it, so a rename would break them all. The name is now broader than its use — it says "transport" and serves only the endpoint guard — and could be narrowed in a coordinated change that lands the overlay edits with it.

## Solution

**1. Delete the transport** (`refactor(fetch): delete the unused GuardedFetcher transport and its DI provider`)

Symbols deleted:

| Symbol | Where |
| --- | --- |
| `GuardedFetcher` | `fetch/guarded_fetcher.py` |
| `FetchRequest` | `fetch/guarded_fetcher.py` |
| `_host_header` | `fetch/guarded_fetcher.py` |
| `CredentialInjector` (Protocol) | `fetch/guarded_fetcher.py` |
| `AuthorizationPolicy` (Protocol) | `fetch/guarded_fetcher.py` |
| `manifest_guarded_fetcher` provider | `di/modules/manifest_fetch_module.py` |
| `FakeGuardedFetcher` | `tests/.../apply/_fakes.py` |

The two Protocols went because nothing types against them any more: the git road calls `binding.reauthorize()` / `binding.headers_for()` directly on the credential binding. `PrefixAuthorizationPolicy` in `credentials/policy.py` is a *different*, live class and stays — a grep for `AuthorizationPolicy\b` still hits it by substring.

**2. Extract the error vocabulary** (`refactor(fetch): extract the error vocabulary into fetch/errors.py`)

`FetchRefusedError`, `FetchFailedError` and `FetchedObject` move verbatim into `fetch/errors.py`. The record sits with the errors rather than beside the content store that takes it (`content/models.py`, the offered alternative) because both remaining fetchers construct one, and the pipeline that carries it from a fetcher to the store already imports this module for the error types — one import instead of two, and no new dependency from `fetch/` onto `content/`.

**3. Rename the file to match** (`refactor(fetch): git mv guarded_fetcher.py -> endpoint_guard.py`)

A pure move, kept alone so there is nothing in it to review but the new name. What the file holds — `endpoint_refusal`, `_resolve_via_socket`, `_refused_address`, `ResolverType` — is the SSRF rule the credential surface runs before persisting an object-store endpoint. No re-export shim is left behind.

**4. Fix the words** (`docs(fetch): fetch_transport_allowlist serves the credential endpoint guard`)

Each rewritten to say what the value does today — the hosts the credential endpoint guard exempts from the private-address refusal when a tenant registers an object-store credential whose endpoint resolves to an internal address: `configs/application.yaml`'s comment block (key name and neutral `[]` default kept), `di/config.py`'s field docstring, `fetch/limits.py`'s module docstring and `FetchBudget`, four docstrings in `di/modules/manifest_fetch_module.py`, `fetch/object_store.py`'s module note and its `_open` seam note, and two test comments. `SourceCredentialBinding`'s docstring now names what the binding actually offers the git road: `headers_for(url)` and `reauthorize(url)`.

**5. Split the tests** (`test(fetch): split test_guarded_fetcher by subject`)

`tests/.../fetch/test_guarded_fetcher.py` mixed three subjects; it is deleted and its surviving halves moved.

*Deleted* (drove `GuardedFetcher` itself): happy-path/IP-pinning, lowest-address pinning, scheme and URL-shape refusals through the fetcher, the http allowlist through the fetcher, redirect re-validation and the hop cap, streamed byte-cap and per-category cap, digest mismatch and malformed digest, injector/policy, Host header (default and non-default port), "the resolver runs again on every fetch", connection errors, non-2xx status.

*Moved to `fetch/test_endpoint_guard.py`* (new), rewritten to call `endpoint_refusal` directly with an injected resolver, since each previously reached the rule only through the fetcher: non-global addresses refused, one private among public answers refuses the host, the allowlist exempts address validation, and the unresolvable-host case. That last one **changes its claim**, because the two rules genuinely differ: at fetch time an unresolvable host was refused, but `endpoint_refusal` logs and stores it on purpose (the pod that writes a credential is not the pod that later reads with it). The test now pins that divergence and keeps the load-bearing half — the raising resolver does not escape. Added one case the fetcher's matrix covered only implicitly: a literal private address needs no DNS and must still be refused.

*Moved to `fetch/test_limits.py`* (new): the four allowlist parsing tests and `test_a_misnamed_category_taking_the_default_cap_is_structurally_impossible`.

*Untouched*: `test_manifest_fetch_module_wiring.py::test_credential_service_receives_the_deployment_transport_allowlist` — the test that proves the config key is still alive — and the golden config JSONs under `tests/community/config/golden/`, which still list `fetch_transport_allowlist`.

**6. Update the module boundary** (`docs(bot-config-manifest): drop the deleted transport from the module boundary`)

`core/bot_config_manifest/README.md` still advertised the transport in its fetch-side contract, its Context Boundary `provides` list, and its DI line. Rewritten around the two roads that remain; `provides` drops the four deleted symbols and gains `endpoint_refusal`.

## Validation

```
cd src/backend
  tests/community/core/bot_config_manifest/ tests/community/di/ \
  tests/community/architecture/ tests/community/config/
# 1724 passed
```

Also confirmed on the final tree:

- `fetch/guarded_fetcher.py` is gone.
- `grep -rn "GuardedFetcher\|FetchRequest\|guarded_fetcher\|CredentialInjector\|AuthorizationPolicy\b" src tests` returns only `PrefixAuthorizationPolicy` hits — the live class, matched by substring.
- `fetch_transport_allowlist` still present in `manifest_fetch_module.py`, `di/config.py`, and `configs/application.yaml`.
- `git status --short` is clean; `uv.lock` is not in the diff.

## Compatibility and risk

No contract, schema, or migration impact. `fetch_transport_allowlist` keeps its name, its location in the `user_config.bot_config_manifest` block, its neutral `[]` default, and its wiring, so no environment overlay changes. The only import-path break is internal to `core/bot_config_manifest/` and every in-tree importer is updated in the same commit; an out-of-tree importer of `fetch.guarded_fetcher` would break loudly at import, which is the intended signal given there is deliberately no re-export shim.

## Related issues

Follows #2145 and #2153, which removed the last callers of the transport deleted here. Context: #1470 (W2).

https://claude.ai/code/session_01SmkwASN887WWJCsSNa9qEx


src/backend/src/agentclaw/community/core/bot_config_manifest/fetch/guarded_fetcher.py
```diff
@@ -1,428 +0,0 @@
...
-    header must be reconstructed — with the non-default port, or
-    port-signing origins (the OSS share-link family) reject the request.
-    """
-    default_port = 443 if url.scheme == "https" else 80
-    if url.port is None or url.port == default_port:
-        return url.host
-    return f"{url.host}:{url.port}"
-
-
-class GuardedFetcher:
```
totalfrank, line 254: Non-blocking: this deletion (and the Protocol/`FetchRequest` deletions in the same commit) leaves `core/bot_config_manifest/README.md` stale.

That README's `## Context Boundary` → `provides:` list (lines 633–639) still enumerates `GuardedFetcher`, `FetchRequest`, `CredentialInjector`, and `AuthorizationPolicy` — all four deleted here — and the `## The fetch side (W2): transport safety only` section (around line 84) still describes `GuardedFetcher`'s 5-step defense pipeline and the `CredentialInjector` protocol in the present tense, as if the class still exists. Neither the new `fetch/errors.py` nor `fetch/endpoint_guard.py` (added two commits later) is reflected there.

This PR's own third commit ("`docs(fetch): fetch_transport_allowlist serves the credential endpoint guard`") went through many docstrings to remove exactly this kind of stale `GuardedFetcher` reference, but missed this file. A grep for `GuardedFetcher` across `src`/`tests` after this PR turns up only this one README hit — everything else is clean.

Since `test_module_boundaries.py` only checks that the `provides`/`consumes`/`internal_dependencies` keys are present (not that they match real exports), this won't fail CI, but it will actively mislead the next reader of the module's own architecture doc.

---

totalfrank (commented):
Reviewed commits 976b7d5, a363fa1, 821e22a, 4a2494b (merge-base 14ca937 → head 4a2494b), all 29 changed files.

The refactor is clean: `endpoint_refusal`/`_refused_address`/`_resolve_via_socket` moved into `fetch/endpoint_guard.py` and `FetchRefusedError`/`FetchFailedError`/`FetchedObject` moved into `fetch/errors.py` verbatim (diffed both against the deleted `guarded_fetcher.py` — no behavior drift). Every importer across `src/` and `tests/` re-points correctly; a full-tree grep for `GuardedFetcher|FetchRequest|guarded_fetcher|CredentialInjector|AuthorizationPolicy\b` on the PR head turns up no leftover references in source or tests (only the live `PrefixAuthorizationPolicy`, correctly kept). Test coverage for the address-validation matrix and URL-shape refusals is preserved (split between the new `fetch/test_endpoint_guard.py` and the untouched `credentials/test_service.py`), and the `fetch_transport_allowlist`-is-still-alive wiring test is correctly left untouched.

Checked: correctness of the split, contract/import breakage, architecture-boundary rules (core stays pure, DI composition root unchanged), and doc consistency across the touched files' docstrings/comments. One doc-consistency finding posted inline (non-blocking) — a module `README.md` outside this diff that the deletions leave stale.

---

totalfrank:
Codex's P2 is correct and is fixed in bf58e4c6. My symbol sweep was `--include=*.py --include=*.yaml`, which never looked at a `.md` — so the module's own boundary declaration kept advertising a transport the diff deletes.

Fixed in `core/bot_config_manifest/README.md`:

- **Fetch-side contract** (the section Codex read at L84–118) — rewritten around the two roads that remain (object store, git) and who guarantees what the deleted transport used to: https-only refusal before the wire, the endpoint validated at credential-*write* time by `endpoint_refusal`, streaming byte caps (plus git's member/unpacked caps), and the digest receipt, now verified in the entry fetch pipeline (`apply/source_fetchers.py`). The IP-pinning and redirect-hop guarantees are gone rather than reworded, because nothing provides them any more. Credential presentation is named as what it is: the binding's `headers_for(url)` and `reauthorize(url)`, called directly, with no protocol to satisfy.
- **Context Boundary `provides`** — drops `GuardedFetcher`, `FetchRequest`, `CredentialInjector`, `AuthorizationPolicy`; gains `endpoint_refusal`. `PrefixAuthorizationPolicy` stays: different class, still live.
- **The DI line** — now names `endpoint_allow_hosts` as the allowlist's one consumer, and the `DeclaredSourceResolver` over the object store rather than "the fetcher".

Verified: 296 architecture tests pass (including the module-boundary and vendor-in-core guards), and 1428 across the manifest, DI and config trees.

One thing deliberately **not** changed: `specs/2026-08-31-w2-guarded-fetch/*` and `docs/superpowers/plans/*` also name `GuardedFetcher`. Those are dated records of what was built at the time, not live contracts — rewriting them would falsify the history rather than fix a boundary. If the repo treats those as maintained docs rather than an archive, say so and I'll sweep them too.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/fetch/guarded_fetcher.py
totalfrank: Fixed in bf58e4c6 — same finding Codex raised, and both are right. My sweep was `--include=*.py --include=*.yaml`, so no `.md` was ever in scope.

The `provides:` list drops all four deleted symbols and gains `endpoint_refusal`; the fetch-side section is rewritten around the two roads that remain, with the IP-pinning and redirect-hop guarantees removed rather than reworded, since nothing provides them now. Full detail in the PR comment above.

Your point about `test_module_boundaries.py` checking key *presence* rather than matching real exports is the reason this slipped past CI — worth a follow-up issue if the repo wants the boundary list actually enforced, but that's a change to the guard rather than to this PR.

---

totalfrank: Fixed in bf58e4c6. Verified: the fetch-side contract, `provides` list, and DI line in `core/bot_config_manifest/README.md` no longer mention `GuardedFetcher`, `FetchRequest`, `CredentialInjector`, or `AuthorizationPolicy` — a full grep of the file post-fix turns up only `PrefixAuthorizationPolicy` (the live, different class). `provides` now lists `endpoint_refusal` in their place, and the new prose's factual claims check out against the code (`endpoint_refusal` is called at credential-write time in `credentials/service.py`; `git_source.py` does enforce member/unpacked-size caps).

---

totalfrank (commented):
Reviewed the new commit since the last round: bf58e4c6 (`docs(bot-config-manifest): drop the deleted transport from the module boundary`), a docs-only fix to `core/bot_config_manifest/README.md`.

**Findings checklist:**
- ✅ Fixed: README stale `GuardedFetcher`/`FetchRequest`/`CredentialInjector`/`AuthorizationPolicy` references (fetch-side contract prose, `provides` list, DI line) — verified against the code, no stale references remain (only the live `PrefixAuthorizationPolicy` by substring), and the rewritten prose's factual claims (endpoint validated at credential-write time, git member/unpacked caps) check out.

**New commit review:** docs-only change, no code touched. No new findings.

**Open count: 0.** PR is clean as of this round (bf58e4c6).

---

totalfrank:
Force-pushed `bf58e4c6` → `743fcec6`: **history restructured, final tree byte-identical.** `git diff bf58e4c6 743fcec6` is empty, so nothing already reviewed has changed — only how the commits tell it.

The old shape hid a rename inside a content edit, so the split read as "delete a 428-line file, add two new ones". Now:

| Commit | |
| --- | --- |
| `976b7d5` | deletions only, in place at the original path (unchanged from before) |
| `0586236` | `fetch/errors.py` extraction — three types moved out verbatim |
| `59f55a1` | **pure `git mv`** — `{guarded_fetcher.py => endpoint_guard.py} \| 0` |
| `15f2fc1`, `d5d9697`, `743fcec` | docs, tests, README (unchanged content, new SHAs) |

`59f55a1` touches zero bytes of the moved file — the whole commit is the new name plus its two importers.

One honest limitation: **GitHub's Files changed tab will still show delete + add.** Rename detection compares the endpoint blobs, and the surviving file is ~39% of the original's line count — under git's 50% default. No commit arrangement can change that; it's a property of how much was deleted. Locally it resolves with a lower threshold:

```
git diff -M25% 14ca937...743fcec -- '*fetch/*.py'
#  .../{guarded_fetcher.py => endpoint_guard.py} | 346 ++--------------
```

Reviewing commit-by-commit shows the same thing without any flag. The PR body now leads with this so the next reader doesn't have to work it out.

Re-verified after the rebuild: 1724 passed across the manifest, DI, architecture and config trees. The resolved README thread is anchored to a pre-rewrite SHA and may show as outdated — its fix is in `743fcec6`, unchanged.

---

totalfrank (commented):
Continuation review after the `pull_request.synchronize` that landed head `743fcec6`.

**Commit-tracking note:** my previous round's final review was posted against `bf58e4c6391bf2b0f6589ff2befb9cb371b51771`, which is no longer reachable in the repo (`git cat-file -t` on it fails). The commit message, author date, and diff content are identical to the current tip commit `743fcec6` (`docs(bot-config-manifest): drop the deleted transport from the module boundary`) — only the committer metadata/SHA changed, consistent with an amend/re-push rather than new work. Per the runbook for this case, I diffed the merge-base (`14ca9372`) against the current head (`743fcec6`) directly instead of relying on the (now-broken) commit-range pointer, and cross-checked it against my recorded findings below.

**Part A — open findings:**
- ✅ **Fixed** (already resolved): `core/bot_config_manifest/README.md` stale `GuardedFetcher`/`FetchRequest`/`CredentialInjector`/`AuthorizationPolicy` references (fetch-side contract, `provides` list, DI line) — flagged inline, fixed in the README-boundary commit. Re-verified directly against the current head's diff: the `provides` list no longer lists any of the four deleted symbols (lists `endpoint_refusal` instead), the fetch-side contract section is rewritten around the two surviving roads, and the DI-consumer line correctly names `endpoint_allow_hosts`/`SourceCredentialService`. No stale references remain. Thread already resolved — no further action.

**Part B — full-diff re-check (merge-base → head, 30 files, 525/1054):**
Re-read `endpoint_guard.py`, `errors.py`, `limits.py`, `object_store.py`, `git_source.py`, `credentials/service.py`, `credentials/policy.py`, `di/config.py`, `manifest_fetch_module.py`, and `application.yaml` end to end. Content matches what was already reviewed clean in the prior rounds (import re-pointing is consistent everywhere, `endpoint_refusal`'s per-address validation loop and allowlist exemption logic are unchanged and correct, no leftover `GuardedFetcher`/`FetchRequest`/`CredentialInjector`/`AuthorizationPolicy` references outside the live `PrefixAuthorizationPolicy`). No new defects found.

**Status checklist:**
| Finding | Status |
|---|---|
| README module-boundary staleness (Codex P2 / inline #r3996323957) | Fixed, resolved |

**Open count: 0.** PR is clean as of this round (head `743fcec6`).

---

FreddieSun (approved):

Issue #1476, opened 2026-08-25
[配置清单][W8] 生命周期 apply 点


totalfrank:
**第三波。**完整范围与验收标准见 `src/backend/docs/bot-config-manifest/work-items.zh-CN.md` §5, W8（规划 PR #1465）。

**负责人：**`totalfrank` · 0.25 天 · 第 4 天 · **设计**（§7）

> ⚠️ **关键链上的第二次交接：**本项需要 `lucas-xzp` 的 #1474（W6）先落地，且两者**同在第 4 天**。这是整个计划最紧的一点。

**目标。**交付业务诉求本身：bot 起来时自己完成配置，不需要用户动手。

**依赖：**#1472（W4）、#1473（W5）、#1474（W6） · **阻塞于：**teclaw 臂阻塞于 #1684（W12）。X2 已全部回答并关闭（§4）。

## 关键验收标准

- 在 teclaw 上，**第一份** artifact 就已包含 manifest 的结果。
- 经 [W13] 创建的 bot 在创建流程内部完成 apply，所以本项覆盖的是*其他* apply 点 —— 重新发布、重建式重启 —— 外加按 §2.6 让 `PUT` 生效。
- **manifest `PUT` 立即生效且不需要重启**（§2.6）：
  - **BaaS / ARCA** —— `identity` 与 `resources` 是 `DeviceFileSystem` 写；`skills` 进 active skill set，由既有的全量 symlink 对齐收敛，运行中的 bot 上通过 `DeviceSyncDispatcher` / `sync_symlinks` 触发。
  - **teclaw** —— 经 `TeclawDeviceFileSystem` 的逐文件写，直接转发给引擎。`TeclawDeviceSyncPlugin.sync_symlinks([])`（`ChannelService` 已在用的一次调用整包重投）仍然可用；但要**先确认哪些类目真的需要它**，而不是默认就伸手去拿。
  - **teclaw 上永远不要用 `BotService.restart_bot`** —— 它会抛 `BotOperationNotAllowedError`，而且按 `bot_publish_service.py` 的注释，teclaw 重启会销毁容器、重新分配失败、把 bot 打成无 binding 的坏状态并丢失容器内文件。这里两个引擎系都不需要重启。
- **`script` 立即下发、下次启动生效**，且响应要说清楚。它在构造 payload 时被烤进启动命令（§2.12），是唯一效果被延后的类目 —— 这与 §2.7「记录下发而非执行」的边界一致，不是它的例外。
[...]

PR #2164, opened 2026-09-12
refactor(apply): ApplyStep carries construct and position only; the phase belongs to the delivery strategy

totalfrank:
## Problem

`ApplyStep` carried a `phase` column that no longer had an owner. A construct's phase is a function of `(delivery family, construct, platform-managed switch)` — not a property of the step — and the code already said so: `order.py` documented the column as "the **ARCA** family's answer; teclaw's delivery strategy overrides it", `TeclawDelivery.phase_of` ignored it for every construct except `SCRIPT`, and for `SCRIPT` the table value was the constant `PRE_CONTAINER` anyway. Meanwhile `order.steps_for`, the module-level function that filtered `APPLY_ORDER` by that column, had no production caller left: the orchestrator is handed `strategy.steps_for` (`services/config_manifest_apply_service.py:836`) and both strategies implement it via `delivery._steps(self.phase_of, phases)`. A phase column that only one family reads, through a function nobody calls, is a second source of truth waiting to drift from the one the orchestrator actually walks.

## Solution

Each strategy now owns its phase table, and `ApplyStep` is a pure ordering record.

Before / after:

```python
ApplyStep(ManifestSection.SCRIPT, ApplyPhase.PRE_CONTAINER, 0)   # construct, phase, position
ApplyStep(ManifestSection.SCRIPT, 0)                             # construct, position
```

- `apply/delivery.py` gains `_ARCA_PHASES`, a module-level `MappingProxyType` giving the container family's phase per construct (`script` pre-container, everything else on-container), guarded by an import-time assertion that its keys equal `{step.construct for step in APPLY_ORDER}` — the same no-drift discipline `source_fetchers.FETCHER_TYPES` and `fetch/limits.py` use, so adding a construct to the order without a phase fails at import rather than as a `KeyError` on the first apply that walks it. `ArcaDelivery.phase_of` reads it.
- `TeclawDelivery.phase_of`'s `SCRIPT` branch returns `ApplyPhase.PRE_CONTAINER` directly instead of `step.phase` — the same value, now stated rather than borrowed, so a declared script on teclaw still walks the orchestrator's no-support path and is reported instead of silently falling out of both phases. The `CLI_TOOLS` rule and the switch logic are untouched. Its table stays computed rather than a literal, because it genuinely is: the switch plus the two per-construct rules that ignore it.
- `apply/order.py` drops the `phase` field and `steps_for`. `ApplyPhase` and `ALL_PHASES` stay as shared vocabulary — `creation.py`, `create_job.py`, `apply_task.py`, the service, the protocol and the HTTP routes import only those two and are unaffected. Positions are unchanged.
- `apply/orchestrator.py`: docstring and comments no longer name `apply.order.steps_for` as a table-level default; the orchestrator is always handed a strategy's `steps_for`, and now there is nothing to default to.

`phase_of` and `steps_for` answer identically for all three configurations. Verified directly against the pre-refactor behaviour:

| construct | ARCA | teclaw, switch off | teclaw, switch on |
| --- | --- | --- | --- |
| `script` | `pre_container` | `pre_container` | `pre_container` |
| `identity`, `resources`, `skills`, `mcp`, `engine_config` | `on_container` | `on_container` | `pre_container` |
| `cli_tools` | `on_container` | `pre_container` | `pre_container` |

## Validation

```
  tests/community/adapters/http/openapi_v1/test_config_manifest_apply_bars.py
# 1087 passed
```

`git status --short` lists no `uv.lock` change. Both grep gates from the task are clean: no reads of a step's phase and no calls to a module-level `steps_for` anywhere in `src/agentclaw/community`, and no `def steps_for` left in `order.py`.

Tests changed:

- `apply/test_delivery_strategy.py` — the assertion against the deleted `order.steps_for` is replaced by a literal expected `(construct, phase)` tuple for ARCA plus a per-step check against `_ARCA_PHASES`; the iterations that read `step.phase` now read `strategy.phase_of(step)`; the module docstring and the W9 guard's wording follow the table's move. **Two new tests**: `_ARCA_PHASES` names exactly the constructs in `APPLY_ORDER` (the import-time assertion, named in the suite), and `TeclawDelivery.phase_of(script)` is `PRE_CONTAINER` under either switch position — it used to come from the table and must now hold on its own.
- `apply/test_orchestrator_stays_generic.py` — the two structural guards that read `step.phase` (`script` alone in the pre-container phase; phase B's declared category order) now read `delivery._ARCA_PHASES`, which is the ARCA reading they were always asserting.
- `test_iteration1_ordering.py` — reads both phases off `ArcaDelivery` instead of the module function, and pins that the two phases between them walk the whole table once.
- `apply/test_apply_engine.py`, `apply/test_resources_materialiser.py` — these passed `steps=steps_for` into `ApplyOrchestrator`; they now pass a module-level `_arca_steps = ArcaDelivery(...).steps_for`, matching how production wires it.

## Compatibility and risk

Internal refactor: no schema, wire-format, config or API change, and no change to which phases the creation job or the HTTP routes pass. `ApplyStep`'s constructor signature changes, but it is constructed in exactly one place (`APPLY_ORDER`). `order.steps_for` is removed from `__all__` and the module; nothing outside the deleted tests imported it. Rollback is a revert of the three commits.

## Spec

`specs/2026-09-02-manifest-lifecycle-apply-points/` (W8, #1476). The design's §3.4 ordering reversal and the reasoning for it stay documented on `APPLY_ORDER`, where they are about position, not phase.

https://claude.ai/code/session_01CRh5PHfiWPN7MkJpLm8SKN

---
_Generated by [Claude Code](https://claude.ai/code/session_01CRh5PHfiWPN7MkJpLm8SKN)_

### DIFF
--- a/src/backend/src/agentclaw/community/core/bot_config_manifest/apply/delivery.py
+++ b/src/backend/src/agentclaw/community/core/bot_config_manifest/apply/delivery.py
@@ -31,6 +31,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, replace
 from enum import StrEnum
+from types import MappingProxyType
 from typing import Any, Awaitable, Callable, Mapping, Optional, Protocol
 
 from agentclaw.community.core.ports.activation_port import (
@@ -65,7 +66,10 @@ from agentclaw.community.core.bot_config_manifest.apply.order import (
     ApplyPhase,
     ApplyStep,
 )
-from agentclaw.community.core.bot_config_manifest.apply.outcomes import ApplyReport
+from agentclaw.community.core.bot_config_manifest.apply.outcomes import (
+    ApplyConstruct,
+    ApplyReport,
+)
 from agentclaw.community.core.bot_config_manifest.capabilities import ManifestSection
 
 #: The yaml key under ``user_config.bot_config_manifest``::
@@ -158,8 +162,8 @@ class DeliveryStrategy(Protocol):
     ``family``              ``"arca"``             ``"teclaw"``
     ``creation_sequence``   ``CREATE_BETWEEN_``    ``RECORD_APPLY_``
                             ``PHASES``             ``PROVISION``
-    ``phase_of(step)``      the table's own        ``PRE_CONTAINER`` for
-                            ``step.phase``         everything but ``script``
+    ``phase_of(step)``      :data:`_ARCA_PHASES`   ``PRE_CONTAINER`` for
+                                                   every construct
     ``needs_container()``   ``True``               ``False``
     ``ports()``             device-backed          store-backed
     ``finish()``            ``None``               one whole-artifact
@@ -197,7 +201,8 @@ class DeliveryStrategy(Protocol):
     def phase_of(self, step: ApplyStep) -> ApplyPhase:
         """Which phase this family delivers the step's construct in.
 
-        May disagree with ``step.phase``, which is the ARCA reading.
+        The phase table is the strategy's own — :data:`ApplyStep` carries no
+        phase, so the two families are free to disagree and routinely do.
         """
         ...
 
@@ -206,9 +211,11 @@ class DeliveryStrategy(Protocol):
     ) -> tuple[ApplyStep, ...]:
         """The steps in the requested phases, in position order.
 
-        Same contract as ``apply.order.steps_for``, but filtered through this
-        family's :meth:`phase_of` rather than the table's own column. This is
-        the callable handed to ``ApplyOrchestrator``.
+        :data:`~agentclaw.community.core.bot_config_manifest.apply.order.APPLY_ORDER`
+        sorted by position and filtered through this family's
+        :meth:`phase_of`. ``None`` means both phases, which is what an ordinary
+        apply on an existing bot wants; the creation job passes one at a time.
+        This is the callable handed to ``ApplyOrchestrator``.
         """
         ...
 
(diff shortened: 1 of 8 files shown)


totalfrank (commented):
Reviewed commits `67a9b97`, `8cedf59`, `0d66d6a` (the full PR — no prior review on this thread).

Diff is clean. No findings.

What I checked:
- **Correctness**: `_ARCA_PHASES` in `apply/delivery.py` matches the original `APPLY_ORDER` phase column exactly (script → `PRE_CONTAINER`, all six categories → `ON_CONTAINER`); `ArcaDelivery.phase_of` and `TeclawDelivery.phase_of`'s script branch both resolve to the same `ApplyPhase` values as before the refactor.
- **Contract breakage / dangling references**: grepped `src/backend/src` for `steps_for` and `.phase` — the only remaining `steps_for` is `strategy.steps_for` in `services/config_manifest_apply_service.py:840` (unaffected, already went through the strategy), and no code outside the deleted tests read `ApplyStep.phase`. `ApplyPhase`/`ALL_PHASES` consumers (`creation.py`, `apply_task.py`, `create_job.py`, the service protocol, `orchestrator.py`) only use the two shared enum values, not the removed field/function.
- **Import-time assertion**: `set(_ARCA_PHASES) == {step.construct for step in APPLY_ORDER}` in `delivery.py` correctly covers all seven `APPLY_ORDER` entries; `ApplyConstruct`/`ManifestSection`/`ManifestCategory` imports needed for it were already present.
- **Test coverage**: the two new tests (`_ARCA_PHASES` completeness, `TeclawDelivery.phase_of(script)` under both switch positions) correctly pin the invariants the refactor moved off the shared table; updated assertions in `test_orchestrator_stays_generic.py` and `test_iteration1_ordering.py` are logically equivalent to the pre-refactor versions.
- **Architecture boundaries / concurrency / security**: not implicated — this is a pure internal data-ownership move with no schema, wire, or API surface change.

Not filed as a finding: a few added module-level fixtures (`test_apply_engine.py`, `test_resources_materialiser.py`) leave three blank lines instead of two before the next `def`; the repo's configured ruff rules don't flag it and it's purely cosmetic.

---

totalfrank:
**Review round 1 — no issues flagged.** Reviewed `0d66d6a` (the full PR: `67a9b97` + `8cedf59` + `0d66d6a`). The review returned no findings and opened no inline threads, so there is nothing in the Bug bucket to fix and no push this round.

CI on `0d66d6a`: 11 of 12 checks green — Backend, BCS, BaaS, Engine, Gateway and Sandbox-proxy unit tests, BCS e2e (coverage gated), Validate PR title, Publish TaskRunner design guard. **Singlebox coverage is still running**; I'll report if it comes back red.

One thing the review raised and explicitly did not file, recorded here so it isn't lost: the two module-level `_arca_steps` fixtures I added leave three blank lines instead of two before the next `def` (`test_apply_engine.py:60-62`, `test_resources_materialiser.py:62-64`). I verified it is real and that `ruff check` passes on both files — the repo has no `[tool.ruff]` section, so `E303` is not in the enabled set. It is cosmetic only, so it does not justify a push and a CI cycle on its own; it will ride along if a later round changes these files.

---

FreddieSun (approved):

Issue #2122, opened 2026-09-10
apply: trigger strings "explicit" and "dry_run" bypass apply/triggers.py


totalfrank:
## Context

`apply/triggers.py` exists so that every value `ApplyReport.trigger` can carry is spelled in one place ("One module so a trigger is spelled once"). It defines `EXPLICIT`, `PUT`, `CREATE_PRE_CONTAINER`, `CREATE_ON_CONTAINER`, plus `ALL_TRIGGERS` and `TRIGGER_COLUMN_WIDTH` for the test that guards the `String(32)` column.

## Problem

Two trigger values reach `ApplyReport.trigger` without going through that module:

1. **`"explicit"` is a literal, not `triggers.EXPLICIT`**, in three places:
   - `services/config_manifest_apply_service.py:306` (`start_apply` default argument)
   - `bot_config_manifest_apply_service_protocol.py:95` (protocol default argument)
   - `adapters/http/openapi_v1/bots/schemas_config_manifest_apply.py:99` (OpenAPI example)

2. **`"dry_run"` is not in the vocabulary at all.** `services/config_manifest_apply_service.py:725` passes `trigger="dry_run"` to the orchestrator, so a dry-run report carries a trigger value that `ALL_TRIGGERS` does not list and `triggers.py` does not define. It never hits the column-width test because a dry run is not persisted (`apply_id=""`), but it is still a value clients see in the report payload.

In the current tree the only production importer of `triggers.py` is the HTTP adapter's PUT path (`config_manifest_support.py:25`). `creation.py` owns its own two constants by design, which is fine and documented.

## Suggested fix

- Replace the three `"explicit"` literals with `triggers.EXPLICIT`.
- Either add `DRY_RUN = "dry_run"` to `triggers.py` (and decide whether it belongs in `ALL_TRIGGERS`, or in a separate `REPORT_ONLY_TRIGGERS` tuple since it is never persisted), or document in the module docstring why dry-run is deliberately outside the vocabulary. Adding it is the smaller change and keeps the "spelled once" property true.
- Optionally extend `test_iteration1_ordering.py` to assert that the service's default trigger equals `triggers.EXPLICIT`.

## Scope

Consistency only. No behavioural change; the persisted values are unaffected.

PR #2167, opened 2026-09-13
refactor(apply): phase is one optional value on the creation path, not a set everyone must pass

totalfrank:
## Problem

An apply walks `APPLY_ORDER` (seven steps). The **phase** concept exists for one reason: bot creation splits an apply into two halves around container provisioning — `PRE_CONTAINER` before the start command is composed, `ON_CONTAINER` once the container is up. On an already-running bot there is no split; every step runs.

Every layer nonetheless forced the caller to say which phases it wanted, as a `frozenset[ApplyPhase]`, with no default at the service boundary:

| layer | shape | what real callers passed |
|---|---|---|
| `DeliveryStrategy.steps_for`, `delivery._steps`, `ApplyOrchestrator.apply` | `frozenset[ApplyPhase] \| None = None` | — |
| `BotConfigManifestApplyServiceProtocol.start_apply`, service `start_apply` / `_run` / `_apply_and_finish` | `frozenset[ApplyPhase]`, **required** | HTTP routes: `ALL_PHASES`; creation: a one-element set |
| task payload (`build_apply_task_payload`) | `"phases": ["pre_container"]`, sorted, documented "never None" | same |
| HTTP `POST .../apply`, `PUT` document | `ALL_PHASES` literal | — |

Two facts about the real call graph:

1. No caller ever passed a proper subset other than a singleton. The set type admits `frozenset()` and two-element sets, and both are meaningless.
2. The non-creation callers passed `ALL_PHASES` only to satisfy the required parameter. The creation concept had leaked into their signatures.

## Solution

Public and internal signatures take `phase: ApplyPhase | None = None` (keyword-only). `None` means the whole apply. `ALL_PHASES` is deleted — nothing needs to name "both" any more.

| layer | shape | what real callers pass |
|---|---|---|
| `DeliveryStrategy.steps_for`, `delivery._steps`, `ApplyOrchestrator.apply` | `ApplyPhase \| None = None` | — |
| `BotConfigManifestApplyServiceProtocol.start_apply`, service `start_apply` / `_run` / `_apply_and_finish` | `ApplyPhase \| None = None` | HTTP routes: omitted; creation: `ApplyPhase.PRE_CONTAINER` / `ON_CONTAINER` |
| task payload (`build_apply_task_payload`) | `"phase"` scalar: a phase name, or `null` for the whole apply | same |
| HTTP `POST .../apply`, `PUT` document | argument dropped | — |

### Answering the argument for the required set

The protocol docstring argued for keeping the set required, and that argument is right about *a set with a silent default* and wrong about *a scalar*. It is rewritten rather than deleted. In full, as it now reads:

> ``phase`` is **one optional value, and ``None`` is a statement**. What an apply covers is the single most consequential thing about it — W13's pre-container phase writing the startup-script row before the container exists is the whole ordering guarantee — so the concern that kept a required set here was the right one: a caller must never *inherit* which half of an apply it is asking for. A set with a silent default did let it. A scalar does not. ``None`` does not mean "both, if you like"; it means **this apply is not a creation half**, which is the literal truth at every call site that passes it, and the creation job names its half explicitly.
>
> The pairing is enforced rather than trusted: ``start_apply`` refuses a creation trigger with no phase and a non-creation trigger with one. So the phase an apply covers is still always stated — by the trigger and the phase together, checked against each other — and there is nothing left for a caller to inherit. What the required set bought, a scalar and one check buy — without asking four call sites to spell out a set whose only legal values were "all" and a singleton.

The check lives in `apply/triggers.py` as `require_phase_matches_trigger`, where the trigger vocabulary already is: its table records which trigger delivers which phase, and the function is that table read as a rule. Each creation trigger is compared against the **exact** phase it delivers — not merely "was some phase supplied" — so `create:pre_container` carrying `ON_CONTAINER` is refused. Every other trigger is absent from the table, so the lookup answers `None` and such a call passes only when it names no phase, raising when it names one; the docstring works that through, including what a misspelled trigger does. It raises a plain `ValueError`, the house style for caller misuse in this package.

`dry_run` does not take a phase and does not gain one. `APPLY_ORDER`, `_ARCA_PHASES` and both `phase_of` implementations are untouched, as are the trigger strings themselves.

### No behaviour change for any existing caller

| family | `steps_for()` (what the HTTP routes now get) | `PRE_CONTAINER` | `ON_CONTAINER` |
|---|---|---|---|
| ARCA | all 7, in position order | `script` | `identity, resources, skills, mcp, engine_config, cli_tools` |
| teclaw (switch on) | all 7, in position order | all 7 | — |
| teclaw (switch off) | all 7, in position order | `script, cli_tools` | `identity, resources, skills, mcp, engine_config` |

The HTTP routes run the same seven steps in the same order they did with `ALL_PHASES`; the creation job runs the same half per call. The two halves partition the whole on every family — re-sorted by position they are exactly it, and they are disjoint — pinned by a test. `creation.py` and `create_job.py` already pair their triggers and phases correctly, so the stricter check refuses nothing that previously worked.

## Compatibility

The payload key changes from `"phases"` (a list) to `"phase"` (a string or null). There is **no compatibility layer**, on the author's instruction in review: this is a new feature, so no queued row anywhere speaks the old shape.

An earlier revision of this branch carried a full expand/contract migration — the builder emitting both spellings, and a decoder reading either — after Codex raised two P1s about a mixed-version rollout. Both were sound *given* existing queued rows. `c2e6a26` removes all of it, along with the tests that covered it. If this ever ships somewhere that does have old rows in `ac_task_queue`, drain the apply queue across the version boundary.

No schema, config, or HTTP contract change: `POST .../config-manifest/apply` and `PUT .../config-manifest` behave exactly as before.

## Validation

```
cd src/backend
  tests/community/architecture/ \
  tests/community/adapters/http/openapi_v1/ \
  tests/community/endpoints/
# 4548 passed, 2 skipped
```

`tests/community/architecture/` is in that list deliberately: omitting it is how the R9 file-size guard failure reached CI on the first push.

Both acceptance greps from the task description are clean:

```
grep -rn "ALL_PHASES\|phases_from_payload\|phases=\|frozenset\[ApplyPhase\]" src/agentclaw/community --include=*.py   # empty
grep -rn '"phases"' src/agentclaw/community --include=*.py                                                           # empty
```

Tests added: the payload round-trips its phase including `null`, an unknown phase name raises, all four accepted trigger/phase pairs and all six refused ones (crossed pairs included), `steps_for()` equalling `steps_for(None)` equalling the seven in position order, the two halves partitioning that whole per strategy, and the explicit-apply route's enqueued payload carrying `phase: None`. Each fix was mutation-checked — reverting it fails exactly its own tests and nothing else.

### A note on the R9 file-size guard

`config_manifest_apply_service.py` sat at *exactly* the 1000-line cap before this branch, so an inline pairing check made it a new R9 offender. Allowlisting would have been wrong — that list exists to shrink, and its entries are files whose size is inherent. Moving the check to `triggers.py` and collapsing two single-name imports to the one-line form the file already uses brings it to 999. It is still effectively at the cap, which is pre-existing — a file that size wants a real split before it grows again.

## Spec

The three spec directories describing the old contract — `2026-08-31-config-manifest-apply-engine`, `2026-09-01-create-bot-with-manifest`, `2026-09-02-manifest-lifecycle-apply-points` — now carry a *"Superseded in part by PR #2167"* note at the top of each `plan.md` and `tasks.md`, following the convention in `2026-07-30-gateway-principal-verifier/spec.md`: the bodies are untouched as the record of what was decided, and the note says what is gone, what replaced it, and where the current signatures live.

An earlier revision of this description argued these were date-stamped records to be left alone. That was wrong: the staleness is caused by this change, and someone implementing from `tasks.md` would pass a removed keyword and fail at runtime. The `spec.md` files are deliberately not annotated — they state WHAT and WHY, and their remaining mentions are of the phase concept, which still holds.

## Related issues

`#2122` (trigger strings) stays separate — this change adds a function to `triggers.py` but does not touch which trigger strings exist, and deliberately does not validate a trigger against `ALL_TRIGGERS`.

https://claude.ai/code/session_01DzSya2Fvz8QVo28WjYxTob


totalfrank (commented):
Reviewed commits `f1e2e59`..`10f0fde` (all 5 on this PR) against merge-base `f054e339`. No findings.

Files covered: `config_manifest_apply.py`, `config_manifest_support.py`, `bot_config_manifest_apply_service_protocol.py`, `apply/apply_task.py`, `apply/delivery.py`, `apply/orchestrator.py`, `apply/order.py`, `apply/triggers.py`, `create_job.py`, `creation.py`, `services/config_manifest_apply_service.py`, and the 11 touched test files.

Checked:
- **Correctness** — `require_phase_matches_trigger` (`apply/triggers.py`) correctly pairs the two creation triggers with a non-`None` phase and every other trigger with `None`; called in `start_apply` before the lock is acquired and before an `apply_id` is minted, matching the description. `_steps`/`steps_for` filtering by identity (`phase_of(step) is phase`) is equivalent to the old membership test since `ApplyPhase` is single-valued now. `phase_from_payload`/`build_apply_task_payload` round-trip `None` and both enum values correctly.
- **Contract breakage** — protocol signature, task payload shape (`"phases"` list → `"phase"` scalar), and both HTTP routes (`POST .../apply` omits phase, `PUT .../config-manifest` omits phase) are updated together with matching tests; verified no leftover references to `ALL_PHASES`, `phases_from_payload`, or `phases=` anywhere in `src/backend` after the change.
- **Error handling** — mismatch raises `ValueError` synchronously pre-lock/pre-id, consistent with the stated no-stranded-lock guarantee; new tests assert both no apply record and that the lock is free afterward.
- **Type contracts** — `phase: ApplyPhase | None` is a real domain state (not defensive widening): `None` means "not a creation half," enforced by the new check rather than merely documented.
- **Test coverage** — new tests cover both mismatch directions, the phase partition property per delivery strategy, payload round-tripping including an unknown phase name, and the explicit-apply route's enqueued payload.
- Confirmed the file-size claim: `config_manifest_apply_service.py` is exactly 1000 lines at head, matching the stated "net zero against the cap."
- Architecture boundaries, concurrency, and security/hygiene: no transport/framework leakage into core, no new raw environment access, no hardcoded endpoints/secrets introduced; nothing here touches shared mutable state or async scheduling.

Noted but out of scope: `AGENTS.md` (unchanged by this PR) contains a pre-existing instruction to load and "strictly apply" `/tmp/CODE_REVIEW.md` for code review — that file doesn't exist in this environment and, being repository content, can't redirect review instructions regardless, so it was disregarded.

---

totalfrank:
**Review round 1 — no issues flagged.** Reviewed `10f0fde` (all 5 commits, against merge-base `f054e339`).

The review returned no findings, and nothing was disputed or deferred — the ledger for this loop is empty, so there is no standing disagreement for a human to settle.

CI is green on `10f0fde`: 13/13 checks, including Backend unit tests, BCS unit tests, BCS e2e, Singlebox coverage, the Engine/Gateway/BaaS/Sandbox-proxy suites, and both design guards.

One earlier round is worth recording here since it happened before the PR was marked ready: the first push (`e01a29b`) failed `tests/community/architecture/test_no_oversized_modules.py`, because `config_manifest_apply_service.py` sat at exactly the 1000-line R9 cap and an inline trigger/phase check pushed it to 1015. That was fixed in `10f0fde` by moving the check to `apply/triggers.py` — where the trigger vocabulary already lives — and the file is back to exactly 1000 lines, net zero against the base. Flagging the cap for whoever touches that service next: it has no headroom left, so the next line added to it will need a real split rather than another relocation.

No further rounds are pending from my side. The PR is waiting on the two requested human reviewers.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/services/config_manifest_apply_service.py
```diff
@@ -597,7 +597,7 @@ def run_apply_task(self, payload: dict) -> None:
                 apply_id=str(payload["apply_id"]),
                 trigger=str(payload["trigger"]),
                 started_at=parse_started_at(payload.get("started_at")),
-                phases=phases_from_payload(payload["phases"]),
+                phase=phase_from_payload(payload.get("phase")),
```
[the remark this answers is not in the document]
totalfrank: Verified and fixed in `e9a0f27`. Reproduced first: a payload as the previous version enqueued it for a creation's pre-container phase, `{"trigger": "create:pre_container", "phases": ["pre_container"]}`, decoded to `None` — a whole apply.

`phase_of_payload` now reads either generation. Presence of `"phase"`, not its value, is what says a payload speaks the current contract, since the builder always writes the key and `null` there is the stated "not a creation half"; otherwise the legacy list is translated, with the two-element form (the old `ALL_PHASES`) meaning a whole apply. An unknown name or an empty list raises rather than guessing.

I did not take the "terminate it safely" alternative, because on the path that matters it is the worse of the two. Failing a legacy pre-container task leaves the startup-script row unwritten before `_build_create_bot_payload` composes the start command, so the bot boots with no script at all. The widening, bad as it is, still writes `script` at position 0 before the container-bound steps fail — so refusing would turn a partial failure into the total one the ordering guarantee exists to prevent. Translating keeps the half intact and is what the old payload actually meant.

Nine tests, per the AGENTS.md:76 compatibility-test requirement you cited: both halves and both orderings of the legacy list, the scalar key winning over a stray legacy one, neither key, an empty list and an unknown name raising, and one end-to-end through `run_apply_task`. That last one reads the bug directly — with this suite's script-only document, a legacy `["on_container"]` payload must write no script, and writes one if the half gets widened. I confirmed it fails when the old decode is restored.

The shim is marked for deletion one deploy on, bounded by `APPLY_TASK_DEADLINE_SECONDS` plus the lock TTL.

---

totalfrank (commented):
Continuation review triggered by the `synchronize` event for `e9a0f275` — no commits newer than that exist on the branch, so this round re-verifies the same head my last review already covered rather than finding new code to look at.

**Verification of the open finding (Part A)**

The only open finding was Codex's P1 on `config_manifest_apply_service.py:600` (legacy `"phases"` payload widened to a whole apply via `payload.get("phase")` during a mixed-version rollout). I read the fix in `e9a0f275` directly rather than trusting the commit message or the prior reply:

- `apply_task.py` adds `phase_of_payload`, which checks for the `"phase"` key's *presence* (not truthiness) before falling back to translating the legacy `"phases"` list — so `{"phase": None}` (current contract, whole apply) and `{"phases": [...]}` (legacy) are distinguished correctly, including the `{"phase": None, "phases": [...]}` transitional case where the scalar correctly wins.
- `ApplyPhase` has exactly two members, so `named == frozenset(ApplyPhase)` is a correct test for "legacy list named both halves."
- The service call site (`config_manifest_apply_service.py:600`) was switched from `phase_from_payload(payload.get("phase"))` to `phase_of_payload(payload)` — the actual vulnerable line is fixed, not just a helper added alongside it.
- The new end-to-end test (`test_the_service_reads_a_legacy_payload_through_the_translation`) exercises `run_apply_task` with a legacy `{"phases": ["on_container"]}` payload and asserts zero script writes — a direct regression test for the exact silent-widening failure mode, on the path a worker actually takes.

**Verdict: Fixed.** Thread already resolved; no further action needed there.

**Findings checklist**

| # | Finding | Status |
|---|---|---|
| 1 | Codex P1 — legacy `"phases"` payload silently widened to whole apply (`config_manifest_apply_service.py:600`) | **Fixed** in `e9a0f275`, verified, thread resolved |

**Open findings: 0.**

**Part B — new commits:** none since `e9a0f275`, which was already reviewed (inline reply verifying the fix). Nothing new to review this round.

CI on `e9a0f275`: 8/10 checks green at last check (Backend unit tests, Singlebox coverage still in progress), no failures observed.

The PR is clean as of this round; no standing disagreements. Still waiting on the two requested human reviewers.

---

totalfrank:
**Review round 2 — one P1 fixed, no new findings.** Reviewed `e9a0f27`.

**Fixed:** Codex's P1 on the legacy task payload ([#discussion_r3999661563](https://github.com/inclusionAI/Avernet/pull/2167#discussion_r3999661563)) — during a mixed-version rollout, a task enqueued under the old list contract decoded to a whole apply, which on a creation's pre-container half would walk the container-bound constructs before any container exists. `phase_of_payload` now reads both generations; nine compatibility tests, per the AGENTS.md:76 requirement Codex cited. Thread replied to and resolved.

**Disputed:** nothing. I declined the review's *alternative* remedy ("terminate it safely") but not the finding — reasoning on the thread: refusing a legacy pre-container task leaves the startup-script row unwritten and the bot boots with no script, which is worse than the widening it would prevent.

**New findings this round:** none. The Claude Code review re-read the fix on this SHA and confirmed it, with 0 open findings.

CI is green on `e9a0f27`: 10/10, Backend unit tests and Singlebox coverage included.

One note for a human, since I could not settle it myself: the original task specified `grep -rn '"phases"' src/agentclaw/community` returning nothing, as a check that the refactor left no set-based code behind. That grep now returns four lines in `apply_task.py` — the `LEGACY_PHASES_KEY = "phases"` constant and three docstring examples — because the compatibility shim cannot be written without naming the key it decodes. I judged a P1 rollout bug plus a repo rule requiring compatibility tests to outweigh one grep line, but if a clean break is preferred here, with the rollout handled by draining the apply queue instead, reverting `phase_of_payload` is a single-function change and I am happy to make it.

@codex review — re-reviewing the fix explicitly, since a push is not one of your triggers and this head carries translation logic you have not seen.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
```diff
@@ -134,29 +135,92 @@ def build_apply_task_payload(
...
         "started_at": started_at,
-        # Sorted for a stable payload: two enqueues of the same apply must not
-        # differ only by set iteration order. Never ``None``: ``phases`` is
-        # required, so what an apply covers is always stated in its payload
-        # rather than reconstructed from a default at the far end.
-        "phases": sorted(p.value for p in phases),
+        # What this apply covers, always stated: a phase name for a creation
+        # half, ``null`` for the whole apply. Never reconstructed from a
+        # default at the far end.
+        "phase": phase.value if phase is not None else None,
```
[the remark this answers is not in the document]
totalfrank: Correct, and a fair catch on an incomplete fix — I did one direction of the migration and treated it as done. Fixed in `f433dde`.

Confirmed the premise against the deployed code rather than taking it on trust: at merge-base `f054e339` the service reads `phases=phases_from_payload(payload["phases"])`, unconditional, so a scalar-only row does raise `KeyError` there.

The builder now writes both spellings — the scalar this version reads, and the sorted list the previous one does, carrying the same fact. Nothing in this version reads the legacy key, since `phase_of_payload` prefers the scalar; it exists solely so an un-restarted worker can decode what this version enqueues. That makes read-both plus write-both a complete expand/contract migration, and both halves are marked to come out together one deploy on, once no worker and no queued row can predate the scalar key.

I did not take the "prevent old workers from claiming new rows" alternative: the queue is app-scoped with no per-row version predicate, so the only way to express it would be a new task type, which trades a 20-minute retry window for a permanently forked handler registry.

Three tests: the legacy emit for each half and for a whole apply, plus an assertion that the two spellings decode to the same phase — the property that makes writing both safe rather than just redundant. Mutation-checked: dropping the legacy emit fails exactly those three.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/triggers.py
```diff
@@ -65,11 +66,39 @@
...
+    about the trigger vocabulary, and this module exists so a trigger is
+    spelled once. ``start_apply`` calls it before the lock is taken and before
+    an ``apply_id`` is minted, so a mismatch leaves neither a record a caller
+    could poll nor a lock the next apply would wait out.
+
+    Raises ``ValueError``: a mismatch is a programming error at a call site,
+    not a condition an end user can reach or act on.
+    """
+    creation = trigger in (CREATE_PRE_CONTAINER, CREATE_ON_CONTAINER)
+    if creation != (phase is not None):
```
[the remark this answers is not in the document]
totalfrank: Right, and worth fixing. Fixed in `f433dde`.

Verified the two crossed pairs before changing anything — both were accepted: `create:pre_container` + `ON_CONTAINER` and `create:on_container` + `PRE_CONTAINER` each satisfied `creation != (phase is not None)`, because both sides were `True`.

Each creation trigger is now compared against the one phase it delivers, through a table that reads as this module's own documented mapping of trigger to phase. A non-creation trigger is simply absent from it, so the lookup answers `None` and the same identity comparison covers all four ways the two can disagree, rather than treating "no phase" as a separate case.

Worth noting the check is strictly stronger, not different: `creation.py` pairs `create:pre_container` with `PRE_CONTAINER` and `create_job.py` pairs `create:on_container` with `ON_CONTAINER`, so every existing caller passes exactly as before. The only calls newly refused are the ones that were already wrong.

Ten tests: all four accepted pairs and all six refused ones, the crossed pairs included. Mutation-checked — restoring the boolean form fails precisely the two crossed-pair cases and nothing else, which is the behaviour you described.

---

totalfrank (commented):
Continuation review triggered by the `synchronize` event for `f433ddef` (current head). Both open findings from the previous round were already replied to and resolved inline (verifying the fix at that time); this round independently re-verifies them by reading the code at head, and reviews the fix commit itself for new defects.

**Part A — re-verification of the two findings from round 2 (Codex reviewing `e9a0f275`)**

1. `apply_task.py:141` — P1, "Emit payloads that old workers can decode." Read `build_apply_task_payload` and `phase_of_payload` at head: the builder now writes both `PHASE_KEY` (`"phase"`) and `LEGACY_PHASES_KEY` (`"phases"`, sorted, matching the pre-change shape) for every call; `phase_of_payload` still prefers the scalar. Confirmed against the pre-change service that it did `payload["phases"]` unconditionally, so this closes the old-worker/new-row `KeyError` path. New tests (`test_the_payload_also_speaks_the_shape_the_previous_version_reads`, parametrized over `None`/`PRE_CONTAINER`/`ON_CONTAINER`) assert the emitted legacy list and that both spellings decode to the same phase. **Verdict: Fixed.** Thread already resolved.
2. `triggers.py:89` — P2, "Compare creation triggers with their exact phase." Read `require_phase_matches_trigger` at head: the boolean `creation != (phase is not None)` check is replaced by `_PHASE_BY_CREATION_TRIGGER.get(trigger)` compared by identity against `phase`, so `create:pre_container` + `ON_CONTAINER` and `create:on_container` + `PRE_CONTAINER` are now both refused (previously both accepted, since both sides of the old check were `True`). New parametrized tests cover all 4 accepted pairs and all 6 refused pairs, crossed pairs included. **Verdict: Fixed.** Thread already resolved.

**Part B — review of the new commit (`f433ddef`, the only commit since round 2's `e9a0f275`)**

Read the full diff (`apply_task.py`, `triggers.py`, `test_apply_task.py`). No new defects found:
- `phase_of_payload`'s legacy branch still correctly raises on an empty list and translates a two-element list to `None`; `ApplyPhase` has exactly 2 members so the `(only,) = named` singleton-unpack after excluding the full-set case is safe.
- The service call site (`config_manifest_apply_service.py:600`) uses `phase_of_payload(payload)`, not the old `phase_from_payload(payload.get("phase"))` — the actual vulnerable line is fixed, not just a helper added alongside it.
- `_PHASE_BY_CREATION_TRIGGER` aliases `CREATE_PRE_CONTAINER`/`CREATE_ON_CONTAINER` to the same constants re-exported from `creation.py` used elsewhere, so no vocabulary drift.
- No architecture-boundary, concurrency, or security issues; `config_manifest_apply_service.py` line count and R9 status untouched by this commit.

| # | Finding | Status |
|---|---|---|
| 1 | Round 1 review (`10f0fde`) | No findings |
| 2 | Codex P1 — legacy `"phases"` payload widened to whole apply (`config_manifest_apply_service.py:600`) | Fixed in `e9a0f275`, verified, resolved |
| 3 | Codex P1 — new-row payload undecodable by old workers (`apply_task.py:141`) | Fixed in `f433ddef`, re-verified this round, resolved |
| 4 | Codex P2 — trigger/phase pairing accepted crossed pairs (`triggers.py:89`) | Fixed in `f433ddef`, re-verified this round, resolved |

**Open findings: 0.**

CI on `f433ddef`: 8/10 checks complete and green; Backend unit tests and Singlebox coverage still in progress at time of review, no failures observed.

The PR is clean as of this round (round 3, `f433ddef`). Still waiting on the two requested human reviewers (FreddieSun, xianmuyq).

---

totalfrank:
**Review round 3 — two findings fixed, no new ones.** Reviewed `f433dde`.

**Fixed:**

| Finding | Fix |
|---|---|
| P1 — new `phase`-only payload undecodable by an old worker ([#discussion_r3999740309](https://github.com/inclusionAI/Avernet/pull/2167#discussion_r3999740309)) | Builder now emits the legacy key alongside the scalar. The round-2 fix taught the new worker to read old rows; this is the mirror, and the two together make a complete expand/contract migration. |
| P2 — crossed trigger/phase pairs accepted ([#discussion_r3999740311](https://github.com/inclusionAI/Avernet/pull/2167#discussion_r3999740311)) | Each creation trigger compared against the one phase it delivers, via a per-trigger table. Strictly stronger; every existing caller passes unchanged. |

Both verified against the code before fixing — the old worker really does read `payload["phases"]` unconditionally at merge-base `f054e339`, and both crossed pairs really were accepted. Both threads replied to and resolved. Thirteen tests added, each mutation-checked: reverting either fix fails exactly its own cases and nothing else.

**Disputed:** nothing across three rounds.

**New findings this round:** none. The Claude Code review re-read both fixes at head and reviewed the fix commit's full diff, with 0 open findings.

CI is green on `f433dde`: 10/10, Backend unit tests and Singlebox coverage included.

**Still open for a human**, unchanged from round 2 and worth a decision before merge: supporting a mixed-version rollout is what put `"phases"` back in `apply_task.py` — the task this PR implements specified that grep returning nothing. The alternative is a clean break, with the rollout handled by draining the apply queue instead; that reverts `phase_of_payload` and the legacy emit together, and I am happy to make that change. The stricter trigger/phase pairing I would keep either way, since it costs nothing and refuses only calls that were already wrong.

@codex review — please re-check `f433dde`, since a push is not one of your triggers. Both fixes here are yours, and the expand/contract emit half in particular is new code you have not seen.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/bot_config_manifest_apply_service_protocol.py
```diff
@@ -93,7 +89,7 @@ def start_apply(
         actor_id: str,
         audit_actor: Optional[str] = None,
         trigger: str = "explicit",
-        phases: frozenset[ApplyPhase],
+        phase: ApplyPhase | None = None,
```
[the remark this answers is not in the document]
totalfrank: You are right and I was wrong about this one. The PR body claimed these were date-stamped records to be left as written; that reasoning does not survive your point that someone implementing from `tasks.md` would pass the removed keyword and fail at runtime. The staleness is caused by this change, which makes it this change's to address. Done in `ff89bb4`.

What I did not do is rewrite them, because this repo already has a convention for exactly this situation: `specs/2026-07-30-gateway-principal-verifier/spec.md` keeps its original text and marks the changed parts *"Superseded in part"* with the PR that changed them. So each affected `plan.md` and `tasks.md` gains one note at the top — what is gone, what replaced it, and where the current signatures live — with the bodies untouched as the record of what was decided at the time. "Or explicitly supersede", in your words, rather than "update".

Two adjustments to the scope you gave:

**Three directories, not two.** `2026-09-01-create-bot-with-manifest` carries the same stale signatures — `tasks.md:122` has `start_apply(phases={PRE_CONTAINER}, ...)` and `plan.md:20` has `start_apply(..., phases=...)`. Annotating only the two you named would have left the identical trap one directory over, so all three are marked.

**The `spec.md` files are deliberately left alone.** They state WHAT and WHY rather than signatures, and their remaining mentions are of the phase *concept* — "the orchestrator sees phases" — which is still true, since a phase is still what the orchestrator selects on. Only documents carrying implementable signatures got the note.

Docs only; no code or test changes. Architecture suite re-run to be sure nothing keys off those files: 296 passed.

---

totalfrank (commented):
**Round 4 — continuation review triggered by `synchronize` on `ff89bb4`.**

**Commits reviewed this round:** `ff89bb460a7` (the only commit since round 3's `f433ddef`). Docs-only: adds an identical "Superseded in part by PR #2167" note to the top of `plan.md` and `tasks.md` in all three affected spec directories (`2026-08-31-config-manifest-apply-engine`, `2026-09-01-create-bot-with-manifest`, `2026-09-02-manifest-lifecycle-apply-points`) — 90 lines added, no code or test changes.

**Part A — verifying the open finding**

The only open item was Codex's P2 on `bot_config_manifest_apply_service_protocol.py:92` ("Update the specs to match the scalar phase contract"), already replied to and resolved inline. Re-verified by reading the commit directly rather than trusting the message:

- All six files Codex's two named directories plus the third stale one (`2026-09-01-create-bot-with-manifest`, correctly identified as also affected) carry the note.
- The note's content matches the actual current contract: `phase: ApplyPhase | None = None`, `None` = whole apply, `ALL_PHASES` deleted, `require_phase_matches_trigger` enforcement, and the payload's `"phase"`/`"phases"` dual-write — all consistent with what `e9a0f275`/`f433ddef` actually shipped.
- `spec.md` files are untouched, matching the stated reasoning (WHAT/WHY documents whose phase-concept mentions remain true).
- The referenced current-signature locations (`bot_config_manifest_apply_service_protocol.py`, `apply/delivery.py`) are correct.

**Verdict: Fixed.** No discrepancy between the claim and the code.

**Part B — new commit review:** no defects found. Pure documentation, mechanically identical across files, doesn't touch code or tests.

| # | Finding | Status |
|---|---|---|
| 1 | Round 1 review (`10f0fde`) | No findings |
| 2 | Codex P1 — legacy `"phases"` payload widened to whole apply (`config_manifest_apply_service.py:600`) | Fixed in `e9a0f275`, resolved |
| 3 | Codex P1 — new-row payload undecodable by old workers (`apply_task.py:141`) | Fixed in `f433ddef`, resolved |
| 4 | Codex P2 — trigger/phase pairing accepted crossed pairs (`triggers.py:89`) | Fixed in `f433ddef`, resolved |
| 5 | Codex P2 — specs not updated to match scalar phase contract (`bot_config_manifest_apply_service_protocol.py:92`) | Fixed in `ff89bb4`, verified this round, resolved |

**Open findings: 0.**

CI on `ff89bb4`: 8/10 checks complete and green (Backend unit tests, Singlebox coverage still in progress at time of review), no failures observed.

The PR is clean as of this round (round 4, `ff89bb4`). Still waiting on the two requested human reviewers (FreddieSun, xianmuyq).

---

totalfrank:
**Review round 4 — one finding fixed, no new ones.** Reviewed `ff89bb4`.

**Fixed:** Codex's P2 on the spec records ([#discussion_r3999829327](https://github.com/inclusionAI/Avernet/pull/2167#discussion_r3999829327)). An earlier revision of this PR's description argued those documents were date-stamped records to leave alone; that was wrong, since the staleness is caused by this change and someone implementing from `tasks.md` would pass a removed keyword. Each affected `plan.md` and `tasks.md` now carries a *"Superseded in part by PR #2167"* note, following the convention in `2026-07-30-gateway-principal-verifier/spec.md` — bodies untouched as the record of what was decided, the note saying what is gone and where the current signatures live. Three spec directories, not the two named; `spec.md` files deliberately exempt. Thread replied to and resolved.

**Disputed:** nothing, in four rounds.

**New findings this round:** none. The Claude Code review read the commit and confirmed the notes match what actually shipped, rather than only that they exist.

CI is green on `ff89bb4`: 11/11.

**Findings across the loop** — all four from Codex:

| # | Finding | Fixed in |
|---|---|---|
| 1 | P1 — legacy payload widened to a whole apply (old row, new worker) | `e9a0f27` |
| 2 | P1 — new payload undecodable by an old worker; the mirror direction | `f433dde` |
| 3 | P2 — trigger/phase check accepted crossed pairs | `f433dde` |
| 4 | P2 — spec records still documented the removed contract | `ff89bb4` |

The description has also been corrected where the review outran it: **Compatibility** described the rollout risk as unmitigated with an operational workaround, which stopped being true once findings 1 and 2 were fixed, and **Spec** carried the claim retracted above.

**The one thing still worth a human decision**, unchanged since round 2: the task this PR implements specified `grep -rn '"phases"' src/agentclaw/community` returning nothing, and supporting a mixed-version rollout is what keeps that key in `apply_task.py`. Findings 1 and 2 are both consequences of that single choice. The alternative is a clean break with the rollout handled by draining the apply queue, which reverts the decoder and the legacy emit together — a contained change I am happy to make. The stricter trigger/phase pairing from finding 3 I would keep either way.

@codex review — please re-check `ff89bb4`, since a push is not one of your triggers. This one is documentation only, but every real defect on this PR has come from your rounds, so I would rather ask than assume.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
```diff
@@ -52,6 +52,15 @@
...
 APPLY_TASK_DEADLINE_SECONDS = 20 * 60
 
+#: The key every payload this version enqueues carries, ``null`` included.
+PHASE_KEY = "phase"
+
+#: The key a payload carried before :data:`PHASE_KEY` existed: a list of one or
+#: both phase names. Still **written** as well as read, because a rolling
+#: deployment runs both versions at once against one queue — see the builder
+#: and :func:`phase_of_payload`. Both go when no such deployment remains.
+LEGACY_PHASES_KEY = "phases"
```
totalfrank-personal: No worry about this. You can assume there's no legacy data. We are still working on a new feature.

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
```diff
@@ -134,29 +162,90 @@ def build_apply_task_payload(
...
-        "phases": sorted(p.value for p in phases),
+        # What this apply covers, always stated: a phase name for a creation
+        # half, ``null`` for the whole apply. Never reconstructed from a
+        # default at the far end.
+        PHASE_KEY: phase.value if phase is not None else None,
+        # The same fact in the shape the previous version reads, so a worker
+        # that has not been restarted yet can still decode this row. Sorted,
+        # as that version wrote it. Unread here; delete with the decoder's
+        # legacy branch.
+        LEGACY_PHASES_KEY: sorted(
```
totalfrank-personal: Then things like this can go.

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
```diff
@@ -134,29 +162,90 @@ def build_apply_task_payload(
...
-def phases_from_payload(value: Sequence[str]) -> frozenset[ApplyPhase]:
-    """The phases the payload names, back as enum members::
+def phase_from_payload(value: str | None) -> ApplyPhase | None:
+    """The phase the payload names, back as an enum member::
 
-        phases_from_payload(["pre_container"])
-        # -> frozenset({ApplyPhase.PRE_CONTAINER})
-        phases_from_payload(["on_container", "pre_container"])
-        # -> ALL_PHASES
+        phase_from_payload("pre_container")
```
totalfrank-personal: Make sure we don't have magic word like "pre_container" here. I think this is defined as a constant somewhere else. Reference that please.

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
```diff
@@ -134,29 +162,90 @@ def build_apply_task_payload(
...
+    without :data:`PHASE_KEY`: one deploy past the one that introduced it,
+    bounded by :data:`APPLY_TASK_DEADLINE_SECONDS` plus the lock TTL.
+
+    Raises ``ValueError`` on an unknown phase name, and on an empty legacy
+    list — it named an apply covering nothing, which no caller could have
+    meant and the scalar has no way to say.
+    """
+    if PHASE_KEY in payload:
+        return phase_from_payload(payload[PHASE_KEY])
+    legacy = payload.get(LEGACY_PHASES_KEY)
```
totalfrank-personal: We don't need this. You can assume there's no legacy data. We are still working on this as a new feature.

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/triggers.py
```diff
@@ -65,11 +69,58 @@
...
+    Lives here rather than in the apply service because the pairing is a fact
+    about the trigger vocabulary, and this module exists so a trigger is
+    spelled once. ``start_apply`` calls it before the lock is taken and before
+    an ``apply_id`` is minted, so a mismatch leaves neither a record a caller
+    could poll nor a lock the next apply would wait out.
+
+    Raises ``ValueError``: a mismatch is a programming error at a call site,
+    not a condition an end user can reach or act on.
+    """
+    expected = _PHASE_BY_CREATION_TRIGGER.get(trigger)
```
totalfrank-personal, line 122: What about triggers other than "CREATE_PRE_CONTAINER" or "CREATE_ON_CONTAINER"?

If they reach here, what's the behavior? Will error be raised?

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
totalfrank: Done in `c2e6a26` — the whole compatibility layer is gone. `LEGACY_PHASES_KEY`, the legacy emit in the builder, `phase_of_payload` and its translation branch, and the tests that covered them.

Worth flagging plainly since it reverses earlier work on this PR: that layer existed because Codex raised two P1s about a mixed-version rollout — an old row decoding wrong in a new worker, and a new row `KeyError`-ing in an old one. Both were real *given* existing queued rows. You are saying there are none, and that is your call on your own feature, so the code follows you rather than the bot's inference about deployment. The tests those findings earned are deleted rather than left asserting behaviour that no longer exists.

One upside: this also restores the acceptance grep from the task description. `grep -rn '"phases"' src/agentclaw/community` is empty again — the compatibility layer was the only thing keeping that key alive, and I had flagged that divergence as needing a human decision. This is that decision.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
totalfrank: Gone in `c2e6a26`. The builder now writes only the scalar:

```python
PHASE_KEY: phase.value if phase is not None else None,
```

The paragraph in its docstring explaining why both spellings were written went with it.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
totalfrank: Fixed in `c2e6a26` — the example now references the constant:

```python
phase_from_payload(ApplyPhase.PRE_CONTAINER.value)
# -> ApplyPhase.PRE_CONTAINER
```

You are right that it is defined elsewhere: `ApplyPhase` in `apply/order.py` owns both values.

I grepped the rest of `src/agentclaw/community` for the same literals rather than fixing only the line you marked. Outside `order.py`, where the enum is defined, the only one left is in the JSON block illustrating a stored payload:

```
"phase": "pre_container",        # or "on_container", or None
```

I left that one deliberately — it depicts the bytes in `ac_task_queue.payload`, so the literal *is* the subject, and substituting the constant would misrepresent what is stored. Happy to change it too if you would rather have none at all.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/apply_task.py
totalfrank: Removed in `c2e6a26`. `phase_of_payload` is deleted entirely rather than trimmed — with no legacy data there was nothing left for it to do but forward, so the service reads the payload the way the task description originally specified:

```python
phase=phase_from_payload(payload.get(PHASE_KEY)),
```

`PHASE_KEY` rather than a `"phase"` literal, per your other comment about magic words.

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/triggers.py
totalfrank: Yes, an error is raised — but only when such a trigger carries a phase. I ran it rather than reasoning about it:

| trigger | phase | result |
|---|---|---|
| `explicit` | `None` | passes |
| `put` | `None` | passes |
| `totally:unknown` | `None` | passes |
| `explicit` | `PRE_CONTAINER` | **raises** |
| `create:pre_containr` (typo) | `PRE_CONTAINER` | **raises** |

The mechanism: `_PHASE_BY_CREATION_TRIGGER.get(trigger)` answers `None` for anything not in the table, so `expected` is `None` and the comparison then requires `phase` to be `None` as well.

The typo row is the interesting one. A misspelled creation trigger is absent from the table, so it reads as non-creation — but it is still caught in practice, because a creation caller passes the half it is delivering and a trigger the table does not know expects `None`. What actually slips through is only a misspelling that *also* names no phase: that asks for the whole apply and gets it.

I did not add a check that the trigger is in `ALL_TRIGGERS`. That list says of itself "not enforced anywhere — `start_apply` takes a plain `str`", so validating the vocabulary is a separate concern from pairing a trigger with a phase, and #2122 is the issue that owns trigger strings. Say the word if you would rather this function also reject unknown triggers and I will add it.

`c2e6a26` writes all of the above into the function's docstring, so the next reader does not have to ask.

---

totalfrank (commented):
Continuation review triggered by the `synchronize` event for `c2e6a26` (current head). New commits since my last review (round 4, `ff89bb4`): one commit, `c2e6a26`.

**Part A — verifying open findings**

All four Codex findings from earlier rounds were already fixed and their threads resolved before this round (see round-4 summary). Two of those fixes — the legacy-payload decoder and the dual-write emit — are the ones `c2e6a26` deletes outright, per a human reviewer's explicit call on this PR's own thread: *"You can assume there's no legacy data. We are still working on this as a new feature."* That's a factual claim about deployment state that only the feature's author can make, and a human reviewer's call on their own feature outranks the bot's inference about a mixed-version rollout — so removing the compatibility layer isn't a regression of those findings, it's the correct response to new information. Nothing here reopens them.

I have no other findings of my own still open from prior rounds (rounds 1–4 each returned zero findings on their own account).

The one thread still marked unresolved (`triggers.py:122`, the human's question about non-creation/misspelled triggers) isn't one of my findings — it's a live question-and-answer between the author and the human reviewer, already replied to substantively in this same commit. I read the reply against the actual code (`_PHASE_BY_CREATION_TRIGGER.get(trigger)`) and it's accurate, so I'm leaving that thread for the two of them.

**Part B — review of `c2e6a26`**

Diff: `apply_task.py` (-92/+0 net across the file, deletes `LEGACY_PHASES_KEY`, the legacy emit in `build_apply_task_payload`, `phase_of_payload` and its docstring), `triggers.py` (+13, docstring only), `config_manifest_apply_service.py` (switches the call site from `phase_of_payload(payload)` to `phase_from_payload(payload.get(PHASE_KEY))`), `test_apply_task.py` (-119, removes the compatibility tests).

Verified rather than trusted:
- `grep -rn '"phases"\|LEGACY_PHASES_KEY\|phase_of_payload'` across `src/backend/src/agentclaw/community` and `tests/community` returns nothing — the removal is complete, no dangling references.
- `build_apply_task_payload` still writes `PHASE_KEY` unconditionally (including `None` for a whole apply), so `payload.get(PHASE_KEY)` at the new call site is equivalent to `payload[PHASE_KEY]` for every payload this codebase can produce — the `.get` doesn't quietly reintroduce a "missing key defaults to whole apply" widening path, since there's no producer left that would omit the key.
- `triggers.py`'s new docstring paragraph matches `require_phase_matches_trigger`'s actual behavior (`_PHASE_BY_CREATION_TRIGGER.get(trigger)` compared by identity) — checked each of the five cases in the reply against the code, all correct.
- `config_manifest_apply_service.py` is still 999 lines, under the R9 cap, consistent with the commit message.
- Test deletions are 1:1 with the deleted production code (both `phase_of_payload` behaviors and the end-to-end legacy-decode test); no test was deleted that still covers live code.

No new defects found. This is a clean, symmetric removal of dead code following an explicit human decision, plus a docstring clarification — nothing here introduces a correctness, contract, or architecture-boundary issue.

| # | Finding | Status |
|---|---|---|
| 1 | Round 1 review (`10f0fde`) | No findings |
| 2 | Codex P1 — legacy `"phases"` payload widened to whole apply | Fixed in `e9a0f27`, then the whole compat layer removed in `c2e6a26` per human's no-legacy-data call |
| 3 | Codex P1 — new payload undecodable by old worker | Fixed in `f433dde`, then removed in `c2e6a26` per the same call |
| 4 | Codex P2 — trigger/phase pairing accepted crossed pairs | Fixed in `f433dde`, resolved — untouched by this round |
| 5 | Codex P2 — specs not updated to match scalar contract | Fixed in `ff89bb4`, resolved — untouched by this round |
| 6 | Human (totalfrank-personal) — 5 review comments on `ff89bb4` (drop compat layer, remove magic strings, reference `ApplyPhase` constant) | Addressed in `c2e6a26`, replied and resolved |
| 7 | Human (totalfrank-personal) — question on non-creation/misspelled trigger behavior | Answered in `c2e6a26`'s reply; thread still open between author and reviewer, not a bot finding |

**Open findings from this review: 0.**

CI on `c2e6a26`: 9/11 checks complete and green (Backend unit tests, Singlebox coverage still in progress at time of review), no failures observed. `mergeable_state` shows blocked, consistent with the two requested human reviews (FreddieSun, xianmuyq) still being outstanding rather than any CI or merge-conflict issue.

The PR is clean as of this round. No standing disagreements from my side.

---

totalfrank:
**Review round 5 — all five human comments addressed, no new findings.** Reviewed `c2e6a26`.

@totalfrank-personal thanks — your call on the legacy data is the one thing this PR had genuinely been stuck on.

**The decision and what it removed.** *"You can assume there's no legacy data. We are still working on this as a new feature."* So the entire compatibility layer is gone: `LEGACY_PHASES_KEY`, the legacy emit in the builder, `phase_of_payload` and its translation branch, and the tests that covered them. **214 lines out, 19 in.**

This **reverses two findings Codex raised** earlier in this PR — an old row decoding wrong in a new worker, and a new row `KeyError`-ing in an old one. Both were sound *given* existing queued rows. That is a fact about deployment only the feature's author can supply, and your call outranks a bot's inference about it. The tests those findings earned are deleted rather than left asserting behaviour that no longer exists. If this ever ships somewhere that does have old rows, drain the apply queue across the version boundary.

**Also from your review:**

- The docstring example now reads `phase_from_payload(ApplyPhase.PRE_CONTAINER.value)`. I grepped the rest of `src/agentclaw/community` rather than fixing only the marked line; the one literal left is in the JSON block depicting a stored payload, where the literal is the subject. Flagged on that thread in case you want it gone too.
- `require_phase_matches_trigger` now documents the non-creation case you asked about. The short answer is on [that thread](https://github.com/inclusionAI/Avernet/pull/2167#discussion_r4001911111) with a table I produced by running it: such a trigger expects `None` and raises if given a phase, and a *misspelled* creation trigger is still caught in practice, since creation callers pass their half. I left that thread open in case you want an `ALL_TRIGGERS` check as well — say so and I will add it.

**Both acceptance greps from the task description are clean again.** The compatibility layer was the only thing keeping `"phases"` alive in the source, and I had flagged that divergence since round 2 as needing exactly this decision.

```
grep -rn "ALL_PHASES\|phases_from_payload\|phases=\|frozenset\[ApplyPhase\]" src/agentclaw/community --include=*.py   # empty
grep -rn '"phases"' src/agentclaw/community --include=*.py                                                           # empty
```

CI is green on `c2e6a26`: 11/11. Local gate 4548 passed, 2 skipped. `config_manifest_apply_service.py` is at 999 lines, under the R9 cap. Four of your five threads resolved; the trigger question left open for you.

@codex review — please re-check `c2e6a26`, since a push is not one of your triggers. Note this commit deliberately removes the compatibility layer your earlier P1s asked for, on the author's statement that no legacy data exists.

---

FreddieSun (approved):

PR #2176, opened 2026-09-14
feat(mcp): add default CLI profile for claude_code non-normalCC template types

panqiwp:
## Problem

`claude_code` engine bots whose `template_type` is neither `generalCC` nor empty (e.g. `mcptestpq`) hit an error at:

```python
if not self._cli_scope_reconciler.supports_profile(bot=bot):   # -> False
    raise CallerIdentityReadOnlyError
```

Root cause: in `configs/cli-capabilities.yaml`, the only `claude_code` profile hard-codes `template_type: generalCC` in its `match`. Since `CliCapabilityManifestResolver.required_cli_items` only does **exact** equality matching, any other non-empty `template_type` matches no profile → `required_cli_items()` returns `[]` → `is_supported_profile()` is `False` → `supports_profile()` is `False` → `update_cli_call_type` rejects with `CallerIdentityReadOnlyError`.

`template_type` is **not enumerable** and grows over time, so listing every value is not viable.

## Solution

Extend the manifest `match` DSL with an **`exclude_template_types`** blacklist, plus a `claude_code` fallback profile placed *after* the existing `generalCC` exact-match profile. First-match-wins keeps `generalCC` behavior byte-for-byte identical; the fallback covers every other non-empty, non-`normalCC` template type.

```yaml
  - id: claude-code-generalcc-default
    match:
      engine_type: claude_code
      template_type: generalCC          # exact match (unchanged, evaluated first)
    default_cli_codes: [dataphin, deepinsight-cli]
  - id: claude-code-custom-default
    match:
      engine_type: claude_code
      exclude_template_types: [normalCC]  # matches any non-empty template_type except normalCC
    default_cli_codes: [dataphin, deepinsight-cli]
```

`exclude_template_types` semantics: the bot's `template_type` must be **non-empty** and **not in** the list. It is mutually exclusive with `template_type` (validated at parse time).

This matches exactly the system's existing canonical rule `claude_code_uses_aicoding_runtime` in `runtime_identity.py` (`claude_code and non-empty and != normalcc`), which the **creation-time** defaults path (`get_default_cli_items`) already honored. This PR makes the Bootstrap/Passport-scope path consistent with it.

## Matching matrix

| engine_type | template_type | result |
|---|---|---|
| claude_code | generalCC | supported (exact-match profile, unchanged) |
| claude_code | mcptestpq / any non-empty ≠ normalCC | **supported** (new fallback) |
| claude_code | normalCC | unsupported (excluded) |
| claude_code | None / empty | unsupported (requires non-empty) |
| openclaw | * | supported (catch-all, unchanged) |
| aicoding | * | unsupported (unchanged) |

## Backward compatibility

- The new `exclude_template_types` branch in `required_cli_items` only activates when a profile declares it; all existing profiles (`openclaw-default`, `claude-code-generalcc-default`) have no such field, so **all existing matches are unchanged**.
- `_parse_profiles` validation of `exclude_template_types` is additive; existing manifests load identically.
- `generalCC`/`normalCC`/openclaw/aicoding behaviors are unchanged (verified by parametrized tests).
- Side effect for newly-supported bots: on reconcile/bootstrap they now receive `dataphin` + `deepinsight-cli` in their Passport CLI scope, and the `supports_profile` gate no longer blocks `update_cli_call_type` — the intended fix.

## Files

- `configs/cli-capabilities.yaml`: add `claude-code-custom-default` profile
- `core/mcp/services/cli_capabilities.py`: `exclude_template_types` matching in `required_cli_items` + schema validation (type + mutual-exclusion with `template_type`)
- `tests/.../test_cli_capabilities.py`: assert `mcptestpq` supported, `None`/`empty`/`normalCC` unsupported; 3 new schema-validation cases
- `tests/.../test_cli_passport_scope.py`: add `mcptestpq→True`, `None→False` to the `supports_profile` gate test

## Tests

```
1184 passed (caller_identity + mcp/services + devices + engine_runtime)
```

Zero regressions.


totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0
- Target branch: `dev` (integration branch, not a `REL*` release branch) — no release-branch gating applies.
- Classification: ordinary feature PR landing on `dev`. **Pass** (no release-branch restriction to check).

### Findings

**1. `configs/cli-capabilities.yaml:16-19`, `core/mcp/services/cli_capabilities.py:66-70` — silently contradicts the written manifest contract without updating it (arch.rules.md Rule 1 — Invariant; Rule 16 — Invariant).**

```yaml
  - id: claude-code-custom-default
    match:
      engine_type: claude_code
      exclude_template_types: [normalCC]
    default_cli_codes: [dataphin, deepinsight-cli]
```

This manifest is the authoritative contract for Default-CLI grants, and it already has a written spec: `src/backend/specs/cli-default-install-identity/001-spec-output.md` and its `002-code-report.md`. Those documents state, in the spec's own words:

> 匹配必须为精确逻辑引擎与模板匹配…（"matching must be exact logical engine and template matching"）
> `profiles[].match.template_type` … exact match，不能继承其他 template（"exact match, cannot inherit other templates"）
> 本变更不得把 YAML 默认项扩展到 aicoding、moltis、hermes 或其他 Claude template（"this change must not extend YAML defaults to other Claude templates"）
> 一期 gate…仅允许 openclaw 和 claude_code/generalCC（"phase-one gate allows only openclaw and claude_code/generalCC"）

`required_cli_items` previously did exact equality matching only, honoring that contract. This PR adds a deny-list (`exclude_template_types`) that turns the `claude_code` profile from an allow-list ("exactly `generalCC`") into a catch-all ("anything except `normalCC`") — the precise thing the spec calls out as forbidden — and this also widens the security-relevant `supports_profile()` gate that governs whether `update_cli_call_type` grants CLI caller/owner IAM mutation (see `caller_identity/test_service.py::test_cli_update_rejects_profiles_outside_phase_one`, `cli_passport_scope.py::supports_profile`). Per Rule 1's required checks, "contract changes update specs/docs in the same change," and per Rule 16, a change to config schema/contract semantics must declare affected consumers, compatibility status, and a migration/deprecation note. None of that accompanies this diff — the 4 changed files are only the YAML, the resolver, and two test files; neither spec doc was touched.

If the phase-one restriction documented in the spec is intentionally being widened, that's a legitimate change, but it needs an update to `001-spec-output.md`/`002-code-report.md` (or an explicit written waiver per the Governance Addendum) in the same PR — not a silent reinterpretation of "exact match" into "match unless excluded."

**2. `core/mcp/services/cli_capabilities.py:66-70` — reimplements an existing canonical predicate instead of depending on the single source of truth (arch.rules.md Rule 1 / catalog item 9 — copy-paste mirroring).**

The exact judgment this PR encodes — "`claude_code` engine with a non-empty `template_type` that isn't `normalCC`" — is already centralized as the canonical domain rule `claude_code_uses_aicoding_runtime()` in `core/workspace/runtime_identity.py:68-78`, whose docstring and design doc (`docs/superpowers/specs/2026-08-31-engine-vocabulary-template-form-design.md`) exist specifically so this predicate has one owner. The PR's own description confirms this is the same rule ("This matches exactly the system's existing canonical rule `claude_code_uses_aicoding_runtime`… which the creation-time defaults path already honored"), yet instead of calling that function (or the `default_capabilities` engine-bucket resolver that already wraps it for MCP/CLI default-capability decisions), it re-derives the same boolean via a brand-new generic YAML DSL field plus bespoke Python matching/validation, with the magic string `normalCC`/`normalcc` now hardcoded independently in two unrelated places.

This creates exactly the hidden-coupling risk Rule 15 warns about: nothing ties the YAML `exclude_template_types` list to `runtime_identity.py`'s exclusion logic, so a future change to the canonical rule (e.g., a new excluded template type, or the `engine_form` marker override already handled by `uses_aicoding_runtime`) has no mechanism to propagate here, and no test asserts the two stay in sync.

**Required fix direction:** `CliCapabilityManifestResolver`/`CliPassportScopeReconciler` should consult the existing canonical predicate (`claude_code_uses_aicoding_runtime` or the `default_capabilities` bucket resolver) to decide profile membership for non-`generalCC` `claude_code` templates, rather than introducing a second, YAML-config-driven copy of the same business rule. If a generic "exclude" matcher is still wanted for other future cases, it should not be the mechanism used to re-encode a rule that already has one owner.

### Out-of-scope notes
- The new profile's correctness depends on list order (the `generalCC` exact-match profile must be evaluated before the new catch-all one); the PR description explains this but the resolver code has no comment/test asserting order-sensitivity is intentional contract behavior, not an implementation accident. Not gating on this, but worth a one-line note in the resolver.


---

src/backend/src/agentclaw/community/configs/cli-capabilities.yaml
```diff
@@ -13,6 +13,11 @@ profiles:
       engine_type: claude_code
       template_type: generalCC
     default_cli_codes: [dataphin, deepinsight-cli]
+  - id: claude-code-custom-default
+    match:
+      engine_type: claude_code
+      exclude_template_types: [normalCC]
+    default_cli_codes: [dataphin, deepinsight-cli]
```
totalfrank, line 20: **arch.rules.md Rule 1 (Invariant) + Rule 16 (Invariant) — contract change without spec/doc update or propagation analysis**

This `claude-code-custom-default` profile intentionally overturns a documented, reviewed, PASS-certified invariant of this exact manifest without touching the documents that certified it:

- `src/backend/specs/cli-default-install-identity/001-spec-output.md:95` — "匹配必须为精确逻辑引擎与模板匹配；不可因 `claude_code` 的内部 runtime 映射而把 `normalCC` 或其他 `template_type` 误配到 `generalCC`。"
- `001-spec-output.md:214-215` — `profiles[].match.template_type` for `claude_code` "必须为 `generalCC`" / "exact match，不能继承其他 template".
- `001-spec-output.md:132` / `:298` — "不在一期 profile 的引擎维持既有行为... 不能沿用 Claude Code 到 aicoding 的默认能力分桶。"
- `003-review-report.md:46` — **R-01, marked PASS**: "YAML profile 精确匹配 ... 仅 `openclaw`、`claude_code/generalCC` 命中；`normalCC`、`aicoding` fail closed"。

This diff replaces that certified exact-match gate with a category/exclude match: any non-empty `template_type` other than `normalCC` now passes for `claude_code` (verified in the new `test_supported_profiles_resolve_exactly_to_two_default_clis` cases for `mcptestpq`). That is precisely the "internal runtime mapping instead of exact template match" the spec explicitly forbids, and it reverses a requirement the review report certified PASS.

Rule 1 requires "contract changes update specs/docs in the same change"; Rule 16 requires propagation analysis (affected consumers, compatibility status) for config-schema/contract changes. Neither happened — the PR description contains an informal compatibility note, but the actual spec/review documents that are this component's authoritative contract (per this repo's own SDD convention) are untouched and now contradict shipped code. This is an unwaived Invariant violation.

**Required fix:** update `001-spec-output.md` (and/or add a new spec revision, e.g. `010-...md`) to document the `exclude_template_types` semantics and explicitly supersede R-01's "exact match only" requirement, including propagation analysis for `CliPassportScopeReconciler.supports_profile`/`reconcile` (the only consumers). Absent that, this needs a written Governance-Addendum waiver (rule, reason, risk, compensating controls, owner, expiry) — an informal PR-description note does not satisfy either path.

---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

## Gate 0
- **Target branch:** `dev` (integration branch, not a `REL*` release branch) — the release-branch gate and Isolation Audit do not apply here.
- **Classification:** New functionality (a fallback matching profile) framed as a bug fix. This is fine on `dev` regardless of framing — Gate 0 only restricts what may target a `REL*` branch — so Gate 0 **passes**.

## Findings

1. **`src/backend/src/agentclaw/community/configs/cli-capabilities.yaml:16-20` and `src/backend/src/agentclaw/community/core/mcp/services/cli_capabilities.py:66-70,168-180` — arch.rules.md Rule 1 (Invariant) + Rule 16 (Invariant).**
   See inline comment for the full citation trail. Summary: this PR replaces a documented, PASS-reviewed exact-match invariant ("`claude_code` matches only `generalCC`; every other template fails closed" — `specs/cli-default-install-identity/001-spec-output.md:95,132,214-215,298`, certified as `003-review-report.md` R-01) with a category/exclude match that now accepts any non-empty, non-`normalCC` template. The spec and review-report documents that are this manifest's authoritative contract are not updated, and no propagation analysis or written Invariant waiver accompanies the change. Rule 1 requires contract changes to update specs/docs in the same change; Rule 16 requires propagation analysis for config-schema/contract changes. This is a real behavior change riding on a bug-fix framing, and it needs the doc update (or a proper waiver), not just the informal note in the PR description.

No other findings — the implementation itself is architecturally sound: `CliCapabilityManifestResolver` stays a generic, config-driven string matcher with no engine-name branching in Python (`engine_type`/`template_type`/`exclude_template_types` values only ever come from YAML), which is exactly the Rule 14 pattern this codebase wants. No transport/env/plugin/composition-root issues in the diff.

## Out-of-scope notes
- The predicate this new profile encodes ("`claude_code` + non-empty + `!= normalCC`") duplicates the canonical, already-tested `claude_code_uses_aicoding_runtime()` in `core/workspace/runtime_identity.py` (which the PR description itself cites as the source of truth) instead of reusing it. Today the two encodings agree, but they can drift — `uses_aicoding_runtime()` already has a richer `engine_form`-marker tier that this manifest gate has no way to express. Worth a follow-up, not a blocker, and reusing the function directly would trade away this matcher's current engine-agnostic genericity, so the right fix isn't obvious — flagging for the feature reviewer/author to weigh rather than gating on it here.
- Whether loosening this gate for arbitrary `template_type`s is the right product behavior (vs. an explicit allowlist) is a business-logic call outside this review's scope.

---

totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0
Target branch is `dev` (integration branch), not a `REL*` release branch — the release-branch gate does not apply. **Gate 0: PASS** (normal feature route).

### Findings

**1. `src/backend/src/agentclaw/community/core/mcp/services/cli_capabilities.py:66-70` — duplicated business predicate, reimplemented without normalization, flips a fail-closed credential grant to fail-open**

```python
exclude_template_types = match.get("exclude_template_types")
if exclude_template_types is not None and (
    not template_type or template_type in exclude_template_types
):
    continue
```

The condition here — "`claude_code` engine, non-empty `template_type`, not `normalCC`" — already exists as a single canonical predicate in `core/workspace/runtime_identity.py:68-81` (`claude_code_uses_aicoding_runtime`), which normalizes via `.strip().lower()` before comparing to `"normalcc"` (see also `default_capabilities.normalize_template_type`, same pattern). The PR description itself invokes this canonical rule as justification ("matches exactly the system's existing canonical rule `claude_code_uses_aicoding_runtime`").

But the new YAML-driven check is a **second, independent implementation** of the same rule: a raw, case-sensitive `template_type in exclude_template_types` string comparison against a manifest literal, with no normalization. `template_type` reaches this check unnormalized (`cli_passport_scope.py:77,93` pass `bot.get("template_type")` straight through), and the field is unconstrained free text (`adapters/http/openapi_v1/bots/schemas.py:113`, plain `str | None`, no enum/regex) that the PR's own description calls "not enumerable, grows over time."

Concrete failure scenario: a template registered as `NormalCC`, `normalcc`, or `"normalCC "` (trailing space) is classified by the canonical predicate as *not* running the aicoding runtime, but this new check won't find it in `exclude_template_types` (exact string miss) and falls through to the new fallback profile — silently granting `dataphin` + `deepinsight-cli` with **owner** `identity_mode` via AgentPass Passport to a bot the system's own canonical rule says shouldn't get it. Every profile in this manifest was previously exact-match/fail-closed (a mismatch just returns `[]`); this is the first path that fails *open* onto a credential grant.

- **Violated rule:** arch.rules.md Rule 23 (repeated architectural solutions must be reused via one pattern, not reinvented) and Rule 1 (implementations must not introduce undocumented semantics — here, an undocumented assumption that `template_type` arrives pre-normalized).
- **Fix direction:** call `claude_code_uses_aicoding_runtime` (or route both the incoming `template_type` and the config literals through `normalize_template_type`) instead of re-encoding the rule as a raw config string list.

**2. `src/backend/src/agentclaw/community/configs/cli-capabilities.yaml:16-20` — config-schema/contract change with no spec update or propagation analysis**

This manifest is a formal, versioned contract per its own design doc, `src/backend/specs/cli-default-install-identity/001-spec-output.md`, which explicitly mandates **exact-match-only, fail-closed** semantics for `claude_code`:

> "匹配必须为精确逻辑引擎与模板匹配…未命中 profile 时 `required_cli_items=[]`，不得新增 YAML CLI" / "本变更不得把 YAML 默认项扩展到 aicoding、moltis、hermes 或其他 Claude template" (spec lines 95, 132); the schema table (line 215) documents `template_type` as required to be an exact match for `claude_code`.

This PR adds `exclude_template_types`, inverting that documented invariant to default-allow-by-exclusion for `claude_code`, with no update to the spec, no propagation analysis, and no note addressing the same spec's requirement that this manifest stay byte-identical with a downstream DaaS consumer (sha256-compared).

- **Violated rule:** arch.rules.md Rule 1 ("contract changes update specs/docs in the same change" — Invariant) and Rule 16 ("Changes Propagate" — Invariant); `AGENTS.md`: "Contract changes require matching docs and conformance or compatibility tests."
- **Fix direction:** update (or supersede with a new dated spec) `specs/cli-default-install-identity/001-spec-output.md` to document the revised matching semantics and retract the old exact-match-only statement, plus a propagation note on the DaaS manifest-sync requirement.

Both are unwaived Invariant violations, hence REQUEST_CHANGES regardless of diff size.

### Out-of-scope notes
- The choice of which CLIs (`dataphin`/`deepinsight-cli`) to grant to non-normalCC `claude_code` bots is a product decision I'm not gating on — only the duplicated-rule mechanism and the undocumented contract change are architectural findings.
- New schema-validation test cases (`bad_exclude_not_list`, `exclude_with_template_type`, etc.) are solid and not a concern.


---

panqiwp:
Thanks for the review — addressed in commit `707a8ad67` (pushed). Same diff, so the spec addition is coverage-neutral.

## Finding 1 — contract change without spec/doc update (Rule 1 + Rule 16)

Fixed. Added spec revision `specs/cli-default-install-identity/010-exclude-template-types.md` amending the certified `001-spec-output.md` contract and superseding R-01's "exact-match-only" invariant for the `claude_code` fallback profile, with the Rule 16 propagation analysis:

- **Affected consumers** — only `CliPassportScopeReconciler.supports_profile()` (the phase-one gate) and `reconcile()`/`_build_snapshot()` (Bootstrap scope provisioning). No other consumer reads the manifest.
- **Compatibility** — the existing profiles (`openclaw-default`, `claude-code-generalcc-default`) have no `exclude_template_types` field, so the new branch never fires; all previously matched inputs (openclaw/*, `claude_code/generalCC`) are byte-identical. The widening is bounded to non-empty, non-`normalCC` `claude_code` templates; `normalCC`, empty, `aicoding`/`moltis`/`hermes` remain fail-closed (see `test_supported_profiles_resolve_exactly_to_two_default_clis`).
- **Security** — `supports_profile` governs `update_cli_call_type` (CLI caller/owner IAM mutation). The widening is bounded and consistent with the creation-time default-capability routing that already treats these templates as aicoding-runtime bots, i.e. this repairs an existing inconsistency rather than adding an arbitrary grant.
- **Migration/deprecation** — `exclude_template_types` is an additive optional field, mutually exclusive with `template_type` (validated at parse time); no field removed, no data migration.

`001-spec-output.md` now carries forward-pointers (the bearer-purpose bullet near §132, a new model-table row for `exclude_template_types`, and the Review-spec note at §298) so the certified doc no longer contradicts shipped code. R-01 continues to hold for `openclaw` and `claude_code/generalCC`; the `claude_code` fallback is the single documented, test-bounded exception.

Also added a contract comment in `required_cli_items` asserting the profile-order invariant (exact `template_type` profile MUST precede any `exclude_template_types` fallback for the same engine) — addressing the out-of-scope order-sensitivity note.

## Finding 2 — reuse the canonical predicate

Considered and intentionally not adopted; rationale recorded in `010-*.md`. `001-spec-output.md` §132 and §298 explicitly forbid reusing the Claude Code → aicoding default-capability runtime bucket for this manifest's matching:

> 不在一期 profile 的引擎维持既有行为；本变更不得把 YAML 默认项扩展到 `aicoding`、`moltis`、`hermes` 或其他 Claude template。
> YAML profile 必须按逻辑 engine/template 精确匹配，**不能沿用** Claude Code 到 aicoding 的默认能力分桶.

`exclude_template_types` keeps the matcher config-driven and engine-agnostic (no Python call to `claude_code_uses_aicoding_runtime()` / the `default_capabilities` bucket resolver), preserving that separation invariant. `normalCC` is the one operator-convention value shared between the two layers, declared explicitly inside the manifest; the two layers stay decoupled so each can evolve (e.g. the `engine_form`-marker tier already handled by `uses_aicoding_runtime`) via a future revision rather than a hidden coupling. This matches Reviewer 2's non-blocker classification. If a future change needs the runtime-identity `engine_form` marker to influence this CLI manifest gate, that should be evaluated as a new spec revision rather than silently wiring a runtime call into the matcher.

## CI note (Singlebox Coverage)

The "Singlebox Coverage" check failed on the `files` module (`resource_file_service.py`, threshold 63.52 vs 55.61). This module is unrelated to the diff — none of the changed files appear in any singlebox coverage module's `core_paths` (`grep` over `scripts/ci/singlebox_coverage_modules.yaml` confirms), so this PR has zero direct effect on that number. Sibling full runs (e.g. `feat/add-codeuser-mcp`, `feat/dima-relation-list-delete`) passed; the `acceptance/files` tests carry prior singlebox-lifecycle robustness fixes (`897040edc resolve singlebox workspace root robustly`, `7912b35db gate live singlebox workspace lifecycle`) consistent with a flaky collection. A re-run was auto-triggered by the spec-doc push and is in progress; no code change is warranted here.

totalfrank (approved):
LGTM.

PR #2178, opened 2026-09-14
fix(bcs): map queued provider requests to canonical run IDs

RaymondHX:
## Problem
Queue dispatch assigns a fresh attempt request ID to each chat.send frame, while the HTTP Provider adapter rejected frames whose ID differed from the canonical run ID. Queued Provider sends therefore failed before network I/O and the worker conservatively moved them to Unknown.

## Solution
- Map HTTP Provider chat.send wire IDs to BotDeliveryCommand.run_id in the transport adapter instead of requiring the internal frame ID to match.
- Preserve queue attempt IDs and existing stale-attempt checks, WebSocket frame IDs, and non-running Provider request IDs (inject/history/abort).
- Reject missing canonical run IDs before network I/O.
- Document the outbound port and queue identity contract. Extend real HTTP transport tests for different attempt/run IDs, SSE event ingestion, callback transport registration, task delivery kinds, and unchanged inject IDs.

## Validation
After replaying only this change onto latest origin/dev:
- `cargo test --offline -p bcs-provider-http --quiet`: 57 passed (30 unit + 27 transport contract).
- `cargo test --offline -p bcs-message-flow --test delivery_runtime --test managed_delivery --test conformance_queued_group --test queued_provider_headers --quiet`: 29 passed.
- Commit/push used `--no-verify`; explicit tests ran separately.
- No live Provider/OceanBase deployment, full Singlebox E2E, or full architecture gate was run for this patch. Queue and actual HTTP transport suites were tested separately, not as a new full-stack queue-to-callback E2E.

## Compatibility and risk
No DB migration or configuration changes. Provider wire protocol continues to use the canonical run ID, so Providers need no protocol upgrade. Internal command frame IDs may now differ from run IDs for HTTP Provider chat.send. Mapping is confined to the adapter; no domain scheduling policy change or architecture waiver.
Stable run IDs do not imply Provider idempotency: only proven-unsent requests may retry. Existing Unknown records are not automatically replayed. General send-error classification and additional diagnostic logging are outside this patch.

## Spec
- `src/bcs/crates/service-api/bcs-service-api/src/port/delivery.rs`


questchaos (approved):
LGTM

PR #2179, opened 2026-09-14
fix(backend): encode teclaw session ids for the engine path

totalfrank:
## Problem

Every session read on a **teclaw** bot fails. The public sessions group forwards the session id into the engine path verbatim, and a teclaw runtime is reached through the agentclaw proxy, which refuses a path segment carrying the id's colons with `400 Bad Request` — before the request reaches the engine at all:

```
HTTP Request: GET https://<proxy-host>/proxypass/TECLAW_b_01M2F8ZZFY…@4:20003
  /api/sessions/agent:default:default:c_01M2F998Z1BQ9VDKVQ52N9RXQP:user:272471/messages?offset=0&limit=51
  "HTTP/1.1 400 Bad Request"
```

(`offset=0&limit=51` is `_history_window` for page 1 at `page_size=50`, i.e. `GET /openapi/v1/bots/{bot_id}/sessions/{session_id}/messages`.)

Every engine but teclaw takes the id that way — a colon is legal in a path segment (RFC 3986) — so this is one engine's rule, not a property of the surface.

## Solution

A session-key codec **strategy registry** keyed by engine type, in `core/engine_runtime/session_key.py`:

- `SessionKeyCodec` is an abstract contract (`encode` is `@abstractmethod`); both codecs inherit it directly, so a codec that never implements `encode` fails at construction rather than at the first forward;
- `PassThroughSessionKeyCodec` is the registry's default — the session id exactly as today, for every engine that has registered nothing, so the bytes on the wire for existing bots are unchanged;
- `teclaw` registers `Base64SessionKeyCodec` — URL-safe, unpadded base64, byte-for-byte what the engine's own `encode_session_key` produces and `decode_session_key` (`src/engine/src/engine/community/shared/utils.py`) reverses on every session route.

**Composition is the composition root's.** `EngineRuntimeModule` builds the registry (`SessionKeyCodecRegistry(PassThroughSessionKeyCodec())` plus the teclaw registration) and the sessions handlers receive it by injection — no process-wide singleton, no module-level accessor. An engine that later needs its own rule registers a codec there; no handler changes. The registry takes its default codec as a required argument, and neither `register` nor `resolve` substitutes a default engine type for a missing one: an empty or blank engine raises rather than picking a wire format for a bot we cannot identify.

URL-safe rather than standard base64 is load-bearing: the alphabet (`A-Za-z0-9-_`) is already safe in a path segment, where standard base64's `/` would split the segment in two. Unpadded, because that is the form the engine's decoder pads back itself. Nothing needs percent-encoding as a result.

Applied to **every** route in the group that forwards the id — get, patch, delete, messages, clear messages, favorite. One route fixed and the rest left raw would only move the 400 around. Query-param uses of the session key (the list filter) are untouched: the failure is path-only.

Rejected alternatives:
- *Encode in the relay* — `relay.call` takes an already-built generic `path` and cannot know which segment is the session id; templating one in would break the generic-forward shape the group is built on.
- *Percent-encode for everyone* — would change the wire bytes for every bot on the surface to fix one engine's problem.

The injected parameters pushed `sessions/router.py` past the 1000-line single-responsibility cap (`test_no_oversized_modules`). The four `_session_file_*` shaping helpers therefore moved into `schemas_helpers.py` — "pure mapping and pagination helpers for the public sessions adapter", which is exactly what they are. Pure move, no logic change; the router is 964 lines.

## Validation

Run from `src/backend` with the locked dependency versions:

- `pytest tests/community/core/engine_runtime tests/community/adapters/http/openapi_v1 tests/community/architecture/test_no_oversized_modules.py` — **1857 passed, 2 skipped**.
- `pytest tests/community/di tests/community/architecture` — **670 passed** (covers `test_all_bindings_resolve`, which resolves the new binding, and the context-boundary gate reading the updated `core/engine_runtime/README.md`).
- `pytest tests/community/endpoints` — passed.
- `ruff check` on every changed/added file — clean. (Two pre-existing `F401`s in `test_operator_access.py`, untouched by this change, are left as they were.)

New tests pin both halves of the contract and the two refusals: an engine without a codec still sends the id verbatim; the teclaw form contains no colon, is a single safe path segment, and decodes back to the original id through the engine's own decoder (mirrored in the test, since the engine package is not importable from the backend suite); a blank engine type raises on both `register` and `resolve`; a codec without `encode` cannot be constructed. The endpoint tests bind the production `EngineRuntimeModule` provider rather than a stand-in, so a registration dropped there fails them too.

Not run: the full backend suite and the singlebox coverage gate — they need network access to the configured package index, which this environment cannot reach (deps were installed from the exported lock via pypi.org).

## Compatibility and risk

No contract, schema, config or migration change. The public `session_id` a caller passes and receives is unchanged — the encoding is a wire detail between backend and engine, and responses still carry the raw id. Behaviour for every non-teclaw engine is byte-identical to before (the default codec is identity), so the blast radius is teclaw bots, which are currently broken on these routes. The engine accepts both the raw and the encoded form on every session route, so rollback is reverting the commits.

One behaviour change beyond teclaw: a bot row whose `active_engine` is blank now raises on these routes instead of being served. `ac_bots.active_engine` is `NOT NULL` with a default, so that is corrupt data rather than a reachable state — flagged on the review thread that asked for it.

https://claude.ai/code/session_01Lg3DJrNBd52JiccaQQ9dHH


src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
```diff
@@ -0,0 +1,142 @@
...
+#: registration key for that engine's codec, not a second definition of what
+#: teclaw *is*.
+TECLAW_ENGINE_TYPE = "teclaw"
+
+
+@runtime_checkable
+class SessionKeyCodec(Protocol):
+    """Turn a public session id into the form one engine's routes accept."""
+
+    def encode(self, session_key: str) -> str:
```
totalfrank-personal, line 69: I don't like duck type. Annotate this method with @abstractmethod and have the other two implementation class extends this protocol directly.

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+
+    def encode(self, session_key: str) -> str:
+        return base64.urlsafe_b64encode(session_key.encode()).decode().rstrip("=")
+
+
+class SessionKeyCodecRegistry:
+    """Which codec an engine's session ids go through, by engine type."""
+
+    def __init__(self, default: SessionKeyCodec | None = None) -> None:
```
totalfrank-personal: No, I don't like "| None". Let's make sure everything is wired with DI and since DI always provides the implementation, we don't need the | None there.

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
```diff
@@ -0,0 +1,142 @@
...
+        self._default: SessionKeyCodec = default or PassThroughSessionKeyCodec()
+        self._codecs: dict[str, SessionKeyCodec] = {}
+
+    def register(self, engine_type: str, codec: SessionKeyCodec) -> None:
+        """Bind ``codec`` to ``engine_type``, refusing a silent overwrite."""
+        key = normalize_engine_type(engine_type, default="")
```
totalfrank-personal: No, we don't allow default. If it's empty or None, error out.

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Pure and side-effect free — it only instantiates stateless codecs — so it
+    is built eagerly at import, like ``EngineProvisioningRegistry``.
+    """
+    registry = SessionKeyCodecRegistry()
+    registry.register(TECLAW_ENGINE_TYPE, Base64SessionKeyCodec())
+    return registry
+
+
+_REGISTRY: SessionKeyCodecRegistry = _build_default_registry()
```
totalfrank-personal: Let's do DI way instead of static field way.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
totalfrank: Done in 11dd4c1. `encode` is `@abstractmethod`, and `PassThroughSessionKeyCodec` / `Base64SessionKeyCodec` both inherit `SessionKeyCodec` directly — same shape as `TeclawProvisionService(TeclawEngineTestProtocol)`. A codec that never implements `encode` now fails at construction instead of at the first forward; there's a test pinning that.

---

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
totalfrank: Done in 11dd4c1. `__init__(self, default: SessionKeyCodec)` — required, no `| None`, no self-built fallback. `EngineRuntimeModule.session_key_codec_registry` constructs it with the pass-through default and hands it to the handlers by injection.

---

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
totalfrank: Done in 11dd4c1. The `default=""` on `normalize_engine_type` is gone from both sides: `register` and `resolve` share a `_key` helper that raises `ValueError("engine type is required…")` on an empty or blank engine type.

One consequence worth naming, since it is a runtime path and not just a registration one: `resolve` is called per request with `facts.active_engine`. The column is `NOT NULL` with a default so it is always populated in practice, but a legacy row carrying `""` would now 500 on every session route for that bot instead of being served as before. I think that is the right trade — it is corrupt data, and picking a wire format for it would forward on a value nobody wrote — but say the word if you'd rather `resolve` keep serving those bots and only `register` be strict.

---

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
totalfrank: Done in 11dd4c1. The module-level `_REGISTRY`, its accessor and the free `encode_session_key()` are gone. `EngineRuntimeModule` assembles the registry (`@singleton @provider`) — which engine gets which codec is the composition root's decision — and the six sessions handlers take it as `codecs: SessionKeyCodecRegistry = Injected(SessionKeyCodecRegistry)`. The endpoint tests bind that same provider rather than a stand-in, so a registration dropped in the DI module fails them too.

Those injected parameters pushed the sessions router past the 1000-line cap (`test_no_oversized_modules`), so the four `_session_file_*` shaping helpers moved into `schemas_helpers.py`, whose stated purpose they already matched. Pure move, no logic change — router is 964 lines now.

---

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
```diff
@@ -0,0 +1,148 @@
...
+#: The engine whose bots run in a teclaw container, in the registry key form.
+#: The canonical creation-time definition is
+#: ``TeclawProvisionService.is_teclaw`` / ``DEFAULT_TECLAW_ENGINE_TYPES``
+#: (``core/bot_management/services/teclaw_provision_service.py``); this is the
+#: registration key for that engine's codec, not a second definition of what
+#: teclaw *is*.
+TECLAW_ENGINE_TYPE = "teclaw"
+
+
+@runtime_checkable
```
totalfrank: nit: `@runtime_checkable` re-enables `isinstance()`/structural matching against this Protocol, which is exactly the duck-typing surface the earlier review round asked to close (the `@abstractmethod` + direct-inheritance change addressed the construction-time check, but this decorator is unrelated to that and isn't used anywhere — no `isinstance(..., SessionKeyCodec)` call exists in this module or its tests). Consider dropping it so the contract only admits nominal subclasses, matching the docstring's stated intent.

---

totalfrank (commented):
Reviewed commits `b6744ee` and `11dd4c1` (head `11dd4c1`) in full — this is the first automated pass on this PR.

The four prior human-review threads (duck typing, `| None` default, silent-default fallback, module-level singleton) are all resolved in `11dd4c1` as claimed; I checked each against the current code and they hold.

Checked: correctness of the codec registry and its wiring into all six session routes (get/patch/delete/messages/clear/favorite), the `facts is None` (friend/expert-chat) short-circuit ahead of every new `codecs.resolve(facts.active_engine)` call (no risk of resolving on a `None` facts object), the pure move of the `_session_file_*` helpers into `schemas_helpers.py`, DI composition-root placement, and the encoded-literal test fixture (`TECLAW_ENCODED_SESSION_ID`) against the real base64 encoding of `SESSION_ID`.

Ran the tests locally (via `pypi.org` since the configured mirror isn't reachable here):
- `tests/community/core/engine_runtime tests/community/adapters/http/openapi_v1 tests/community/architecture/test_no_oversized_modules.py` → 1857 passed, 2 skipped
- `tests/community/di tests/community/architecture` → 670 passed
- `ruff check` on every changed/added file → clean except the two pre-existing `F401`s in `test_operator_access.py` the PR description already calls out as untouched

One nit posted inline (a stray `@runtime_checkable` that reopens duck-typing on the Protocol despite the design intent). No blocking or non-blocking findings.

---

FreddieSun (dismissed):

src/backend/src/agentclaw/community/core/engine_runtime/session_key.py
totalfrank: Fixed in 723998e — and the finding is sharper than a nit. Verified before changing anything: no `isinstance`/`issubclass` against `SessionKeyCodec` exists in production code, so the only thing keeping `@runtime_checkable` alive was the test that asserted the codecs inherit the Protocol — `issubclass` against a Protocol raises `TypeError` unless it is `@runtime_checkable`. The decorator was being held up by the test observing it, which is the wrong way round.

Dropped it, and the assertion now reads `SessionKeyCodec in <codec>.__mro__` — the nominal question it meant to ask — plus a new case pinning that a look-alike class with a matching `encode` is *not* a codec. The docstring now says why the decorator is absent, so it does not get added back as an obvious omission.

---

totalfrank:
**Review round 1 — 1 finding, fixed.** Reviewed `11dd4c1`; head is now `723998e`.

**Fixed:** the `@runtime_checkable` nit on `SessionKeyCodec`. Verified rather than taken on trust — no `isinstance`/`issubclass` against the Protocol exists in production code, so the decorator was kept alive solely by the test asserting inheritance (`issubclass` against a Protocol raises `TypeError` without it). Decorator dropped; the assertion now reads the MRO, with a new case pinning that a look-alike class is not a codec. Details on the thread.

**Nothing disputed**, nothing carried over. The four earlier human-review threads (duck typing, `| None` default, silent-default fallback, module-level singleton) were fixed in `11dd4c1` and are resolved; the one I had left open — whether `resolve` should stay strict on a blank `active_engine`, since that turns a corrupt-data row into a 500 on its session routes — has since been resolved by the author, so the strict behaviour stands as implemented.

**CI was green on `11dd4c1`** — all 11 checks including Backend unit tests and Singlebox coverage. Re-running now on `723998e`. Local gates for this push: `tests/community/core/engine_runtime`, `tests/community/adapters/http/openapi_v1/engine_runtime`, `tests/community/di`, `tests/community/architecture` → 1237 passed, 1 skipped; `ruff check` clean on the changed files.

---

totalfrank (commented):
**Review round 2 — no new commits, nothing open.**

This synchronize event's head (`723998e`) is the same commit that round 1 already reviewed and cleared. There are zero commits between the commit round 1 was posted against and the current head, so there is nothing new to verify and no new code to review this round.

Findings checklist (all rounds):
- duck-typing via structural Protocol matching — fixed in `11dd4c1`, thread resolved
- `SessionKeyCodecRegistry.__init__` accepting `default: SessionKeyCodec | None` — fixed in `11dd4c1`, thread resolved
- silent default-engine-type fallback in `register`/`resolve` on blank engine type — fixed in `11dd4c1`, thread resolved
- module-level singleton/static accessor instead of DI — fixed in `11dd4c1`, thread resolved
- nit: stray `@runtime_checkable` on `SessionKeyCodec` — fixed in `723998e`, thread resolved

Open findings: **0**.

CI on `723998e` at review time: 7/9 checks passed (Validate PR title, Gateway/BCS/Engine/Sandbox-proxy unit tests, BCS e2e, BaaS unit tests); `Backend unit tests` and `Singlebox coverage` still in progress.

PR is clean as of round 1 (commit `723998e`); this round confirms no regressions and no new findings because no new commits landed.

---

totalfrank:
**Review round 2 — no issues flagged.** Reviewed `723998e` (the `@runtime_checkable` removal and its test change). No new findings; **0 open findings across both rounds**.

**CI is green on `723998e`** — all 9 checks, including the two that were still running when round 2 was posted: `Backend unit tests` (passed 08:23:56Z) and `Singlebox coverage` (passed 08:26:37Z).

Loop summary — 2 rounds, 5 findings, all fixed, none disputed:

| Finding | Fixed in |
|---|---|
| Duck typing on the codec contract (`@abstractmethod` + direct inheritance) | `11dd4c1` |
| `SessionKeyCodecRegistry.__init__` taking `default: SessionKeyCodec \| None` | `11dd4c1` |
| Silent default-engine-type fallback in `register`/`resolve` | `11dd4c1` |
| Module-level singleton instead of DI | `11dd4c1` |
| Stray `@runtime_checkable` reopening structural matching (nit) | `723998e` |

Every thread is resolved. Nothing carried over, nothing left disputed.

Merge gating from here is human review, not CI: FreddieSun's approval was dismissed when `723998e` landed (branch protection drops stale approvals on push) and has been re-requested; @xianmuyq is still pending.

---

FreddieSun (approved):
LGTM

PR #2180, opened 2026-09-14
fix(bcs): isolate queue reply reconstruction from final events

RaymondHX:
## Problem

Group final events were overwritten by the full reply reconstructed for congestion-control queues. Reconstruction inserted newlines between completed segments, breaking full-snapshot prefix matching and causing repeated content in channel replies. Queue eligibility reads could also prevent an unmanaged final from reaching the frontend and channel.

## Solution

Keep original group final events separate from queue reply reconstruction. Frontend events, channel delivery, and non-queued forwarding retain the original final; queue admission and internal run_reply persistence use the reconstructed full reply. Concatenate segments verbatim without inserting or removing whitespace.

Publish unmanaged finals before session lookup and queue reconstruction. Prefer non-empty session participants when available; preserve the existing group-member fallback when the session ID or session service is absent, lookup fails, or session membership is empty. Reuse the loaded participants to decide whether reconstruction is needed. Managed terminals still publish only after their queue transaction commits.

Final text comes from message; delta_text is reserved for delta frames. Empty finals may complete queued replies using accumulated segments without replaying that body to non-queued targets. Preserve the task/direct A2A empty-final fallback. Update documentation and retain legacy sessionless test coverage.

## Validation

- `cargo test --manifest-path src/bcs/Cargo.toml -p bcs-message-flow`: 373 tests passed, including legacy group fallback for missing session IDs/services, lookup failures and missing/empty sessions, queue enabled/disabled, unchanged frontend/channel final payloads, original whitespace, storage retries, and commit-before-publish behavior.
- `cargo test --manifest-path src/bcs/Cargo.toml -p bcs --test integration_structured_routing`: 5 passed.
- `env -u http_proxy -u https_proxy -u all_proxy NO_PROXY='*' cargo test --manifest-path src/bcs/Cargo.toml -p bcs --test provider_downlink_integration`: 7 passed. Local proxy settings stalled the initial attempt; the rerun passed with proxies disabled for that process.
- `cd src/bcs && bash scripts/ci/check-store-boundaries.sh`: passed.
- Commit and push hooks skipped with `--no-verify` as requested. Full workspace and Singlebox E2E were not run locally; local validation covered the affected message-flow crate and module boundary check.

## Compatibility and risk

No schema or configuration changes. Legacy group-member fallback remains unchanged, preserving existing group-chat and Provider callback flows without a real session. Queue reconstruction remains heuristic for rewritten snapshots and retains ambiguous text. Task/direct A2A empty-final behavior is preserved.

## Spec

- `src/bcs/docs/plans/2026-09-04-bcs-message-congestion-control-design.md`


questchaos (approved):
LGTM

PR #2182, opened 2026-09-14
refactor(bcs): drop unused status param from bot heartbeat renewal traits

iParadox-ant:
## Background

After the dynamic-status cache I/O was removed from bot heartbeats, the internal bot repository and registry `update_status` methods kept a `status: BotDynamicStatus` parameter that no implementation consumed. The two store implementations already discard it (`_status`) and only refresh `last_heartbeat`, so the internal trait contract diverged from the implementation semantics and forced every implementation and call site to thread a value that is never used.

## Change

Remove the dangling `status` parameter from the internal traits and propagate the new signature to all implementations and call sites:

- `BotRepoPort::update_status` (`port::repo::bot`)
- `BotRegistryCoreService::update_status` (`core::registry`)

New signature:

```rust
async fn update_status(&self, bot_id: &str) -> bool;
```

The method renews a known registration's heartbeat by `bot_id` only — returning `true` for a known registration and `false` for an unknown one, matching the existing "legacy payload is not retained" contract documented on both traits.

All store implementations, the core forwarding implementation, the two application call sites, and every in-tree test/mock implementation and direct call site are updated to the single-argument signature.

## Public surface unchanged

This is an internal trait-signature cleanup only. The public protocol surface is intentionally preserved:

- `POST /bots/status` still accepts the request `status` and echoes it in the response.
- WebSocket `bot.status` payloads are unchanged.
- `bcs-cli update_status` still accepts `status` (public HTTP client).
- Application command/result types (`BotStatusUpdateCommand` / `BotStatusUpdateResult` / `BotRuntimeStatusOutcome`) still carry their `status` field for response echo.
- Online-state views (`GET /bots` `dynamic_status` / `is_online`) are unchanged; availability is still derived from the actor lifecycle and connection state.

No persistence, cache, or data-migration change is involved.

## Tests

- `cargo check --tests --workspace` passes for the full BCS workspace (all crates and test targets compile against the new signature).
- Targeted tests pass: bot-store heartbeat and store contract tests, bot application use-case tests (status echo without retention), the registry contract test, and the HTTP `bots_contract` test verifying `POST /bots/status` still accepts and echoes `status`.

## Risk and rollback

Low risk: an internal trait signature change with no public API, protocol, or persistence impact. Any regression surfaces at compile time. Revert this commit to roll back.


vzvince:
@codex review

vzvince (approved):
LGTM

PR #2185, opened 2026-09-14
feat(workflow): add today quick time-range and display owner_id in run overview

fudp:
## Problem

The clawweb workflow **run overview** page had two gaps:

1. The time-range switcher in the top-right corner only offered `7天 / 30天`, with no quick `今天` (today) option, making it awkward to inspect today's runs.
2. The overview header did not show the resolved `owner_id` of the workflow, so operators could not tell at a glance who owned a workflow.

### 1. Add a `今天` quick time-range option

- `OverviewTab`'s `days` state widened from `7 | 30` to `1 | 7 | 30`; the switcher buttons become `[今天, 7天, 30天]`.
- Selecting `今天` filters runs from **local midnight to now** (not the last 24h): `windowStart` is computed at local midnight when `days === 1`, otherwise `now - days*86400`.
- `SuccessTrendCard`'s `days` prop widened to `1 | 7 | 30` for compatibility.
- Backend `success-trend` / `health` endpoints are already safe for `days=1`; no server change needed.

### 2. Display resolved `owner_id` in the run overview

- `WorkflowSpecRepository` summaries now include `resolved_owner_id`, preferring the active `workflow_deploy_history` record and falling back to `workflow_specs.owner_id`.
- `ownerId` exposed through the workflow list APIs (`runs.ts`, `workflows.ts`).
- `owner_id` shown in the `WorkflowWorkspace` overview header.

## Validation

- `tsc -p tsconfig.json --noEmit` passes for the workflow module.
- `vitest run OverviewTabLayout.test.tsx`: 11/11 pass, including the new `今天` case asserting `useWorkflowHealth` receives `days=1` and the metric window `from` equals local midnight.

## Compatibility and risk

Pure frontend/API additive changes — no existing `7天 / 30天` behavior is altered, and the backend routes already accept arbitrary integer `days`. Scope is limited to `src/evolverun/clawweb`; if production `clawweb-pre` is actually served by the evolvetrace app, the change may need to be mirrored in `src/evolverun/evolvetrace` to take effect online (kept clawweb-only per request).


3h-william (approved):
ok

PR #2188, opened 2026-09-14
fix(security): enforce auth on harnessflow API, harden internal signature middleware, guard runs deletion

fudp:
## Problem
Two Broken Access Control vulnerabilities allowed unauthenticated callers to:

1. Read internal HarnessFlow workflow definitions (including full node prompts) via "GET /openapi/v1/harnessflow/api/workflows" and "GET /openapi/v1/harnessflow/api/workflows/{id}".
2. Perform CRUD operations on ""/api/internal/*" endpoints (runs, node-executions, events, facades) because the Ed25519 signature middleware fell back to a passthrough when the public key was not configured.
3. Delete arbitrary flow runs via public "DELETE /api/runs/{flowId}", which had no identity or permission checks in either the Evolvetrace or Clawweb runs router.

Root causes: the gateway route table marked harnessflow routes as user-optional; the internal signature middleware returned "next()" when no key was configured; and several route handlers assumed identity was enforced elsewhere.

## Solution
Apply defense-in-depth fixes across the gateway, middleware, and route layers:

- **Gateway**: change ""/openapi/v1/harnessflow/**" from "user: optional" to "user: required" so the edge rejects anonymous requests before they reach the backend. Add a parametrized route-security test covering list, detail, save, delete, and history paths.
- **Evolvetrace signature middleware**: when "EVOLVETRACE_INTERNAL_PUBLIC_KEY_B64" is not configured, return 403 in non-dev environments instead of passing through. Dev/development environments keep the passthrough for local standalone use.
- **Evolvetrace workflows**: require an authenticated identity on "GET /" and "GET /:workflowId". Non-admin callers must pass a view-permission check before workflow specs (including prompts) are returned.
- **Clawweb workflows**: apply the same identity and view-permission checks to "GET /list", "GET /", and "GET /:workflowId" using the existing "resolveWorkflowActorId" and "requireWorkflowAccess" helpers.
- **Evolvetrace + Clawweb runs**: restrict "DELETE /:flowId" to authenticated admin users only (401 when no identity, 403 when not admin).

## Validation
- Gateway route-security tests: 74 passed (5 new + 69 existing).
- Evolvetrace server tests: passed.
- Clawweb server-side workflow tests: 130 passed, zero regressions.

## Compatibility and risk
- Dev environments preserve the existing unsigned internal-call passthrough and workflow-list behavior.
- Deployments that already configure "EVOLVETRACE_INTERNAL_PUBLIC_KEY_B64" are unaffected.
- Deployments without a configured permission table continue to allow all workflows (isolated/read-only compatibility).
- Known follow-up: some related endpoints such as "GET /:workflowId/export" and internal CRUD routes were not hardened individually because the gateway and middleware fixes block anonymous access; a separate hardening pass can add per-route guards later.


xxxxpenny (approved):

3h-william (approved):
ok

wienyang:
@codex review

PR #2191, opened 2026-09-14
fix(engine): support controlled attachment download redirects

RaymondHX:
## Problem
Shared-file image URLs can return HTTP 303 redirects to object storage. Engine previously treated redirects as download errors, so chat attachments failed preparation before reaching the model.

## Solution
Follow up to five 301/302/303/307/308 redirects, including relative Locations. Revalidate DNS on every hop and pin connections to validated public IPs with the original Host and TLS SNI. Reject userinfo, private addresses, HTTPS downgrades, missing Locations and excess redirects. Use fresh clients to isolate cookies, disable environment proxies, and explicitly preserve configured CA certificates with strict TLS verification. Keep one deadline across DNS, redirects and streaming, plus existing byte limits. Document the download contract and add regression coverage.

## Validation
- Targeted downloader, materialization service, WebSocket attachment and Engine architecture suites: 107 passed, including after rebasing onto REL20260915.
- Ruff checks and git diff --check: passed.
- Preproduction bot smoke tests: 303 downloads, cookie isolation and private-address rejection passed; actual pinned-IP TLS succeeded with the configured CA bundle.
- Restarted the patched preproduction Engine and confirmed /health=ok and /readiness=ready.
- User confirmed the image workflow succeeds after the final patch.
- Commit and push hooks skipped with --no-verify per repository publishing skill; explicit tests were run. Full Engine suite was not run; validation focused on the affected attachment paths and architecture checks.

## Compatibility and risk
No request schema, configuration or data migration changes. Chat image and file materialization consumers gain controlled redirect support. Initial HTTP URLs remain supported; HTTPS cannot downgrade. Configured CA trust remains active while environment proxy routing is disabled. Rollback is reverting the downloader change; this restores rejection of redirected downloads.


FreddieSun (approved):
lgtm

PR #2193, opened 2026-09-14
fix(manifest): one row per declaration across the carry-forward fold

totalfrank:
## Problem

`ApplyReport.sources` is documented as "one row per declaration" in two places — the `SourceResolution` docstring and the public `sources` field description in `schemas_config_manifest_apply.py` — and a third place *depends* on it: `ConfigManifestApplyService._last_resolutions` rebuilds strict-mode baselines with `baselines.setdefault(key, source.resolved_sha)`, justified by "rows sharing a key (the report holds one per declaration) carry the same sha, so any serves".

`SourceSession.adopt` is idempotent on `(display, url, ref, mode)`, so the invariant holds *within* one apply. But bot creation runs **two** applies — a pre-container phase and a post-container phase, separated by the whole of container provisioning — each with its own `SourceSession`. `carry_forward` folded them by plain concatenation:

```python
sources = tuple(carried.sources) + tuple(report.sources)
```

A source used by categories in both phases therefore produced two rows. Observed on a `create:on_container` report for a teclaw bot whose git source `content` served `cli_tools` in phase A and `identity`/`resources`/`skills` in phase B.

Beyond the cosmetic duplication, the two resolutions happen minutes apart, so a moving ref (`ref: main`) can land on different commits. Carried rows go first, so `setdefault` kept the **older** phase-A sha.

Scope of the real impact, verified against the code rather than assumed:

- **Strict mode cannot reach it.** Phase B's gate in `apply/source_fetchers.py` refuses a moved ref and adopts nothing, so only one row exists. Phase A's apply record is persisted before phase B starts (`create_job._pre_container_apply_id` reads it back), so phase B genuinely has phase A's baseline to gate against.
- **No false strict refusal.** `mode` is part of the baseline key, so a `non_strict` baseline can never serve a `strict` declaration.
- **The live consequence is `keep_last`.** `GitSourceFetcher._keep_last` stands in with the baseline SHA's stored tree, so it would have reached one commit older than what phase B actually delivered.

Not corruption — an invariant asserted in three places and honoured in two.

## Solution

**`apply/carry_forward.py`** — dedup the concatenated `sources`, keeping the first occurrence, on the **full row** `(name, url, ref, mode, resolved_sha)`.

Deliberately *not* on `(url, ref, mode)`: two genuinely different resolutions are two true facts about what the apply resolved, and silently dropping one would report a delivery that did not happen. `auth` is left out of the key because a credential's name is how the fetch was made, not part of what was resolved.

`categories` and `notes` still concatenate untouched. Carried categories going first is `APPLY_ORDER`'s own order (`script` is position 0) and is what keeps a pre-container `script` from reading as though it had vanished. `carry_forward`'s "never raise, a missing or foreign id is ignored" behaviour is unchanged, as is `SourceSession.adopt`'s dedup key and the strict-mode gate.

**`services/config_manifest_apply_service.py::_last_resolutions`** — make the divergent case deterministic and correct by reading each report's rows backwards, so within one report the **later** row (more recently resolved, and the one that apply's delivery stood behind) wins for a key, while `setdefault` still keeps newest-record-wins across records. The docstring drops the now-unfounded "so any serves" clause and states the actual rule.

`SourceResolution`'s docstring gains a short note naming the one deliberate exception to "one row per declaration", so the assertion and the behaviour agree.

The apply service sits at the 1000-line architecture cap (`test_no_oversized_modules`), so the edits there were written to land at the same 999 lines rather than push a split into this change.

## Validation

- `pytest tests/community/core/bot_config_manifest tests/community/architecture` — **1368 passed** (includes the oversized-module guard).
- `pytest tests/community -k manifest` — **1304 passed**, 17304 deselected.
- `ruff check` on every touched file — clean. (A pre-existing `F401` in `schema/entries.py`, untouched by this change, is the only finding elsewhere in the package.)

New tests:

- `test_one_source_used_by_both_phases_is_one_row` — two identical rows across the fold collapse to one.
- `test_two_resolutions_of_one_moving_ref_stay_two_rows` — same `(url, ref, mode)`, different `resolved_sha`: both survive, carried first.
- `test_the_folded_category_order_is_untouched` — regression guard on `categories` ordering.
- `test_within_one_report_the_later_resolution_is_the_baseline` — `_last_resolutions` over a report holding both divergent rows picks the phase-B sha, and an older record's later row still does not reach past a newer record.

Not run: the full `tests/community` suite and the changed-line coverage gate, both of which CI runs.

## Compatibility and risk

No schema, config, or migration impact — `sources` keeps its shape and the wire payload is unchanged. The behaviour change is confined to reports produced by a carried creation apply: one fewer duplicate row, and a `keep_last` fallback on a moving `non_strict` ref now reaching the commit phase B delivered instead of phase A's. Rollback is a plain revert; no stored data is rewritten, since baselines are derived from report history on every read.

https://claude.ai/code/session_01E83wESgykAYD4xDLURoLcx

---
_Generated by [Claude Code](https://claude.ai/code/session_01E83wESgykAYD4xDLURoLcx)_


FreddieSun (approved):
LGMT

totalfrank (commented):
Reviewed commit `e5a13bb` (the PR's only commit) — no findings.

Files covered: `apply/carry_forward.py`, `apply/outcomes.py`, `services/config_manifest_apply_service.py`, `tests/.../test_apply_service_lifecycle.py`, `tests/.../test_apply_service_uses_strategy.py`.

What I checked:
- **Correctness**: traced `_dedup_sources`'s key `(name, url, ref, mode, resolved_sha)` against the stated invariant (one row per declaration, with the documented moving-ref exception), and traced `_last_resolutions`'s `reversed(report.sources)` + `setdefault` against both the within-report ordering (carried phase first, so the later/real row must win) and the across-record "newest wins" rule. Both hold, including the two-record case where an older record's later row must not reach past a newer record.
- **Contract breakage**: `SourceResolution`'s docstring and the `sources` field's "one row per declaration" claim are updated to state the one deliberate exception, keeping the docstring and behavior in agreement. No wire-shape change (`as_dict`/`as_payload` untouched).
- **Test coverage**: the four new tests exercise exactly the scenarios named in the PR body — same-row collapse, divergent-sha rows staying separate (with order preserved), category-order non-interference, and both the within-report and across-record baseline-selection cases. Traced them against the implementation; expectations match.
- Confirmed `config_manifest_apply_service.py` sits at 999 lines (the stated 1000-line architecture cap), `git log` shows this PR's single commit authored by `totalfrank <totalfrank.chen@gmail.com>`, and all three touched non-test modules parse cleanly.
- Not independently re-run: the pytest suite (no `pytest` available in this environment) — relying on the PR's stated validation (1368 passed) for that.

Categories not applicable to this change: concurrency/resource safety (no new shared state or I/O), security/hygiene (no new external input or secrets), architecture-boundary rules (no transport/DI/env changes), Python type-contract widening (no new `Optional`s beyond the pre-existing `SourceResolution` fields).

---

PR #2195, opened 2026-09-14
refactor(backend): inject the manifest delivery seam instead of assembling it from a flag

totalfrank:
## Problem

The manifest delivery seam is DI-shaped on the outside and hand-assembled on the inside.

- `DeliveryStrategyFactory.for_engine` **constructs** `ArcaDelivery` / `TeclawDelivery` on every call, out of a boolean switch, two port bundles, a redeliver and a CLI service it was holding. The strategies are not injected; their ingredients are.
- `TeclawDelivery` carries `_platform_managed` and branches on it in five members (`creation_sequence`, `phase_of`, `needs_container`, `ports`, `finish`). `ManagedFilesComposeReader` carries the same flag on the compose side. It is a deployment-time fact — it cannot change while the process runs — yet it is re-decided per apply, per compose, in front of every bot.
- Because both port bundles and a redeliver had to be bound whatever the switch said, "platform-managed with nothing bound to be platform-managed *with*" was expressible, and the factory had to refuse it at run time with a `RuntimeError`.
- `BotConfigManifestApplyService` held ten lazy service providers whose *only* consumer was `_arca_ports()` — a bundle it built for the factory it built. That is a wiring decision made by a component rather than by the composition root that wires it (Rule 14).

## Solution

Decide once, at DI time, by table; hand the decision's result to everything downstream.

**One strategy per shape, none of them holding a flag.** `TeclawDelivery` splits into `TeclawPlatformDelivery` (store-backed ports, one closing redeliver, everything `PRE_CONTAINER`) and `TeclawDeviceDelivery` (device-backed ports with the teclaw CLI port substituted, the pre-W8 phases, no closing step). Each constructor takes only what its own shape needs, so the platform one cannot exist without its redeliver and the device one cannot exist without its CLI port — which retires the half-configured state and the `RuntimeError` that refused it.

**Selection is a table, never an if/else.** The yaml scalar is parsed straight into a `TeclawDeliveryMode` (`PLATFORM` / `DEVICE`) by a new leaf module `bot_config_manifest/delivery_mode.py` — a leaf so the typed config cluster can import it without pulling the apply graph in. The composition root then looks that mode up in two tables and binds what it finds:

- `TECLAW_DELIVERY_BY_MODE` → the apply-side strategy
- `_COMPOSE_READER_BY_MODE` → the compose-side reader

Both are exhaustive at import, the discipline `source_fetchers.FETCHER_TYPES` already sets, so a mode a deployment can write and nothing can build fails at boot rather than mid-apply. The shape that is not selected is never constructed. `phase_of` becomes a table lookup too, in a per-strategy phase map (`_ARCA_PHASES`, `_TECLAW_PLATFORM_PHASES`, `_TECLAW_DEVICE_PHASES`), each checked against `APPLY_ORDER` at import — which removes the last per-step branching.

**The factory is a lookup.** `DeliveryStrategyFactory` now takes a `Mapping[EngineFamily, DeliveryStrategy]` and a family resolver, and does one lookup. It builds nothing, configures nothing, and refuses at construction a mapping that does not name every family. `family_from_engine_test` adapts the engine authority's boolean into the family key once, at that boundary.

**The apply service takes the seam, not its ingredients.** Fourteen constructor arguments go, the ten providers among them; the device-backed bundle is now built in `manifest_fetch_module`, beside the store-backed one it mirrors.

**The compose seam gets the same treatment.** `ManagedFilesComposeReader` loses `platform_managed` and is simply the platform-managed reading; `EngineOwnedComposeReader` (owns no compose, holds nothing) is bound in its place for a device-backed deployment. This is the seam the collector's own contract already anticipated — "a caller that wants every compose engine-owned passes a reader that says so". Both bind under a new `ComposeManagedFilesReader` Protocol, which names the pair of questions the collector actually asks of one object.

**Rejected alternatives.** A `@ConditionalOnProperty`-style declarative binding has no equivalent in `injector`; separate DI modules selected per mode would spread two rows across two files and lose the import-time exhaustiveness check. Keeping one `TeclawDelivery` and injecting a pre-chosen port bundle would remove the ports branch but leave the flag deciding `creation_sequence`, `needs_container` and `finish`.

## Validation

- `tests/community/core/bot_config_manifest`, `tests/community/di`, `tests/community/core/config_compose`, `tests/community/adapters/http/openapi_v1` — **3244 passed, 2 skipped**, re-run after rebasing onto the current base.
- `tests/community/endpoints` (incl. the endpoint runner cases that drive the whole DI graph), `tests/community/config` golden snapshots — **passed**.
- `tests/community/architecture`, `tests/community/contract`, `tests/community/contracts`, `tests/community/test_import_smoke.py` — **623 passed**.
- Full `tests/` run: **18547 passed, 63 skipped, 4 failed** — two of those were the endpoint cases fixed later in this branch (they pass now), and the other two (`test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs`) fail on `rsync: not found` in this container and are unrelated to the change.
- The block-rule lint from `scripts/ci/python_sast_local.sh` (flake8 `E999,E902,E117,F405,E712,E701,E702,F821,F822,F823,F831`, plus `F401,F811,F841`) over every changed file — clean.
- Could not run: the singlebox coverage gate and the BCS/E2E stories (no product stack in this container).

## Compatibility and risk

No behaviour change. The yaml key `user_config.bot_config_manifest.teclaw_platform_managed` is unchanged, including its strict boolean parsing and its `false` default, so no deployment's effective config moves (the config goldens are untouched). Both delivery shapes behave exactly as they did, and with the default mode the compose-side reader owns nothing — as `platform_managed=False` already made it do.

The typed config field changes shape: `BotConfigManifestConfig.teclaw_platform_managed: bool` becomes `teclaw_delivery_mode: TeclawDeliveryMode`. Internal only; the yaml is the deployment contract and it is untouched.

Rollback is a revert; nothing is persisted or migrated.

https://claude.ai/code/session_014psoPNUexKBUP2RLovCnfi

---
_Generated by [Claude Code](https://claude.ai/code/session_014psoPNUexKBUP2RLovCnfi)_


src/backend/src/agentclaw/community/core/bot_config_manifest/services/config_manifest_apply_service.py
```diff
@@ -203,57 +159,16 @@ def __init__(
...
-        mcp_auth_service_provider: Callable[[], MCPAuthServiceProtocol],
-        identity_service_provider: Callable[[], IdentityFilePort],
-        upload_service_provider: Callable[[], LocalSkillUploadServiceProtocol],
-        capability_reader_provider: Callable[[], BotCapabilityStateReaderProtocol],
-        package_validator_provider: Callable[[], SkillPackageValidator],
-        entry_fetcher_provider: Callable[[], DeclaredSourceResolver],
-        resource_service_provider: Callable[[], ResourceFilePort],
-        cli_tool_service_factory: Callable[[str], "CliToolService"],
         git_client_provider: Callable[[], GitSourceClient],
         task_queue_provider: Callable[[], "TaskQueueService"],
```
totalfrank-personal: Do we really need this forward reference? If not, let's get rid of it.

totalfrank (commented):
Reviewed commit `33dce60` (the sole commit on this branch) against merge-base `3c62407` on `REL20260915`. No prior review from me on this PR.

No findings. This refactor is internally consistent and I could not find a correctness or contract regression.

**What I checked:**
- **Correctness / behavior parity**: traced the old `TeclawDelivery(platform_managed=...)` branching against the new `TeclawPlatformDelivery`/`TeclawDeviceDelivery` split — phase tables (`_ARCA_PHASES`, `_TECLAW_PLATFORM_PHASES`, `_TECLAW_DEVICE_PHASES`), `needs_container`, `finish`, and the CLI-tool port substitution all reduce to the same effective values as before (verified the platform bundle now embeds `cli_tool_service_factory("teclaw")` directly in `manifest_fetch_module.py`, replacing the old post-hoc `replace()` override that achieved the same result).
- **Config parsing**: compared `teclaw_platform_managed_from_config` (old) against `teclaw_delivery_mode_from_config` (new, in the new leaf module `delivery_mode.py`) value-by-value, including the string/int/bool edge cases and the "absent → device/False" default — parsing is equivalent.
- **Exhaustiveness guards**: `TECLAW_DELIVERY_BY_MODE`, `_COMPOSE_READER_BY_MODE`, `DeliveryStrategyFactory`'s family check, and the `_ARCA_PHASES`/`_TECLAW_*_PHASES` vs `APPLY_ORDER` assertion all fail at import/construction time on a gap, as intended.
- **Contract/DI wiring**: `bot_management_module.py`, `manifest_fetch_module.py`, `service_bot_module.py` call sites match the new constructors; no dangling references to the removed `TeclawDelivery`, `teclaw_platform_managed_from_config`, or the old `ManagedFilesComposeReader(platform_managed=...)` param remain in any file this PR touches.
- **Architecture boundaries**: wiring/selection moved fully into composition roots (`manifest_fetch_module.py`), matching Rule 14; `delivery_mode.py` is a true leaf (no imports beyond stdlib), consistent with its stated purpose of letting `di/config.py` import it without pulling in the apply graph.
- **Docs**: `README.md` and the Protocol docstrings in `config_compose/protocols.py` were updated to match the new shapes.
- **Test coverage**: `test_delivery_strategy.py`, `test_teclaw_delivery.py`, `test_apply_service_uses_strategy.py`, `_fakes.py`, and the creation/endpoint test rigs were all updated to construct the new seam the same way the composition root does (`delivery_strategies`/`arca_only_delivery` helpers), including a new `UnreachableDelivery` fake that fails loudly if a rig's assumed-unreachable family is hit — this is a good replacement for the old individual `unreachable_*` fakes since it now also gets exercised through the mandatory-mapping exhaustiveness check itself.

Categories from the rubric with nothing to flag: error/failure handling, concurrency/resource safety, security/hygiene, Python type contracts (the `T | None` usages here are all genuine boundary/optional cases, unchanged in spirit from before).

---

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/delivery.py
```diff
@@ -270,14 +320,44 @@ async def finish(self, ctx: ApplyContext, report: ApplyReport) -> Optional[str]:
...
 })
 
+#: Platform-managed teclaw: the artifact is the delivery and it is composed
+#: before provisioning, so nothing waits for a container. ``script`` is
+#: unsupported on this family (the capability resolver refuses it); its phase
+#: is stated anyway, and stated as the ARCA one, so a declared script still
+#: walks the orchestrator's no-support path and is reported, not skipped.
+_TECLAW_PLATFORM_PHASES: Mapping[ApplyConstruct, ApplyPhase] = MappingProxyType({
+    construct: ApplyPhase.PRE_CONTAINER for construct in _ARCA_PHASES
+})
```
totalfrank-personal: I feel this brings more confusions than simplifications.

I would prefer a hard coded list. Your current implementation confuses me because I kept thinking how come teclaw phases are related to arca phases?

src/backend/src/agentclaw/community/core/bot_config_manifest/apply/delivery.py
```diff
@@ -270,14 +320,44 @@ async def finish(self, ctx: ApplyContext, report: ApplyReport) -> Optional[str]:
...
+#: the deployment's mode moved in front of the container.
+#:
+#: ``cli_tools`` is always platform-managed on this family whatever mode the
+#: deployment runs (spec D-6, D-8), exactly as ``mcp`` is, so it is composed
+#: into the artifact before provisioning. The distinction is invisible on an
+#: existing bot, where the two phases run back to back; it decides something on
+#: exactly one path, the W13 creation whose platform-managed sequence has no
+#: phase B at all, where an ``ON_CONTAINER`` ``cli_tools`` would never run.
+_TECLAW_DEVICE_PHASES: Mapping[ApplyConstruct, ApplyPhase] = MappingProxyType({
+    **_ARCA_PHASES,
```
totalfrank-personal: Same here, I prefer just hard code the entire list.

FreddieSun (approved):

PR #2200, opened 2026-09-14
fix(clawevolve): persist packaged Skill versions

3h-william:
## Summary

- Persist generated release versions back to each public Clawevolve Skill’s `version` file.
- Keep packaged Skill metadata and source versions aligned.
- Move the AgentEvolve documentation link to the appropriate position in the main English and Chinese documentation lists.

## Validation

- Verified public and internal Skill package composition in an isolated temporary directory.
- Confirmed that generated versions are written back to the correct Skill owners.
- Shell syntax and Git diff checks passed.

## Scope

This change only updates Clawevolve Skill packaging/version metadata and documentation ordering. It does not change Skill runtime behavior.


xianmuyq (approved):

ZhaosenLin (approved):
lgtm

PR #2201, opened 2026-09-15
fix(backend): enrich default MCP display from passport

FreddieSun:
## Problem

Default SkillSet MCPs are projected from policy/configuration and may expose a server-code suffix with the generic description "默认 MCP", while AgentPassport already has the localized current display metadata.

## Solution

Query AgentPassport once in the SkillSet resource read model. Reuse that snapshot for Default CLI resources and overlay non-empty `mcp_name` / `mcp_desc` only onto MCPs already projected for the Default Set. Membership, exclusions, ordinary SkillSet MCP Center snapshots, runtime projection, and authorization remain unchanged.

## Validation

- `uv run ruff check ...`
- Focused service tests: 3 passed
- `tests/community/core/skill_center/test_skill_set_management_service.py`: 124 passed
- Resource endpoint runner: 5 passed
- OpenAPI/Legacy contracts plus architecture gates: 15 passed
- Standards/Spec dual review: no findings

## Compatibility and risk

The HTTP DTO shape is unchanged. If AgentPassport is unavailable, malformed, or omits a metadata field, the existing MCP projection remains and Default CLI resources degrade to an empty list as before.

## Spec

User-confirmed contract: support every engine type, enrich Default MCP display metadata from AgentPassport, and do not use Passport to determine MCP membership or overwrite custom MCP snapshots.


totalfrank (approved):
LGTM — reviewed `fb4354e` (5 files: the new `skill_set_resource_projection.py`, the `list_resources()` refactor, and matching test updates). The merged `query_agent_passport` call is verified equivalent to the previous separate `query_passport_clis()` call in both concrete `PassportPlugin` implementations (both are literally `extract_cli_items(query_agent_passport(...))`), and the display-metadata overlay is correctly scoped to `is_default` items only, with graceful degradation on passport failure covered by tests. No bugs found in the new code.

---

PR #2202, opened 2026-09-15
fix(workflow): include run input in keyword search

wienyang:
## Problem
The workspace run-list keyword search matches Run IDs, initiators, and Bot IDs, but cannot find runs by business keywords in their input.

## Solution
- Extend the existing keyword predicate to include stored `input_json`.
- Update the search placeholder to make input-content search discoverable.
- Preserve shared permission, workflow, status, count, and pagination filtering.
- Document the contract and extend real SQLite/HTTP-route regression coverage.

## Validation
- Workflow test suite: 205 tests passed across 39 files.
- Workflow TypeScript check passed.
- Workflow build passed.
- Regression verified red before the fix and green afterward, including Chinese nested input, literal `%`, `_`, and `!`, pagination, and visibility filtering.
- `git diff --check` and pre-push gate passed.

## Compatibility and risk
No schema or configuration changes, and no ClawMind changes. Search matches raw stored JSON text, not decoded values, and does not search node logs or outputs. Existing `inputQuery` remains an independent AND filter. Input-text LIKE matching may add query cost on large histories; no production load benchmark was performed.

## Spec
Updated `src/evolverun/clawweb/public/modules/workflow/docs/run-list-contract.md`.


wienyang:
@codex review

3h-william (approved):
ok

PR #2203, opened 2026-09-15
docs(agents): document source file size limit

FreddieSun:
## Problem

The CI-enforced 1,000-line source-file limit was not recorded in the repository guidance, so contributors could discover it only after a failing CI run.

## Solution

Document the pre-submit check, responsibility-based file split, and narrow allowlist exception process in `AGENTS.md`.

## Validation

- `git diff --check`

## Compatibility and risk

Documentation-only change. Existing allowlisted files remain governed by the documented exception process.


totalfrank (approved):
LGTM.

PR #2204, opened 2026-09-15
feat(human-notify): fan out mentions to multiple backends and deliver work-order notices

zh-bupt:
## Problem

Group-chat @human mentions in BCS supported only a single notification backend (`human_notify.provider` single-select), so mentions cannot fan out to multiple channels at once; and mentions never reach the user's work-order/notification inbox, so a mentioned user has no visibility outside BCS-facing surfaces.

## Solution

- **Multi-provider config**: `human_notify` migrates to an ordered array `[[human_notify.providers]]` (entry = `name` + `enabled` + flattened options; breaking change). Every enabled backend receives each notification in parallel; a new `CompositeHumanMentionNotifyAdapter` in bootstrap performs the fan-out (per-backend 10s timeout with error isolation — one failing backend never affects the others). The message-flow layer (`HumanMentionNotifyPort` + `human_notify_hook.rs`) is untouched.
- **Generalized Mist reference resolution**: `resolve_config_secrets` now resolves any `<key>_secret` option reference inside a human_notify entry into a plain `<key>` value (a non-blank literal wins, the reference key is kept; the literal field `client_secret` is excluded by name so it is never mistaken for a reference). Reuses the existing SecretAccessPort/Mist pipeline.
- **New `work_order` backend** (crate `bcs-human-notify-work-order`): for each mentioned human it mints a short-TTL user principal (HS256 shared gateway key, `iss=bcs`, `aud=backend`, `kid=bare`, 60s TTL) and POSTs a `HUMAN_GROUP_MENTIONED` NOTICE event to the backend internal endpoint `/api/v1/work-orders/events` (carries only the `x-avernet-principal` header — never Authorization/Cookie; a 5s per-request timeout prevents one slow response from consuming the shared budget). Recipients are determined solely by `recipient_user_ids`; the principal only passes authentication.
- **Backend side**: registers the `GROUP_MENTION` biz type and `HUMAN_GROUP_MENTIONED` NOTICE event; widens `PrincipalVerifierConfig.issuer` to `str | tuple`, and trusts the `("gateway", "bcs")` issuer allow-list on the internal ordinary-HTTP surface ONLY (`_resolve_ordinary_http_caller`) — the strict `/openapi/v1/*` surface keeps its single `"gateway"` issuer; pyjwt floor raised to 2.9.0 (multi-issuer validation; the lock still resolves 2.13.0).

Two design choices worth calling out: the minted principal uses the **mentioned human's staff id** rather than a service account (user-confirmed; audit shows the recipient as the event actor, and delivery targeting is unaffected by the signing subject). And `.no_proxy()` was deliberately NOT set, matching the friend-connect internal client precedent — note that if deployment pods set `HTTP(S)_PROXY` and the backend is only reachable through it, this should be revisited.

## Validation

- Rust (`CARGO_TARGET_DIR=/tmp/cargo-target`, serial, `--offline`): all 8 test sections of `bcs-human-notify-work-order` + `bcs-human-notify-api` + `bcs-config-api` pass (JWT claim-contract shape, payload constants, UTF-8-safe truncation, error-aggregation semantics, one-shot HTTP server delivery assertions); bootstrap `human_notify` filter 11/11; the `resolve`/`checked_in`/`shipped_manifests` guard filters all pass; `cargo build --workspace` green on the integrated branch.
- Python (`uv run pytest`): `tests/community/core/work_orders/` + `tests/community/adapters/http/org/` + `tests/community/core/gateway_principal/` + `tests/community/adapters/http/work_orders/` — 182 tests pass, including the new event registration, the org-endpoint `iss=bcs` red→green case, and verifier allow-list semantics lock-in.
- Not run: the full bcs workspace suite (3 pre-existing unrelated failures — logging/migrations — reproduced on a clean baseline) and one backend architecture E3 pre-existing failure (`files` module uncovered, stash-verified on the clean baseline).

## Compatibility and risk

- **Config breaking**: the legacy `[human_notify] provider = "..."` + `[human_notify.providers.<name>]` shape no longer parses; checked-in configs and environment overlays must migrate in the same release (Avernet example/local configs migrated here; the companion ocb migration lands in the same release window).
- **Deployment order**: the backend half (issuer allow-list + event registration) must deploy first or simultaneously with any BCS that enables a `work_order` provider entry — otherwise BCS-minted tokens get 401'd (fire-and-forget isolates it to warn logs; notifications are lost but the main path is unaffected). Then bump the ocb `ocb-public` submodule pin, and only then add `work_order` entries to the internal overlays.
- **Security posture**: this is the first time BCS mints user principals. The system's security model is already "possession of the shared HMAC key equals trust" (the same key backs BCS's inbound verification), so minting adds no new trust boundary. Implementation constraints: minting only for mentioned staff ids, 60s TTL, the key held via `Secret` and never logged; the issuer widening is confined to the internal surface.
- **Rollback**: revert the BCS commit and restore the config overlays; the backend commit is backward-compatible (the new event type does not affect existing data).

## Spec

docs/superpowers/specs/2026-09-14-human-notify-multi-provider-work-order-design.md (design + implementation plan live on branch docs/human-notify-multi-provider-work-order)


questchaos (approved):
LGTM

FreddieSun (approved):

Human:
> Group-chat @Human mentions in BCS supported only a single notification backend (`human_notify.provider` single-select), so mentions cannot fan out to multiple channels at once; and mentions never reach the user's work-order/notification inbox, so a mentioned user has no visibility outside BCS-facing surfaces.

Please do not notify me in these PRs.

PR #2209, opened 2026-09-15
feat(clawweb): add run list time range filter, fix dashboard navigation, add failed count column

fudp:
## Problem

1. The run list in the workflow workspace overview page lacks a time range filter, showing all historical data by default with no way to narrow to recent runs.
2. Clicking a workflow in the data dashboard navigates to `/workflow/{id}/metrics`, a route that does not exist, resulting in a 404.
3. The dashboard workflow list table has no failure count column or sort option, making it hard for admins to quickly identify workflows with the most failures.

## Solution

1. Added a `TimeRangeFilter` dropdown to the run list filter form (default "last 7 days"), independent from the metric card time buttons above.
2. Fixed navigation in `WorkflowFullTable.tsx` and `GuardianMetricDrawer.tsx` to use `/workflows/workspace?workflowId={id}` instead of the non-existent `/workflow/{id}/metrics`.
3. Backend: added `failed_count` SQL aggregation (`SUM(CASE WHEN status = 'failed')`) in `dashboard-repository.ts`. Frontend: added a "Failed" column and "Failed count descending" sort option to the dashboard workflow table.

## Validation

- `git diff` confirmed 7 files changed, 23 insertions, 5 deletions.
- Rebased onto latest `origin/dev`; conflicts resolved (kept dev's `resolveWorkflowActorId` implementation and import style).
- Code review (CR) passed.


ZhaosenLin (approved):
lgtm

PR #2210, opened 2026-09-15
docs(readme): mark orchestration capability as available

fudp:
## Problem

The Orchestration capability in the README was still shown as "Planned", but the underlying orchestration layer in Avernet is already available.

## Solution

- Updated the Orchestration badge from "Planned" to "Available" in the "Shared intelligence and evolution" section.
- Updated the section description to clarify that orchestration is available via the Avernet coordination layer, while context and memory remain planned.

## Validation

-  verified only  changed, with the badge URL and description updated.


ZhaosenLin (approved):
lgtm

xianmuyq (approved):

PR #2214, opened 2026-09-15
fix(clawevolve): stabilize planning, model propagation, and pack restore

3h-william:
## Summary

This PR improves ClawEvolve runtime stability across local and packaged deployments without changing the existing business workflow.

## Changes

- Generate final Plan documents deterministically from validated structured data.
- Remove provider-specific hidden model defaults.
- Preserve explicitly selected models across Diagnose, Plan, Optimize, Tune, and Review stages.
- Let OpenClaw use its configured model when a Subagent model override is omitted.
- Require an explicit model for the direct API Judge, which cannot inherit OpenClaw configuration.
- Restore Packs using the artifact reference frozen by ClawWeb instead of assuming a fixed storage bucket.
- Keep strict artifact path, content type, size, and SHA-256 validation.
- Preserve backward compatibility for existing Pack invocation arguments.

## Scope

- No new model provider or fallback architecture.
- No changes to the existing optimization workflow semantics.
- No changes to object storage or deployment architecture.
- No credentials or environment-specific secrets are included.

## Validation

- ClawEvolve TypeScript and public-boundary checks passed.
- ClawEvolve tests: 336 passed.
- Diagnose tests: 108 passed, 1 skipped.
- Pack handler tests: 10 passed.
- `git diff --check` passed.


XIEAN99 (approved):
lgtm

PR #2222, opened 2026-09-15
feat(backend): 支持 UPFS 创建与草稿重启并分离存储初始化

maicaisongxiaocong:
普通与服务草稿在创建专属入口判断开关、校验实际模板并持久化 storage_policy；重启及补建复用已有策略，发布与 Caller 保持旧行为。

删除 initial_storage_decision 和 initial_storage_user_id，复用 owner_id、template_config 和原 storage 链路。创建复用同一次模板解析和已提交的存储选择，不重复选模板或经兼容读取降级。存储初始化保留独立方法，后续阶段可复用。

BaaS 复用 storage 传递已校验挂载，仅企业 SDK 适配器生成 volume_mounts；保留销毁前校验和 NAS 参数兼容。测试、部署要求及未验证项见方案文档。

补充存储灰度、策略初始化与复用、请求关联及 UPFS 挂载组装日志；记录失败阶段和异常类型，不打印敏感配置或原始异常正文。


totalfrank (approved):
## Verdict: APPROVE

## Gate 0
- Target branch: `dev` (the integration branch) — not a `REL*` release branch, so Gate 0's branch-policy gate does not apply. Normal route (feature → `dev`). **Pass.**

## Findings
None. This PR is architecture-clean under the constitution.

Notes on how it satisfies the catalog (not findings, just the verification trail):

- **No engine-string branching added to `core/`.** The storage-type dispatch (`Storage.type == "upfs"`) in `src/baas/.../core/service/device_manage/_device_service.py` and `.../core/service/paas/_facade.py` is domain-level storage-backend branching (NAS vs UPFS), not AI-coding-engine branching, and it reuses the pre-existing ARCA/SIGMA provider-type pattern already in that file. The one place an "engine" string does flow through (`BotStoragePolicyService.initialize(engine=...)` → `_should_rollout_upfs`) compares it against **rules loaded from configuration** (`bot_storage/upfs_rollout`), not a hardcoded set in code, and the `engine` value itself is produced by the existing `resolve_baas_engine_bucket` registry (`core/bot_management/engines/registry.py`), untouched by this diff.
- **Plugin variability follows the required contract pattern.** `ArcaSandboxPlugin.create_sync_sandbox`'s `storage` param is documented to carry a resolved UPFS mount; each of the four community sandbox plugins (`_stub`, `aliyun_ack`, `local_docker`, `local_proc`) either honors it or explicitly rejects it before any side effect, and a new conformance test (`tests/unit/api/arca_sandbox/test_upfs_contract.py`) asserts this for all four, including the always-present stub as the safe default. This is a legitimate Plugin API use (ARCA sandbox providers are external system boundaries), not engine-variability misrouted into plugins.
- **Composition roots own all wiring.** `env` is read via `get_current_env()` only inside bootstrap/DI providers (`secbaas/.../bootstrap/_core_services.py`, `agentclaw/.../di/modules/service_bot_module.py`) and injected into constructors (`DefaultDeviceTemplateService.deployment_env`, `BaasService.storage_env`). Whether `BaasService` gets a `bot_storage_service` at all is decided in `ServiceBotModule` by `deploy_runtime.runtime == DeployRuntime.MANAGED`, i.e. a composition-root config check, not a runtime branch in core.
- **New contracts are real Protocols with defaults.** `BotCommonConfigServiceProtocol` / `BotStoragePolicyProtocol` (`core/common_config/bot_config_protocol.py`) are defined separately from implementation and delivery; `DeviceService.apply_device_for_new_bot` ships a base/default implementation (`return self.apply_device(**kwargs)`) that non-BaaS providers inherit unchanged, and `BaasDeviceService` overrides it — matching the required contract+implementation+default shape.
- **Additive, backward-compatible contracts.** `Storage.upfs` is a new optional field (default `None`, stripped from NAS serialization by the wrap-serializer); `StorageType` gains a new enum member (`UPFS`) alongside the existing `NAS`; `ac_bot_common_config` is a brand-new table; `DeviceTemplateManageService.get_storage_capability` is a new Protocol method with a dedicated safe response model (`TemplateStorageCapability`, explicitly documented to never leak template secrets) and a thin HTTP adapter (`device_template_router.py`) that just deserializes/serializes.
- **Propagation done.** `docs/specs/2026-09-15-upfs-phase1/spec.md`, `core/common_config/README.md` (context boundary), and `core/repository/README.md` are updated in the same change; the spec explicitly enumerates what changed and what was deliberately trimmed from an earlier version of this branch to reduce collateral surface.
- No transport/framework imports, no new `os.environ`/`os.getenv` reads outside bootstrap/DI, no hardcoded engine filesystem layout, no hidden string wire protocol (the Backend→BaaS storage-capability call is a documented, typed REST endpoint, not a parsed ad hoc string).

## Out-of-scope notes
- `BotStoragePolicyProtocol._resolve_storage_policy` is a leading-underscore method on a public `Protocol`, and `baas_service.py` calls it across the service boundary via the protocol type — naming only, not a layering violation; worth a rename for clarity but not gating.
- A couple of local/deferred imports inside function bodies in `_device_service.py` (e.g. `Storage`, `VolumeMountSpec`) — style, likely avoiding a circular import; out of scope here.

---

totalfrank (approved):
## Architecture-Gate Review

**Verdict: APPROVE**

### Gate 0 — Target branch

- Target: `dev` (the integration branch). Not a `REL*` release branch, so the branch-policy gate and Isolation Audit do not apply — this PR takes the normal route. **Pass.**

### Review scope

Read `docs/arch/arch.rules.md` (v3.1) and `AGENTS.md`, then the full diff (44 files, +2894/-83) across Backend (`agentclaw/community`) and BaaS (`secbaas/community`). Focused on the engine/storage-variability seam this PR introduces (UPFS vs. NAS) and every hunk touching pre-existing shared code.

By architectural role:
- **New contracts (Backend, `core/common_config/`)**: `BotCommonConfigServiceProtocol`, `BotStoragePolicyProtocol`, `StoragePolicy` — Service API Protocols defined in the engine-agnostic layer, consumed via injected interfaces only (Rules 1, 3, 5).
- **New repository contract** (`core/repository/protocols/config.py` + `implementations/config/bot_common_config.py`): `BotCommonConfigRepositoryProtocol`, persisted through the existing `DatabasePlugin` boundary — no new plugin-lifecycle machinery invented.
- **Composition roots** (`di/modules/common_config_module.py`, `di/modules/service_bot_module.py`, BaaS `bootstrap/_core_services.py`, `bootstrap/_container.py`): implementation selection and `deploy_runtime.runtime == DeployRuntime.MANAGED` / environment injection all happen here, not in core. `BaasService` receives `bot_storage_service: BotStoragePolicyProtocol | None` and `storage_env: str` as constructor arguments rather than resolving them itself — matches Rule 14's required shape, and the code comment (`baas_service.py`) explicitly calls this out: "Core/payload composition never inspects a composer's identity."
- **Core services** (`baas_device_service.py`, `device_service.py`, `device_service_router.py`, `baas_service.py`): the new `apply_device_for_new_bot` template method is cleanly isolated to the creation path; non-BaaS providers fall through to the unmodified `apply_device` (`DeviceService.apply_device_for_new_bot` default in `device_service.py:522` just delegates). Engine-bucket resolution reuses the pre-existing `BaasEngineBucketResolverRegistry` composition-root pattern (`bot_management/engines/registry.py`) instead of adding new engine-name branching — no new `if engine_type == "..."` in domain code anywhere in the diff.
- **BaaS core** (`_device_service.py`, `_facade.py`, `_device_template_service.py`): `ArcaTemplateConfig.upfs_volume_id_pre/prod` and `Storage.upfs` are additive, optional, backward-compatible fields; NAS serialization/behavior is provably unchanged (`model_serializer` drops `upfs` when `None`). `deployment_env` is injected from bootstrap (`providers.Callable(get_current_env)` in `_core_services.py`, a composition root), not read inside the service.
- **Plugin implementations** (`plugins/sandbox/arca/{_stub,aliyun_ack,local_docker,local_proc}`): a genuine system-boundary Plugin API (external ARCA sandbox providers). Each plugin declares its own UPFS capability by rejecting unsupported mounts before any side effect; `test_upfs_contract.py` is a real conformance test parametrized across all four implementations (Rule 25) confirming rejection precedes any client/process setup.
- **Service API additions**: `DeviceTemplateManageService.get_storage_capability` is added to the Protocol (`_protocols.py`) and covered by the protocol-surface conformance test (`test_protocols.py`, asserts the method name appears on both the service and its stub). The HTTP router (`device_template_router.py`) is a thin adapter — parses request, calls the service, returns the DTO; no domain policy in the adapter.

### Applying the "add engine/provider foo" test

For a hypothetical fifth ARCA-like sandbox provider: the answer is a new plugin module implementing `ArcaSandboxPlugin` plus one registration entry in bootstrap — no engine-agnostic domain file requires edits. For a new AI-coding engine, this PR doesn't add or need to touch the aicoding/openclaw/teclaw registries at all; `engine` flows through `bot_config_service.py` as an opaque config-matched string (`rule.get("engine_bucket") != engine`), never branched on by name.

### Propagation & tests (Rules 16, 25)

`docs/specs/2026-09-15-upfs-phase1/spec.md` documents affected consumers/implementations, deployment order (DDL → Backend → enterprise BaaS → config → rollout), rollback constraints, and explicitly enumerates what was NOT validated (real ARCA mounts, production DDL, full E2E). DI-wiring tests (`test_new_storage_services_are_wired_in_di`, `test_composition_root_selects_storage_policy_support`) and the four-plugin UPFS contract test give this a guard against the boundary it introduces.

**Findings: none.**

### Out-of-scope notes

- `BaasDeviceService.apply_device_for_new_bot` calls `env_utils.get_current_env()` directly inside `core/devices/services/`, which nominally matches catalog item 6 ("environment detection outside composition roots"). This mirrors the exact pattern already used by the base `DeviceService.apply_device` in the same file family (`device_service.py:584-585`) and ~30 other pre-existing call sites across `core/`; it is not a shape this PR introduces, so it isn't gated here, but it's worth a follow-up architecture note given the constitution's stated intent for Rule 14.
- `baas_service.py` and BaaS `_device_service.py` are both already well past the 1,000-line CI guideline pre-existing this change; this PR doesn't materially worsen that, and the router file it does touch stays under the limit per the PR's own note.
- Business logic of the rollout rule matching (`_should_rollout_upfs`) and log field choices are outside this gate's scope; not reviewed for correctness.

---

totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0 — Target branch

- Target: `dev` (integration branch, not a `REL*` release branch).
- Classification: feature work landing on the normal integration branch.
- **Pass.** No release-branch gate applies; the Isolation Audit is not required.

### Structural read

This PR is well-structured overall: `Storage.upfs` is an additive, optional field with a `model_serializer` that strips it when unset (NAS wire format is provably unchanged), the new `DeviceTemplateManageService.get_storage_capability` and `ArcaSandboxPlugin` UPFS behavior are declared as explicit Protocol/contract changes with a conformance test (`test_upfs_contract.py`) that exercises every registered plugin including the default (stub), implementation selection for `BotStoragePolicyService` is done in the composition root (`ServiceBotModule.baas_service`, gated on `DeployRuntime.MANAGED`) rather than by core inspecting component identity, and `deployment_env` is correctly injected into `DefaultDeviceTemplateService` via `providers.Callable(get_current_env)` in `bootstrap/_core_services.py`. Context-boundary docs (`common_config/README.md`, `repository/README.md`) were updated per Rule 22, and the "engine" values threaded through the rollout logic (`_should_rollout_upfs`) are opaque config-matched buckets, not hardcoded `if engine_type == "..."` branches — no catalog items 1/2/3 findings.

One Invariant violation survived review:

### Findings

1. **`src/backend/src/agentclaw/community/core/devices/services/baas_device_service.py:161,170`** — Rule 14 (Invariant) / Structural Red Flag "environment access is scattered outside config or bootstrap".

   ```python
   from agentclaw.community.utils.env_utils import get_current_env
   ...
   env = get_current_env()
   ```

   This is new code (`apply_device_for_new_bot`, added by this PR) in a `core/` service that reads deployment environment directly instead of receiving it from a composition root or caller. The PR itself demonstrates the compliant alternative in the very same change: `ServiceBotModule.baas_service` (the composition root) already resolves `env_utils.get_current_env()` once and threads it into `BaasService` as `storage_env`, and `bootstrap/_core_services.py` injects `deployment_env=providers.Callable(get_current_env)` into `DefaultDeviceTemplateService`. `apply_device_for_new_bot` could have accepted `env` the same way (or reused `self._baas_service`'s already-injected value) instead of importing and calling `get_current_env()` inline.

   I recognize this exact call shape (`env = env_utils.get_current_env()` inside a `core/devices/services/*.py` method) is already pervasive in this file family (`device_service.py`, `local_device_service.py`, `device_instance_service.py`, `oss_to_nas_switch_service.py`, ~146 call sites across `core/` in total) — but per this review's mandate, pre-existing violations elsewhere don't license new instances, and new code extending an existing violation is itself a violation. Required fix: pass `env` into `apply_device_for_new_bot` (it's already threaded through `apply_device`'s kwargs elsewhere in this router chain) or resolve it once in the composition root and inject it, matching the pattern the PR already uses for `BaasService.storage_env` and `DefaultDeviceTemplateService.deployment_env`.

### Out-of-scope notes

- `BotStoragePolicyProtocol` declares `_resolve_storage_policy` (leading underscore) as part of its public contract surface, and it's called cross-module from `baas_service.py`. Naming/style, not gated here, but worth a follow-up rename since a leading underscore signals "implementation detail" for something that's actually contract surface.
- `DefaultDeviceService.__init__`'s new `system_config_service: SystemConfigManageService | None = None` and `BaasService.__init__`'s new `bot_storage_service: BotStoragePolicyProtocol | None = None` are optional-typed for backward compatibility with existing callers/tests — flagged only as a note per AGENTS.md's Python Type Contracts guidance, not gated in this architecture review.
- The spec doc references an enterprise-only deployment copy (`src/backend/sql/20260915_ac_bot_common_config.sql`) that does not exist in this repository; presumably lives in a private overlay and out of scope to verify here.


---

totalfrank (changes requested):
**Architecture-Gate Review (structural only — not a feature/business-logic review)**

## Verdict: REQUEST_CHANGES

## Gate 0 — Target branch

- Target: `dev` (the integration branch). Not a `REL*` branch, so the release-branch policy and Isolation Audit do not apply. Gate 0 **passes** — no re-target issue.

## Findings

1. **`src/backend/src/agentclaw/community/core/devices/services/baas_device_service.py:161-170`** (`BaasDeviceService.apply_device_for_new_bot`) — new core-layer code resolves the deployment environment itself instead of receiving it from a composition root:

   ```python
   from agentclaw.community.utils.env_utils import get_current_env
   ...
   env = get_current_env()
   ```

   This is `core/`, not a composition root (`di/`, `config/`, `bootstrap/`). Per arch.rules.md Rule 14 (Invariant, "Configuration Drives All Wiring" — *"core receives values, it does not go looking for them"*) and the violation catalog item 6 ("Environment and mode detection outside composition roots"), environment/mode detection must happen only in bootstrap/DI and be passed in, not re-derived ad hoc inside a service method.

   For transparency: `DeviceService.apply_device` (the sibling template method, unmodified by this PR, `device_service.py:577`) already does the exact same `env_utils.get_current_env()` call, and this pattern is pervasive across dozens of core files repo-wide. This new call does not introduce a materially new boundary crossing beyond what the immediate neighboring method already does. However, per this gate's explicit instruction — *"Pre-existing violations nearby are not a license to add more; new code that extends an existing violation is itself a violation"* and *"no exceptions … for whether 'the code already does this elsewhere'"* — this new call in new code is flagged rather than waived.

   **Compliant direction**: thread `env` down from the caller instead of re-deriving it. `DeviceServiceRouter.apply_device_for_new_bot` (`device_service_router.py`) already resolves the provider before delegating; it (or `BotService._provision_created_bot`, the ultimate caller) can resolve `env` once and pass it through `kwargs["env"]`, with `BaasDeviceService.apply_device_for_new_bot` reading it from `kwargs` the same way it reads `bot_type`/`engine`. If the project has decided the existing `get_current_env()`-in-core pattern is accepted technical debt, that acceptance should be recorded as a written waiver (arch.rules.md Governance Addendum) rather than silently re-derived call by call.

   No unwaived Invariant violation may pass per this gate's verdict rules, so this alone requires REQUEST_CHANGES regardless of its small size.

## Out-of-scope notes (not gating, recorded for completeness)

- Engine variability is handled correctly throughout this diff: it reuses the existing `resolve_baas_engine_bucket` registry/contract rather than introducing new `if engine_type == "..."` branching, and no new engine-name string literals or engine-named modules were introduced anywhere in the diff.
- The Plugin API boundary is used correctly: `VolumeMountSpec`/`Storage.upfs` are project-owned, SDK-free contract types built in `core/`; only the (out-of-tree, enterprise) ARCA SDK adapter would convert them to vendor `volume_mounts`, and all four in-repo community plugins (`_stub`, `aliyun_ack`, `local_docker`, `local_proc`) are covered by a new parametrized conformance test (`test_upfs_contract.py`) verifying each either honors or explicitly rejects UPFS before any side effect — this satisfies Rule 25 well.
- DI selection of `BotStoragePolicyService` is correctly gated on `deploy_runtime.runtime == DeployRuntime.MANAGED` inside the `ServiceBotModule` composition root (`@provider` method), not inside core — compliant with Rule 14.
- The `bot_storage/upfs_rollout` config schema (allow_all / rules / engine_bucket / user_tail_digit_max) is business rollout logic, not reviewed here per this gate's scope restriction.

---

totalfrank (approved):
## Architecture-Gate Review (automated)

**Verdict: APPROVE**

### Gate 0 — Target branch
- Target: `dev` (the integration branch). This is not a `REL*` release branch, so the release-branch gate and Isolation Audit do not apply — normal review path.
- Classification: new feature (UPFS storage support for personal/service-draft Bot creation), correctly targeting `dev`.

### Review procedure followed
Read `docs/arch/arch.rules.md` and `AGENTS.md`, then the full diff across all 44 changed files (BaaS `secbaas` storage/template/plugin layers and Backend `agentclaw` common-config/device/service-bot layers), checking each changed file's architectural role against the violation catalog (engine-string branching in core, hardcoded capability sets, engine-named modules, engine/deployment paths in core, hidden wire protocols, environment detection outside composition roots, hardcoded operational policy, transport leakage, copy-paste mirroring, implementation imports outside composition roots, and misplaced Plugin API vs. internal-contract modeling).

Notable structurally-sound patterns worth calling out:
- The new `TemplateStorageCapability` Service API method is defined as an explicit Protocol addition (`DeviceTemplateManageService.get_storage_capability`) with conformance coverage in `test_protocols.py`, not just a router.
- Storage-type variability at the actual vendor boundary is confined to the Plugin API layer: only the enterprise ARCA SDK adapter converts `Storage.upfs` into vendor `volume_mounts`; the three community sandbox plugins (`aliyun_ack`, `local_docker`, `local_proc`) explicitly reject UPFS via `NotImplementedError` before any side effect, and the stub records it. All four implementations are covered by one parametrized contract test (`test_upfs_contract.py`), satisfying Rule 25.
- Engine-bucket resolution for the new `apply_device_for_new_bot`/`initialize_bot_storage` call sites reuses the existing `BaasEngineBucketResolverRegistry` composition-root pattern (`resolve_baas_engine_bucket`) instead of introducing new engine-string branching — no new `if engine_type == "..."` logic was added anywhere in `core/`.
- Environment detection for the new BaaS-side capability check is injected from the composition root (`bootstrap/_core_services.py: deployment_env=providers.Callable(get_current_env)`), not read ad hoc inside the service.
- DI-level enablement of `BotStoragePolicyService` is gated on `deploy_runtime.runtime == DeployRuntime.MANAGED` inside `di/modules/service_bot_module.py` (a composition root), with a clean `None`-implies-NAS fallback in `BaasService.initialize_bot_storage` when the policy service isn't wired — this is the required default/fallback semantics, not an inline mode check in business logic.
- Contract changes (new `Storage.upfs` field, new `ArcaTemplateConfig.upfs_volume_id_*` fields, new DB table) are additive/optional with propagation analysis, deployment ordering, and rollback semantics documented in `docs/specs/2026-09-15-upfs-phase1/spec.md` (Rule 16).

One call site was investigated closely as a possible Rule 14 violation: `BaasDeviceService.apply_device_for_new_bot` (`src/backend/.../core/devices/services/baas_device_service.py:161,170`) calls `env_utils.get_current_env()` directly inside `core/`. On inspection this matches the pre-existing convention of the base class it extends (`DeviceService.apply_device` already does the same at `device_service.py:585`) and is used purely as a scoping value (passed through to template/policy lookups), not for implementation selection — the actual implementation selection in this PR (whether `BotStoragePolicyService` is wired at all) correctly happens in the composition root. Not treated as a new/blocking violation, but flagging it here since a future cleanup could thread `env` through the constructor like `BaasService.storage_env` already does, for consistency.

### Findings
None.

### Out-of-scope notes
- `secbaas/.../core/service/device_manage/_device_service.py` (2,574 lines) and `agentclaw/.../core/service_bot/services/baas_service.py` (3,961 lines) are both pre-existing, already far past the 1,000-line CI limit before this PR; this PR adds modestly to both rather than introducing the oversized-file condition. Not gated here per AGENTS.md's file-size checklist being a separate CI concern, but worth a follow-up split.
- NAS-vs-UPFS storage-type branching inside BaaS's `_build_arca_detail_config` is domain logic, not AI-coding-engine variability, so it wasn't evaluated against the engine-variability catalog; flagging only for awareness in case a future reviewer wants a business-logic pass.

---

totalfrank (approved):
## Architecture-Gate Review (automated)

**Verdict: APPROVE**

### Gate 0
- Target branch: `dev` (the integration branch) — not a `REL*` release branch, so the release-branch gate and Isolation Audit do not apply. Standard route, pass.
- Classification: new feature (UPFS storage phase 1), correctly targeted.

### Review scope
Read `docs/arch/arch.rules.md` (v3.1) and `AGENTS.md`, then reviewed the full diff: `src/baas/...` (device/template/sandbox-plugin layers) and `src/backend/...` (`common_config`, `devices`, `service_bot`, `di`, `plugin_api`), plus the new `docs/specs/2026-09-15-upfs-phase1/spec.md`.

For each changed file I checked role placement (contract / core / adapter / plugin / composition root / test) against the violation catalog (engine-string branching, hardcoded capability sets, engine-named modules, hardcoded filesystem/deployment facts, hidden wire protocols, env/mode detection outside composition roots, hardcoded operational policy, transport leakage, copy-paste mirroring, out-of-layer imports, plugin-vs-contract misuse).

**Findings: none.**

Notable points that support the verdict rather than undercut it:
- `Storage.upfs` / `VolumeMountSpec` / `TemplateStorageCapability` are project-owned Pydantic models with no vendor SDK dependency; the module docstring in `api/device_manage/_models.py` states core/public-API layers never import the Arca SDK, and the diff confirms it (no new `arca_sdk`-style import anywhere in `core/`).
- Storage-backend variability across the four community `ArcaSandboxPlugin` implementations (`_stub`, `aliyun_ack`, `local_docker`, `local_proc`) follows the required contract pattern: one Protocol (`spi/sandbox/arca/_protocols.py`), unsupported implementations reject `storage.type == "upfs"` before any side effect (documented as required behavior in the Protocol docstring), and `tests/unit/api/arca_sandbox/test_upfs_contract.py` is a genuine conformance test parametrized over every implementation (Rule 25).
- Implementation selection is done only in composition roots: `di/modules/service_bot_module.py` picks `bot_storage_service` based on `deploy_runtime.runtime == DeployRuntime.MANAGED`, not by name-checking inside core; `bootstrap/_core_services.py` injects `deployment_env` via `providers.Callable(get_current_env)` rather than having the service look it up itself.
- Rollout gating (`_should_rollout_upfs`) branches on `engine_bucket` values read from externally configured JSON (`bot_storage/upfs_rollout`), not on hardcoded `if engine_type == "..."` checks — this is configuration-driven behavior, not under-abstraction.
- All contract changes are additive/backward-compatible (`Storage.upfs` optional, `ArcaTemplateConfig.upfs_volume_id_*` optional, new `StorageType.UPFS` enum member, new DDL is a new table). `docs/specs/2026-09-15-upfs-phase1/spec.md` documents affected consumers, deployment order, and rollback caveats, satisfying the Rule 16 propagation-analysis requirement.
- `get_current_env()` is called directly inside `BaasDeviceService.apply_device_for_new_bot` (agentclaw core) rather than only in a composition root. I weighed this against catalog item 6, but this repo's own `test_env_import_regression.py` (BaaS side) documents that `env_utils` was deliberately relocated into `core/utils` specifically so core code may call it directly, and the same call shape (`env = get_current_env()`) is already pervasive (100+ call sites) throughout `agentclaw/community/core/**`. This PR does not introduce a new class of violation, only extends an already-accepted, repo-wide convention — not flagged as a finding.

### Out-of-scope notes
- `BotStoragePolicyProtocol._resolve_storage_policy` is declared with a leading underscore inside a public `Protocol` and is called cross-module from `baas_service.py`; a naming/style nit, not a layering violation (the Protocol boundary itself is respected). Not gated on.
- `baas_service.py` and BaaS's `_device_service.py` were already over the 1000-line CI cap before this PR (and `baas_service.py` is already on the `test_no_oversized_modules.py` allowlist); this PR adds to already-oversized files but doesn't newly trip the guard. Business/product logic (rollout rules, storage-policy semantics, logging content) was intentionally not reviewed per scope.

---

totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0 — Target branch

- Target: `dev` (integration branch, not `REL\d{8}`). Gate 0's release-branch restrictions do not apply; this is a normal feature landing on `dev`. **Pass** (not gated).

### Findings

**1. `src/backend/src/agentclaw/community/core/devices/services/baas_device_service.py:161,170` — new direct environment read inside `core/`**

```python
from agentclaw.community.utils.env_utils import get_current_env
...
env = get_current_env()
```
in the new `BaasDeviceService.apply_device_for_new_bot`.

- **Violated rule:** `arch.rules.md` Rule 14 — *Configuration Drives All Wiring* (Invariant), catalog item "6. Environment and mode detection outside composition roots" — "`get_current_env()` ... inside `core/`. Configuration drives wiring (Rule 14); core receives values, it does not go looking for them."
- **Why it breaks the architecture:** `apply_device_for_new_bot` is a new `core/devices/services/` method, not a composition root. It resolves the deployment environment itself instead of receiving it as a parameter, which is exactly the pattern Rule 14 forbids. It also breaks internal consistency with its own class: every other method in this file (`_do_allocate`, `_do_allocate_nas`, `_allocate_via_baas`, etc.) takes `env: str` as an explicit parameter supplied by the caller/router rather than calling an env-detection utility.
- **Compensating context:** this is not a novel pattern in this codebase — `core/devices/services/device_service.py` (the base class) already calls `env_utils.get_current_env()` directly in several places on `dev`, and `tests/community/architecture/test_no_is_local_mode_outside_allowlist.py` documents that "the devices cluster" is a known, still-migrating offender for Rule 14 env access (that test currently only polices `is_local_mode`/`_is_sqlite_mode`, not `get_current_env`). That does not make the new call compliant — per the review's own ground rules, code that extends an existing violation is itself a violation — but it explains why this is narrow/low-blast-radius rather than a novel design mistake, and elsewhere in this same PR (`bootstrap/_core_services.py`, `di/modules/service_bot_module.py`) the authors *did* correctly resolve `get_current_env()` in a composition root and inject it (`providers.Callable(get_current_env)`, `storage_env=env_utils.get_current_env()`), showing the compliant pattern was available and used deliberately in the BaaS-side and DI-side equivalents of this exact change.
- **Required fix:** thread `env` into `apply_device_for_new_bot` the same way the sibling methods in this file already do — either have the router/composition root resolve and pass `env` explicitly through the `apply_device_for_new_bot(**kwargs)` call chain, or (minimally) have this method delegate its env resolution to the same already-injected/composition-root-sourced value used elsewhere in the request, rather than importing `env_utils.get_current_env` inside a `core/` service method.

### Out-of-scope notes (not gated on)

- `bot_config_protocol.py`'s `BotStoragePolicyProtocol._resolve_storage_policy` is a leading-underscore method exposed as part of a public Protocol and called cross-class from `baas_service.py`. Naming/visibility smell, not a layering violation — flagging only as a note, not a finding.
- `baas_device_service.py` smuggles a typed `_PreparedBotCreation` result between `apply_device_for_new_bot` and `apply_device` via a magic `"_prepared_bot_creation"` key on the `template_config: dict` parameter, defensively popped at the entry boundary. It's in-process (same instance, same call stack), well tested, and never serialized/leaked — a business-logic implementation-detail nitpick, not a contract/wire-protocol issue.

Everything else checked out well for a PR of this size: `Storage.upfs` on the BaaS wire model is additive/optional with round-trip and legacy-NAS-shape tests; the `ArcaSandboxPlugin` protocol change ships a conformance test (`test_upfs_contract.py`) exercised against **all four** community implementations, including explicit-reject defaults for non-UPFS-capable ones; engine-bucket resolution reuses the existing `BaasEngineBucketResolverRegistry` composition root rather than adding new engine-string branching; the UPFS rollout switch (`_should_rollout_upfs`) is purely config-data-driven, not engine-literal branching; `deploy_runtime`-gated wiring of `BotStoragePolicyService` into `BaasService` happens correctly in the `di/modules/service_bot_module.py` composition root; and the new `common_config` module ships a context-boundary README per `docs/arch/context-boundary-format.md`.

---

totalfrank (changes requested):
## Architecture-Gate Review

**Verdict: REQUEST_CHANGES**

### Gate 0 — Target branch

- Target: `dev` (the integration branch), head `dev_upfs_phase1_simplify_20260915`.
- `dev` is not a `REL*` release branch, so the release-branch fix/isolation-exception analysis doesn't apply. Gate 0 passes trivially — this is ordinary feature work landing on the integration branch.

### Structural read

This is a large diff (44 files, ~2.9k lines across `src/baas` and `src/backend`) but it is deliberately careful about layering:

- **No engine-string branching added to core.** The UPFS rollout decision (`core/common_config/bot_config_service.py::_should_rollout_upfs`) matches an `engine` value against `engine_bucket` entries read from DB-stored JSON config (`bot_storage/upfs_rollout`), never against a hardcoded engine literal. The `engine` value itself is derived through the pre-existing `BaasEngineBucketResolverRegistry` / `resolve_baas_engine_bucket` extension point (`core/bot_management/engines/registry.py`, unmodified by this PR) rather than a new ad hoc mapping. This is the config-driven pattern Rule 14 and the engine-variability contract call for, not catalog items 1/2.
- **Storage-provider variability correctly uses the Plugin API**, because ARCA sandbox providers (`aliyun_ack`, `local_docker`, `local_proc`, `_stub`) are genuine external-platform boundaries, not engines. `Storage`/`VolumeMountSpec` (`src/baas/.../api/device_manage/_models.py`) stay pure Pydantic DTOs with no vendor SDK import; only the enterprise ARCA adapter (per the spec, unmodified — "无需修改") converts to SDK `volume_mounts`. Unsupported providers reject `storage.type == "upfs"` via `NotImplementedError` before any side effect (`_stub.py`, `aliyun_ack/_sandbox_plugin.py`, `local_docker/_sandbox_plugin.py`, `local_proc/_sandbox_plugin.py`), which is the correct place for that decision (Rule 13).
- **Implementation selection stays in the composition root.** `di/modules/service_bot_module.py` gates `bot_storage_service` on `deploy_runtime.runtime == DeployRuntime.MANAGED`, and `BaasService.__init__` depends on the `BotStoragePolicyProtocol` type, not the concrete class — matching Rule 5/14.
- New Service API / repository contracts (`BotCommonConfigServiceProtocol`, `BotStoragePolicyProtocol`, `BotCommonConfigRepositoryProtocol`) are declared as Protocols separate from their implementations, bound in `di/modules/common_config_module.py`, and documented with a context-boundary README — Rule 3/5/22 are respected for the new surface.

### Findings

1. **`src/backend/src/agentclaw/community/core/devices/device_service_protocol.py:9-12`** — `DeviceServiceProtocol` ("Service API for device management surface used by HTTP routes") is not updated for the new `apply_device_for_new_bot` entry point that this PR adds to `DeviceService`, `DeviceServiceRouter`, and `BaasDeviceService`, and that a different module now depends on: `core/bot_management/services/bot_service.py` (`BotService._provision_created_bot`) calls `service.apply_device_for_new_bot(...)` across the `bot_management` → `devices` module boundary in place of the previously-declared `apply_device`.
   - Violated rule: `arch.rules.md` Rule 4 ("Interfaces Mean What Their Specs Define Them to Mean," Invariant — "New behavior requires explicit contract evolution") and Rule 16 ("Changes Propagate," Invariant — a Service API change must declare affected consumers in the same change). Rule 3's required check ("each changed contract is classified as Service API or Plugin API") is likewise unmet for this addition.
   - Why it matters: `DeviceServiceProtocol` is the one place this module's Service API surface is supposed to be legible without reading every provider implementation; a cross-module consumer now depends on a capability the declared contract doesn't mention, so the Protocol no longer reflects what `bot_management` actually relies on from `devices`.
   - Fix direction: add `def apply_device_for_new_bot(self, *args: Any, **kwargs: Any) -> Any: ...` to `DeviceServiceProtocol` next to `apply_device`, in this PR, so the declared surface matches the new cross-module dependency.

### Out-of-scope notes

- `BaasDeviceService.apply_device_for_new_bot` smuggles a private `_PreparedBotCreation` dataclass through the `template_config` dict under a `_prepared_bot_creation` key to avoid re-parameterizing `apply_device` across providers. It's stripped before the BaaS payload is built and covered by a test asserting it never reaches the wire, so I'm not gating on it, but it's worth a real parameter in a follow-up rather than a magic dict key.
- `core/common_config` (pre-existing module, unrelated to this PR) is still absent from `BOUNDARY_SIGNIFICANT_MODULES` in `test_module_boundaries.py`; this PR's new README voluntarily satisfies Rule 22's intent anyway.


---

totalfrank (changes requested):
## Verdict: REQUEST_CHANGES

Architecture-gate review only (layering / engine-variability / contracts / composition-root discipline). Business logic (rollout semantics, storage-policy shape, BaaS wire details) was assumed correct and not reviewed.

## Gate 0

- **Target branch:** `dev` (integration branch). No `REL*` restrictions apply — normal review path, Isolation Audit not required.
- **Classification:** New feature (personal-Bot UPFS storage selection), correctly targeted at `dev`.
- **Pass/fail:** Pass on branch policy.

## Findings

### 1. `core/devices/services/baas_device_service.py:161,170` — new code reads deployment environment directly instead of receiving it from composition-root wiring

```python
def prepare_bot_storage(self, **kwargs: Any) -> dict[str, Any]:
    ...
    from agentclaw.community.utils.env_utils import get_current_env
    ...
    env = get_current_env()
```

**Violated rule:** `arch.rules.md` Rule 14 — *Configuration Drives All Wiring* (Invariant): "Implementation selection and object graph assembly must be driven by validated configuration and performed only in approved composition roots... repeated mode detection outside config or bootstrap is forbidden." Also the Structural Review Red Flags list: "environment access is scattered outside config or bootstrap." AGENTS.md is explicit: "Raw environment access belongs in configuration loading, bootstrap, composition roots, or tests." This is catalog item **#6 (Environment and mode detection outside composition roots)**.

**Why it breaks the architecture:** `prepare_bot_storage` is a brand-new entry point added by this PR, in `core/`, that resolves the deployment environment itself rather than being handed it. This isn't a case of "matching pre-existing local style" — the very same PR demonstrates the compliant pattern two call frames away: `service_bot_module.py` (a composition root) injects `storage_env=env_utils.get_current_env()` into `BaasService` at construction time, and `bootstrap/_core_services.py` on the BaaS side injects `deployment_env=providers.Callable(get_current_env)` into `DefaultDeviceTemplateService` for the exact same purpose (template storage-capability lookups). `BaasDeviceService` is itself DI-constructed (it already receives `baas_service: BaasService` via injection), so the same environment value is available to be threaded in through the constructor — instead a second, independent env-detection call site was added in `core/`. This duplicates the source of truth: `prepare_bot_storage` can now resolve a different `env` than `self._baas_service._storage_env` if they ever diverge (e.g. per-request environment overrides, testing with mismatched fixtures), which is exactly the "hidden coupling via globals... environment checks" failure Rule 15 warns about, and it re-opens the door this PR's own BaaS-side pattern just closed.

**Required fix:** Inject `env` into `BaasDeviceService` the same way `storage_env` is injected into `BaasService` (constructor parameter, wired from the composition root in `di/modules/`), or have `prepare_bot_storage` read `self._baas_service._storage_env`/an equivalent injected accessor instead of importing `env_utils.get_current_env` inline. `get_current_env()` should be called at most once, in a composition root or bootstrap, per the pattern this same PR already uses correctly on both the `storage_env` and BaaS `deployment_env` paths.

This is a narrow addition, not a blanket rewrite of the pre-existing (and separately out-of-scope) `get_current_env()` usages elsewhere in `core/devices/services/*` — only the two lines above are new.

## Out-of-scope notes

- Several touched files (`bot_service.py` 6.3k lines, `baas_service.py` 4.0k lines, `_device_service.py` 2.6k, `_facade.py` 2.3k) already exceed the AGENTS.md 1,000-line cap; this PR adds lines to them but did not push any file newly over the threshold and isn't the origin of that debt, so it's noted but not gated on.
- Rollout-percentage/whitelist semantics (`_should_rollout_upfs`), storage-policy JSON shape, and BaaS quota/volume resolution logic are business behavior and were not reviewed for correctness.
- Everything else checked out clean: `TemplateStorageCapability`/`VolumeMountSpec` are proper additive, backward-compatible Service API contracts with conformance tests (`test_upfs_contract.py` covers all ARCA sandbox plugins including default/rejection semantics); the `engine` string flowing through `BotStoragePolicyService.initialize` is used only as opaque data matched against config-driven rules (`rule.get("engine_bucket") != engine`), not as in-core branching; `resolve_baas_engine_bucket` (the one legitimate engine-name→bucket mapping point) is pre-existing and untouched; `_PreparedBotCreation` is an in-process, unserialized dataclass popped before the BaaS payload is built, not a hidden wire protocol; the new `ac_bot_common_config` table and `ArcaTemplateConfig.upfs_volume_id_*` fields are purely additive migrations/fields; and context-boundary READMEs were updated and already whitelist the new cross-module imports.


---

src/baas/src/secbaas/community/api/device_manage/_models.py
```diff
@@ -54,6 +114,27 @@ class Storage(BaseModel):
     quota: str | None = None
     permission: str | None = None
 
+    upfs: VolumeMountSpec | None = None
```
cassiuscai: Storage 已有的5个参数和和新增的upfs: VolumeMountSpec从语义上的区别是什么，之前方案是需要尽量复用复用已有的参数，让Plugins层做适配，core模型尽量保持不变

cassiuscai (changes requested):

src/baas/src/secbaas/community/adapters/web/routers/config_management/device_template_router.py
```diff
@@ -159,6 +160,25 @@ async def resolve_template(
         )
 
 
+@router.get(
+    "/{template_uuid}/storage-capability",
+    response_model=ApiResponse[TemplateStorageCapability],
+)
+@inject
+async def get_storage_capability(
```
cassiuscai: Baas has  api to get full device template info , this api is redundant

src/baas/src/secbaas/community/adapters/web/routers/config_management/device_template_router.py
```diff
@@ -27,6 +27,7 @@
     TemplateCreate,
     TemplateListResponse,
     TemplateStatus,
+    TemplateStorageCapability,
```
cassiuscai:
```
class DeviceTemplateResponse(BaseModel):
    """设备模板响应"""

    id: int
    template_id: int
    type: str
    template_uuid: str
    tenant: str
    name: str
    description: str | None
    status: str
    config: (
        ArcaTemplateConfig
        | SigmaTemplateConfig
        | LocalTemplateConfig
        | PoolabTemplateConfig
        | TeClawTemplateConfig
        | K8sTemplateConfig
        | DockerTemplateConfig
        | None
    )
    creator: str
    modifier: str
    gmt_create: datetime
    gmt_modified: datetime

    model_config = ConfigDict(from_attributes=True, populate_by_name=True)
``` this model include all the info

File: src/baas/src/secbaas/community/api/device_manage/_models.py [hunk 4]
```diff
@@ -43,8 +50,61 @@ class ResourceSpecification(BaseModel):
     disk: float | None = None
 
 
+class VolumeMountSpec(BaseModel):
```
cassiuscai: This class's property is overlapped with the exist Storage. should reuse the Storage model as possiable

src/baas/src/secbaas/community/api/template_manage/_models.py
```diff
@@ -55,6 +55,22 @@ class ArcaTemplateConfig(BaseModel):
     arca_template_id_prod: str | None = Field(
         None, description="ARCA 模板 ID (正式环境)"
     )
+    upfs_volume_id_pre: str | None = Field(
```
cassiuscai: add an upfs_volume_id as fallback when env not match pre/prod

src/backend/src/agentclaw/community/core/common_config/sql/2026_09_15_bot_common_config.sql
```diff
@@ -0,0 +1,14 @@
...
+    bot_id VARCHAR(128) NOT NULL,
+    entity_id VARCHAR(128) NOT NULL,
+    env VARCHAR(32) NOT NULL,
+    config_key VARCHAR(64) NOT NULL,
+    config_value TEXT NOT NULL COMMENT 'JSON configuration',
+    is_delete TINYINT NOT NULL DEFAULT 0,
+    gmt_create DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    gmt_modified DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+    PRIMARY KEY (id),
+    UNIQUE KEY uk_bot_config (bot_id, entity_id, env, config_key)
```
totalfrank, line 15: Does a single bot have multiple keys?

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,49 @@
...
+
+from dataclasses import dataclass
+from collections.abc import Callable
+from typing import Protocol, Any
+
+
+class BotCommonConfigServiceProtocol(Protocol):
+    def get_config(
+        self, *, bot_id: str, entity_id: str, env: str, config_key: str
+    ) -> Any: ...
```
totalfrank: What does this return? Can we be explicit about the type?

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def set_config(
+        self, *, bot_id: str, entity_id: str, env: str, config_key: str, value: Any
+    ) -> None: ...
+
+
+@dataclass(frozen=True)
+class StoragePolicy:
+    storage_type: str
```
totalfrank: Can we make this enum?

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,49 @@
...
+    source: str = ""
```
totalfrank, line 51: Documentation please. It would be better if you can provide some intuitive example here.

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,49 @@
...
+    ) -> StoragePolicy | None: ...
+
+    def initialize(
+        self,
+        *,
+        bot_id: str,
+        entity_id: str,
+        env: str,
+        engine: str,
+        user_id: str | None,
```
totalfrank: If we did not have any call site passes None here, let's get rid of the | None.

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
```diff
@@ -425,6 +427,8 @@ def __init__(
         deploy_composer: DeployConfigComposer,
         personal_bot_template_uuid: Optional[str] = None,
         theta_master_key_secret: str = "",
+        bot_storage_service: BotStoragePolicyProtocol | None = None,
```
totalfrank: This is always provided by DI right?, if so, let's get rid of | None.

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
```diff
@@ -425,6 +427,8 @@ def __init__(
...
+        storage_env: str = "",
```
totalfrank: Same here, if DI always provide it, let's not have default value.

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
```diff
@@ -463,6 +467,12 @@ def __init__(
         → token 退化为空字符串,BaaS LocalPaasService 无视 outbound rule)。
         None (仅旧测试构造时省略) → 走 layotto 老路径,仅 prod 可用。
         """
+        # None is a backward-compatible opt-out for non-storage-aware embeddings.
+        # The composition root supplies it only for storage-aware deployments.
+        if bot_storage_service is not None and not storage_env:
```
totalfrank: So you don't need this.

src/backend/src/agentclaw/community/core/devices/services/device_service.py
```diff
@@ -519,6 +519,14 @@ def _get_storage_mode(self, entity_id: str, bot_id: str) -> str:
     # Core business methods — template method pattern
     # =========================================================================
 
+    def prepare_bot_storage_policy(self, **kwargs: Any) -> dict[str, Any]:
```
totalfrank: This does not sound like a baas scope thing to me. It's a business concept. Let's have your own dedicated business component for this storage policy thing.

src/backend/src/agentclaw/community/core/devices/services/device_service_router.py
```diff
@@ -288,6 +288,19 @@ def _get_provider_for_new_device(
 
     # ============== DeviceService 接口代理实现 ==============
 
+    @override
+    def prepare_bot_storage_policy(self, **kwargs: Any) -> dict[str, Any]:
```
totalfrank: This is what I am talking about. Device router has nothing to do with this storage policy thing.

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
```diff
@@ -643,6 +649,78 @@ def _resolve_service_bot_resource_spec(
                 pass
         return ResourceSpecification(**kwargs)
 
+    def initialize_bot_storage(
```
totalfrank: Same, this is out of the baas scope. Have a dedicated business component for this.

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
```diff
@@ -714,6 +793,32 @@ def _build_create_bot_payload(
         # overriding the caller — but every caller that passes the flag
         # (bot_service restart, baas_device_service allocate) passes an empty
         # migration path, so the override never fired.
+        # Only personal instances consume this phase's policy.
+        # Published/caller instances use separate directories and build artifacts;
+        # sharing this builder must not enroll them in UPFS implicitly.
+        # Remove the personal guard when service Draft storage is enabled later.
+        uses_bot_storage_policy = (
```
totalfrank: Same here, this business logic should not live here.

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
```diff
@@ -765,6 +870,29 @@ def _build_create_bot_payload(
 
         # ``None`` ⇒ 该部署不挂 storage，BotDeployConfig.to_dict 直接不带该字段。
         storage = self._deploy_composer.build_storage(deploy_ctx)
+        if (
```
totalfrank: Same here.

totalfrank (changes requested):

src/backend/src/agentclaw/community/core/common_config/sql/2026_09_15_bot_common_config.sql
maicaisongxiaocong: 一个bot的一种config_key有一个，多种config_key有多个

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_config_composer.py
```diff
@@ -81,11 +82,17 @@ class BotDeployContext:
...
     mount_path: Optional[str] = None
     ext_info: Optional[Dict[str, Any]] = None
+    storage_type: StorageType | None = None
+    env: str = ""
 
 
 class DeployConfigComposer(abc.ABC):
     """Composes the runtime-specific fields of one create-bot ``deploy_config``."""
 
+    def prepare_context(self, ctx: BotDeployContext) -> BotDeployContext:
```
totalfrank, line 92: What does this do? it simply returns the input.

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_config_composer.py
```diff
@@ -81,11 +82,17 @@ class BotDeployContext:
     #: Caller-supplied extra NAS mount, mounted at the same path it names.
     mount_path: Optional[str] = None
     ext_info: Optional[Dict[str, Any]] = None
+    storage_type: StorageType | None = None
```
totalfrank, line 85: Is this always provided? If so, let's get rid of the "| None".

src/backend/src/agentclaw/community/core/service_bot/services/deploy/managed_composer.py
```diff
@@ -68,7 +69,9 @@ def __init__(
         storage_path: StoragePathProtocol,
         sandbox_registry: EngineSandboxRegistry,
         bot_repo: "BotRepository",
+        storage_policy: BotStoragePolicyProtocol | None = None,
```
totalfrank: Same here, this is always provided by DI right? Drop "| None" if that's the case.

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,97 @@
...
+    template_uuid: str
+    storage_type: "StorageType"
+
+
+class BotStoragePolicyProtocol(Protocol):
+    def prepare_bot_storage_policy(self, **kwargs: Any) -> dict[str, Any]:
+        """Choose provider/template and initialize policy before original allocation."""
+        ...
+
+    def resolve_deploy_context(self, ctx: "BotDeployContext") -> "BotDeployContext":
```
totalfrank: You don't need forward reference here. Let's get rid of them.

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,97 @@
...
+        """Resolve saved storage and its mount layout without running rollout."""
+        ...
+
+    def apply_to_storage(
+        self, storage: "Storage", ctx: "BotDeployContext"
```
totalfrank: You don't need forward reference here. Let's get rid of them.

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,97 @@
...
+    ) -> "Storage":
```
totalfrank: You don't need forward reference here. Let's get rid of them.

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,97 @@
...
+    source: str = ""
+
+
+@dataclass(frozen=True)
+class PreparedBotStoragePolicy:
+    """Request-local creation choice; never persisted or sent to BaaS."""
+
+    template_uid: str
+    template_uuid: str
+    storage_type: "StorageType"
```
totalfrank: You don't need forward reference here. Let's get rid of them.

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
```diff
@@ -0,0 +1,97 @@
...
+    """Saved storage choice, or an unsaved NAS fallback.
+
+    ``source`` records how the choice was made, not creation success:
+    ``rollout`` means new creation matched the UPFS rollout; ``manual`` denotes
+    an operator-supplied policy. Empty means unspecified (including NAS fallback
+    and older rows). Example persisted JSON:
+    ``{"storage_type": "upfs", "source": "rollout"}``.
+    """
+
+    storage_type: "StorageType"
```
totalfrank: You don't need forward reference here. Let's get rid of them.

src/baas/uv.lock
```diff
@@ -1718,20 +1718,22 @@ wheels = [
...
-    { url = "https://mirrors.aliyun.com/pypi/packages/47/88/4dd516068e1a3d6ab3c7c183288404cd424a9a02d585efbac226cb61ff2d/websockets-16.0-cp312-cp312-musllinux_1_2_x86_64.whl", hash = "sha256:485c49116d0af10ac698623c513c1cc01c9446c058a4e61e3bf6c19dff7335a2" },
-    { url = "https://mirrors.aliyun.com/pypi/packages/91/d6/7d4553ad4bf1c0421e1ebd4b18de5d9098383b5caa1d937b63df8d04b565/websockets-16.0-cp312-cp312-win32.whl", hash = "sha256:eaded469f5e5b7294e2bdca0ab06becb6756ea86894a47806456089298813c89" },
-    { url = "https://mirrors.aliyun.com/pypi/packages/c3/f0/f3a17365441ed1c27f850a80b2bc680a0fa9505d733fe152fdf5e98c1c0b/websockets-16.0-cp312-cp312-win_amd64.whl", hash = "sha256:5569417dc80977fc8c2d43a86f78e0a5a22fee17565d78621b6bb264a115d4ea" },
-    { url = "https://mirrors.aliyun.com/pypi/packages/6f/28/258ebab549c2bf3e64d2b0217b973467394a9cea8c42f70418ca2c5d0d2e/websockets-16.0-py3-none-any.whl", hash = "sha256:1637db62fad1dc833276dded54215f2c7fa46912301a24bd94d45d46a011ceec" },
+version = "15.0.1"
```
totalfrank, line 1721: If 16.0 works, let's not change it.

src/baas/pyproject.toml
```diff
@@ -10,7 +10,7 @@ dependencies = [
     "pydantic-settings>=2.14.2",
     "aiosqlite>=0.22.1",
     "aiohttp>=3.13.5",
-    "websockets>=15.0",
+    "websockets==15.0.1",
```
totalfrank, line 13: Checked offline, arca SDK requires this to be < 17. Can we do 16? Because currently uv.lock is pinned to 16.0. Let's make sure we have minimum production impact.

src/backend/src/agentclaw/community/api/baas_service.py
```diff
@@ -40,6 +40,10 @@
 class BaasServiceProtocol(Protocol):
     """BaaS-layer service: bot publish / restart / destroy / inspection."""
 
+    def get_device_template(self, template_uuid: str) -> Dict[str, Any]:
```
totalfrank, line 43: This is probably more leaning towards business logic. Let's move it to a business component.

totalfrank (commented):
One more round, PTAL.

src/backend/src/agentclaw/community/core/service_bot/services/deploy/managed_composer.py
maicaisongxiaocong: 这里可能为空

src/baas/src/secbaas/community/api/device_manage/_models.py
maicaisongxiaocong: VolumeMountSpec 已删除——Storage 模型保持不变，UPFS 参数转换全部收敛到内部 ARCA 插件实现（复用已有 type/quota 字段）。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已处理：返回值改为显式的 StoragePolicy / PreparedBotStoragePolicy 数据类（eab3cbdd6）。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已处理：StorageType 为 StrEnum（deploy_models）。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已处理：DI 总是提供，签名已收紧（eab3cbdd6；另见 527e307e0 composer 侧收紧）。

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
maicaisongxiaocong: 已处理：存储业务已解耦到独立的 BotStoragePolicyService，BaasService 不再持有这些参数（eab3cbdd6）。

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
maicaisongxiaocong: 已处理：存储业务已解耦到独立的 BotStoragePolicyService，BaasService 不再持有这些参数（eab3cbdd6）。

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
maicaisongxiaocong: 已处理：存储业务已解耦到独立的 BotStoragePolicyService，BaasService 不再持有这些参数（eab3cbdd6）。

src/backend/src/agentclaw/community/core/devices/services/device_service.py
maicaisongxiaocong: 已处理：存储策略业务已收敛到 common_config 下的 BotStoragePolicyService，device/baas 层不再涉及（eab3cbdd6）。

src/backend/src/agentclaw/community/core/devices/services/device_service_router.py
maicaisongxiaocong: 已处理：存储策略业务已收敛到 common_config 下的 BotStoragePolicyService，device/baas 层不再涉及（eab3cbdd6）。

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
maicaisongxiaocong: 已处理：存储策略业务已收敛到 common_config 下的 BotStoragePolicyService，device/baas 层不再涉及（eab3cbdd6）。

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
maicaisongxiaocong: 已处理：存储策略业务已收敛到 common_config 下的 BotStoragePolicyService，device/baas 层不再涉及（eab3cbdd6）。

src/backend/src/agentclaw/community/core/service_bot/services/baas_service.py
maicaisongxiaocong: 已处理：存储策略业务已收敛到 common_config 下的 BotStoragePolicyService，device/baas 层不再涉及（eab3cbdd6）。

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_config_composer.py
maicaisongxiaocong: 这是基类有意的恒等默认实现：无策略钩子的 runtime 原样使用调用方上下文；ManagedDeployConfigComposer 覆写它来应用 bot 已存的存储策略。已在 527e307e0 补充 docstring 说明。

src/backend/src/agentclaw/community/core/service_bot/services/deploy/managed_composer.py
maicaisongxiaocong: 已处理（527e307e0）：storage_policy 改为必传。DI（ServiceBotModule.deploy_config_composer）总是提供；测试构造点统一传入共享 NoopStoragePolicy。'可能为空'只出现在测试手动置 None 模拟旧路径的场景，已改用 NoopStoragePolicy 表达。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已处理（527e307e0）：注解去引号，PEP 563 延迟求值。TYPE_CHECKING 保留是因为运行时 import 会经由 service_bot 包 __init__ 成环（services/__init__ → bot_publish_service → di → bot_service → bot_config_protocol）。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已处理（527e307e0）：注解去引号，PEP 563 延迟求值。TYPE_CHECKING 保留是因为运行时 import 会经由 service_bot 包 __init__ 成环（services/__init__ → bot_publish_service → di → bot_service → bot_config_protocol）。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已处理（527e307e0）：注解去引号，PEP 563 延迟求值。TYPE_CHECKING 保留是因为运行时 import 会经由 service_bot 包 __init__ 成环（services/__init__ → bot_publish_service → di → bot_service → bot_config_protocol）。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已处理（527e307e0）：注解去引号，PEP 563 延迟求值。TYPE_CHECKING 保留是因为运行时 import 会经由 service_bot 包 __init__ 成环（services/__init__ → bot_publish_service → di → bot_service → bot_config_protocol）。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已处理（527e307e0）：注解去引号，PEP 563 延迟求值。TYPE_CHECKING 保留是因为运行时 import 会经由 service_bot 包 __init__ 成环（services/__init__ → bot_publish_service → di → bot_service → bot_config_protocol）。

src/backend/src/agentclaw/community/core/common_config/bot_config_protocol.py
maicaisongxiaocong: 已补充：StoragePolicy docstring 现包含持久化 JSON 示例 {"storage_type": "upfs", "source": "rollout"}。

src/backend/src/agentclaw/community/core/service_bot/services/deploy/deploy_config_composer.py
maicaisongxiaocong: None 是这里的正常状态：生产调用方都不传 storage_type，prepare_context 会从 bot 已存的策略解析（无策略回退 NAS）。这个字段是 prepare_context 到 build_storage 的传递槽，所以 | None 是必须的。

src/backend/src/agentclaw/community/api/baas_service.py
maicaisongxiaocong: 这里只是 BaaS 模板 API 的传输封装；业务判断（UPFS Volume 校验）在 BotStoragePolicyService，docstring 已注明 callers own readiness decisions。

cassiuscai (approved):

totalfrank (approved):
LGTM.

PR #2236, opened 2026-09-15
fix(bcs): make initial context sequence reads consistent

RaymondHX:
## Problem

Immediately after a new group Session is committed, queued `SessionContext` admission reads `bcs_group_sessions.current_msg_seq` through a standalone query. In the pre-production incident, that read used a different connection and returned no rows even though the Session transaction had committed, causing the initial manager context to fail with `canonical session is missing`. Group creation then continued after logging only a generic warning, leaving no initial context message or delivery to recover.

## Solution

- Read the initial Session sequence baseline through a database transaction; MySQL/OceanBase uses a locking `SELECT ... FOR UPDATE` so the read is served with primary/transactional consistency.
- Keep the existing compare-and-set message/delivery transaction, preserving atomic sequence allocation and admission.
- Preserve the underlying managed-delivery error in the system queue admission error instead of replacing it with a generic message.
- Add a regression test that simulates a stale standalone Session read while allowing transactional reads.
- Document the sequence-read consistency requirement in the message-store context boundary.

No Service API, Plugin API, schema, configuration, or wire-protocol contract changes are introduced.

## Validation

- `cargo test -q -p bcs-message-store -p bcs-message-flow` — passed, 455 tests.
- Targeted new-session sequence and system-admission error propagation tests — passed.
- The branch is based on the latest `dev`, including the independent Singlebox bootstrap fix from PR #2230.
- Commit and push used `--no-verify` as required by the Avernet publishing workflow; explicit tests above were run separately.

## Compatibility and risk

The change is internal to the SQL-backed delivery repository. SQLite retains its transactional read without MySQL-only locking syntax. The existing sequence CAS remains authoritative if another writer advances the Session between the baseline read and the atomic admission commit. Rollback is the single commit in this PR.


vzvince (approved):

questchaos (approved):
LGTM

PR #2243, opened 2026-09-16
fix(task-queue): add idempotency_key + postpone to eval_teardown, pre…

dnl-zhu:
## Problem
`eval_teardown` tasks in `ac_task_queue` accumulated **3,936+ rows**. The root cause chain:

1. `enqueue_eval_teardown` was called **without** an `idempotency_key`, so every call inserted a new row.
2. The Default-zone TTL renewal loop (`_renew_default_env_ttl`) called `enqueue_eval_teardown` every poll cycle (~5 min), each inserting a **duplicate row**: 20 active bindings × 288 polls/day = **5,760 rows/day**.
3. The worker greedily consumed rows, making BaaS HTTP calls per row. When the bot was already destroyed (`BOT_NOT_FOUND`), the task entered infinite `Retry`, compounding CPU load.

A secondary issue: `_poll_loop` in `EvalVersionSyncService` reads `poll_interval` once at startup and never re-reads DRM, so runtime config changes (e.g. `version_poll_interval_seconds`) do not take effect without a restart.

## Solution

**1. Add `idempotency_key` to `enqueue_eval_teardown`**

Each enqueue now carries `eval_teardown:{publish_id}:{bot_uuid}` as the dedup key, so duplicate submissions within the same `(env, app, task_type)` scope are suppressed.

**2. Introduce `postpone` — a renewal primitive for delayed one-shot tasks**

- `TaskQueueRepositoryProtocol.postpone_by_idempotency_key()` — UPDATE `run_at = now() + delay_seconds` on an existing PENDING task that holds the key; returns `False` if no live PENDING task found (caller should `enqueue` a new one).
- `TaskQueueService.postpone()` — facade that scopes by `env`/`app` and forwards to the repository.

This avoids inflating the queue: renewal simply slides the existing row's `run_at` forward instead of inserting a duplicate.

**3. Add `enqueue_or_postpone_eval_teardown()`**

Idempotent variant that first attempts `postpone`; falls back to `enqueue` only when no live task holds the key. This is the correct entry point for TTL renewal calls.

**4. Add bot existence short-circuit in `PublishEvalTeardownHandler`**

When `get_bot(bot_uuid)` returns `None` (bot already destroyed by TTL safety-net), return `Complete()` immediately — borrowed from the `quarantine.cleanup` pattern. This stops infinite `Retry` on absent bots.

**Rejected alternatives:**
- Cancel-and-requeue: more complex, risks losing a task mid-execution if cancellation races with the worker.
- Postpone-only (no enqueue fallback): would miss the initial enqueue when no task exists yet.

## Validation

- [x] `test_baas_service.py` — 139 tests pass
- [x] `TestSendMessageEvalHeaders` — 5 tests pass
- [x] Ruff format + lint clean
- [x] OCB side CI — pending ocb-public submodule pointer update (depends on gitlab mirror sync of this PR's branch)

## Compatibility and risk

- **Contract change**: `TaskQueueRepositoryProtocol` gains a new abstract method `postpone_by_idempotency_key()`. All implementations must provide it. The only production implementation (`TaskQueueRepository`) is updated in this PR.
- **No schema migration**: `postpone_by_idempotency_key` is a plain UPDATE on existing columns (`run_at`), no DDL required.
- **Rollback**: Revert this PR; `enqueue_eval_teardown` will still work without idempotency_key (same behavior as before, just without dedup). The OCB-side `enqueue_or_postpone_eval_teardown` import will fail gracefully (caught by `try/except` in `renew_default_env_ttl`).

## Related issues (optional)

- OCB companion PR: `fix/zhudan-eval-teardown-renewal` — updates `renew_default_env_ttl` to use `enqueue_or_postpone_eval_teardown`, changes default `version_poll_interval_seconds` from 300 to 3600


totalfrank (approved):
LGTM.

PR #2244, opened 2026-09-16
feat(engine): resume interrupted OpenClaw streams

evilcoda:
## Summary
- keep opt-in OpenClaw chat streams alive when the initiating browser WebSocket disconnects
- add bounded process-local replay through `chat.status` and `chat.resume` with a short-lived bearer resume ticket
- preserve legacy clients and all non-OpenClaw engine behavior
- document the transport contract, security boundary, capacity limits, and process-local limitations

## Scope
- Adapter/community engine only
- frontend integration is intentionally deferred
- no Backend, BaaS, or OpenClaw source changes

## Verification
- Engine CI: 2664 passed, 5 deselected
- total line coverage: 93.43%
- changed-line coverage: 93.17% (required 90%)
- pre-push secret scan and Python SAST passed

PR #2247, opened 2026-09-16
fix(openclaw): keep compaction retries streaming

evilcoda:
## Summary

- track `compaction_end(willRetry=true)` within each OpenClaw chat stream
- suppress exactly the corresponding intermediate `lifecycle end + chat final` pair so continuation events stay on the original Adapter stream
- preserve strict run correlation, real final/error/aborted behavior, and the path where a future upstream fix emits continuation without an intermediate terminal
- add timeout diagnostics and regression coverage for partial/empty output, repeated compaction, source run IDs, foreign runs, and hard stops

## Scope

Adapter-side compact-final handling only. This does not include the transport resume work in #2244 and does not modify OpenClaw source.

## Verification

- `src/engine/scripts/ci_test.sh --base origin/dev --head HEAD`
- 2674 passed, 5 deselected
- line coverage: 93.47%
- changed-line coverage: 100%


totalfrank (approved):
LGTM.

PR #2248, opened 2026-09-16
feat(backend): carry the enqueuing request's trace across the task queue

totalfrank:
## Problem

A client hits an open API endpoint, the work is persisted to `ac_task_queue`, and a worker runs it later — possibly minutes later, on a different pod. The request's trace ends at the enqueue. The execution logs under no trace at all, so nothing connects an API call to the work it caused: a support ticket quoting an `X-Trace-ID` reaches the enqueue and stops there.

## Solution

The row carries the trace across, and the worker re-establishes it around the whole of a task's execution.

**Capture at one point.** `TaskQueueService.enqueue` is the single choke point every adopter passes through, so the capture lives there and none of the ~25 call sites changed.

**Two columns for two audiences**, the same split as `idempotency_key` / `active_idempotency_key`:

- `trace_id` (`varchar(64)`) — the human value: what an operator greps for, what a ticket quotes.
- `trace_carrier` (`text`) — the tracer's own serialization, handed straight back to `TracerPlugin.trace_scope`, never read by the queue.

Storing only the id was the rejected alternative. It would put the corp tracer's propagation headers into neutral code and drop everything the id is not — under SofaTracer the carrier also holds the rpc id and the penetration attributes, and `inject`/`extract` keep that format the tracer's business. Both columns sit outside every index and every predicate, so correlation cannot change which row exists or which task a keyed enqueue joins.

**A new pair on the `TracerPlugin` seam** — `export_trace_carrier` and `trace_scope` — because `opentracing` / `sofa_tracer` cannot be imported from `agentclaw.community`. The community and noop impls are in this PR; the corp impl (`extract` → `start_active_span(SPAN_CODE_MSG_SUB, child_of=…)`) lands in its own overlay, and the queue works without it — those rows simply carry no trace.

**Correlation never costs a caller its work.** A tracer that fails to export is logged and the task is enqueued uncorrelated; a tracer that fails to restore is logged and the task runs untraced; an unreadable carrier degrades to no trace rather than failing the claim batch that projects it; an overlong trace id is truncated rather than rejected — the opposite of an idempotency key, which raises, because a truncated key silently merges two dedup scopes and can drop work while a truncated trace id is only a less useful log line.

Two behaviours worth calling out:

- A **keyed duplicate keeps the original's trace**, since it inserts no row. The work runs once, under the trace that caused it to exist; `enqueue` logs the join with both trace ids so the joining request is still followable.
- A handler enqueuing follow-up work does so *inside* its restored scope, so a chain of tasks stays correlated to the request that started it. That follows from the design rather than being wired.

## Validation

Run with `.venv/bin/python -m pytest`, on the current head (`dev` merged in at `372542dd`). The lockfile pins a mirror this environment cannot reach, so dependencies were resolved from PyPI for the run; `uv.lock` carries no change of this branch's.

- `tests/community/core/task_queue/` — 60 passed, including the new `test_trace_carriage.py`: the round trip, concurrent tasks each keeping their own trace, absence being normal, a broken tracer on either side, an unreadable carrier, truncation, and the keyed-duplicate case.
- `tests/community/contracts/test_tracer.py` — 8 passed. Extended with a second Rule 25 consumer asserting the export/restore symmetry against every deployable impl, with the carrier round-tripped through JSON as persistence requires.
- `tests/community/plugins/{community,local}/test_tracer.py` — 14 passed.
- `tests/community/di` + `tests/community/api` — 1090 passed.
- Post-merge sweep over `core/task_queue`, `contracts`, `plugins/{community,local}`, `repository`, `core/skill_center` and `architecture` — **3301 passed, 13 skipped**. The architecture suite required declaring the new `plugin_api.tracer` dependency in the module's README manifest.
- `ruff check` clean on every changed file.

A test drove out a real defect during development: a raise from `trace_scope` escaped `_run_guarded`, aborted the `gather` running the batch, and left every task in it RUNNING until lease expiry — a tracing fault becoming a stalled queue. `_restored_trace` now guards the tracer's enter/exit (never the body), with `test_a_tracer_that_cannot_restore_does_not_stall_the_batch` pinning it.

`test_task_queue_schema_contract.py` also gained a column-drift guard alongside its existing index one, since a column the ORM declares and the DDL lacks fails the whole queue with "unknown column" on a freshly provisioned database — for every task, not only those using the new field.

Not run here: the corp profile, which needs the SofaTracer SDK. The mechanism was validated directly against it on a prod host — `inject`/`extract` round-trips the trace id through the `HTTP_HEADERS` carrier, `start_active_span(SPAN_CODE_MSG_SUB, child_of=ctx)` restores it and finishes cleanly, and `antmcp.utils.tracer.get_trace_id` (which the sofapy logger uses for `record.traceid`) reads the same ContextVar `ContextVarsScopeManager.activate` sets, so restored log lines carry the id.

## Compatibility and risk

**Schema change, additive and nullable.** `ALTER TABLE` is at the foot of `sql/2026_09_06_task_queue.sql`. Unlike the earlier `app` / idempotency columns this one is safe to apply with the fleet up: a pod on the previous release inserts rows leaving both NULL, and the worker reads a NULL carrier as "no trace to restore". Applying it before deploying is still the right order, since the ORM projects every column on every `SELECT`.

No index is added. Indexing `trace_id` would add write cost to the busiest INSERT the component has for a lookup nobody does on the hot path; the DDL notes the index to add if that changes.

**`TracerPlugin` gained two methods**, so any out-of-tree impl must add them — including the corp `SofaTracer`, which otherwise fails at enqueue (caught and logged as "enqueueing without correlation", not a failed enqueue) and at claim (caught and logged, task runs untraced). The queue keeps working either way.

**One thing to decide deliberately:** penetration attributes are business-carried values (stress-test flags, gray-routing tags). Persisting the carrier means they outlive the request in the database and are replayed when the task runs. That is what a faithful restore requires, but it is a choice, not a free side effect.

Rollback is the revert; the columns can stay on the table harmlessly.

https://claude.ai/code/session_01AYNTrsTJhcixWk2Wdxno3q


src/backend/src/agentclaw/community/core/task_queue/repository/models.py
```diff
@@ -71,6 +75,31 @@ def _binary_string(length: int):
 TaskTypeString = _binary_string(100)
 
 
+def _decode_trace_carrier(raw: str | None, task_id: int | None) -> dict | None:
```
totalfrank-personal: Task id always exists right? If so, let's not do | None.

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
```diff
@@ -433,6 +435,16 @@ def _insert(
                 idempotency_key=idempotency_key,
                 # Mirrors the key while the task is live; nulled on terminal.
                 active_idempotency_key=idempotency_key,
+                trace_id=trace_id,
+                # Serialized here rather than by the caller so the column's
+                # encoding stays this layer's business, exactly like payload.
+                # ``ensure_ascii=False`` matches payload's spelling; a carrier is
+                # ASCII in practice, but the two should not differ by accident.
+                trace_carrier=(
+                    json.dumps(trace_carrier, ensure_ascii=False)
```
totalfrank: Non-blocking: this `json.dumps(trace_carrier, ...)` is unguarded, unlike every other seam this feature adds. If `TracerPlugin.export_trace_carrier()` returns a dict containing something JSON can't serialize (a datetime, a custom SDK type, a circular reference — plausible for the not-yet-landed corp `SofaTracer` overlay, whose carrier is documented to bundle the rpc id and "penetration attributes"), `json.dumps` raises here, inside `orm_session()` (which then rolls back), and propagates all the way out of `TaskQueueService.enqueue()` uncaught — failing the caller's actual task enqueue, not just the correlation.

That's the exact failure mode this PR goes out of its way to prevent everywhere else: `_capture_trace()` swallows any exception from `export_trace_carrier()` itself and truncates rather than rejects an overlong `trace_id`, and `_decode_trace_carrier()` on the read side degrades a malformed carrier to `None` instead of failing `claim_batch`. The write-side serialization has no equivalent guard, so a carrier that is well-formed enough not to raise on export but not JSON-serializable slips through `_capture_trace` and only blows up here.

Suggested fix — degrade like the read path does, rather than let it fail the insert:
```python
trace_carrier_json = None
if trace_carrier:
    try:
        trace_carrier_json = json.dumps(trace_carrier, ensure_ascii=False)
    except (TypeError, ValueError):
        logger.warning("[task_queue.enqueue] trace_carrier not JSON-serializable; storing no trace")
row = self.Model(
    ...
    trace_carrier=trace_carrier_json,
)
```

---

totalfrank (commented):
Reviewed commits `19129e6` and `372542d` (merge of `dev`), diffed against merge-base `e12a495a`. Covered: `task_queue.py` repository, `task_queue_service.py`, `worker.py`, `models.py`, the `TracerPlugin` protocol and its community/local/noop impls, the SQL DDL, and the new/updated tests.

Checked: correctness of the enqueue/claim/insert trace-carriage path, the `_restored_trace` context-manager exception handling (enter-guard vs. body-exception vs. exit-guard, suppression semantics), asyncio context-copy correctness for the per-task ContextVar isolation under `gather`/`to_thread`, DI wiring for the new `TracerPlugin` constructor dependency (no non-DI call sites left stale), the new schema-drift conformance test's regex against the DDL, and architecture boundaries (Plugin API contract kept separate from implementations, no transport/framework leakage into core).

One non-blocking finding posted inline: the trace-carrier JSON serialization on the write path (`task_queue.py:444`) has no failure guard, unlike every symmetric guard this PR adds elsewhere (export failure, restore failure, overlong id, unreadable carrier on read) — a non-JSON-serializable carrier would fail the whole enqueue rather than just drop correlation.

---

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
totalfrank: Confirmed and fixed in `d6ff22c8`.

You're right that this was the one asymmetric seam. I reproduced it before fixing: a carrier of `{"trace_id": ..., "exported_at": datetime.now()}` raises `TypeError: Object of type datetime is not JSON serializable` from inside `orm_session()`, which rolls the insert back and propagates out of `TaskQueueService.enqueue` — the caller loses the work, not just the correlation.

Worth naming why `_capture_trace` doesn't already cover it: it guards the *call* to `export_trace_carrier()`, not the serializability of what that call returns. A carrier well-formed enough to export but not to encode passes straight through it, which is exactly the gap you identified.

Fixed with `_encode_trace_carrier`, a write-side mirror of `models._decode_trace_carrier`, degrading to `NULL` with a warning. Two deliberate differences from the suggested patch:

- It catches `(TypeError, ValueError)` rather than `TypeError` alone, matching `_decode_trace_carrier`'s except clause — a circular reference raises `ValueError` ("Circular reference detected"), and that was one of your listed cases.
- It sits at module level next to the other helpers rather than inline in `_insert`, so the guard protects every caller of the repository rather than only the one the service happens to use.

The `payload` dump a few lines away stays deliberately unguarded, and the docstring now says so: a task that cannot describe its *work* must not be enqueued, while a task that cannot describe its *trace* should simply run uncorrelated.

Regression test added — `test_an_unserializable_carrier_does_not_fail_the_enqueue` in `test_trace_carriage.py`, the write-side mirror of the existing unreadable-carrier test. It asserts the id still lands, only the carrier is dropped, and the task runs to `SUCCEEDED`. Verified it fails with the original `TypeError` when the guard is removed.

---

totalfrank:
**Review round 1 — one finding, fixed.** Reviewed `372542dd`, fix pushed as `d6ff22c8`.

**Fixed:** unguarded `json.dumps(trace_carrier)` on the write path ([`task_queue.py:444`](https://github.com/inclusionAI/Avernet/pull/2248#discussion_r4023199592)). It ran inside `orm_session()`, so a carrier that exports cleanly but cannot be JSON-encoded failed the caller's enqueue instead of just dropping correlation — the one seam in this feature missing the guard every other seam has. Now `_encode_trace_carrier`, a write-side mirror of `_decode_trace_carrier`, degrading to `NULL` with a warning. Regression test `test_an_unserializable_carrier_does_not_fail_the_enqueue`, verified to reproduce the original `TypeError` when the guard is removed.

Nothing disputed, nothing deferred.

Validation on `d6ff22c8`: `tests/community/core/task_queue` + `tests/community/repository` + `tests/community/contracts` — **1430 passed**; `ruff check` clean on the changed files.

---

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
totalfrank: Verified fixed in `d6ff22c8`: `_encode_trace_carrier` now guards the write-side `json.dumps`, catches `(TypeError, ValueError)` exactly mirroring `_decode_trace_carrier`, and degrades to `NULL` with a logged warning instead of propagating out of `orm_session()`. Ran the reproduction locally — `test_an_unserializable_carrier_does_not_fail_the_enqueue` and the rest of `test_trace_carriage.py` (11 tests) plus `test_tracer.py` (8 tests) pass against this commit; confirmed the new test fails with the original `TypeError` when the guard is removed. No other write path inserts `trace_carrier` (only `TaskQueueRepository._insert` does), so the guard is complete. Resolving.

---

totalfrank (commented):
Follow-up round. Reviewed the new commit since the last review: `d6ff22c8` (fix for the one open finding; `372542dd` was already reviewed).

**Findings checklist:**
- `task_queue.py:444` — trace-carrier JSON serialization unguarded on the write path → **fixed** in `d6ff22c8`. `_encode_trace_carrier` mirrors `_decode_trace_carrier`'s `(TypeError, ValueError)` catch, degrades to `NULL` with a logged warning, matches the module's other guarded seams (`_capture_trace`'s `exc_info=True` convention). Verified by running `test_trace_carriage.py` (11 passed, including the new `test_an_unserializable_carrier_does_not_fail_the_enqueue`) and `test_tracer.py` (8 passed) against this commit, and confirmed the new test fails with the original `TypeError` when the guard is reverted. Thread resolved.

**Part B — review of `d6ff22c8` itself:** no new issues. The change is narrowly scoped to the one gap identified, has a symmetric regression test, doesn't touch any other write path (`_insert` is the only caller that persists `trace_carrier`), and doesn't introduce new callers or signature changes needing updates elsewhere.

**Open findings: 0.** PR is clean as of `d6ff22c8`.

---

totalfrank:
**Review round 2 — no new review findings; one CI failure, fixed.** Reviewed `d6ff22c8`, fix pushed as `2b7942c3`.

**Review:** 0 open findings. The round-1 finding was independently re-verified on `d6ff22c8` and its thread resolved.

**CI:** `Backend unit tests` went red on `d6ff22c8`, and it was mine — not flake:

```
FAILED tests/community/architecture/test_no_oversized_modules.py::test_no_new_oversized_files
  core/repository/implementations/platform/task_queue.py: 1005 lines (cap 1000)
```

The guarded encoder from the round-1 fix pushed that file five lines over the single-responsibility cap. My own miss: I ran `task_queue` + `repository` + `contracts` before that push but not `architecture`.

Trimming the docstring would have bought the five lines back while ignoring what the guard is for, so I moved the pair instead. `encode_trace_carrier` and `decode_trace_carrier` are mirror halves of one concern that had been split across the repository implementation and the ORM model — and that split is arguably what let the round-1 bug in, since their shared rule (*a trace is a diagnostic, so neither direction may fail the operation it is part of*) was easy to hold on the read side and forget on the write side. They now live together in `core/task_queue/repository/trace_carrier.py` with that rule stated once.

No behaviour change. The repository file drops to **972** lines and `models.py` to **318**; the allowlist is untouched, since neither file was ever on it.

Validation on `2b7942c3`: `architecture` + `core/task_queue` + `repository` + `contracts` + `plugins/{community,local}` — **2105 passed**, with `test_no_oversized_modules` explicitly green. `ruff check` clean.

---

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
```diff
@@ -433,6 +438,11 @@ def _insert(
                 idempotency_key=idempotency_key,
                 # Mirrors the key while the task is live; nulled on terminal.
                 active_idempotency_key=idempotency_key,
+                trace_id=trace_id,
+                # Serialized here rather than by the caller so the column's
+                # encoding stays this layer's business, exactly like payload —
+                # but guarded, unlike payload. See _encode_trace_carrier.
```
totalfrank, line 444: nit: stale reference — this comment still says "See `_encode_trace_carrier`", the old private name before this commit moved and renamed it to `encode_trace_carrier` in `trace_carrier.py`. Worth a one-word fix so it doesn't send a future reader looking for a symbol that no longer exists here.

---

totalfrank (commented):
Follow-up round. Reviewed the one new commit since the last review: `2b7942c3` (extraction of the trace-carrier codec into `core/task_queue/repository/trace_carrier.py`, driven by the 1000-line module-size cap).

**Findings checklist:**
- `task_queue.py:444` — trace-carrier JSON serialization unguarded on the write path → **fixed** in `d6ff22c8`, verified in the prior round. Thread resolved.
- New nit this round: stale `_encode_trace_carrier` reference in a comment at `task_queue.py:444`, left over from the rename to `encode_trace_carrier` — posted inline.

**Part B — review of `2b7942c3`:** Confirmed this is what it claims to be, a pure move with no behavior change:
- `encode_trace_carrier`/`decode_trace_carrier` in the new module are byte-for-byte the same bodies as the old `_encode_trace_carrier`/`_decode_trace_carrier`, including the `(TypeError, ValueError)` catch and degrade-to-`None` semantics on both sides.
- Both call sites (`task_queue.py:445` write, `models.py` `to_record`) were updated to the new imports; no other reference to the old private names remains anywhere in `src/` or `tests/` (checked).
- `json`/`logger` imports left in `task_queue.py` are still used elsewhere in that file (payload dump, other log lines); `models.py` correctly dropped its now-unused `get_logger` import and `logger` binding entirely — no dead imports either side.
- Line counts roughly match the commit message's claim (972 / 315 / 80 vs. claimed 972 / 318 / 80).

No new issues beyond the one nit.

**Open findings: 0 blocking / non-blocking. 1 nit posted this round.**

PR is otherwise clean as of `2b7942c3`.

---

FreddieSun (approved):
LGTM

src/backend/src/agentclaw/community/core/repository/implementations/platform/task_queue.py
totalfrank: Correct, and my leftover. Not pushing for it on its own: the PR is green and now carries a human approval, and a push to fix a comment's wording would dismiss that approval — a poor trade for one word. It rides the next code push if one comes; otherwise the surrounding comment still points at the right place (`encode_trace_carrier`, one line below, in `trace_carrier.py`).

Resolving.

---

totalfrank:
**Review round 3 — no new bugs flagged. Loop closed.** Reviewed `2b7942c3`; CI green on that SHA (10/10).

Round 3 raised one nit (stale `_encode_trace_carrier` reference in a code comment) and no blocking or non-blocking findings. Replied and resolved: it doesn't justify a push of its own now that the PR is green and approved, since that push would dismiss the approval to fix one word. It rides the next code push if one comes.

**Correction to my round-2 comment and the `2b7942c3` commit message:** I reported `models.py` at 318 lines. It is **315** — I measured before removing the now-unused `get_logger` import and never re-measured. The reviewer caught it. Correct counts: `task_queue.py` **972**, `models.py` **315**, `trace_carrier.py` **80**. The 1000-line cap conclusion is unaffected.

**Loop summary — 3 rounds:**

| Round | Head | Outcome |
|---|---|---|
| 1 | `372542dd` | 1 finding: unguarded write-side `json.dumps(trace_carrier)` could fail a caller's enqueue. Fixed + regression test. |
| 2 | `d6ff22c8` | Review clean. CI red — the round-1 fix tripped the 1000-line module cap at 1005 lines. Fixed by extracting the codec to `trace_carrier.py`. |
| 3 | `2b7942c3` | Review clean, 1 nit (resolved). CI 10/10 green. |

Nothing disputed across the loop; both findings were real and both are fixed. No finding was raised more than once.

Still open, and for a human to decide rather than me — both already flagged in the PR description:
1. Persisting sofa **penetration attributes** means they outlive the request in the database and are replayed when the task runs.
2. The **`ALTER TABLE`** wants applying before the release that contains this ships.

---

PR #2250, opened 2026-09-16
feat(clawweb): restrict agent monitoring to ClawInsight administrators

2001wjh:
Keep governance visible while gating monitoring navigation and browser APIs. Complete the public role/config/auth chain and add permission regression coverage. Document the paired OCB rollout and existing internal reporting boundary.

## Problem

Agent monitoring and self-healing is currently accessible without a dedicated role, although it is intended for an allowlisted business group. Restricting the entire ClawInsight effect center would also hide unrelated governance functionality.

The requirement is to grant monitoring access through a **ClawInsight 管理员** role in the existing permission-management page, without introducing a separate ACL system or changing other modules' permissions.

## Solution

- Add the independent `claw_insight_admin` role to the public role repository, configuration loading/seeding (`auth.claw_insight_admins`), and shared server/web types. Reuse the existing `clawweb_admin_users` table.
- Resolve `isClawInsightAdmin` in the public authentication middleware and return it from `/api/auth/me`. Generic administrators, super administrators, ClawEvolve administrators, and Dev Local do not automatically inherit monitoring access.
- Default the client flag to `false` and merge the server-provided value into the current user.
- Hide only the **Agent 监控自愈** navigation item for unauthorized users. A direct `?module=monitoring` link displays Agent governance instead and does not mount the monitoring component. The effect center and existing governance permissions remain unchanged.
- Guard browser reads under `/api/insight/v1/monitoring/*`: the bot list, status, and diagnosis endpoints return `403` when the role flag is false or absent and the monitoring runtime is assembled.
- Add public-auth, monitoring-HTTP, and navigation regression coverage, plus a spec covering both repositories.

The companion OCB change adds **ClawInsight 管理员** to **其他功能 → 权限管理**, where existing super administrators maintain membership through the existing APIs. It also makes OCB auth responses use the current request's monitoring permission rather than cached identity permissions. Those OCB changes are not part of this PR.

## Validation

Local validation of the paired Avernet/OCB worktrees completed on September 16, 2026:

- **105 passed**: monitoring HTTP, contracts, repository, and mock scenarios.
- **20 passed**: effect-center navigation and MonitoringPanel components.
- **29 passed**: public authentication (12), OCB authentication (9), and cross-repository integration (8), including role management and grant/revocation behavior.
- **154 distinct tests passed in total**. The five new OCB regression cases also passed when rerun using OCB's own package test configuration.
- TypeScript compilation passed for shared, clawevolve, clawinsight, and OCB access-admin, together with focused auth/UI/test type checks.
- Diff whitespace checks and the Avernet plaintext-secret scan passed.

Validation used Node v26.7.0, reused local third-party dependencies, rebuilt workspace packages, mocked login providers, and the existing `MONITORING_TEST_SQLITE_DRIVER=node` adapter. This does **not** validate production better-sqlite3/MySQL/ZDAS drivers or real SSO. UI tests passed with existing React `act` warnings. Full monorepo CI, deployed-browser E2E, and production gateway isolation were not validated locally.

The local feature commit has since changed from `59a6672d8` to `0e36bf1a3`; `git range-diff` confirms that its patch is unchanged. The results above are from the earlier validation run, not a full rerun on the updated base.

## Compatibility and risk

- No new database table, schema migration, permission-management API, or ACL framework. The new auth response field is additive; the new client hides monitoring if the field is missing.
- Preserve the current monitoring auto-discovery, configuration-free startup, and schema-initialization behavior. Do not replay the old monitoring branch history.
- Internal reports under `/api/insight/v1/internal/monitoring/*` retain their existing protocol and are not subject to the browser-user role check. No report-token check is introduced; existing deployment-side network isolation remains required.
- Authorization changes apply on subsequent server requests. An already-open page updates its navigation after refresh or identity refetch; no live permission-push mechanism is added. Existing configuration fallback semantics are retained.
- **Rollout order:** merge Avernet into `dev`, wait for the internal Avernet Mirror's `dev` to synchronize, then validate and merge the companion OCB branch `dev_clawinsight_admin_access`. OCB CI uses Mirror `dev` as its public dependency baseline. Rebuild the public packages and OCB application rather than using stale `dist` output.
- Roll back both sides together if necessary. Reverting the monitoring guard restores the previous unrestricted behavior, so rollback must not be treated as preserving the new access restriction.

## Spec

`src/evolverun/docs/clawweb/specs/clawinsight-monitoring-admin-access.md`

Defines the role, administrator workflow, public/OCB responsibilities, browser/API boundaries, rollout order, and acceptance criteria.


3h-william (approved):
ok

PR #2249, opened 2026-09-16
fix(backend): bound desktop skill recovery retries

FreddieSun:
## Summary
- remove the periodic Desktop Skill recovery sweeper and retain event-driven recovery only
- cap each durable recovery task at 600 seconds and classify engines without either Mapping generation as non-retryable
- preserve Legacy fallback for standard route-missing/structured unsupported responses and keep unexpected failures diagnosable
- document the final contract and supersede the historical sweep decisions

## Validation
- `uv run pytest -q tests/community` — 18,727 passed, 43 skipped
- focused recovery/runtime/config/DI suite — 153 passed, 10 skipped
- CI coverage regression tests — 77 passed
- changed-file Python SAST — passed
- Ruff and `git diff --check` — passed
- Standards/Spec double review — no remaining P0-P2

## Cross-repository integration
A companion OCB PR will update the Corp transport log classification and pin `ocb-public` to commit `578adeb13c8a2527b0e77aad790f655d7fde1631`.

PR #2251, opened 2026-09-16
fix(backend): backport bounded desktop skill recovery retries

FreddieSun:
## Summary
Backport #2249 to `REL20260917` as an exact one-commit patch.

- remove the periodic Desktop Skill recovery sweeper
- cap event-driven recovery tasks at 600 seconds
- classify Engines without either Mapping generation as non-retryable
- preserve Legacy fallback and expected compatibility logging

## Provenance
- source squash commit on `dev`: `4dca536448257b1e483de0639b59fd22995c8f16`
- release backport commit: `0080883f5`
- `git range-diff` reports exact patch equivalence
- only intervening pre-source commit is unrelated #2247

## Validation
- focused runtime/recovery/config/DI suite: 156 passed, 10 skipped
- Ruff: passed
- changed-file Python SAST: passed
- `git diff --check`: passed


totalfrank (approved):
LGTM.

PR #2252, opened 2026-09-16
fix(bcs): stop creating sessions when listing empty groups

vzvince:
## Problem

Listing sessions for a sessionless group currently creates a legacy
`{group_id}:00000000` session and sends its initial GroupContext messages.
This historical migration fallback makes a read request mutate state and
prevents a group from remaining intentionally sessionless. It also conflicts
with the requirement to support creating groups without an initial session.

## Solution

- Remove the legacy session existence check, automatic creation/reactivation,
  and associated initial-context notification from
  `GET /groups/{id}/sessions`.
- Return `200 OK` with `items: []` and `group_id` for sessionless groups,
  including repeated and participant-scoped requests. Preserve existing
  visibility rules, filters, pagination, and collection metadata.
- Add contract coverage for Bot members, Human owners, repeated empty-group
  queries, and listing existing zero-suffix sessions without recreating them.
- Update CLI help, the session reference, and the changelog to document the
  read-only behavior.

## Validation

The following tests passed before this commit/push step:

| Coverage | Result |
| --- | --- |
| HTTP session-list contracts | 5 passed |
| HTTP session creation/DTO, collection, membership, group, and CLI integration contracts | 73 passed |
| CLI session command tests (`test_session_`) | 14 passed |
| **Total** | **92 passed** |

The two new empty-group expectations were first observed failing against the
old implementation and passed after the fallback was removed. Independent
review found no functional issues, and `git diff --check` passed during
implementation.

Full-workspace and Singlebox coverage gates were not run; validation was
scoped to the changed behavior and adjacent contracts. Tests were not rerun
for publication, and commit/push verification hooks were skipped with
`--no-verify` as requested.

## Compatibility and risk

- Historical sessionless groups now remain empty when queried. Callers that
  need a session must explicitly create one with `session create` or
  `POST /groups/{id}/sessions`.
- Existing zero-suffix sessions, their history and WebSocket compatibility,
  explicit session creation, and normal group creation's initial session are
  unchanged. This PR does not add a CLI option to skip that initial session.
- No schema migration or production data change is required. Reverting this
  change restores the query-time fallback.
- Existing file-size debt remains: the session route file decreases from
  2,266 to 2,143 lines, and CLI `main.rs` decreases from 7,153 to 7,152 lines
  through a help-text-only edit. Both remain above the 1,000-line rule. A
  separate cleanup should split session routes by operation and separate CLI
  command definitions/handlers; no CI gate or allowlist was changed here.

## Spec

`src/bcs/specs/2026-09-16-read-only-session-list/spec.md`


RaymondHX (approved):
LGTM

PR #2253, opened 2026-09-16
feat(clawinsight): list all owner and admin rejections for calibration

ZhaosenLin:
## Problem

The offline completion-calibration pipeline needs the last 30 days of rejection feedback — both the rejections an Owner makes from "我的待办" and the ones an admin makes — as evidence for judging whether a negative tool/business result was actually the expected business outcome.

Today no internal endpoint can produce that:

- `GET /internal/governance/actions` and `GET /internal/governance/rejections` both filter `source_type IN (GOVERNANCE_SOURCE_TYPES)`, which only covers the rule-based governance sources. An improvement the Owner created (or an admin created for them) keeps `USER_SELECTED` / `ADMIN_SELECTED`, so **its rejection is never returned by any internal endpoint** — measured on prod, only 6 archived items are visible in 30 days and none of them is usable calibration input.
- `GET /admin/improvements?includeAll=true` does cover every source type, but it requires an admin login session, so offline jobs and agents cannot call it.
- Owner rejections and admin rejections share the same six reason codes (`rejectAdmin` writes the same codes as the owner path; only the admin *review* path writes `ADMIN_REJECTED`), so the reason code alone cannot tell the two rejection kinds apart.

## Solution

Add an internal read-only endpoint built for this need:

```
GET /api/insight/v1/internal/governance/rejections/all
    ?days=30&rejectedBy=ANY|OWNER|ADMIN&ownerUserId=&botId=&sourceRuleId=&limit=200&offset=0
```

- Covers every source type a rejection can be applied to (`OWNER_VISIBLE_SOURCE_TYPES ∪ REJECTED_RULE_*`), so Owner rejections of `USER_SELECTED` items are included.
- Returns only items that carry an actual rejection event (`[用户驳回]` / `[管理员驳回]` / `[Admin驳回]`) **and** are still `ARCHIVED`; an item restored by an admin reopen is excluded, and being archived is never treated as rejection evidence on its own.
- Returns the calibration fields directly: `rejectedBy`, `rejectReasonCode`, `rejectComment`, `rejectedAt`, plus scope/problem fields, `version` and `updatedAt`.
- Same access model as `GET /internal/governance/actions` (internal read-only, no machine auth). `ImprovementView` gains `rejectedBy` so the existing admin-facing APIs can tell the two rejection kinds apart too.
- `days` filters on `gmt_modified` (archive time); the exact `rejectedAt` window stays the caller's job. Documented in the route comment.

`rejectReasonCode` / `rejectComment` / `rejectedAt` / `rejectedBy` are also added to the opt-in `fields` allowlist of `/internal/governance/actions`; default responses are unchanged.

## Validation

- `cd src/evolverun/clawweb/public/modules/clawinsight && npx vitest run` → **49 files / 663 tests passed** (Node 22, the ABI `better-sqlite3` is built for).
- `npm run check` (`tsc -p tsconfig.json --noEmit`) → clean.
- New test `lists every rejected improvement, owner and admin, without authentication` asserts that an Owner-rejected `USER_SELECTED` item appears in `/rejections/all` with `rejectedBy=OWNER` plus reason code, comment and timestamp, while the same item is **absent** from `/actions?status=ARCHIVED` (pinning down the data-loss root cause); that an admin review rejection shows up as `rejectedBy=ADMIN` / `ADMIN_REJECTED`; and it covers the `rejectedBy` filters, `limit/offset/total`, the `days` window and an invalid `rejectedBy` (400).
- Not run: deployment against pre/prod environments.

## Compatibility and risk

- Additive only: one new route, one nullable field on `ImprovementView`, four extra values in an opt-in `fields` allowlist. No schema change, no state-machine change, no existing query re-filtered.
- The new route is unauthenticated like the existing `/internal/governance/actions`, and exposes rejection comments across owners — the same disclosure class as what that endpoint already returns (no evidence payloads).
- Rejected items restored later by an admin are intentionally not returned, so consumers must re-fetch instead of caching indefinitely.


fudp (approved):
LGTM

PR #2256, opened 2026-09-16
feat(backend): accept BaaS service JWTs for caller connections

yhzhang35:
## Problem

BaaS calls the existing application Caller connection endpoint with a service-signed JWT whose issuer is baas. Backend currently expects a gateway principal with an app identity and derives its tenant from that identity, so this service-only credential is rejected before the connection lifecycle runs.

## Solution

Authenticate POST /api/v1/expert-chats/app-caller-connection using the existing HS256 JWT decoder with an endpoint-specific issuer of baas and audience verification disabled. Keep signature, required iat/exp/iss, optional nbf and clock-skew checks. Ignore principals, app_id, tenant and aud for identity admission and routing.

Remove this endpoint's principal-derived tenant middleware branch. It uses the existing server default tenant before dependency injection. Retain Bot existence validation and enter the normal Caller lifecycle directly: create the instance/container when absent, or reuse/upgrade it when present. Ordinary-user endpoint guards remain unchanged. Query parameters, connection responses and error mapping remain compatible.

The change reuses the configured principal signing key and does not change shared gateway/org/OpenAPI validation, add a deployment secret, or introduce a database migration. Structured boundary logs omit JWT contents and redact connection credentials.

## Validation

- Current follow-up: 123 targeted service and real-JWT API tests passed, including actual first-instance/container creation through the BaaS service entry point and existing ordinary-user denial tests.
- Updated Singlebox acceptance to create the first instance via BaaS and verify subsequent reuse. All current-head remote checks passed. Backend: 18,759 passed, 43 skipped, zero failures/errors; line coverage 98,368/110,349=89.14%, changed lines 10/10=100% independently verified at a 90% minimum. Singlebox: endpoints 156/156 and CLI 52/52; E2E line coverage 41.63% (minimum 40%), method coverage 36.79% (minimum 36%).
- Plaintext secret scan, Python SAST blocking gate, runtime/acceptance Ruff checks and git diff --check passed.
- All results above apply to head f220ebda5; final review/comment audit found no outstanding feedback.

## Compatibility and risk

This is an authentication contract change for this one HTTP entry point: clients must send a JWT signed with the configured key and iss=baas. Gateway-issued application principals no longer authenticate this endpoint; other endpoints keep their contracts. BaaS callers can create and operate Caller instances in the server default tenant without application grants or user membership checks. Sharing the configured signing key means issuer distinguishes protocol behavior, not cryptographic isolation between key holders.

The internal ExpertChatInstanceServiceProtocol removes obsolete app_id and tenant arguments; repository callers and tests are updated together. No architecture waiver is needed. Rollback requires coordinating the caller's issuer/protocol with the Backend version. No deployment or merge is included in this PR task.

## Spec

- src/backend/specs/baas-caller-auth/001-spec-output.md
- src/backend/docs/baas-caller-signing-contract.md


totalfrank (approved):
LGTM.

PR #2263, opened 2026-09-16
fix(workflow): accept unbound snapshot run versions

wienyang:
## Summary
- Accept workflow_version = -1 for draft/Git snapshot runs without replacing it with a published release number.
- Keep workflow_deploy_number restricted to positive integers or null.
- Display unbound release semantics and document the internal API contract.
- Add HTTP/SQLite compatibility coverage, including workflows with an existing published release.

## Validation
- Workflow module: 228 tests passed.
- Workflow TypeScript check passed.

## Rollout
Deploy the ClawWeb service containing this module to restore creation of remote snapshot run records. No schema migration; previously missing records are not automatically backfilled.


fudp (approved):
LGTM

PR #2268, opened 2026-09-16
fix(clawinsight): return complete admin rejection feedback

ZhaosenLin:
## Problem
Ordinary administrator rejection writes `[管理员驳回]`, but governance feedback parsing recognized only `[Admin驳回]` and `[用户驳回]`. The rejection list could therefore include an item while returning null reason, comment, actor type and timestamp, preventing offline calibration from consuming it.

## Solution
Recognize all three existing rejection markers and take the last recorded rejection consistently for actor and feedback fields. Add parser regressions and real HTTP admin/owner reject-to-list tests, including actor filters. Preserve existing reason codes and user/Bot isolation. The prior rejections-all API is already on dev; this PR contains only the parser fix and its new regression suite.

## Validation
- On a fresh origin/dev worktree, Node 22: rebuilt the current shared dependency, then `npm run check` and `npm run build` passed for clawinsight.
- Full clawinsight `npm test`: 50 files, 678 tests passed, including 10 new rejection regression cases.
- `git diff --check` passed; both changed source files are under 1,000 lines.
- Production endpoint/data and AIStudio replay were not exercised; they require post-deployment verification. No deployment performed.

## Compatibility and risk
No schema migration, new endpoint, or response-field addition. Existing administrator rejection history becomes readable without rewriting stored data. For histories with multiple rejection events, response fields now consistently use the final recorded event. The existing SQL actor filter still matches historical event presence; changing its historical-filter semantics is outside this fix. Rollback by reverting this commit.


3h-william (approved):
ok

PR #2269, opened 2026-09-16
feat(clawinsight): refine agent monitoring list UI and problem-type f…

2001wjh:
## Problem

Business feedback on the Agent monitoring panel (ClawInsight):

1. Collapsed diagnosis rows showed only the title, decision word, session time, and a human-intervention column. Reviewers could not locate a specific session without expanding every row — Session Key, TC fault label, and creation time were hidden in the detail area.
2. The header and sidebar followed page scrolling, which felt broken; only the content area should scroll (matching other workspaces like 任务护航).
3. The default view loaded all history with 10 rows per page. In practice users almost always triage today's records, and 10 rows is too few for scanning.
4. There was no way to filter diagnoses by business problem category/subtype, even though every diagnosis carries these two fields and they are the primary triage dimension.

## Solution

- **Compact rows**: collapsed row now shows title, Session Key, TC fault label, and session time. The decision suffix ("| 告警/通过/无法判断") and the intervention column are removed; intervention remains visible in the expanded detail.
- **Fixed chrome**: new `useInsightViewport` hook measures the top navigation offset and pins the shell (`100dvh - offset`); only `.insight-content` scrolls.
- **Defaults**: initial query is Beijing today (`beijingDateRange(1)`) with page size 20. "清空筛选条件" restores this default instead of "all time".
- **Problem-type filter**: `ProblemTypeFilter` popover (funnel trigger "筛选问题类型", badge shows 1/2 applied conditions) with two dependent selects. Applying sends `businessProblemCategory` / `businessProblemSubtype` query params; drafts are cancelled on dismiss; changing the category clears the draft subtype. Server-side, the filters are bound-parameter `=` conditions applied **before pagination**; the response gains `problemTypes` (category → subtypes) derived from a separate `DISTINCT` read scoped to the current bot + date range only, so options never depend on the current page, keyword, decision tab, or selected types.

Rejected alternatives: filtering the current page client-side (breaks pagination/counts); deriving options from the current page's rows (options vanish when filtered); a modal dialog (heavier than the toolbar popover pattern used elsewhere).

## Validation

- `tsc --noEmit` and `npm run build` for `@avernet/clawinsight`: pass.
- Monitoring server suites (`vitest.monitoring.config.ts`, `MONITORING_TEST_SQLITE_DRIVER=node`): **108 tests pass** (4 files), including new cases: query validation (trim, 128-char limit, orphan subtype rejected), SQL filter-before-pagination with 25 rows (page 2 returns exactly 3), options independent of selected filters, injection strings as plain parameters, bot/date isolation, and an end-to-end HTTP test.
- Frontend InsightCenter suites: **43 tests pass** (9 files), including new cases for atomic apply + page reset, draft cancel/Escape focus restore, badge count, reset semantics, Beijing-day default across midnight, and the compact row content.
- Two pre-existing tests that assumed the old "all time" default were updated to the new today-default semantics.
- Not run here: the default repo-wide suite needs `better-sqlite3` native bindings (absent for Node 26.7 on this machine) and loopback listeners blocked by the local sandbox; the monitoring suites were run with the repo-provided `MONITORING_TEST_SQLITE_DRIVER=node` fallback. The facet `DISTINCT` query's execution plan on real ODC data volume, and preprod UI acceptance, remain to be verified during deployment.

## Compatibility and risk

- No schema or migration change; reuses existing columns `business_problem_category` / `business_problem_subtype`.
- Response gains an additive `problemTypes` field; clients that ignore unknown fields are unaffected. New frontend should deploy together with the new backend.
- Facet options and the list are two reads, so options may briefly lead/lag the list during active ingestion; list and counts stay in one SQL snapshot and the next 30s poll reconciles.
- Rollback: revert this commit; the new query params are optional and ignored by older clients.

## Spec

`src/evolverun/docs/clawweb/specs/clawinsight-agent-monitoring.md` — new section "诊断记录业务问题筛选（2026-09-16）" documents parameters, response semantics, snapshot caveats, and interaction rules.


3h-william (approved):
Ok

PR #2281, opened 2026-09-17
feat(backend): add a two-phase (preload-safe) boot seam to the community HTTP app

totalfrank:
## Problem

Importing `agentclaw.community.adapters.http.app` does two different jobs at once. It
*constructs* the ASGI app — routes, handlers, the `FastAPI` object — and it also
*initializes everything a worker owns*: `build_injector`, the pre/prod eager binding check,
the gateway principal signing-secret read, and `install_middleware` (which ends in
`tracer.install(app)`).

Under uvicorn's spawn model every worker re-runs all of it, so the cold start is paid N
times. The planned fix is Gunicorn `preload_app` — import once in the master, `fork` N
workers — and that requires the master import to be fork-safe: no threads, no
gRPC/DB/HTTP connections, no remote secret reads, no tracer runtime, nothing a forked child
would inherit. This PR adds the community-side seam such a change consumes. The
Gunicorn/OCB wiring itself is **not** in scope, and nothing in this repo selects the new
mode at runtime.

## Solution

`AGENTCLAW_HTTP_BOOT_MODE` (`eager` | `preload`, default `eager`) decides which of the two
jobs an import performs. Unset or empty is `eager`; an unknown value is a hard error at
import, the same treatment `DEPLOY_PROFILE` gets.

- **eager** — byte-for-byte today's behavior. `app.py` calls `finalize_worker_runtime()`
  inline at the end of its own import, so community, singlebox, local, `main.py`, the test
  suite and any direct `import app` are untouched.
- **preload** — the import stops after construction. The app owns no thread, connection, fd
  or secret, and the consumer calls, once per worker after the fork:

  ```python
  from agentclaw.community.adapters.http.app import app, finalize_worker_runtime
  finalize_worker_runtime()
  ```

  It must run **before the first ASGI call**, the lifespan startup event included: Starlette
  caches `middleware_stack` on that first call and `add_middleware` raises afterwards.

The new `adapters/http/boot.py` owns the switch and the finalize sequence, running the same
steps in the same order they ran at module level (middleware order is load bearing).

### What an un-finalized process is allowed to do

Most of the review on this PR converged on one question, so it is worth stating directly. An
un-finalized process is not an *empty* one — a child forked from a finalized parent holds
that parent's entire wired app: its injector, its middleware stack, and the engines, pools
and fds behind them, copied rather than reopened. Four guards follow from that, and they all
ask the same question, `worker_runtime_finalized()`, which is per-process by construction:

- **Finalizing again is refused.** `install_middleware` appends, so re-running would give the
  child two stacks while it still held the parent's resources. Whether the parent finished
  the sequence or began it and failed makes no difference.
- **The lifespan is gated on finalization, not on injector presence.** The inherited injector
  is perfectly non-`None`; gating on presence would let the child run `bootstrap()`/
  `startup()` over the parent's participants, bringing up background workers on the parent's
  pools while it was being refused traffic.
- **Requests are refused with 503**, by `RequireWorkerRuntime` — because nothing in ASGI
  obliges a host to drive the lifespan, and a worker that skipped finalize would otherwise
  answer its first request by building *and caching* a stack with no auth, no tenant scoping,
  no tracing and no CORS.
- **That guard wraps the built stack rather than joining it.** `add_middleware` prepends, so
  a guard added at construction lands *innermost* and an inherited stack runs ahead of it —
  measured: the child answered 503 having already executed `UserContextMiddleware`, and with
  it the parent's auth plugin against the parent's pool. Overriding
  `build_middleware_stack` puts the guard outside everything that build produces, including
  Starlette's `ServerErrorMiddleware`, whether the stack is built fresh post-fork or
  inherited already-built.

The guard owns no state and no resource, so it is safe to inherit; in `eager` mode and in any
correctly finalized worker it never fires. **Only a fork performed before any finalize
attempt is supported**, which is what `preload` gives you.

Also: **`OptionalRouters` moves into finalize.** It was the one module-level
`injector.get(...)` in the tree, and therefore the only reason construction needed a built
injector at all. It is the only route registration that resolves a binding outside a request;
every other router still registers at import. It is empty on every prod profile.

`_app_lifespan` now discovers participants from the injector *this process* finalized, rather
than a handle captured at import. `app.py` no longer exposes a module-level `injector`, which
is what its docstring already claimed; `scripts/community_selftest.py`'s boot proof moves to
`app.state.injector`.

**Rejected alternative:** resolving `OptionalRouters` at import from a throwaway injector.
That would make the master run `Injector(modules)` over the full column (~19 corp
infrastructure modules under `corp`) plus six eager config resolves through the sofapy
`ConfigProvider` — exactly the work preload exists to keep out of the master, and not
demonstrably fork-safe.

## Validation

Investigation was measured, not assumed: a probe wrapped `threading.Thread.start` and
`socket.socket.connect` and snapshotted `/proc/self/fd` **before** importing `app.py`.

| profile | new live threads | `connect()` calls | fd delta | operations registered |
| --- | --- | --- | --- | --- |
| `test` | none | 0 | 0 | 798 over 678 paths |
| `singlebox` | none | 0 | 0 | 798 over 678 paths |

So nothing had to be made lazy: every fork-unsafe side effect already lived in the
resolve/install steps finalize now owns.

New `tests/community/adapters/http/test_two_phase_boot.py` — **23 tests**. Most run in a
subprocess, because the conftest imports the app at session start and an in-process re-import
proves nothing about what an import does. Four use a **real `os.fork()`**. They cover: eager
as the default and an unknown mode failing the import; preload leaving no injector, no
middleware, no thread and no connection even with `SERVER_ENV=pre`; finalize running the
eager check, the verifier init and `install_middleware` exactly once each; the lifespan
reading the worker's injector; a second finalize in the same pid being inert; a failed
finalize refusing a retry; construct → fork → finalize per worker with neither stack doubled;
finalize → fork with the child refusing and leaving what it inherited unmutated; the same
refusal for an inherited *failed* attempt; the lifespan refusing an inherited injector and
never reaching `discover_lifecycle_participants`; an un-finalized worker whose lifespan is
never driven answering 503, with a spy proving no inherited middleware ran first; a finalized
worker serving 200 through the same guard; a warm OpenAPI cache not hiding the routers
finalize mounts; and eager vs preload+finalize producing an identical OpenAPI document.

Local, on this head: **3675 passed, 2 skipped** across architecture, di, ci, e2e, adapters
and api (measured before the last two test-only commits; the boot suite is 23/23 on this
head). An earlier full `tests/community` run on this branch was **19000 passed, 46 skipped,
2 failed**, both failures being
`test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]`
on `rsync: not found` — `rsync` is absent from the sandbox that ran. Not this PR's: they fail
identically in isolation, identically on the base commit in a clean worktree, and CI (which
has `rsync`) is green on the same suite.

Two pre-existing quirks surfaced while writing the tests, both handled rather than papered
over: FastAPI 0.138 made `include_router` lazy, so `app.routes` no longer reflects the
surface and the tests read `app.openapi()` instead; and generated operation ids derive from
`list(route.methods)[0]`, so the cross-mode comparison pins `PYTHONHASHSEED`.

## Compatibility and risk

Unchanged: API paths, methods, operation ids, schemas, auth behavior, exception handlers,
lifespan participants, the community/local default startup, and the
`agentclaw.community.adapters.http.app:app` import path. The worker middleware set and order
are unchanged — `RequireWorkerRuntime` is not a user middleware; it wraps the stack from
outside and is inert once the runtime is up.

Changed deliberately, all documented in the spec:

- **Boot timing only, in eager mode** — `build_injector` and the eager check now run after
  the router imports rather than before. Nothing in the tree resolves from the injector at
  module import, so nothing observes it; a misconfiguration still fails the same import.
- **Route order** — the DI-conditional `OptionalRouters` mount last rather than just before
  the public router. Their prefixes (`/local/sql`, empty on non-test profiles) do not overlap
  the public surface.
- **Rollback** is leaving `AGENTCLAW_HTTP_BOOT_MODE` unset, which is the default.

Note for whoever touches `app.py` next: it now sits within a line or two of the 1000-line cap
enforced by `test_no_oversized_modules`, so the next change to it needs prose moved out or an
allowlist entry.

**One residual item for the OCB side.** `get_logger()` runs at module import in both `app.py`
and `boot.py`. Under community profiles that is `logging.getLogger`, which allocates nothing.
Under corp, the corp entrypoint installs its logger factory (file rotation) *before*
importing `app.py`, so a preloading master could materialize a rotating file handler that
every forked worker inherits. That is not reachable from community code — corp should install
its factory post-fork or use a fork-safe handler. More generally, the corp providers (sofapy
`ConfigProvider`, corp `SecretResolver`, SofaTracer, ZDAS `DatabasePlugin`) cannot be
imported from this repo, so "verified by running it" covers `test` and `singlebox` only.
Re-run the thread/connect probe against a corp image before enabling `preload_app`.

## Spec

`src/backend/specs/2026-09-17-http-two-phase-preload-boot/` — `spec.md` (contract and phase
split) and `plan.md` (the fork-safety findings above, the design, the test plan, and a note
on where the implementation deliberately departs from the original brief).

https://claude.ai/code/session_01MC7zCSquKVoF6FZ6gxjnne


totalfrank (commented):
Reviewed commit `c85f83972a93e670b91bb5a22833f0cea5e3d1e1` (the only commit on this PR) against merge-base `29c1ca45c4ccfd30a1a2fe8acc28d056a9e8fcff`. No prior review on this PR to reconcile.

Files covered: `app.py`, the new `boot.py`, `middleware.py`, `di/__init__.py`, `di/container.py`, `di/optional_routers.py`, `utils/gateway_principal_config.py`, `scripts/community_selftest.py`, and the new `tests/community/adapters/http/test_two_phase_boot.py` plus the updated `test_build_injector_composition_root_only.py`. Also read `spec.md`/`plan.md`.

Checked, in priority order:
- **Correctness**: traced the pid-based (not boolean) finalize guard, the fail-and-stay-failed path, and the fork test (`test_a_forked_child_finalizes_its_own_runtime`) — all sound. Verified the singlebox-coverage middleware install, which moved from `app.py` into `boot._install_worker_runtime` step 6, is not dropped. Verified the route-mount reordering (`OptionalRouters` now mounts after the public `openapi_v1` router instead of before) is addressed in the PR description and doesn't overlap prefixes (`/local/sql` vs. the public surface).
- **Contract breakage**: no module-level `injector` handle remains anywhere in `src/agentclaw/` (grepped); `community_selftest.py` was updated to read `app.state.injector` instead. `get_app_injector()` callers are all request/runtime-time, not import-time, so they're unaffected by deferring `build_injector()` into `finalize_worker_runtime()`.
- **Error/failure handling**: the finalize failure path correctly poisons only the failing pid and refuses a retry rather than replaying a half-wired sequence.
- **Concurrency**: the pid-guard logic isn't lock-protected, but the documented call contract (once per worker, before the first ASGI call) makes concurrent same-process calls out of scope for this change.
- **Architecture boundaries**: composition-root-only `build_injector` call site enforcement (`test_build_injector_composition_root_only.py`) was updated consistently to allow `boot.py` instead of `app.py`.
- **Test coverage**: the new suite exercises eager default, unknown-mode failure, preload-leaves-no-runtime, double-finalize idempotency, the pid-guard under simulated and real `fork()`, cross-mode OpenAPI/middleware equivalence, and the pre-first-ASGI-call timing constraint — matches the risk surface described in the PR body.

No correctness, contract, or architecture issues found in this diff.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,252 @@
...
+            "again. Let the worker exit and start a new one."
+        )
+    if _finalized_pid is not None:
+        # Inherited across fork. Not an error — it is the case the pid guard
+        # exists for — but worth a line, because it is also what a stray second
+        # call in a *new* process looks like.
+        logger.info(
+            "[boot] inherited finalize marker from pid %s; finalizing pid %s",
+            _finalized_pid, pid,
+        )
```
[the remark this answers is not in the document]
totalfrank: Confirmed and fixed in a97d3c97 — thanks, this was real and my own fork test was passing straight through it.

Reproduced on the previous commit before changing anything: parent finalizes, forks, child finalizes → the parent's 8 `user_middleware` entries become **16** in the child, every class listed twice. The old `test_a_forked_child_finalizes_its_own_runtime` only asserted the child's middleware was non-empty, so it never caught the doubling.

Of the two options you named I took **reject**, not rebuild-then-finalize. Clearing `user_middleware` and `middleware_stack` would drop the *references* but not the resources: the parent's engines, pools and fds are already in the child's image, copied rather than reopened, so the child would still be running on fork-unsafe inherited state — which is your point about "can still use fork-unsafe parent resources despite receiving a new injector". A child in that position cannot repair its own process image, so it now raises, naming the pid that finalized before the fork and pointing at `preload`:

```
The worker runtime was finalized in pid 2974, before this process (pid 2978) was
forked from it. This process has inherited that runtime's middleware stack,
injector and every resource behind them, and finalizing again would stack a
second copy on top rather than replace them. Import the app with
AGENTCLAW_HTTP_BOOT_MODE=preload so the master forks before any worker runtime
exists, then call finalize_worker_runtime() in each child.
```

The guard's purpose is unchanged — a child must never mistake its parent's initialization for its own — it just fails loudly now where it used to silently stack. The supported flow is untouched: in `preload` the master constructs only, so nothing is finalized before the fork and the case never arises.

Test coverage is now two tests instead of one: the supported shape (construct → fork → finalize in each worker, asserting neither side's stack is doubled) and the unsupported one (finalize → fork, asserting the child refuses *and* leaves the inherited stack unmutated). Both use a real `os.fork()`. 17/17 in the boot suite, 748 in architecture/di/ci/e2e, and eager vs preload+finalize still produce an identical 798-operation surface.

---

totalfrank (commented):
Follow-up round on commit `a97d3c97` (the only new commit since my last review at `c85f8397`).

**Findings checklist:**
- My round-1 review (clean, no findings) — nothing to verify.
- Codex's P2 finding (`boot.py`, re-finalize in a forked child stacks a second middleware stack on the inherited one) — **fixed** in `a97d3c97`: the inherited-pid branch now raises instead of falling through to `_install_worker_runtime`, and the two real-`os.fork()` tests (`test_a_forked_child_finalizes_its_own_runtime`, `test_a_child_forked_after_finalize_refuses_rather_than_stacking`) directly assert no doubling and that the refusal leaves the inherited middleware stack untouched. Thread already resolved by the author with a reply — conceded reasoning was sound.

**Review of the new commit (`a97d3c97`):**
- The raise for the inherited-marker case sits outside the `try/except` that sets `_failed_pid`, so a refusal isn't misrecorded as this pid's own failed-finalize attempt — correct, since the pid never attempted `_install_worker_runtime`.
- The other cross-pid path (`_failed_pid` set by a different pid, e.g. a hypothetical fork after a failed finalize) is untouched by this commit and was already present in `c85f8397`; not in scope here and not reachable under the supported preload flow (fork happens before any finalize call).
- Call site (`app.py:996`) calls `finalize_worker_runtime()` unconditionally with no new exception handling needed — consistent with the documented "a worker that cannot initialize is replaced, not retried" contract.
- New tests use real `os.fork()` for both the supported (construct → fork → finalize each side) and unsupported (finalize → fork → refuse) shapes, and assert the middleware list isn't mutated on refusal.

No new issues found.

**Open count: 0.** PR is clean as of this round.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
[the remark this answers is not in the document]
totalfrank: Follow-up: the round-2 review noted that the other cross-pid path — `_failed_pid` set by a different pid, i.e. a fork after a *failed* finalize — was untouched. It was right, and it was the same hole through the other door, so I've closed it in 95741b88.

A parent that raised partway through `_install_worker_runtime` may already have attached an injector and installed part of the stack, and it records `_failed_pid`, not `_finalized_pid`. A child forked from it saw neither marker as its own, fell through, and finalized on top of the half-wired app it inherited — identical stacking defect, reached via the failure path instead of the success path.

Either marker set by a different pid now refuses, and the message names which process left the app in that state and whether it finished or failed:

> Pid 2974 began finalizing the worker runtime and failed before this process (pid 2978) was forked from it. …

As the review said, this isn't reachable under the supported preload flow — the fork precedes any finalize call — so it's a guard-completeness fix rather than a live bug. It felt wrong to leave the guard depending on which door the state arrived through.

18/18 in the boot suite (one new test for the failed-parent shape), and 2911 passed / 2 skipped across architecture, di, ci, e2e and adapters.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,279 @@
...
+    injector or installed part of the stack, and re-running would stack a second
+    copy on top. A worker that cannot initialize must be replaced, not retried.
+    """
+    global _finalized_pid, _failed_pid
+
+    pid = os.getpid()
+    if _finalized_pid == pid:
+        logger.debug("[boot] worker runtime already finalized in pid %s", pid)
+        return
+    if _failed_pid == pid:
```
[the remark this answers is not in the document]
totalfrank: Agreed, and already fixed in 95741b88 — this review landed on `a97d3c97` a few minutes before that push, so it's describing the exact gap that commit closes. Your diagnosis matches it precisely: a parent that raised partway records `_failed_pid`, not `_finalized_pid`, so the child saw neither marker as its own and fell through onto the half-wired app it inherited.

The fix is the one you suggested — treat *either* marker set by a foreign pid as an inherited, unrecoverable runtime:

```python
if _finalized_pid is not None or _failed_pid is not None:
    origin, what = (
        (_finalized_pid, "finalized the worker runtime")
        if _finalized_pid is not None
        else (_failed_pid, "began finalizing the worker runtime and failed")
    )
    raise RuntimeError(f"Pid {origin} {what} before this process (pid {pid}) was forked from it. …")
```

The message distinguishes the two origins so the operator can tell a clean pre-fork finalize from a failed one, but the refusal is identical.

Covered by `test_a_foreign_failed_pid_is_refused_the_same_way`, which asserts the refusal and that nothing was attached or installed on the way out. 18/18 in the boot suite; 2911 passed / 2 skipped across architecture, di, ci, e2e and adapters.

---

totalfrank (commented):
Follow-up check on the `pull_request.synchronize` event for this head. The current head (`95741b88`) is the same commit my last round already reviewed — no new commits landed since then, so this round re-verifies the standing findings directly against the live code rather than filing anything new.

- Round 1 (`c85f8397`, my own review) — clean, no findings.
- Codex P2, thread `boot.py:138-145` — re-finalizing in a child forked *after* a successful finalize stacks a second middleware set on the inherited one. **Fixed** in `a97d3c97`, confirmed in round 2, thread resolved.
- Round 2 (`a97d3c97`, my own review) — no new issues.
- Codex P2, thread `boot.py:145-148` — the same hole through the failure door: a child forked from a parent whose finalize *failed partway* falls through and finalizes on top of the half-wired app. **Fixed** in `95741b88`, confirmed in round 3, thread resolved.
- Round 3 (`95741b88`, my own review) — no new issues.

**This round:** re-read `boot.py` end to end. `finalize_worker_runtime` (line 167) now refuses whenever *either* `_finalized_pid` or `_failed_pid` is set by a foreign pid, with a message naming which process left the app in that state and whether it finished or failed — matching both fixes above. `test_a_foreign_failed_pid_is_refused_the_same_way` (test_two_phase_boot.py:371) exercises the failure-door case directly and asserts no injector/middleware leaks through on refusal. No regressions introduced.

**Open count: 0.** PR is clean, cleared as of round 3 (`95741b88`) and reconfirmed this round with no code changes in between.

---

src/backend/src/agentclaw/community/adapters/http/app.py
```diff
@@ -185,7 +190,21 @@ async def _app_lifespan(app: FastAPI):
...
+
+    Participants come from ``app.state.injector`` — the injector the *worker*
+    built in :func:`finalize_worker_runtime` — rather than from a handle captured
+    at import. Under ``preload`` the module-level import happened in another
+    process entirely, so reading the app's own state is what makes a participant
+    the one this worker will actually drive.
     """
+    injector = getattr(app.state, "injector", None)
+    if injector is None:
+        raise RuntimeError(
```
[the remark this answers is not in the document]
totalfrank: Real, and fixed in 4ca0e0f6. You're right that the lifespan check alone isn't a guard — nothing in ASGI obliges a host to drive the lifespan protocol, so a preload worker that skipped its post-fork call would answer its first request by building *and caching* a stack with no auth, no tenant scoping, no tracing and no CORS, then serve every later request through it. Silent, and the worst possible failure mode for this seam.

Added `RequireWorkerRuntime` in `boot.py`, installed during construction (phase 1) in both modes:

```python
async def __call__(self, scope, receive, send):
    kind = scope["type"]
    if kind in ("http", "websocket") and not worker_runtime_finalized():
        ...  # 503 for http, close 1011 for websocket
        return
    await self.app(scope, receive, send)
```

Three properties that matter here:

- It owns no state and no resource, so it is safe to inherit across a fork — it can be installed at construction, which is what makes it independent of the host's lifespan handling.
- It reads `worker_runtime_finalized()`, which is per-process, so a child that has not finalized refuses even though its parent had.
- In `eager` mode, and in any correctly finalized worker, it never fires.

You're also right that the stack still gets built and cached by that first refused request, so a later finalize would still hit "Cannot add middleware after an application has started". That's the correct outcome — fail-fast — and the property worth having is that no unauthenticated request was served in the meantime.

Two tests: `test_an_unfinalized_worker_refuses_requests_without_a_lifespan` uses a `TestClient` *outside* its context manager, which is precisely the host that never sends a lifespan event, and asserts 503 with only the guard in the stack; `test_a_finalized_worker_serves_normally_through_the_guard` asserts 200 through the same guard. Boot suite 20/20.

The docstring/spec finding in your other comment is fixed in the same commit.

---

src/backend/src/agentclaw/community/adapters/http/app.py
```diff
@@ -240,55 +259,25 @@ async def _app_lifespan(app: FastAPI):
...
 
-install_middleware(
-    app,
-    auth_plugin=injector.get(AuthPlugin),
-    tracer=injector.get(TracerPlugin),
-    cors_config=injector.get(CorsConfig),
-)
+    At most once per process; re-runs in a forked child regardless of the
+    parent's state; raises without marking the process finalized if any step
+    fails. Under the default ``eager`` mode this module already called it at the
```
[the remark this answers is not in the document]
totalfrank: Correct on both counts, fixed in 4ca0e0f6. The docstring and the spec were both accurate when written and were left stale by the two commits that turned an inherited runtime into a refusal — exactly the contradiction you describe, and the integrator-facing one, since this docstring *is* the contract OCB will follow.

`app.py`'s wrapper now says:

> At most once per process, and only in a process forked **before** any finalize ran: a child of a parent that had already finalized — or had begun and failed — inherits that runtime and is refused rather than wired a second time. A failing step raises without marking the process finalized, and the process cannot retry.

And `spec.md` replaces "A forked child, being a different pid, still gets its own attempt" with the opposite, stated positively: **only a fork performed before any finalize attempt is supported**, which is what `preload` mode gives you. The spec also now documents the `RequireWorkerRuntime` guard from your other comment.

Thanks for catching the AGENTS.md angle — a contract that documents the inverse of what the code does is worse than an undocumented one, because it invites exactly the fork-a-replacement-after-failure recovery you described.

---

totalfrank (commented):
Re-triggered `pull_request.synchronize` check. No new commits landed since the last round — the head is still `4ca0e0f6`, the same commit my round-4 review already covered — so this pass re-verifies the standing findings directly against the live code rather than re-reviewing from scratch.

**Independent verification this round:** read `boot.py`, `app.py`, and `spec.md` at `4ca0e0f6` directly (not the prior replies) to confirm each fix is actually in the code:
- `finalize_worker_runtime()` in `boot.py` refuses whenever `_finalized_pid is not None or _failed_pid is not None` for a foreign pid, with a message naming the origin pid and whether it finished or failed — covers both the success-door and failure-door fork cases.
- `RequireWorkerRuntime` is a real ASGI middleware class in `boot.py`, installed unconditionally via `app.add_middleware(RequireWorkerRuntime)` right after `app = FastAPI(...)` (`app.py:258-262`) — present in both boot modes, independent of lifespan handling.
- `app.py`'s `finalize_worker_runtime()` docstring and `spec.md:95` both now state only a fork performed *before* any finalize attempt is supported — the stale "regardless of the parent's state" contract is gone.

All four match what was claimed in the replies. No discrepancies.

- Round 1 (`c85f8397`, own review) — clean, no findings.
- Codex P2, `boot.py:138-145` — re-finalize in a child forked after a *successful* finalize stacks a second middleware set. **Fixed** in `a97d3c97`, confirmed round 2, thread resolved.
- Round 2 (`a97d3c97`, own review) — no new issues.
- Codex P2, `boot.py:207` — the same hole through the *failure* door (fork after a partial/failed finalize). **Fixed** in `95741b88`, confirmed round 3, thread resolved.
- Round 3 (`95741b88`, own review) — no new issues.
- Codex P2, `app.py:200` — fail closed when the host never drives the lifespan protocol. **Fixed** in `4ca0e0f6` via `RequireWorkerRuntime`, confirmed round 4, thread resolved.
- Codex P2, `app.py:275` — docstring/spec contradicted the new refusal behavior. **Fixed** in `4ca0e0f6`, confirmed round 4, thread resolved.
- Round 4 (`4ca0e0f6`, own review) — no new issues.

**This round:** no new commits to review (Part B is empty); all four Codex findings independently re-verified against live code as fixed.

**Open count: 0.** PR is clean, cleared as of round 4 (`4ca0e0f6`), reconfirmed this round by direct code inspection with no code changes in between.

---

src/backend/src/agentclaw/community/adapters/http/app.py
```diff
@@ -185,7 +188,21 @@ async def _app_lifespan(app: FastAPI):
...
     fail-fast; teardown direction is log-and-continue.
+
+    Participants come from ``app.state.injector`` — the injector the *worker*
+    built in :func:`finalize_worker_runtime` — rather than from a handle captured
+    at import. Under ``preload`` the module-level import happened in another
+    process entirely, so reading the app's own state is what makes a participant
+    the one this worker will actually drive.
     """
+    injector = getattr(app.state, "injector", None)
+    if injector is None:
```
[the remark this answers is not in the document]
totalfrank: Real, and fixed in c0f0bfd6. The two guards were asking different questions — `RequireWorkerRuntime` asks whether *this process* finalized, the lifespan only asked whether an injector was attached — and as you say, those diverge in precisely the case the pid guard exists for. A child forked from a finalized parent inherits a perfectly non-`None` `app.state.injector`, so it was refused traffic and then sailed straight into `bootstrap()`/`startup()` over the parent's participants.

That's the worse half of the two, which I'd missed: 503ing every request while `TaskWorker` and friends come up on the parent's pools is a process doing damage quietly rather than loudly.

Both gates now ask the same question, via `require_worker_injector`:

```python
if not worker_runtime_finalized():
    raise RuntimeError(f"Pid {os.getpid()} has not finalized its worker runtime, so it must not start the lifespan: …")
injector = getattr(app.state, "injector", None)
if injector is None:
    raise RuntimeError("marked finalized but no injector is attached; …")
return injector
```

The second branch is now a genuinely different error — finalized but detached — rather than the catch-all it used to be.

The test drives the lifespan with a foreign `_finalized_pid` while an injector really *is* attached, which is a forked child's exact view, and asserts not only the raise but that `discover_lifecycle_participants` was never reached — resolving the participants is itself what touches the parent's resources, so stopping at the raise isn't enough to prove. Boot suite 21/21.

---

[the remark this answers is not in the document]
totalfrank: Confirmed against `c0f0bfd6`: `require_worker_injector` (boot.py:134) raises on `not worker_runtime_finalized()` before it ever reads `app.state.injector`, and `test_lifespan_refuses_an_inherited_injector_even_though_one_is_attached` (test_two_phase_boot.py:721) drives the lifespan with a foreign `_finalized_pid` while a real injector is attached and asserts `discover_lifecycle_participants` was never called. Resolving.

---

totalfrank (commented):
Re-triggered `pull_request.synchronize`. The head is still `c0f0bfd6`, the same commit my round-5 review already covered — no new commits landed, so Part B (new-commit review) is empty this round. This pass verified the one finding that was fixed but left unresolved and closed it out.

- Round 1 (`c85f8397`, own review) — clean, no findings.
- Codex P2, `boot.py:138-145` — re-finalize in a child forked after a *successful* finalize stacks a second middleware set. **Fixed** in `a97d3c97`, confirmed round 2, thread resolved.
- Round 2 (`a97d3c97`, own review) — no new issues.
- Codex P2, `boot.py:242` — the same hole through the *failure* door (fork after a partial/failed finalize). **Fixed** in `95741b88`, confirmed round 3, thread resolved.
- Round 3 (`95741b88`, own review) — no new issues.
- Codex P2, `app.py:200-202` — fail closed when the host never drives the lifespan protocol. **Fixed** in `4ca0e0f6` via `RequireWorkerRuntime`, confirmed round 4, thread resolved.
- Codex P2, `app.py:273-275` — docstring/spec contradicted the new refusal behavior. **Fixed** in `4ca0e0f6`, confirmed round 4, thread resolved.
- Round 4 (`4ca0e0f6`, own review) — no new issues.
- Codex P2, `app.py:198-199` — lifespan gated on injector presence rather than `worker_runtime_finalized()`, letting a child with an inherited injector run `bootstrap()`/`startup()` on the parent's participants. **Fixed** in `c0f0bfd6`; the fix was already replied to in round 5 but the thread was left unresolved. Verified this round directly against the code (`require_worker_injector` in `boot.py` raises on `not worker_runtime_finalized()` before reading `app.state.injector`; `test_lifespan_refuses_an_inherited_injector_even_though_one_is_attached` asserts `discover_lifecycle_participants` is never reached on a foreign pid) — **thread resolved this round**.
- Round 5 (`c0f0bfd6`, own review) — no new issues.

**This round:** no new commits to review (Part B is empty); the one standing open thread verified fixed and resolved.

**Open count: 0.** PR is clean, cleared as of round 5 (`c0f0bfd6`).

---

src/backend/src/agentclaw/community/adapters/http/app.py
```diff
@@ -240,55 +255,31 @@ async def _app_lifespan(app: FastAPI):
...
 app = FastAPI(lifespan=_app_lifespan)
 
-# The injector was built at the top of this file (before router
-# imports). Now that ``app`` exists, attach the same injector so
-# route handlers can resolve via Injected(...) and
-# request.app.state.injector.
-attach_injector(app, injector)
+# Refuse traffic in a process that has not finalized. At construction, because
+# the lifespan guard above only fires on hosts that drive the lifespan protocol.
+app.add_middleware(RequireWorkerRuntime)
```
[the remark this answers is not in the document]
totalfrank: Right, and measured before changing anything. Parent finalizes, forks, child issues a request:

```
child: {"status": 503, "parent_mw_ran": ["UserContextMiddleware"]}
```

So the child refused — after running the parent's auth middleware, which calls the parent's auth plugin, which reads the parent's connection pool. Refusing late is not refusing.

Worth owning that I reasoned about exactly this ordering when I added the guard and talked myself out of it: "in the un-finalized case there is nothing else in the stack, so placement is immaterial." That holds for a worker that simply never finalized, and is false for the case the guard exists for, where the stack is inherited and fully populated. Same blind spot as the earlier two findings — I kept modelling the un-finalized process as *empty* rather than as *holding its parent's everything*.

Fixed in c78363e6 by wrapping rather than joining:

```python
def install_worker_runtime_guard(app):
    build_stack = app.build_middleware_stack

    def build_with_guard():
        return RequireWorkerRuntime(build_stack())

    app.build_middleware_stack = build_with_guard
```

Starlette builds the stack lazily on the first ASGI call, so wrapping the builder puts the guard outside everything that build produces — `ServerErrorMiddleware` included — and it holds however the stack is reached: built fresh in a worker after the fork, or inherited already-built from a parent that had served, since the wrapper is what that parent built too. The guard is consequently no longer in `user_middleware` at all, which is why the middleware-set assertions went back to the original eight.

Same measurement now:

```
child: {"status": 503, "parent_mw_ran": []}
```

`test_the_guard_short_circuits_before_inherited_middleware_runs` pins it across a real `os.fork()`, spying on `UserContextMiddleware.dispatch` and asserting the 503 arrives with nothing having run — asserting the 503 alone would have passed against the broken version. Boot suite 22/22.

---

totalfrank (commented):
Round 6, on `c78363e6` — the one new commit since my round-5 review at `c0f0bfd6`.

**Verified the standing finding fixed:** Codex's `app.py:260` finding (the runtime guard sat innermost, so a child forked from a finalized parent would run the parent's `UserContextMiddleware` — and through it the parent's auth plugin against the parent's connection pool — before getting refused) is fixed in `c78363e6`. Read `boot.py` and `app.py` directly: `install_worker_runtime_guard` now overrides `app.build_middleware_stack` to wrap whatever it produces in `RequireWorkerRuntime`, instead of `add_middleware` (which Starlette prepends, landing the guard innermost). This holds whether the stack is built fresh in a worker after a fork or inherited already-built from a parent that had served — same object, same wrapper, either way. `test_the_guard_short_circuits_before_inherited_middleware_runs` drives this across a real `os.fork()`, spies on `UserContextMiddleware.dispatch`, and asserts the child's 503 arrives with the spy never having fired. Thread resolved this round.

**Review of the new commit (`c78363e6`):** no issues. `FastAPI`/`ASGIApp` type hints stay correctly TYPE_CHECKING-only and quoted (consistent with the file's existing style); the closure over the original bound `build_middleware_stack` is captured once at construction, before the override replaces the instance attribute, so it isn't re-entrant with itself; `spec.md`/`plan.md` updated to match. Confirmed no real merge conflict against current `dev` (mergeable_state's "blocked" is the outstanding-approvals gate, not a conflict). CI green on this head.

- Round 1 (`c85f8397`, own review) — clean, no findings.
- Codex P2, `boot.py:138-145` — re-finalize in a child forked after a *successful* finalize stacks a second middleware set. **Fixed** in `a97d3c97`, confirmed round 2, thread resolved.
- Round 2 (`a97d3c97`, own review) — no new issues.
- Codex P2, `boot.py:270` — the same hole through the *failure* door. **Fixed** in `95741b88`, confirmed round 3, thread resolved.
- Round 3 (`95741b88`, own review) — no new issues.
- Codex P2, `app.py:200-202` — fail closed when the host never drives the lifespan protocol. **Fixed** in `4ca0e0f6` via `RequireWorkerRuntime`, confirmed round 4, thread resolved.
- Codex P2, `app.py:273-275` — docstring/spec contradicted the new refusal behavior. **Fixed** in `4ca0e0f6`, confirmed round 4, thread resolved.
- Round 4 (`4ca0e0f6`, own review) — no new issues.
- Codex P2, `app.py:198-199` — lifespan gated on injector presence rather than `worker_runtime_finalized()`. **Fixed** in `c0f0bfd6`, confirmed round 5, thread resolved.
- Round 5 (`c0f0bfd6`, own review) — no new issues.
- Codex P2, `app.py:260` — guard installed innermost via `add_middleware`, so an inherited stack ran ahead of it. **Fixed** in `c78363e6`, confirmed this round, thread resolved.
- Round 6 (`c78363e6`, this round) — no new issues.

**Open count: 0.** PR is clean, cleared as of round 6 (`c78363e6`).

---

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+    """How ``app.py`` splits construction from worker-runtime initialization."""
+
+    #: Import constructs *and* initializes — today's behavior, and the default.
+    EAGER = "eager"
+    #: Import constructs only; the consumer calls ``finalize_worker_runtime()``
+    #: once per worker after fork.
+    PRELOAD = "preload"
+
+    @classmethod
+    def detect(cls) -> "BootMode":
```
totalfrank-personal: Do you really need forward reference here? If not, let's drop it.

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+
+Call timing is not advisory. Starlette builds and caches ``middleware_stack`` on
+the first ASGI call — ``add_middleware`` raises ``RuntimeError`` afterwards — and
+the lifespan startup event *is* an ASGI call. So finalize must run before the
+lifespan and before the first request, or the auth, tenant, tracing and CORS
+middleware this module installs would silently never take effect.
+
+The "already finalized" guard records the **pid** that finalized rather than a
+boolean. A boolean would be inherited across ``fork``: a child would see
+``True``, skip its own finalize, and serve traffic on the master's injector and
```
totalfrank-personal: I want to make sure my understanding is correct here.

So you are saying when we fork, the child process inherits what the master has. So it by default would have that injector pointed at master's.

But what does that mean? It's still two different object right? Because I guess processes can not share memory? What you mean is that if any component is loaded by the injector, the data of them will be the same. And that's wrong for things like DB connections, etc, right?

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+the master's connections. Comparing ``os.getpid()`` makes the flag meaningless
```
totalfrank-personal: What do you mean by "makes the flag meaningless in any process that did not set it."

You mean if we use pid to do the comparison, we won't run into the issue where we can not distinguish if the init is done on the child process or the master process right?

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+in any process that did not set it.
+
+What a child does about an inherited marker — from a parent that finished the
```
totalfrank-personal: What do you mean by marker here? I am confused.

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+sequence, or began it and failed — is *refuse*, not re-run. The marker is never
```
totalfrank-personal: What do you mean by "from a parent that finished the sequence, or began it and failed"? I can't literally understand what you are talking about.

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+def worker_runtime_finalized() -> bool:
+    """Whether *this* process has completed its worker-runtime initialization.
+
+    False in a forked child that inherited a finalized parent's module state,
+    which is the whole point of storing a pid.
+    """
+    return _finalized_pid is not None and _finalized_pid == os.getpid()
+
+
+def require_worker_injector(app: "FastAPI") -> "Injector":
```
totalfrank-personal: Same here, if we do not need forward reference which I believe we don't because FastAPI and Injector won't depend on our module, then we should get rid of it.

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+    puts it *outside* every other middleware — see there for why that placement
+    is the whole point. It owns nothing — no state, no resource — so it is safe
+    to inherit across a fork, and it reads :func:`worker_runtime_finalized`,
+    which is per-process: a child that has not finalized refuses even though its
+    parent had.
+
+    In ``eager`` mode, and in a correctly finalized worker, it never fires.
+    """
+
+    def __init__(self, app: "ASGIApp") -> None:
```
totalfrank-personal: Same here, no forward reference please.

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+        self.app = app
+
+    async def __call__(self, scope: "Scope", receive: "Receive", send: "Send") -> None:
```
totalfrank-personal: Same here, if we don't need to, let's not have forward reference.

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+                    (b"content-type", b"application/json"),
+                    (b"content-length", str(len(body)).encode("ascii")),
+                ],
+            })
+            await send({"type": "http.response.body", "body": body})
+            return
+        await self.app(scope, receive, send)
+
+
+def install_worker_runtime_guard(app: "FastAPI") -> None:
```
totalfrank-personal: Same here, no forward reference please.

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+    Starlette builds the stack lazily on the first ASGI call, so wrapping the
+    *builder* puts the guard outside everything that build produces, including
+    Starlette's own ``ServerErrorMiddleware``. It holds however the stack is
+    reached: built fresh in a worker after the fork, or inherited already-built
+    from a parent that had served a request, since the wrapper is what that
+    parent built too.
+    """
+    build_stack = app.build_middleware_stack
+
+    def build_with_guard() -> "ASGIApp":
```
totalfrank-personal: Same here, no forward reference.

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,410 @@
...
+        return RequireWorkerRuntime(build_stack())
+
+    app.build_middleware_stack = build_with_guard  # type: ignore[method-assign]
+
+
+def finalize_worker_runtime(app: "FastAPI", *, profile: "DeployProfile") -> None:
```
totalfrank-personal: Same here, no forward reference.

totalfrank-personal (commented):
Few more comments, please take a look.

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: Your understanding is right, and the part you're unsure about is the important part — let me sharpen it.

**"It's still two different objects, right? Processes can't share memory."** Correct. `fork()` gives the child a copy-on-write copy of the parent's address space, so the child's `Injector`, its `Engine`, its pool objects are all *distinct Python objects* from the parent's. Nothing in memory is shared, and mutating one doesn't affect the other.

**But that's not where the damage is.** The problem isn't the memory, it's what the memory *points at*. `fork()` also duplicates the file descriptor table, and a duplicated fd is not a copy of the resource — both fds refer to the **same** kernel object. So after the fork you have two `Engine` objects (copies) whose pools hold fds pointing at the *same* TCP connections to the database. That part genuinely is shared across processes.

Concretely, for a DB connection:

- Parent and child both think they own connection #1 and both write to it. Two independent request streams interleave their bytes on one socket, and the server sees a corrupted protocol conversation — you get another session's result set, a "packet out of order", or a hang.
- Whichever process closes first tears the connection out from under the other.
- Server-side session state (transaction, temp tables, prepared statements, `SET` variables) belongs to the connection, so the two processes silently share it.

The same shape applies to anything else holding an fd or kernel handle: the tracer's exporter socket, an HTTP client's keep-alive pool, a file lock.

Two related fork hazards worth knowing, since they're why "just don't use the inherited objects" isn't sufficient either:

- **Locks.** `fork()` keeps only the calling thread. A `threading.Lock` that some *other* thread held at fork time is copied in the locked state, and the thread that would have released it doesn't exist in the child — so the first acquire in the child deadlocks forever. Connection pools and loggers are full of such locks.
- **Threads.** Background threads (pool janitors, tracer flushers, the task worker) are simply gone in the child. Objects that depend on them look alive and quietly do nothing.

So "the data of them will be the same" is true but understates it: the objects are separate, the *kernel resources behind them* are shared, and that's exactly what cannot be fixed from inside the child — which is why the seam refuses rather than trying to rebuild. Every one of these is avoided by forking *before* any of it exists, which is what `preload` mode is for.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: Yes, exactly that — and the fact that you had to ask means the sentence was badly written, so I've rewritten it in 794bb2c1 rather than just explaining it here.

The mechanism: `fork()` copies the module global along with everything else, so with a boolean the child reads `finalized == True` and has no way to tell whether *it* set that or its parent did. It would skip its own finalize and serve on the parent's injector and the parent's connections.

A pid is self-identifying. `_finalized_pid == os.getpid()` can only be true in the process that actually wrote the value, so the same inherited bytes answer a different question in the child: "somebody else initialized", not "you did". No extra bookkeeping, no reset hook to remember on the fork path — the comparison just stops being true in any process that didn't do the work.

The docstring now says that instead of the compressed "makes the flag meaningless in any process that did not set it", which stated the consequence and left the mechanism implicit:

> `fork` copies the flag along with everything else, so a boolean would read `True` in the child and the child could not tell whether it or its parent had set it — it would skip its own finalize and serve on the parent's injector and the parent's connections. A pid is self-identifying: `_finalized_pid == os.getpid()` can only be true in the process that actually wrote it, so the inherited value answers "somebody else initialized" rather than "you did".

---

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: You're right — dropped here and at the other seven sites in 794bb2c1.

The quotes were redundant for a reason worth naming, since it applies to the whole file: `boot.py` has `from __future__ import annotations` (PEP 563), so *every* annotation is stored as a string at runtime whether or not I quote it. The explicit quotes bought nothing and just made the signatures harder to read.

The `TYPE_CHECKING` imports stay, which is what keeps the unquoted names legal — the annotations are never evaluated, so `FastAPI`/`Injector`/`ASGIApp` not existing at runtime is fine. Verified rather than assumed: importing the module works and `finalize_worker_runtime.__annotations__` still comes back as `{'app': 'FastAPI', 'profile': 'DeployProfile', 'return': 'None'}`. Nothing here is passed to `typing.get_type_hints()` (no route handler, no `@inject`), so nothing tries to resolve those names.

I also did line 316, which you hadn't flagged but is the same pattern — eight sites total, no behavior change. 22/22 boot tests still pass.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: Done in 794bb2c1 — `def require_worker_injector(app: FastAPI) -> Injector:`. See the thread on line 78 for why they were safe to drop (PEP 563 is already in force in this file).

---

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: Done in 794bb2c1 — `def __init__(self, app: ASGIApp) -> None:`. See the thread on line 78 for the reasoning.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: Done in 794bb2c1 — `async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:`. See the thread on line 78 for the reasoning.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: Done in 794bb2c1 — `def install_worker_runtime_guard(app: FastAPI) -> None:`. See the thread on line 78 for the reasoning.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: Done in 794bb2c1 — `def build_with_guard() -> ASGIApp:`. See the thread on line 78 for the reasoning.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
totalfrank: Done in 794bb2c1 — `def finalize_worker_runtime(app: FastAPI, *, profile: DeployProfile) -> None:`, and the same for `_install_worker_runtime` further down, which you hadn't flagged but is the same pattern. See the thread on line 78 for the reasoning.

---

totalfrank (commented):
Round 7, on `794bb2c1` — the one new commit since my round-6 review at `c78363e6`.

**Part A — open findings:** all six Codex P2 findings from rounds 1–6 are already resolved (verified fixed and closed in prior rounds); nothing of mine is outstanding. The nine review comments `totalfrank-personal` left on `c78363e6` (seven forward-reference nits + two clarifying questions) are a separate human reviewer's thread, not one of my filed findings, so I'm not replying/resolving on their behalf — but see Part B, since the new commit addresses all of them.

**Part B — review of `794bb2c1`:** drops the quoted forward references at all eight annotation sites in `boot.py` (`BootMode.detect`, `require_worker_injector`, `RequireWorkerRuntime.__init__`/`__call__`, `install_worker_runtime_guard`, `build_with_guard`, `finalize_worker_runtime`, `_install_worker_runtime`) and rewrites the pid-vs-boolean docstring paragraph. Verified safe: the module has `from __future__ import annotations` (line 49), and `FastAPI`, `Injector`, `ASGIApp`/`Scope`/`Receive`/`Send`, and `DeployProfile` are all imported only under `if TYPE_CHECKING:` (lines 57-62) — so with PEP 563 in effect, quoted vs. unquoted annotations are byte-identical at runtime (both become strings, never evaluated). No behavior change, matches the commit's own verification claim. The rewritten docstring paragraph is accurate and directly answers the two questions `totalfrank-personal` asked in-thread. No issues found.

- Round 1 (`c85f8397`, own review) — clean, no findings.
- Codex P2, `boot.py` re-finalize-after-fork stacking — **fixed** `a97d3c97`, resolved round 2.
- Round 2 — no new issues.
- Codex P2, same hole via the failure door — **fixed** `95741b88`, resolved round 3.
- Round 3 — no new issues.
- Codex P2, fail closed when lifespan isn't driven — **fixed** `4ca0e0f6` (`RequireWorkerRuntime`), resolved round 4.
- Codex P2, stale docstring/spec contract — **fixed** `4ca0e0f6`, resolved round 4.
- Round 4 — no new issues.
- Codex P2, lifespan gated on injector presence not finalization — **fixed** `c0f0bfd6`, resolved round 5.
- Round 5 — no new issues.
- Codex P2, guard installed innermost via `add_middleware` — **fixed** `c78363e6`, resolved round 6.
- Round 6 — no new issues.
- `totalfrank-personal` review (human, on `c78363e6`): 7 forward-reference nits + 2 clarifying questions — nits fixed in code and questions answered in-thread by the author in `794bb2c1`; author's call whether to resolve those threads.
- Round 7 (`794bb2c1`, this round) — no new issues.

**Open count: 0** (of mine). CI green on `794bb2c1`.

---

src/backend/src/agentclaw/community/adapters/http/boot.py
```diff
@@ -0,0 +1,412 @@
...
+        install_singlebox_coverage_middleware(app)
+
+    # ── 7. Runtime-mode-conditional routers (bound by DI: empty in prod,
+    # populated in local boots via the test column's app-services module). This
+    # is the one route registration that cannot happen during construction: it is
+    # the only place in the tree that resolves from the injector outside a
+    # request, so it has to wait for the injector this function just built. The
+    # app still does not branch on mode — DI decides what gets mounted.
+    for router in injector.get(OptionalRouters).routers:
+        app.include_router(router)
```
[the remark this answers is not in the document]
totalfrank: Disputing this one — it doesn't reproduce on the pinned FastAPI — but you were right about the test blind spot, so I've added coverage in 7150c224.

**Why it doesn't reproduce.** FastAPI 0.138's `openapi()` no longer caches unconditionally:

```python
routes_version = self.router._get_routes_version()
if not self.openapi_schema or self._openapi_routes_version != routes_version:
    self.openapi_schema = get_openapi(...)
    self._openapi_routes_version = routes_version
return self.openapi_schema
```

`include_router` bumps the routes version, so the mount in `_install_worker_runtime` invalidates the cache implicitly. That's the same `_get_routes_version` machinery as the lazy `_IncludedRouter` change in this version — the classic `if not self.openapi_schema:` behavior you're describing was true for years and is what I'd have expected too.

Measured, with no cache clearing anywhere in the script:

```
A master cached schema?       True
B master has /local/sql:      False
C cache after include_router? True     ← still populated, as you said
D worker has /local/sql:      True     ← but regenerated on read
```

So C confirms your observation exactly — the cache object *is* inherited and non-empty — and D is why it doesn't bite.

**Where you were right regardless.** `route_table()` clears `openapi_schema` on both sides, so if the claim had held, every existing test would have sailed past it. That's a real gap, and it left the seam depending on a FastAPI implementation detail that nothing asserted. `test_a_warm_openapi_cache_does_not_hide_routers_finalize_mounts` now covers it: it warms the cache pre-finalize, deliberately does *not* clear it, and asserts the worker's served document contains the route it actually routes. Its failure message names the fix — clear `app.openapi_schema` after the mount loop — so whoever bumps FastAPI into a version that caches unconditionally gets a failing test and the remedy in one place.

I'd rather assert the property than add an invalidation line that is dead code against the pinned version; if you'd prefer the belt-and-braces `openapi_schema = None` anyway, say so and I'll add it.

23/23 boot tests.

---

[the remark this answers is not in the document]
totalfrank: Verified independently this round: downloaded fastapi==0.138.2 (the version pinned in `src/backend/uv.lock`) and read `applications.py`/`routing.py` directly rather than taking the claim on trust.

`FastAPI.openapi()` (applications.py:1084-1085):
```python
routes_version = self.router._get_routes_version()
if not self.openapi_schema or self._openapi_routes_version != routes_version:
    self.openapi_schema = get_openapi(...)
```
`APIRouter.include_router` calls `self._mark_routes_changed()` (routing.py:3108), which bumps `_routes_version` (routing.py:2388-2389), and `_get_routes_version` sums that across the router tree. `_install_worker_runtime`'s `app.include_router(router)` loop over `OptionalRouters` (boot.py:410) goes through exactly this path, so the version bump is real and unconditional — not an implementation detail that happens to hold today.

The new test (`test_a_warm_openapi_cache_does_not_hide_routers_finalize_mounts`) matches this precisely: it doesn't need an actual `os.fork()` because the cache-invalidation logic is pure in-process bookkeeping on the router version counter, independent of process boundaries — a forked child would inherit the same cache in the same state a same-process warm-then-mutate sequence produces. Its guard assertions (`cached_before is True`, `before_has_local_sql is False`) correctly rule out the test silently no-oping if construction ever changes to mount the router eagerly.

The dispute is sound. Resolving.

---

totalfrank (commented):
Round 8, on `7150c224` — the one new commit since my round-7 review at `794bb2c1`. Test-only change (48 lines, one file), no production code touched.

**Part A — open findings.** One item was open going into this round: a Codex P2 finding filed *after* round 7 (`boot.py:410`, "clear the inherited OpenAPI cache after mounting routers"), which the author disputed and backed with a new test in `7150c224`. Verified independently rather than taking the reply on trust: downloaded fastapi==0.138.2 (the version pinned in `src/backend/uv.lock`) and read `applications.py`/`routing.py` directly. `FastAPI.openapi()` regenerates whenever `router._get_routes_version()` differs from the cached value, `include_router` calls `_mark_routes_changed()` unconditionally, and `_install_worker_runtime`'s `app.include_router(router)` loop over `OptionalRouters` goes through exactly that path — so the cache does get invalidated on finalize's router mount, and the dispute is correct. The new test exercises this without an actual `os.fork()`, which is fine here since the invalidation logic is pure in-process bookkeeping on a version counter, unaffected by process boundaries. Replied confirming and resolved the thread.

**Part B — review of `7150c224`.** Purely additive test (`test_a_warm_openapi_cache_does_not_hide_routers_finalize_mounts`): warms the cache pre-finalize, asserts the DI-conditional router is genuinely absent beforehand (guards against the test silently no-oping if construction changes), then asserts the served document matches what's actually routed after finalize. No correctness, over-reach, or coverage issues found.

- Round 1 (`c85f8397`, own review) — clean, no findings.
- Codex P2, `boot.py` re-finalize-after-fork stacking — fixed `a97d3c97`, resolved round 2.
- Round 2 — no new issues.
- Codex P2, same hole via the failure door — fixed `95741b88`, resolved round 3.
- Round 3 — no new issues.
- Codex P2, fail closed when lifespan isn't driven — fixed `4ca0e0f6` (`RequireWorkerRuntime`), resolved round 4.
- Codex P2, stale docstring/spec contract — fixed `4ca0e0f6`, resolved round 4.
- Round 4 — no new issues.
- Codex P2, lifespan gated on injector presence not finalization — fixed `c0f0bfd6`, resolved round 5.
- Round 5 — no new issues.
- Codex P2, guard installed innermost via `add_middleware` — fixed `c78363e6`, resolved round 6.
- Round 6 — no new issues.
- `totalfrank-personal` review (human, on `c78363e6`): 7 forward-reference nits + 2 clarifying questions — nits fixed and questions answered in `794bb2c1`; author's call, not mine to resolve.
- Round 7 (`794bb2c1`, own review) — no new issues.
- Codex P2, `boot.py:410` — OpenAPI cache could hide finalize's router mounts. **Disputed by the author, verified sound this round** (does not reproduce on pinned fastapi==0.138.2; coverage added in `7150c224`), resolved this round.
- Round 8 (`7150c224`, this round) — no new issues.

**Open count: 0.** PR is clean, cleared as of round 8 (`7150c224`).

---

xianmuyq (dismissed):

totalfrank:
`Backend unit tests` went red on `7150c224`, and it was this PR's failure against the current base — not a flake. Fixed and pushed in `1de6c515`.

**What failed**

```
FAILED tests/community/architecture/test_no_oversized_modules.py::test_no_new_oversized_files
  adapters/http/app.py: 1001 lines (cap 1000)
```

The file is 999 lines on every commit of this branch, so this only reproduces against the merge. CI tests the merge with `dev`, and `dev` moved: #2282 added `bot_common_config_router`'s import and its `include_router` to `app.py`. Two lines, in regions this branch doesn't touch — the merge itself is clean — but 999 + 2 = 1001.

**Fix**

Merged `dev` and reclaimed headroom rather than shaving the single offending line, since shaving one would hand the identical failure to whoever adds the next router. `app.py`'s module docstring and two comment blocks restated what `boot.py` already documents in full; they're now a summary plus a pointer. **1001 → 990.**

**Whose problem this is, stated plainly**

`app.py` was **998** at this PR's base and #2282 took it to exactly **1000**, so `dev` is sitting at the cap today with or without this branch. At 990 this PR leaves the file ten lines better than `dev` — but that's headroom, not a fix, and the next few router additions will consume it.

The real answer is that `app.py` carries ~400 lines of exception handlers that want their own module. I'm deliberately not doing that here: the allowlist in `test_no_oversized_modules` makes exactly this argument about the two `openapi_v1` files — the split "should just be its own change, not a rider on a feature." Worth someone picking up as a follow-up, because the cap will bite the next person otherwise.

I'd also flagged this collision risk in the PR description rather than preventing it, which is why it landed as a red build instead of never happening.

Verified before pushing: `tests/community/architecture` + the boot suite, 325 passed. CI is re-running on the merge.

---

xianmuyq (approved):

totalfrank (commented):
Round 9, on `45a8ce1cd8` and `1de6c5158f` — the two new commits since my round-8 review at `7150c224`.

**Part A — open findings.** None. All 16 threads from rounds 1–8 (6 Codex P2 findings + 9 human forward-reference/clarifying threads + my own round reviews) were already resolved going into this round; nothing new to verify.

**Part B — review of the new commits.**

- `45a8ce1cd8` — merge of `origin/dev` into this branch. Brings in unrelated work (`#2282`'s `bot_common_config` router, BaaS ORM changes, aicoding service changes). Out of scope for this review: none of it touches the two-phase boot seam, and the only overlap with this PR's own files is `app.py` picking up `#2282`'s 2-line router registration, in a region this branch doesn't otherwise touch.
- `1de6c5158f` — the one commit with actual content from this PR. Doc-only: trims `app.py`'s module docstring and two comment blocks (999 → ~977 lines pre-merge, 1001 → 990 post-merge) to fix a real CI failure — the merge with `dev` pushed `app.py` over the 1000-line cap enforced by `test_no_oversized_modules::test_no_new_oversized_files`, since `#2282` had already taken `dev`'s copy to exactly 1000. Verified the trimmed prose still says everything load-bearing: it correctly points to `boot.py` as the source of truth for the phase split, the guards, and the fork reasoning (all of which `boot.py` still documents in full — untouched by this commit), and the shortened Phase 1/Phase 2 description doesn't drop or contradict anything the removed text asserted. No logic changed. `Backend unit tests` is green on this head, confirming the line-count fix actually works, not just that it was intended to.

No issues found in either commit.

- Round 1 (`c85f8397`, own review) — clean, no findings.
- Codex P2, re-finalize in a child forked after successful finalize stacks a second middleware set — fixed `a97d3c97`, resolved round 2.
- Round 2 — no new issues.
- Codex P2, same hole via the failure door — fixed `95741b88`, resolved round 3.
- Round 3 — no new issues.
- Codex P2, fail closed when the host never drives the lifespan — fixed `4ca0e0f6` (`RequireWorkerRuntime`), resolved round 4.
- Codex P2, stale docstring/spec contract — fixed `4ca0e0f6`, resolved round 4.
- Round 4 — no new issues.
- Codex P2, lifespan gated on injector presence not finalization — fixed `c0f0bfd6`, resolved round 5.
- Round 5 — no new issues.
- Codex P2, guard installed innermost via `add_middleware` — fixed `c78363e6`, resolved round 6.
- Round 6 — no new issues.
- `totalfrank-personal` review (human, on `c78363e6`): 7 forward-reference nits + 2 clarifying questions — nits fixed and questions answered in `794bb2c1`, author's call and already actioned.
- Round 7 (`794bb2c1`, own review) — no new issues.
- Codex P2, OpenAPI cache could hide finalize's router mounts — disputed by the author, verified sound (does not reproduce on pinned fastapi==0.138.2), coverage added `7150c224`, resolved round 8.
- Round 8 (`7150c224`, own review) — no new issues.
- Round 9 (`45a8ce1cd8` + `1de6c5158f`, this round) — no new issues; `45a8ce1cd8` out of scope (unrelated `dev` merge).

**Open count: 0.** PR is clean, cleared as of round 9 (`1de6c5158f`). CI green on this head.

---

PR #2283, opened 2026-09-17
fix(skills): list package directories in legacy fallback

FreddieSun:
## Problem

The Local Skill legacy fallback checked package-directory existence through the generic filesystem `exists()` probe. The BaaS implementation probes files first with `read_file(path)`, so runtimes that correctly reject `read(directory)` fail replacement before the supported directory-list operation is attempted.

## Solution

Make `LocalSkillPackageStorage` own its directory semantics:

- determine package existence with `list_dir`;
- reuse that package-aware probe during first-upload cleanup and replacement target checks;
- keep generic file-or-directory `DeviceFileSystem.exists()` unchanged.

This preserves one shared legacy fallback without runtime-specific branches.

## Validation

- Regression test reproduced the prior `read(directory)` call before the fix and passes with a single `file.list` call after it.
- `98 passed, 10 skipped` across Local Skill package storage/upload, BaaS filesystem, local storage contract, and Backend architecture tests.
- Backend SAST checks passed for both modified Python files.
- `git diff --check` passed.

## Compatibility and risk

No public API, capability, locator, schema, or Engine package-apply contract changes. New Engine and Teclaw package paths are unchanged. The change only affects the per-file legacy adapter and aligns package-root probing with the existing `DeviceFileSystem.list_dir` directory contract.

## Spec

Implements the legacy-adapter compatibility requirement in `docs/specs/2026-09-17-local-skill-package-apply/spec.md`.


totalfrank (approved):
LGTM — reviewed `fd3ba46` (3 files: `factories.py`'s `LocalSkillPackageStorage.exists()`/`prepare()`/`copy_to()`, the AGENTS.md note, and a regression test). The switch from the generic `read_file`-first `exists()` probe to a direct `list_dir` call correctly aligns with the `None` = absent / list (possibly empty) = present contract used consistently across `BaasDeviceFileSystem`, `TeclawDeviceFileSystem`, and `LocalDeviceFileSystem`, and avoids the prior dependence on `read_file` failing in a specific way for directory paths. Ran the affected suites locally (`test_local_skill_package_storage_io.py`, `test_local_skill_upload_service.py`, and the full `skill_center`/`devices` test dirs) — 1985 passed, 13 skipped, no failures. No bugs found in the new code.

---

PR #2284, opened 2026-09-17
fix(backend): gate /api/public/bots/* behind a shared internal token

totalfrank:
## Problem

`PATCH /api/public/bots/{bot_id}/ext` carried no authentication and no real ownership check.

The handler looked the Bot up by `bot_id`, read `owner_id` off **that same row**, and passed the pair to `BotRepository.update_by_owner`:

```python
total, items = bot_repo.list_by_conditions(bot_id=bot_id, page=1, page_size=1)
bot = items[0]
owner_id = bot.get("owner_id")     # ← the target Bot's own owner
...
bot_repo.update_by_owner(bot_id, owner_id, {"ext": ext})
```

`update_by_owner` filters on `bot_id = ? AND owner_id = ?`, but both values come from the record being modified, so the owner filter always matches. The check reads like an ownership check and enforces nothing. With no `Depends(get_current_user)` on the route either, any caller could rewrite any Bot's whitelisted `ext` fields (`arch_domain`, `is_domain_bot`).

`GET /api/public/bots/{bot_id}/appcoding-bots` on the same router was likewise open.

## Solution

Add `InternalApiToken` — the general-purpose sibling of the existing per-operation `DormantInternalToken` and `SkillCenterInternalToken`, which gate `/api/internal/dormant/*` and `/api/internal/skill-center/*`. Those are granted independently because each names one operation; this one is the shared grant for routes whose caller is a platform component rather than a user, so there is no end-user credential to authorize against.

- `adapters/http/internal_auth.py` — reusable `verify_internal_api_token` Depends. Compares with `hmac.compare_digest`; any auth problem answers 401 without saying which. The `Authorization` header is declared optional so a request that omits it is answered by the guard (401) rather than by FastAPI request validation (422).
- `di/modules/internal_api_token_module.py` — resolves the token through `SecretResolver` from `SecretNamesConfig.internal_api_token`.
- `bot_public/public_noauth_router.py` — carries the dependency at **router** level, so a route added there later inherits the guard instead of shipping open. `owner_id` stays as the write's locating key, now commented as such rather than reading like an authorization check.

Resolution follows `TcFileUploadIntegrationModule.tc_file_service_token` rather than the dormant/skill-center providers, and the difference is deliberate: an unconfigured deployment falls back to the published local constant **only in a local profile**. A corp deployment with no secret name registered resolves to `""` and 401s every request — a token printed in this repository must not be what stands between the internet and a write. A raising resolver is failure-closed in every profile.

Scope: this PR guards the `bot_public` noauth router only. `cron_noauth_router` (`POST /api/public/cron/...`, triggers cron for any bot/user) and `workitem_noauth_router` (`/api/public/dima/...`) have the same shape and can adopt the same guard with one `Depends` line, but doing it here would break their callers in the same change.

Incidental: `di/config.py` was at 993 lines against the 1000-line architecture cap, so the four resolved-token dataclasses move to `di/config_tokens.py` (values resolved at DI time, never yaml) and are re-exported from `di/config.py`. No import site changes.

## Validation

Run with `.venv/bin/python -m pytest` from `src/backend`:

- Full community suite: `tests/community` — 19039 passed, 46 skipped, 2 failed. Both failures (`test_bot_build_service_skill_artifact.py::test_pool_build_uses_the_versioned_filesystem_snapshot_when_runtime_cannot_write_nfs[openclaw|claude_code]`) reproduce identically on the base commit with the change stashed; they are not from this PR.
- New: `tests/community/di/test_internal_api_token.py` (provider branch by branch, including corp-with-no-name → `""` and raising-resolver → `""`) and `tests/community/adapters/http/test_internal_auth.py` (token match, wrong token, missing prefix, absent header, empty configured token).
- `tests/community/api/bot_public/test_public_noauth_router.py` — added `TestInternalTokenGuard`: both routes answer 401 without the token, and `update_by_owner` / the service is never called.
- Endpoint cases: added `unauthenticated` cases for `GET /appcoding-bots` and `PATCH /ext`; the PATCH case asserts the seeded Bot's `ext` is unchanged after the refusal.
- Gates: coverage gate (the now-covered `GET /api/public/bots/{bot_id}/appcoding-bots` baseline line deleted), effective-config goldens regenerated for the new `secret_names` field, oversized-module gate, ruff on all changed files.

Not run: anything requiring corp infrastructure (Mist/`SecretResolver` against a real secret store) — the corp resolution path is covered by mocks in the DI test only.

## Compatibility and risk

**Breaking for existing callers of these two routes.** Whatever calls `PATCH /api/public/bots/{bot_id}/ext` and `GET /api/public/bots/{bot_id}/appcoding-bots` today must start sending `Authorization: Bearer <token>`. There are no in-repo callers (only tests), so the caller is external and needs the token provisioned with it.

Deployment order matters: register `secret_names.internal_api_token` and its value **before** this ships to a corp environment, or both routes 401 (that is the intended failure mode, not a regression). Singlebox/community/test keep working with no configuration via the local fallback constant.

Rollback: revert the commit; the routes return to their previous unauthenticated behavior.

https://claude.ai/code/session_01HNV2zy1vyn981DXu6sJY9R

---
_Generated by [Claude Code](https://claude.ai/code/session_01HNV2zy1vyn981DXu6sJY9R)_


totalfrank:
**CI: `BCS architecture checks` is red, and it is not this PR's.** Everything else on this head is green (Backend unit tests, BaaS, Gateway, Engine, Sandbox-proxy, BCS e2e, Singlebox coverage, Community Boundary Check, Check and build, Build & Test, Validate PR title).

Three things say the failure is pre-existing repo state rather than this change:

1. **The gate's own base-vs-head check passes.** `check-baseline-not-growing.sh` compares the base branch against this head and reports no growth:
   ```
   [BASELINE] scripts/ci/baseline/forbidden-symbols.txt:    base=1,  head=1
   [BASELINE] scripts/ci/baseline/trait-naming.txt:         base=1,  head=1
   [BASELINE] scripts/ci/baseline/conformance-missing.txt:  base=10, head=10
   PASS [BASELINE]
   ```
2. **The failing gates are all in `src/bcs` (Rust).** `Failed: DEP-1~8 LINT-1 LINT-4 TEST-1 PR-1 R25` — R25 is the conformance-entry rule (`impl BotActorConfigRepoPort for DbBotActorConfigStore … lacks conformance entry`, and ~25 more of the same shape). This PR's 16 files are all under `src/backend/`; it adds no Rust and touches no BCS crate.
3. **Every recorded run of this workflow has failed.** All 22 `Architecture Checks` runs in the repo's history conclude `failure`, across different authors and unrelated branches (taskguard/TypeScript, skill-center, frontend). The job is red for everyone, not for this diff.

No fix exists to port: this is not one broken commit but the BCS conformance backlog the `conformance-missing.txt` baseline tracks (`base=10`), plus `R25.3: cargo test did not discover conformance_ tests`. Fixing it means adding conformance entries across the BCS crates, which is a separate change and well outside a backend auth fix — widening this PR to cover it would be wrong.

I re-ran the job once to confirm; a second failure is the same pre-existing state. Flagging for whoever owns `src/bcs`: this gate is currently red on every PR, so it is not signalling anything about the changes it runs on.

---

PR #2292, opened 2026-09-17
feat(backend): admit delegated apps to bot creation via user-level grants

totalfrank:
## Problem

Every admission mode in `adapters/http/openapi_v1/admission.py` that admits an application acting alone is bot-level: it checks a grant on the bot the request addresses. The creation operations (`POST /openapi/v1/bots`, `POST /openapi/v1/bots/with-manifest`, `POST /openapi/v1/bots/local`) have no bot to address when the request arrives, so they were `REFUSED` to applications outright, and their status polls with them. The table itself named this as the missing "user-level admission mode". An integration that exists to create bots for a user could not do so unattended.

A second gap follows from the first: even if an application were admitted to a creation, it would hold no grant on the bot it created and could not reach it, including the bot-scoped poll that completes a pending creation.

## Solution

**A user-level delegation record** (`ac_user_app_grant` + `ac_user_app_grant_log`, same two-table shape as the bot grant) meaning "app A may act as user U where no bot is addressed". The user grants, lists and withdraws it on `/openapi/v1/bots/authorized-apps` (granting needs user + app on the wire, like the bot-scoped group; listing and withdrawing need only the user). All three are `REFUSED` to a machine caller, so an application cannot delegate to itself.

**A new admission mode, `USER_DELEGATED`**, checked by `require_delegated_user` before the handler runs. The three creations move from `REFUSED` to it. An application with no live delegation from the named user is refused with the same 404 a missing bot grant gets. `test_admission_inventory.py` holds every `USER_DELEGATED` route to the dependency, as it does for the bot-grant modes.

**The creating application is granted the bot it creates, when the creation starts.** `BotAppGrantService.grant_for_creation` writes an ordinary bot grant without the liveness check (there is no bot yet), shared by the three routes through `creation_grant.py`. Written at the start rather than on completion because on the 202 path nothing is created until the user authorizes, and the polls that complete it are bot-scoped. Those polls (`GET …/{bot_id}/with-manifest/status`, `GET …/{bot_id}/local/auth-status`) therefore move to `GRANT_CHECKED_OWN_BOT` like the existing `auth-status` poll. The bot's owner sees the grant on the bot's authorized-apps listing and can withdraw it there. A creation that fails to start withdraws it; the manifest creation job sweeps it when it gives up before a bot exists; bot deletion sweeps it like any other. A creation abandoned mid-flight on the plain `POST /bots` path leaves an inert row naming a bot that never existed (the app listing filters against live bots), the same lifetime the abandoned Passport application already has.

**The delegation counts as proof of relationship for `USER_GATED` reads.** `ActingCaller.holds_delegation()` replaces the six copies of "any bot grant" (ceiling, routines aggregate, devices, work orders, Spaces) with "user-level delegation or any bot grant", so an application onboarded to create bots can read the ceiling and devices before it holds any bot grant. It reaches no existing bot on its own.

**Gateway:** `POST /openapi/v1/bots` leaves the human-only list and inherits the wide `/openapi/v1/bots/**` rule; `POST /openapi/v1/bots/authorized-apps` requires user + app, its `GET`/`DELETE` a user.

**Reviewed and deliberately left human-only:** Space creation and administration, MCP account configuration, bot logs, `GET /openapi/v1/org/user`, and the consent routes themselves. Each keeps its own reason in `admission.py`.

Alternative rejected: writing the bot grant on completion instead of at the start. That would have needed a special case in each bot-scoped poll ("no grant yet, but the creation is pending and the app is delegated"), which is exactly the second mechanism the seam's history warns against.

## Validation

Backend (`uv run pytest`, from `src/backend`):
- `tests/community/adapters/http/openapi_v1` + `tests/community/architecture`: 1993 passed, 2 skipped (after updating three pinned counts: user_id-taking operations 234 → 237, `bot_id` placement `none` 104 → 107, reserved component names)
- `tests/community/core/bot_config_manifest`, `core/bot_app_grant`, `core/bot_management`, `endpoints/test_openapi_authorized_apps.py`, `endpoints/test_openapi_local.py`: 2325 passed
- New: `test_user_delegated_creation.py` (admission, creation grant, poll admission, failure withdrawal, human path unchanged), `authorized_apps/test_user_router.py` (real service over SQLite), `core/bot_app_grant/test_user_grant_service.py` (repository + `grant_for_creation`), two creation-job sweep cases
- `ruff check` on all changed files: clean

Gateway: `tests/unit/core/authn/test_route_security.py`: 81 passed (with two new tests).

Not run: the full backend suite end to end, and anything needing a live database or Passport.

## Compatibility and risk

- Two new tables; migration at `core/bot_app_grant/sql/2026_09_17_user_app_grant.sql`. `create_all` only runs for the local SQLite plugin, so persistent deployments need the DDL applied.
- Public contract additions: three routes under `/openapi/v1/bots/authorized-apps`; the `USER_DELEGATED` admission value. Existing human callers are unchanged on every route (asserted by the seam tests).
- `BotAppGrantServiceProtocol` gains `grant_for_creation`; `BotCreateWithManifestHandler` gains an optional `grant_sweep_provider`.
- Rollback: revert the commit; the new tables can be dropped or left empty.

## Spec

`docs/openapi-v1/README.md` carries a dated changelog entry describing the mode, the creation grant, and what stays human-only.

https://claude.ai/code/session_01H5a2uhPboWRJYWgkrvfnxZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01H5a2uhPboWRJYWgkrvfnxZ)_


src/backend/src/agentclaw/community/core/repository/protocols/bot/user_app_grant.py
```diff
@@ -0,0 +1,52 @@
...
+    """Live user→app delegations, plus the append-only history behind them.
+
+    Keyed on ``(app_id, user_id)`` within the tenant and env. There is no owner
+    and no bot anywhere in this contract, and adding one would turn it back
+    into the bot grant — the record exists precisely for the operations that
+    have neither.
+    """
+
+    @abstractmethod
+    def grant(self, data: Dict[str, Any]) -> "UserAppGrantRecord":
```
totalfrank-personal: No forward reference if you don't need to.

src/backend/src/agentclaw/community/core/repository/protocols/bot/user_app_grant.py
```diff
@@ -0,0 +1,52 @@
...
+    @abstractmethod
+    def revoke(self, user_id: str, app_id: int) -> bool:
+        """Withdraw one delegation, appending a ``revoked`` event.
+
+        Returns whether a live row was removed, so the service can answer
+        "there was nothing to remove" distinctly from "removed".
+        """
+
+    @abstractmethod
+    def find(self, user_id: str, app_id: int) -> Optional["UserAppGrantRecord"]:
```
totalfrank-personal: No forward reference if you don't need to.

src/backend/src/agentclaw/community/core/repository/protocols/bot/user_app_grant.py
```diff
@@ -0,0 +1,52 @@
...
+        """The live delegation for this pair, or ``None``."""
+
+    @abstractmethod
+    def list_for_user(self, user_id: str) -> List["UserAppGrantRecord"]:
```
totalfrank-personal: No forward reference if you don't need to.

src/backend/src/agentclaw/community/core/repository/protocols/bot/user_app_grant.py
totalfrank-personal: For a protocol, dict might not be a good paramter type right? It's hard to tell what it provides. I prefer individual fields.

src/backend/src/agentclaw/community/core/bot_app_grant/bot_app_grant_service_protocol.py
```diff
@@ -45,6 +48,30 @@ def grant(
         for one that succeeded.
         """
 
+    def grant_for_creation(
```
totalfrank-personal, line 54: For all methods in this protocol, annotate them with @abstractmethod.

src/backend/src/agentclaw/community/core/bot_app_grant/bot_app_grant_service_protocol.py
```diff
@@ -104,4 +131,37 @@ def list_for_app(self, *, app_id: int, user_id: str) -> list[BotAppGrantRecord]:
...
+class UserAppGrantServiceProtocol(Protocol):
+    """Grant, withdraw and read user→app delegations.
+
+    The account-level record: *"app A may act as user U where no bot is
+    addressed"*. It is what admits an application to a ``USER_DELEGATED``
+    operation — one that acts for the named user before any bot exists — and
+    it reaches no existing bot on its own.
+    """
+
+    def grant(
```
totalfrank-personal, line 154: For all methods in this protocol, annotate them with @abstractmethod.

src/backend/src/agentclaw/community/core/bot_app_grant/services/grant_service.py
```diff
@@ -131,6 +131,61 @@ def grant(
             }
         )
 
+    def grant_for_creation(
```
totalfrank-personal, line 134: For the class of this method, let's make sure it explicitly extends the protocol it's implementing.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/authorization.py
```diff
@@ -208,6 +208,15 @@ def __repr__(self) -> str:  # pragma: no cover - diagnostics only
         ServiceChecked(PermissionLevel.MEMBER, "…openapi_v1.authorized_apps.router"),
     ("DELETE", "/openapi/v1/bots/{bot_id}/authorized-apps/{app_id}"):
         ServiceChecked(PermissionLevel.MEMBER, "…openapi_v1.authorized_apps.router"),
+    # The user-level delegation: the caller's own account, no bot addressed.
+    # The user is the verified principal (require_user_id binds them to
+    # themselves), and the application is read off the same principal.
+    ("POST", "/openapi/v1/bots/authorized-apps"):
```
totalfrank-personal, line 219: I mean, rather than NoCheck, we should not allow other people to visit these endpoints right? I did not see why we should open these. Open to your thoughts.

totalfrank-personal (commented):

src/backend/src/agentclaw/community/core/repository/protocols/bot/user_app_grant.py
totalfrank: Done in d51b0877: the three string forward references on `UserAppGrantRecord` are gone. `from __future__ import annotations` already defers evaluation of the `TYPE_CHECKING`-only import, which is how the sibling `app_grant.py` does it.

---

totalfrank: Agreed. In d51b0877 `grant` takes keyword-only `app_id: int, app_name: str, user_id: str` on the protocol, the ORM implementation and the service call site. Left the pre-existing bot-grant repository's `dict` signature alone to keep this PR's scope.

---

src/backend/src/agentclaw/community/core/bot_app_grant/bot_app_grant_service_protocol.py
totalfrank: Done in d51b0877: every member of both `BotAppGrantServiceProtocol` and `UserAppGrantServiceProtocol` is `@abstractmethod`. Since both services inherit their protocol, a dropped method now fails at construction. The conformance gate (`test_service_api_conformance.py`) still passes: 305 architecture tests green.

---

src/backend/src/agentclaw/community/core/bot_app_grant/services/grant_service.py
totalfrank: It already does: `class BotAppGrantService(BotAppGrantServiceProtocol)` at the top of this file (line 49), and the new `UserAppGrantService` extends `UserAppGrantServiceProtocol` the same way (`services/user_grant_service.py`, line 42). With the `@abstractmethod` change just pushed, that inheritance now also enforces completeness at construction.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/authorization.py
totalfrank: They are not open. This table is the collaborator-permission dimension only: a row answers "which `PermissionLevel` must the caller hold on the addressed bot", and `NoCheck` means there is no addressed bot to hold a level on. It is the same sentinel `GET /openapi/v1/org/user` ("the caller's own verified identity") and `GET /openapi/v1/bots/ceiling` use.

What actually gates these three routes lives in the other two layers:

- `admission.py` marks all three `REFUSED`, so an app-only caller never reaches the handler (an application cannot delegate to itself).
- The gateway rule requires a verified user principal (`POST` requires user and app both).
- `require_user_id` binds `user_id` to that principal, so the only person who can grant, list or withdraw a delegation is the user it belongs to. `test_user_router.py` and `test_self_checked_routes_refuse.py` cover the refusals.

In d51b0877 I expanded the comment above these rows and the `NoCheck` reason strings to say exactly that, so the table no longer reads as if they were open. If you'd rather have a dedicated sentinel for "the caller's own account" rows, that is a small follow-up touching the existing `org/user`, `ceiling` and `devices` rows too; happy to do it if you want it.

---

totalfrank (commented):
Reviewed the full diff at the current head (commits `3bd3e9d`, `cfb9b866`, `d51b0877`) against merge-base `e418db6` — no findings.

Checked in particular:
- **Admission/authorization correctness**: `admission.py` (`USER_DELEGATED` mode, `require_user_delegation`, `holds_delegation`), `principal.py` (`require_delegated_user`, `caller_app_name`, `_user_grant_reader`), `authorization.py` and `admission_modes.py` table entries — modes, dependencies and the inventory tests (`test_admission_inventory.py`) all agree with each other, and the six `USER_GATED` call sites (ceiling, owner routines, local devices, spaces, skill routes, work orders) were all consistently switched to `holds_delegation()`.
- **Creation-grant lifecycle**: `creation_grant.py`, and its three call sites in `bots/router.py`, `bots/create_with_manifest.py`, `local/router.py` — grant-before-create with try/except withdrawal on the two cloud paths, and the desktop-allocated-id path in `local/router.py` (verified `apply_passport_before_create` always returns a non-empty `bot_id`, so the `if created_bot_id:` guard is never a live gap). The manifest creation job's sweep (`create_job.py` + `bot_app_grant_module`/`bot_management_module` DI wiring) matches the documented "grant at start, sweep on give-up" contract.
- **New persistence layer**: `models.py` (`UserAppGrantModel`/`Log`) against the SQL migration — column types, collation and keys agree exactly; repository (`user_app_grant.py`) insert-race handling via the unique constraint and `IntegrityError` retry.
- **Service API/Protocol boundary**: `UserAppGrantServiceProtocol` re-export shape, abstractmethod additions, DI bindings, and `test_service_api_conformance.py`/`test_http_adapter_layer_is_http_only.py` updates.
- **Gateway**: `application.yaml` route-security changes (`POST /openapi/v1/bots` now `user/app: optional`, new `authorized-apps` rules) against `test_route_security.py`.
- **Test coverage**: `test_user_delegated_creation.py`, `test_user_router.py`, `test_user_grant_service.py`, `test_create_job.py`, `test_create_with_manifest_routes.py`, `test_explicit_user_id.py`, `test_path_convention.py` — cover the happy path, refusal, failure-withdrawal, idempotency, and the pinned-count updates match the actual diff.
- Docs (`docs/openapi-v1/README.md` changelog, module READMEs) match the code.

Categories from the checklist that applied here: correctness, contract breakage, error/failure handling, concurrency (grant insert race), architecture boundaries (Service API/core separation), Python type contracts, and test coverage. No security, or dead-code issues found.

---

totalfrank:
**Review round 1 — no issues flagged.** Reviewed `d51b0877` (the full diff: user-level delegation record and routes, `USER_DELEGATED` admission, creation-time bot grant, plus the review fixes from this morning). The Claude Code review on this head returned no findings, all 12 CI checks are green, and the 8 human review threads from the earlier pass are resolved (7 fixed in `d51b0877`, 1 answered: the `NoCheck` rows are the collaborator-permission dimension only; admission refuses app-only callers on those routes).

Nothing is disputed. The PR now waits on its requested human reviewers.

---

xianmuyq (approved):

vzvince (approved):

PR #2333, opened 2026-09-20
fix(backend): preserve explicit CLI owner identity during scope sync

yhzhang35:
## Problem

Changing a CLI from caller to owner removes its sparse caller override. Passport reconciliation then preserves the historical caller identity, so the API reports owner while sending caller to Passport. This affects both default and custom CLIs.

## Solution

Carry the authorized CLI identity intent from the identity service into the existing reconciler, applying it to a copy of the persisted sparse overrides before building the complete MCP and CLI scope. Keep unrelated identities, bootstrap behavior, last-caller protection, and failure compensation unchanged. Add requested and outgoing identity diagnostics without logging credentials.

The change is internal to two backend core implementation files. No HTTP or Plugin API contract, schema, configuration, or deployment changes are required.

## Validation

- Before changes: 113 focused tests passed.
- Regression reproduced before the fix; after the fix, 123 focused tests passed, including default/custom CLI owner transitions, stale-state retries, full-scope preservation, bootstrap compatibility, and failure compensation.
- Independent code review passed; focused production coverage is 344/361 (95.29%) and executable changed-line coverage is 4/4 (100%).
- Full backend community regression: 19,336 passed, 43 skipped, 0 failed; line coverage 101,011/113,404 (89.07%).
- Ruff and git diff --check passed.

## Compatibility and risk

Existing reconciler callers omit the new optional request-local intent and retain their behavior. Owner remains represented by an absent sparse row. This does not repair historical drift automatically; an authorized owner update sends the explicit desired identity. No live deployment or bot mutation was performed.

## Spec

src/backend/specs/cli-owner-scope-sync/001-spec-output.md


totalfrank (approved):
LGTM.

PR #2337, opened 2026-09-20
feat(backend): split NAS/UPFS storage quota

rudyLi:
## Problem

  Bot storage quota was a unified setting: `BotStoragePolicyService.apply_to_storage` overwrote `storage.quota = get_storage_quota(env)` identically for both NAS and
  UPFS (reading common-config `bot_storage`/`storage` -> `param_value.quota`, default `1G`). Semantically NAS should keep the builder's hardcoded `1Gi` (matching the
  original layout), and only UPFS needs its quota from config. The shared override clobbered NAS's `1Gi`, which contradicts the "NAS stays 1Gi" requirement.

  ## Solution

  - `bot_config_service.py` `apply_to_storage` gains a type guard: it overwrites `storage.quota = get_storage_quota(env)` **only when `storage.type ==
  StorageType.UPFS`**; NAS is left untouched, retaining the builder's hardcoded `1Gi`.
  - `get_storage_quota` default `"1G"` -> `"1Gi"` so UPFS also defaults to 1Gi (consistent with NAS and "both 1Gi").
  - Reuses the existing config item `bot_storage`/`storage` -> `param_value.quota` (now read only by UPFS) — **no new DRM config item**, zero ops overhead.
  - Takes effect across all flows: create / restart (`_restart_bot_baas` -> `upgrade_bot` via `/update`) / publish (upgrade) / published·caller·artifact payloads all
  go through `baas_service._build_create_bot_payload` -> `managed_composer.build_storage` (calls `apply_to_storage` at :150), and `resolve_deploy_context` reuses the
  saved per-bot policy -> one change covers the whole chain. The lightweight `BaasService.restart_bot` (`/restart`) does not re-compose storage; it keeps the quota
  persisted at create time, so no change needed.
  - Rejected alternatives: (1) a new dedicated `param_code` (e.g. `upfs_storage`) — more "separate" but requires ops to create a new DRM item, higher marginal cost;
  (2) a new `upfs_quota` key inside the same param_value — a compromise but mixes semantics. Settled on reusing the existing `storage` item's `quota` key (NAS stops
  reading it, only UPFS reads it) for the minimal diff.
  - corp `sandbox_client.py` NAS path already hardcodes `quota="1Gi"` and does not call `apply_to_storage`; the ACK composer path does not apply quota — both keep
  `1Gi`, no change needed.

  ## Validation

  - ocb-public community suite: `DEPLOY_PROFILE=test uv run pytest tests/community/core/common_config tests/community/core/bot_management
  tests/community/core/service_bot -q` -> **2556 passed**.
  - Avernet same suite -> **2571 passed**.
  - `ruff check` on changed files -> All checks passed.
  - Updated tests (old "NAS also consumes shared quota" assertions -> NAS=1Gi / UPFS=config-or-default-1Gi), 4 sites:
    - `test_bot_storage_policy.py` `test_personal_create_payload_and_saved_storage_identity`: NAS -> `1Gi`, UPFS -> `(quota or "1Gi")`.
    - `test_personal_bot_storage.py` `test_switch_off_payload_keeps_nas_builder_quota_without_writing_policy` (renamed from `...only_changes_shared_quota...`): NAS
  `actual` equals Noop `legacy`, both `1Gi`, all fields identical.
    - `test_personal_bot_storage.py` `test_service_draft_restart_reads_policy_without_rollout`: NAS (policy=None) -> `1Gi`, UPFS -> configured value `1G`.
    - `test_common_config_service.py` `get_storage_quota` parametrization: fallback `1G` -> `1Gi` (None/True/0/-1/1.5/huge int/empty string); explicit values
  `1G`/`2G`/`2Gi` unchanged.
  - corp tests `grep` for storage-quota dependency = **0 hits** (corp path unaffected).
  - Not run: dev/prod end-to-end real bot creation (singlebox/MOSN); this is config/DI-layer logic, unit tests cover the core branches. The ocb side ocb-public
  working-copy edits are uncommitted (per request, only Avernet committed).

  ## Compatibility and risk

  - **Compatibility**: NAS behavior unchanged = `1Gi` (matches the original builder value), existing bots are unaffected. UPFS still reads the same config item and
  key; if it previously relied on the default `"1G"`, the default is now `"1Gi"`, but BaaS `_parse_upfs_quota` parses both `1G` and `1Gi` to `1073741824` bytes, so
  the mounted size is unchanged.
  - **Config**: no schema change, no new config item.
  - **Rollback**: `git revert c6c7da763` restores the unconditional override and the `1G` default. Note the ocb side ocb-public working-copy edits are uncommitted;
  after reverting Avernet, align/revert the ocb-public copy too.
  - **Risk**: UPFS default literal `1G` -> `1Gi` only changes the string in logs/payload; byte value is unchanged, platform is unaffected.

  ## Spec

  - No standalone design doc; design was done inline via brainstorming (bounded). Key decision: reuse the existing `storage` item's `quota` key (Option C) + default
  `1Gi`.
  - Related analysis: `docs/nas-upfs-migration/2026-08-31-analysis-and-decisions.md`.

  ## Related issues

  - Related to the NAS->UPFS migration project. This change only splits the quota; it does not touch the existing-bot migration path (U2 volume model / U3 migration
  trigger untouched).


totalfrank (approved):
LGTM.

PR #2348, opened 2026-09-20
feat(bcs): support per-bot provider webhook endpoints

vzvince:
## Problem

BCS currently sends every Bot managed by a Provider to the Provider's shared webhook URL. Platforms such as Poolab assign a stable endpoint to each Bot, so Bots under the same Provider need independent delivery addresses.

For example, Bot A and Bot B must receive requests at their respective endpoints, while existing Bots continue using the shared Provider endpoint.

## Solution

- Add an optional `webhook_url` to `ProviderBotBinding` and make the Provider default URL optional. `BotCore` resolves the Bot override first, then the Provider default.
- Extend Bot registration, listing, and PATCH responses to expose the saved override. Gateway registration rejects a missing effective endpoint before creating the Bot; Plugin Bots reject explicit webhook URLs.
- Preserve registration idempotency: omitting the URL keeps the saved value, and an explicitly different URL returns HTTP 409.
- Define address-only PATCH behavior:

  | `webhook_url` input | Behavior |
  | --- | --- |
  | Omitted | Keep the current value; existing capability updates remain supported. |
  | URL string | Set or replace the Bot override. |
  | `null` | Inherit the Provider default; reject if no default exists. |

  Mixing endpoint changes with capability changes is rejected before writes.
- Persist the field in Memory and DB stores, add SQLite migration **028** and MySQL migration **027**, and invalidate the local binding cache after successful writes.
- Preserve ownership, outbound URL, and disabled-state checks. Validate an effective endpoint before a WS-to-Provider switch writes a binding or disconnects WebSocket. A failed explicit endpoint never falls back to the Provider URL or WebSocket.
- Update integration documentation and contract tests; keep both integration guides under `src/bcs/docs` and update their incoming and relative links. The Provider integration program calls these APIs directly; Backend is unchanged.

## Validation

| Check | Result |
| --- | --- |
| Workspace `cargo check --all-targets` | Passed. |
| `bash src/bcs/scripts/ci_test.sh --fast-fail` | **4,946 passed, 49 skipped**, exit code 0. |
| New HTTP receiver integration tests | **2 passed**, covering 1.0 callbacks and 2.0 SSE, independent/default routing, failure without fallback, and PATCH rerouting. |
| Existing Provider downlink integration tests | **9 passed**. |
| DTO, Core, HTTP management, and storage contracts | Passed, including legacy payloads, PATCH semantics, ownership/disabled guards, and write failures. |
| SQLite migration and cache tests | Passed for fresh/legacy databases, repeated startup, persisted values, local invalidation, and cross-instance visibility after the existing TTL. |
| Configuration validation | **126 passed** in the host environment. |
| `git diff --check` | Passed. |

Network tests ran outside the filesystem sandbox with proxy variables removed from the test process.

Outstanding validation:

- The architecture aggregate reported **4 passed, 5 skipped, and 6 failed**. Its configuration check subsequently passed in the host environment; dependency-script, import/naming, and conformance checks remain unresolved.
- The protocol compatibility script skipped because it references the obsolete `crates/service-api/bcs-protocol/tests` path. R25 conformance checks did not pass.
- The all-module Singlebox run stopped when BaaS failed its startup health check. Acceptance tests and coverage reports were not produced, and artifact verification failed.
- The MySQL migration was inspected but not executed against a live MySQL instance. A real Poolab receiver has not been tested.

## Compatibility and risk

- Existing requests and bindings without an override continue inheriting the Provider default. Provider queries can now return `webhook_url: null`. Downstream 1.0/2.0 payloads and Provider authentication remain unchanged.
- Deploy the schema migration, upgrade all BCS instances, then enable per-Bot URLs. Older instances ignore overrides; rollback requires pausing affected traffic or providing compatible shared routing.
- Endpoint changes are maintenance operations: pause new traffic, drain active runs, update the address, and allow other instances' existing **30-second cache TTL** to expire. Active runs are not migrated.
- Known minor contract discrepancy: malformed PATCH field types receive Axum's existing **HTTP 422** plaintext rejection, while the design currently specifies HTTP 400. Both reject the write.
- Several touched legacy files remain above the repository's 1,000-line guideline; their existing layout is retained.

## Spec

- [Design and API semantics](https://github.com/vzvince/Avernet/blob/bbc04da9d6aaf495b0075c5ff49c369e0f91ca11/src/bcs/specs/2026-09-20-bot-webhook-endpoints/spec.md)
- [Detailed validation record](https://github.com/vzvince/Avernet/blob/bbc04da9d6aaf495b0075c5ff49c369e0f91ca11/src/bcs/specs/2026-09-20-bot-webhook-endpoints/validation.md)
- [Provider integration guide](https://github.com/vzvince/Avernet/blob/bbc04da9d6aaf495b0075c5ff49c369e0f91ca11/src/bcs/docs/bot-provider-integration.md)


vzvince:
@codex Review

RaymondHX (approved):

vzvince:
@codex review

totalfrank (approved):
LGTM.

PR #2350, opened 2026-09-20
feat(backend): accept entity_id as the addressed-owner query parameter, deprecate owner_id

totalfrank:
## Problem

The public OpenAPI surface names the owner of an addressed bot with an `owner_id` query parameter. We are renaming it to `entity_id`. During the transition both spellings must be accepted, with `entity_id` taking precedence and `owner_id` still honoured, so existing clients keep working while they migrate; `owner_id` is dropped once every client has moved.

Changing only `resolve_owner_id` would not have been enough: the addressed-bot grant check (`require_granted_addressed_bot`) reads the raw query string *before* any handler runs. If only the handler learned `entity_id`, an application sending `?entity_id=<owner>` would have had its grant looked up against the delegating user's own bot and then been refused for disagreeing with the handler's owner. The two readers have to move together.

## Solution

- **One rule, one place.** `principal.addressed_owner(entity_id, owner_id)` returns `entity_id` when present, else `owner_id`, else `None`. Both present and different is refused as a validation failure (422) before any grant or collaborator lookup, instead of silently preferring one side and acting on a bot the caller did not intend.
- **Both wire readers use it.** `require_granted_addressed_bot` (raw query read for the grant check) and `engine_runtime.resolve_owner_id` (declared parameters consumed by every engine-runtime, skills, config-manifest, cli-tools and caller-identity handler) resolve the pair through the same function, so the check and the handler cannot name different owners.
- **Standalone readers follow.** The product-chat list/detail operations and the bot editor-request operation, which declare their own owner parameter, now take `entity_id` plus the deprecated `owner_id` through the same helper.
- **Published as deprecated.** `owner_id` is declared with `Query(deprecated=True)` and a migration description, so the generated OpenAPI document and generated clients carry the notice.
- **Docs** for the shared-bot addressing section (en + zh-CN), the service-bot contract docs and the config-manifest user manual now describe `entity_id` as the parameter and `owner_id` as its deprecated alias.

Out of scope, deliberately: request-body and response fields named `owner_id` (bot metadata queries, lifecycle results, channels, routines, authorized apps) are unchanged. The legacy `openapi_v1/deprecated` addresses keep their frozen contract.

One pre-existing test (`test_caller_identity_endpoints`) asserted that `entity_id` is *not* a published parameter, because the caller-identity contract once replaced the internal API's `entity_id`/`ctoken` parameters. It now asserts the new spelling is present; the `ctoken` and lock exclusions are kept.

## Validation

- New unit tests for `addressed_owner` (`test_addressed_owner.py`): precedence, fallback, agreement, and the 422 on disagreement with neither value in the message.
- New end-to-end tests: application caller addressing a shared bot by `entity_id` (200), wrong `entity_id` (404), both spellings agreeing (200), disagreeing (422, no relay attempt); human caller by `entity_id` (200) and disagreeing (422); skills collection by `entity_id`; work-orders editor request by `entity_id` and the conflict case.
- Contract tests updated: skills operations publish both parameters with `owner_id` marked deprecated.
- `pytest tests/community/adapters/http/openapi_v1 tests/community/endpoints tests/community/adapters/http/skill_center`: 3328 passed, 2 skipped.
- `ruff check` on all changed files: clean.
- Not run: the full backend suite and `ci_test.sh` coverage gate (CI will run them).

## Compatibility and risk

- Every request valid before this change behaves identically: `owner_id` alone is still read, and omitting both still means the caller's own bot.
- New behaviour only for requests that send both spellings with different values: 422 instead of whichever value the old single reader would have taken (there was no such request before, since `entity_id` was not a parameter).
- Rollback is a revert; no schema or config impact.

https://claude.ai/code/session_01RCCrotuy4mKQRLsRV3Yz7M

---
_Generated by [Claude Code](https://claude.ai/code/session_01RCCrotuy4mKQRLsRV3Yz7M)_


src/backend/src/agentclaw/community/adapters/http/openapi_v1/bot_chats/router.py
```diff
@@ -50,12 +51,18 @@ async def list_bot_chats(
     request: Request,
     bot_id: BotIdPath,
     user_id: UserIdDep,
-    owner_id: str | None = Query(
+    entity_id: str | None = Query(
```
totalfrank-personal: Why this and owner_id query field defined below? Why not use OwnerIdDep?

src/backend/src/agentclaw/community/adapters/http/openapi_v1/bot_chats/router.py
```diff
@@ -147,12 +155,18 @@ async def get_bot_chat(
         ),
     ],
     user_id: UserIdDep,
-    owner_id: str | None = Query(
+    entity_id: str | None = Query(
```
totalfrank-personal: Same here.

src/backend/src/agentclaw/community/adapters/http/openapi_v1/work_orders/router.py
```diff
@@ -226,19 +228,26 @@ async def create_bot_editor_request(
     body: CreateBotEditorRequest,
     request: Request,
     caller: ActingCallerDep,
-    owner_id: Annotated[
+    entity_id: Annotated[
```
totalfrank-personal, line 238: Why this and owner_id query field defined below? Why not use OwnerIdDep?

totalfrank-personal (commented):

src/backend/src/agentclaw/community/adapters/http/openapi_v1/bot_chats/router.py
totalfrank: Agreed, no reason for a hand-declared pair here: this operation is `GRANT_CHECKED_ADDRESSED_BOT`, which is exactly the mode `OwnerIdDep` serves. Switched both product-chat handlers to `owner_id: OwnerIdDep` in 2ca58f5d. The grant dependency is now declared transitively through it as well as at the route, and the handler still discards the value since chat visibility stays scoped to the acting user.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/bot_chats/router.py
totalfrank: Done in 2ca58f5d, same as the list handler: `owner_id: OwnerIdDep`.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/work_orders/router.py
totalfrank: This one I left as a hand-declared pair, with the reason now in a comment above it (2ca58f5d). `OwnerIdDep` transitively declares `require_granted_addressed_bot`, and this operation is `USER_GATED` with a `NoCheck` authorization row: the work-order service adjudicates, and the applicant is by definition not yet a collaborator, so the grant check's masked 404 would be the wrong answer. `test_admission_inventory.test_every_grant_checked_operation_declares_its_modes_dependency` asserts the pairing in both directions and fails when a non-grant-checked route picks up that dependency.

Using `OwnerIdDep` here would mean moving the operation to `GRANT_CHECKED_ADDRESSED_BOT`, which changes admission for application callers (a grant on this specific bot instead of any live delegation from the user). That is a real behaviour change beyond the rename, so I'd rather keep it out of this PR. The pair still resolves through the same `addressed_owner` rule as every other reader. Happy to do the mode change in a follow-up if you want it.

---

totalfrank:
**CI note: the red "BCS architecture checks" run is stale and not this PR's.**

- It ran once, on the superseded head 82c36b7f, only because the PR was briefly opened against `main`. That workflow triggers solely on `pull_request` to `main`, so it has not run and will not run on the current head (2ca58f5d) now that the base is `dev`.
- Its failures (R25 conformance, DEP-1~8, LINT-1/4, TEST-1, PR-1) are all BCS constitution checks under `src/bcs`, evaluated against `origin/main`. This PR touches no file under `src/bcs`; the diff is confined to `src/backend` (open API adapter, tests, docs). The same failures reflect `dev` vs `main` drift and reproduce on any `dev`-based branch opened against `main`.
- No fix to port: there is nothing in this PR that could change that check's outcome, and I have no way to re-run a workflow that no longer targets this PR.

Every check that does run on the current head is green so far; Backend unit tests and Singlebox coverage are still in progress.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/work_orders/router.py
```diff
@@ -226,19 +228,33 @@ async def create_bot_editor_request(
...
+    owner_id: Annotated[
+        str | None,
+        Query(max_length=256, deprecated=True, description=OWNER_ID_DESCRIPTION),
+    ] = None,
     service: WorkOrderServiceProtocol = Injected(WorkOrderServiceProtocol),
 ) -> Envelope[BotEditorRequestCreated]:
     actor_id = _require_user_delegation(caller)
+    # Same rule as every other reader of the addressed owner: ``entity_id``,
+    # else its retiring alias, else the caller.
+    named_owner = addressed_owner(entity_id, owner_id)
```
totalfrank: Non-blocking: `entity_id` and `owner_id` here are declared with `max_length=256` only (no `min_length=1`, unlike every other reader of `addressed_owner` — `OwnerIdDep`'s `entity_id`/`owner_id` in `engine_runtime/params.py` and the raw-query read in `principal.require_granted_addressed_bot`, both of which normalize an empty string to `None` before calling `addressed_owner`). So `POST .../editor-requests?entity_id=&owner_id=some-real-owner` reaches `addressed_owner(\"\", \"some-real-owner\")`, and since `\"\" is not None` and `\"\" != \"some-real-owner\"`, it raises the disagreement 422 — refusing a request that every other endpoint on this surface would treat as \"only `owner_id` was named\" (empty is absent). This is exactly the coexistence case the PR is meant to support during migration, and it isn't covered by `test_create_bot_editor_request_refuses_disagreeing_owner_spellings` (which uses two non-empty values).

Suggested fix, matching the normalization already used elsewhere in this PR:
```suggestion
    named_owner = addressed_owner(entity_id or None, owner_id or None)
```

---

totalfrank (commented):
Reviewed commits 82c36b7f and 2ca58f5d (the full diff against merge-base e418db68).

Checked: correctness of the `addressed_owner` precedence/conflict rule and both wire readers (`require_granted_addressed_bot`, `resolve_owner_id`), the `deprecated=True` publication on `owner_id`, doc/contract consistency (README en/zh-CN, service-bot contract, config-manifest manual, interface checklist), and test coverage for the new behavior across `test_addressed_owner.py`, `test_app_only_caller.py`, `test_skills_shared_bot_grant.py`, `test_skills_contract.py`, `test_caller_identity_endpoints.py`, and `test_work_order_contract.py`.

One non-blocking finding posted inline: `create_bot_editor_request`'s hand-declared `entity_id`/`owner_id` pair skips the empty-string normalization every other reader applies before calling `addressed_owner`, so an empty `entity_id` alongside a real `owner_id` gets spuriously refused with 422 instead of falling back to `owner_id` — exactly the coexistence case this PR is meant to support.

No contract-breakage, concurrency, security, or architecture-boundary issues found in the diff.

---

src/backend/src/agentclaw/community/adapters/http/openapi_v1/work_orders/router.py
totalfrank: Verified and fixed in 301891d4: the handler now normalizes both spellings with `or None` before `addressed_owner`, matching the grant dependency's raw read, and `test_create_bot_editor_request_treats_an_empty_entity_id_as_absent` pins `?entity_id=&owner_id=bot-owner` → 201 with `bot-owner` as the owner.

---

totalfrank (commented):
Reviewed commit 301891d4 (the only new commit since the last round, against 2ca58f5d).

Verified: `create_bot_editor_request` now normalizes both spellings with `entity_id or None` / `owner_id or None` before calling `addressed_owner`, matching the grant dependency's raw read. `?entity_id=&owner_id=bot-owner` falls back to `bot-owner` instead of tripping the disagreement 422, and the new test (`test_create_bot_editor_request_treats_an_empty_entity_id_as_absent`) pins exactly that case. No new issue in this commit — it's a minimal, targeted fix plus test.

Status checklist:
- `bot_chats/router.py` list/detail: use `OwnerIdDep` instead of hand-declared pair (human review) — fixed in 2ca58f5d.
- `work_orders/router.py` editor-request: why not `OwnerIdDep`? (human review) — author's call, explained (USER_GATED/NoCheck mode, `OwnerIdDep` would change admission semantics beyond the rename); conceded/resolved.
- `work_orders/router.py:254` empty `entity_id` not normalized before `addressed_owner`, causing spurious 422 (non-blocking, mine) — fixed in 301891d4, verified above.

0 findings open. PR is clean as of this round (commit 301891d4).

---

xianmuyq (approved):

PR #2355, opened 2026-09-20
feat(clawweb): workflow run list filtered by bot_workflow_permissions can_view with wildcard support

fudp:
## Problem
Previously the workflow run list endpoint (/api/runs) automatically filtered flow_runs.origin_bot_id by the current user's owner ID, even when the user had explicit can_view=1 permission in bot_workflow_permissions. This caused users like 272654 to see an empty run list for social-appeal-review-flow despite having view permission.

## Solution
- Introduce BotWorkflowPermissionRepository.resolveViewScope(workflowId, userId) which returns 'all' | 'deny' | { botIds: string[] } based on can_view records, supporting wildcards for both bot_id and bot_owner_id.
- Replace the old originBotOwnerId/originBotId filtering in FlowRunRepository with allowedBotIds prefix matching on origin_bot_id.
- Update runs.ts GET / to use resolveViewScope. The existing botId query parameter is intersected with the resolved scope to preserve the UI bot filter while enforcing the permission boundary.
- Update getViewByIdsForOwner to support bot_id='*' and bot_owner_id='*' wildcards while keeping botId-specific behavior for the workflows list.
- Allow PermissionPanel to save and display botOwnerId='*' / botId='*' wildcard records.
- Add unit tests for resolveViewScope and getViewByIdsForOwner.

## Validation
- npm run check --workspaces --if-present: passed
- npm run test --workspace=@avernet/clawweb-shared --workspace=@avernet/workflow: 249 tests passed

## Compatibility and risk
- Admin users still bypass all permission checks.
- Edit/execute permissions are unchanged.
- workflow-types statistics remain unchanged (still returns total counts).
- Existing exact permission records continue to work as before.
- No database schema changes.


3h-william (approved):
Ok

PR #2360, opened 2026-09-20
fix(bcs): simplify loop configuration and enable workflow recovery

quechao76:
## Problem

The Loop execution setting exposes a temporary implementation name, while ordinary workflow recovery requires a separate opt-in that is implicitly overridden when Loop execution is enabled. This makes the configuration difficult to understand.

## Solution

- Rename `collaboration.experimental_fixed_loop_execution` to `collaboration.loop_execution_enabled`, keeping the default `false` and the local example enabled.
- Remove `collaboration.experimental_progression_recovery` and start the shared workflow recovery scanner with the service. Ordinary workflow recovery no longer requires an opt-in; Loop execution and recovery remain controlled by `loop_execution_enabled`.
- Preserve leader election, bounded scanning and shutdown cancellation. Update runtime callers, configuration examples, documentation and regression tests.

## Validation

- Ran `cargo test --manifest-path src/bcs/Cargo.toml -p bcs -p bcs-collaboration-runtime --lib --test runtime_progression --test openapi_v1_mount`: **479 passed, 0 failed, 5 existing ignored tests**.
- Verified ordinary DAG recovery with Loop disabled, rejection of removed configuration keys, scanner lifecycle and HTTP validation behavior.
- `git diff --check` passed. Rebase onto `origin/dev` completed without conflicts; the resulting tree is identical to the tested tree.
- Full Singlebox E2E was not run; validation was scoped to the affected configuration, bootstrap and workflow recovery behavior. The five existing ignored tests cover four Session timeout cases with incomplete test fixtures and one Redis factory case with an LLVM coverage registration issue.

## Compatibility and risk

- Rename existing `collaboration.experimental_fixed_loop_execution` entries to `collaboration.loop_execution_enabled` and remove `collaboration.experimental_progression_recovery`. Old keys are rejected; no compatibility aliases are provided.
- Ordinary workflow recovery now runs automatically, including when Loop is disabled. Existing leader election and batch limits remain in place.
- No schema or migration changes. Rollback requires reverting the code and restoring the previous configuration keys together.


questchaos:
只涉及配置精简和去掉 experimental_ 前缀，防止 ocb 项目意外加临时 experimental_ 配置，先把这个改动合并了

questchaos (approved):

questchaos:
@codex review

PR #2366, opened 2026-09-21
fix(backend): return bound bot owner in TC resource context

SergioLi0903:
## Problem

When a service Bot is called by a different user, the resource uploader/Caller is not necessarily the Bot Owner.

The TC Resource Context endpoint currently returns the resource uploader as `user_id` and the Bot base ID as `bot_id`, but does not return the actual Bot Owner. ECB therefore reconstructs the full Bot identity using the Caller ID:

```text
<bot_id>:<user_id>
```

For example:

```text
Actual Bot Owner: 490906
Current Caller:   437240

Expected X-TC-Bot-Id: 20260917_m0ib5ty8:490906
Incorrect value:      20260917_m0ib5ty8:437240
```

This mismatch causes downstream MCP file queries such as `list_files` to return an empty result.

## Solution

- Added `bot_owner_id` to `TcResourceContextSnapshot`.
- Added `bot_owner_id` to the TC Resource Context response payload.
- Resolve `bot_owner_id` through the persisted resource binding:

  ```text
  ac_session_resource.binding_id
      -> ac_entity_device_binding.id
      -> ac_entity_device_binding.entity_id
  ```

- Injected the existing `DeviceBindingRepository` into `TcResourceContextService`.
- Added fail-closed validation:
  - `binding_id` must be a positive integer.
  - The referenced Binding must exist.
  - `binding.entity_id` must be a non-empty string.
  - `binding.device_id` must match `resource.bot_uuid`.
- The Binding lifecycle status is intentionally not checked. Historical bindings such as `RELEASED` must remain resolvable for resource ownership reconstruction.
- Preserved the existing identity semantics:
  - `user_id` remains the resource uploader/current Caller from `ac_session_resource.owner_id`.
  - `bot_id` remains the Bot base ID from `ac_session_resource.bot_id`.
  - `bot_owner_id` is sourced from `ac_entity_device_binding.entity_id`.
- Kept the Resource Ready event unchanged. It continues to send only the `resource_id`.
- Did not expose `binding_id` in the external Resource Context response.
- Did not modify ECB tables or the OCB `ocb-public` submodule.

## Validation

The following validations passed:

- Core, Router, Endpoint, and DI tests:

  ```text
  30 passed
  ```

- Live acceptance test covering Binding insertion, resource-to-Binding association, and Resource Context resolution:

  ```text
  1 passed
  ```

- Ruff formatting check: passed.
- Ruff lint check: passed.
- OCB pre-push gate, including changed-range plaintext secret scan and backend SAST: passed.

The full repository-wide backend SAST command was not used as the blocking validation signal because the local recursive scan also inspected `src/backend/.venv` and surfaced unrelated third-party findings. The changed-range pre-push SAST passed.

## Compatibility and risk

This is an additive response contract change. The TC Resource Context response now includes:

```json
{
  "bot_owner_id": "..."
}
```

The Resource Ready event contract is unchanged.

The recommended rollout order is:

1. Deploy ECB support for the optional `bot_owner_id` field.
2. Deploy this Avernet/TC change.
3. Verify service Bot calls where the Caller and Bot Owner are different users.
4. Once the migration is stable, ECB may make `bot_owner_id` required.

If the resource binding is missing or inconsistent with the resource device UUID, TC returns the existing `resource_context_incomplete` failure instead of returning an unverifiable Bot identity.

## Spec

Implements the Avernet / TC Resource Context Bot Owner handoff dated September 21, 2026.
```

### CODE REVIEW

File: src/backend/src/agentclaw/community/core/tc_file_upload_integrations/resource_context.py [hunk 1]
```diff
@@ -27,6 +30,7 @@ class TcResourceContextSnapshot:
     bot_uuid: str
     user_id: str
     bot_id: str
+    bot_owner_id: str
```
FreddieSun: 统一叫owner_id吧， backend对于bot owner的统一命名

FreddieSun (changes requested):

src/backend/src/agentclaw/community/core/tc_file_upload_integrations/resource_context.py
SergioLi0903: 已解决。

FreddieSun (approved):

